PR #10, opened 2026-04-25
Remove Linux binary from the repository

zh:
`spinel_parse` is a Linux binary, product of doing `make`, Should not be in the repository. Without doing `make clean` after initial clone, compilation is failing on non-Linux platforms:

```make
$ make
=== Bootstrap Step 1: parse ===
./spinel_parse spinel_codegen.rb build/codegen.ast
./spinel_parse: ./spinel_parse: cannot execute binary file
make: *** [spinel_codegen] Error 126
```


crowell:
add the file to .gitignore to prevent it from coming back in the future

PR #11, opened 2026-04-25
 Integer-keyed string-valued hash support (sp_IntStrHash)                                                                          

xl350r:
**Problem Observed**                                                                                                                     

Spinel emitted `sp_StrStrHash` for all hash literals, including those with integer                                                
  keys like `{3 => "Fizz", 5 => "Buzz"}`. At runtime, `sp_str_hash()` dereferenced                                                  
  the integer key value as a `const char *` pointer, causing an immediate SIGSEGV.
  
  **Original ruby**
                                                    
  ```
  map = {3 => "Fizz", 5 => "Buzz", 7 => "Bazz"}

(1..1000000).each do |x|
	output = "#{x} => "
	map.keys.each do |k|  
	  if x % k == 0  
	    output = output + map[k]
	  end 
	end  
	puts output if output[-1].downcase == "z"
end 
```

**Failing Generated C code with segfault**
```
/* Generated by Spinel AOT compiler */
#include "sp_runtime.h"
static const char *sp_sym_to_s(sp_sym id){(void)id;return "";}


typedef struct{const char**data;mrb_int len;}sp_Argv;
static sp_Argv sp_argv;

int main(int argc,char**argv){
  sp_argv.len=argc-1;sp_argv.data=(const char**)malloc(sizeof(const char*)*(argc>1?argc-1:1));{int _i;for(_i=0;_i<sp_argv.len;_i++)sp_argv.data[_i]=sp_str_dup_external(argv[_i+1]);}
    SP_GC_SAVE();
    sp_StrStrHash *lv_map = NULL;
    SP_GC_ROOT(lv_map);
    mrb_int lv_x = 0;
    const char *lv_output = ("\xff" + 1);
    SP_GC_ROOT(lv_output);
    const char *lv_k = ("\xff" + 1);
    SP_GC_ROOT(lv_k);

    sp_StrStrHash *_t1 = sp_StrStrHash_new();
    sp_StrStrHash_set(_t1, 3, ("\xff" "Fizz" + 1));
    sp_StrStrHash_set(_t1, 5, ("\xff" "Buzz" + 1));
    sp_StrStrHash_set(_t1, 7, ("\xff" "Bazz" + 1));
    lv_map = _t1;
    sp_Range _t3 = sp_range_new(1, 1000000);
    for (lv_x = _t3.first; lv_x <= _t3.last; lv_x++) {
      lv_output = sp_sprintf("%lld => ", (long long)lv_x);
      sp_StrArray * _t4 = sp_StrStrHash_keys(lv_map);
      SP_GC_ROOT(_t4);
      for (mrb_int _t5 = 0; _t5 < sp_StrArray_length(_t4); _t5++) {
        lv_k = sp_StrArray_get(_t4, _t5);
        if ((sp_imod(lv_x, lv_k) == 0)) {
          lv_output = sp_str_concat(lv_output, sp_StrStrHash_get(lv_map, lv_k));
        }
      }
      const char * _t6 = sp_str_sub_range(lv_output, -1, 1);
      SP_GC_ROOT(_t6);
      if ((strcmp(sp_str_downcase(_t6), ("\xff" "z" + 1)) == 0)) {
        { const char *_ps = (const char *)(lv_output); if (_ps) { fputs(_ps, stdout); if (!*_ps || _ps[strlen(_ps)-1] != '\n') putchar('\n'); } else putchar('\n'); }
      }
    }
  return 0;
}
```

**New generated C code**
```
/* Generated by Spinel AOT compiler */
#include "sp_runtime.h"
static const char *sp_sym_to_s(sp_sym id){(void)id;return "";}


typedef struct{const char**data;mrb_int len;}sp_Argv;
static sp_Argv sp_argv;

int main(int argc,char**argv){
  sp_argv.len=argc-1;sp_argv.data=(const char**)malloc(sizeof(const char*)*(argc>1?argc-1:1));{int _i;for(_i=0;_i<sp_argv.len;_i++)sp_argv.data[_i]=sp_str_dup_external(argv[_i+1]);}
    SP_GC_SAVE();
    sp_IntStrHash *lv_map = NULL;
    SP_GC_ROOT(lv_map);
    mrb_int lv_x = 0;
    const char *lv_output = ("\xff" + 1);
    SP_GC_ROOT(lv_output);
    mrb_int lv_k = 0;

    sp_IntStrHash *_t1 = sp_IntStrHash_new();
    sp_IntStrHash_set(_t1, 3, ("\xff" "Fizz" + 1));
    sp_IntStrHash_set(_t1, 5, ("\xff" "Buzz" + 1));
    sp_IntStrHash_set(_t1, 7, ("\xff" "Bazz" + 1));
    lv_map = _t1;
    sp_Range _t3 = sp_range_new(1, 1000000);
    for (lv_x = _t3.first; lv_x <= _t3.last; lv_x++) {
      lv_output = sp_sprintf("%lld => ", (long long)lv_x);
      sp_IntArray * _t4 = sp_IntStrHash_keys(lv_map);
      SP_GC_ROOT(_t4);
      for (mrb_int _t5 = 0; _t5 < sp_IntArray_length(_t4); _t5++) {
        lv_k = sp_IntArray_get(_t4, _t5);
        if ((sp_imod(lv_x, lv_k) == 0)) {
          lv_output = sp_str_concat(lv_output, sp_IntStrHash_get(lv_map, lv_k));
        }
      }
      const char * _t6 = sp_str_sub_range(lv_output, -1, 1);
      SP_GC_ROOT(_t6);
      if ((strcmp(sp_str_downcase(_t6), ("\xff" "z" + 1)) == 0)) {
        { const char *_ps = (const char *)(lv_output); if (_ps) { fputs(_ps, stdout); if (!*_ps || _ps[strlen(_ps)-1] != '\n') putchar('\n'); } else putchar('\n'); }
      }
    }
  return 0;
}
```
  
                                                                                              
  **Files changed**                                                                                                                 
                                                                                                                                    
  `lib/sp_runtime.h`  : Add `sp_IntStrHash`, a hash type for integer keys and string                                                 
  values. Uses fibonacci hashing (`k * 11400714819323198485ULL`) and a                                                              
  `mrb_bool* used[]` sentinel array (integers have no NULL sentinel). Supports `set`,
  `get`, `has_key?`, `length`, `keys` returning `sp_IntArray`, `values` returning                                                   
  `sp_StrArray`, and insertion-order iteration via an `order[]` index.                                                              
                                                                                                                                    
  `spinel_codegen.rb` :  Wire `int_str_hash` type inference throughout: literal                                                      
  creation, `[]` lookup, `[]=` assignment, `has_key?`/`length`/`keys`/`values`/`fetch`,                                             
  `keys.each` (block param typed as `mrb_int`), `each |k,v|` iteration, and string                                                  
  `+=` via `sp_str_concat`.                                                                                                         
                                                                                                                                    
  `test/bm_int_str_hash.rb` : Regression test covering the above. 76/76 pass.

### CODE REVIEW

**Comment by matz** [comment]
Thanks for the report and the draft fix — the underlying bug is real and important. `{3 => "Fizz"}` segfaulting because keys are dereferenced as `const char *` is an obvious wart that needs an `sp_IntStrHash` or equivalent.

I tested the PR locally and the source is correct: running `ruby spinel_codegen.rb` on a clean `build/codegen.ast` with this branch produces the right output, and `test/bm_int_str_hash.rb` passes.

**However, after a clean bootstrap (gen2 == gen3), the self-hosted compiler regresses on `benchmark/bm_ao_render.rb`.** Symptom: `lv_z` (a local Float from `z = Math.sqrt(r)`) miscompiles to a literal `\x01` byte:

```c
/* expected: */ lv_ry = (((lv_x * _t56->y) + (lv_y * _t57->y)) + (lv_z * _t58->y));
/* actual:   */ lv_ry = (((lv_x * _t56->y) + (lv_y * _t57->y)) + (\x01 * _t58->y));
```

cc rejects this with `error: stray '\1' in program`. The breakage shows up after just the first commit of the three (`feat(runtime,codegen): add sp_IntStrHash`). On a fresh clone:

```
$ git checkout master && git cherry-pick <commit-1>
$ make bootstrap        # gen2 == gen3 OK
$ make bench
ERR:  bm_ao_render
```

This is not a runtime issue in `sp_IntStrHash` itself — `bm_ao_render` doesn't use any integer-keyed hashes. The newly-added codegen Ruby (~162 lines spread across `infer_method_name_type`, `scan_features`, `scan_locals`, `compile_each_block`, etc.) is correct under CRuby but, when self-compiled, triggers a latent miscompilation in an unrelated spot.

We've hit this same family of self-host bug before (`@ivar.push` inside loops, `@ivar << long_concat_chain`) and it's recorded as a known issue. The PR isn't doing anything wrong per se; it just adds enough ivar-mutating codegen to cross whatever threshold trips the self-host compiler.

Two ways forward — happy to do whichever you prefer:

1. **You investigate which specific construct in the diff trips the bug**, and we land a more surgical version. A useful narrowing: `git bisect run` against `make bench | grep bm_ao_render` should isolate it within minutes once you know the symptom. Things to try removing first: the `scan_locals` special case for `int_str_hash.keys.each` (lines ~10545+), and any large `if/else` chain you added inside an `@ivar`-mutating method.

2. **We hold this PR until the underlying self-host bug is properly diagnosed**. That diagnosis is overdue and not your responsibility — we'll do it as a separate task. If you can wait, this PR's diff would likely apply cleanly once the root cause is fixed.

Either way, your patch is valuable as it stands, since it's the right design and the test case is solid. Let us know how you'd like to proceed.

xl350r:
Took a swing at fixing the regression (option 1).
                                                                                                                                                   
Spinel emitted `sp_StrStrHash` for all hash literals, including those with integer keys like `{3 => "Fizz", 5 => "Buzz"}`. At runtime, `sp_str_hash()` dereferenced the integer key value as a `const char *` pointer, causing an immediate SIGSEGV.                                                                    
                                                                                                                                                      
  **Files changed**                                                                                                                                   
1. lib/sp_runtime.h: the grow function now iterates the slot array directly for O(N) rehash, rather than looking up each insertion-order key via linear scan. 
                                                                                                                                       
 2. spinel_codegen.rb: fixes the miscompilation reported by matz — the string += handler was emitting sp_str_concat whenever the left-hand variable was typed as string, without checking the right-hand value type. Passing a non-string (int, float) to sp_str_concat is undefined behaviour and was producing stray bytes in unrelated output. The handler now requires both sides to infer as string; mismatched types fall back to +=. This should fix the \x01 bytes reported in bm_ao_render.rb.
                                                                                                                                                                                                                                                                                                            
  **limitations**
                                                                                                                                                      
  `fetch` evaluates the key expression twice in the `has_key?`/`get` ternary. Safe
  for integer literals and constant expressions, which are the only key types valid
  for `sp_IntStrHash`. Fixing it via a C temp variable is not possible in the current
  expression-returning codegen context without corrupting the self-hosted build.     
                                                                                                                                                      
  String `+=` via `sp_str_concat` is applied to local variables only. Instance and
  global variable `+=` on strings is deferred; patching `InstanceVariableOperatorWriteNode`                                                           
  during self-compilation causes ivar resolution to emit raw `@name` into the
  generated C.

PR #15, opened 2026-04-25
Support instance_eval { block } in DSL receivers

khasinski:
## Question first

I had a go at lifting one of the README's listed limitations (`instance_eval`) and would value your read on whether this is the right shape before going further. Happy to close this if the approach does not fit the project.

## What this does

A new pass between class collection and return-type inference detects `recv.instance_eval do ... end` where `recv` is a local previously bound to `ClassName.new`. The block body is lifted into a synthetic instance method `__sp_ieval_<N>` on that class, and the call is rewritten to dispatch to it. After that, normal user-method codegen takes over: `self` is the receiver, `@ivars` and `attr_accessor` setters work without changes, type inference flows through the existing pipeline.

89 lines added in `spinel_codegen.rb`, no existing code changed.

## Verified

- Bootstrap closes (`gen2.c == gen3.c`).
- `make test`: 74 pass, 0 fail, 1 error (the `gvar` error is pre-existing on master).
- A small Sinatra-style routing DSL and an HTML builder DSL both compile and produce identical output to CRuby.

## Deliberate v1 limits

- Receiver must be a top-level local from `ClassName.new`. Receivers held in ivars, parameters, or returned from methods are not tracked yet.
- Block takes no parameters.
- No closure capture of outer locals.

## The actual question

Two things I would like your view on:

1. Is the synthetic-method-lifting approach the right direction, or would you prefer a different mechanism (for example a parse-time rewrite, or treating `instance_eval` as a first-class node)?
2. If the direction is right, would you be open to follow-ups that lift the v1 limits one at a time (wider receiver tracking, then block params, then closure capture)? Each step is a separate, contained change.

Thanks for considering, and no rush.

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -250,6 +250,13 @@ class Compiler
 
     # Symbol type Phase 2 Step 1: intern table (infrastructure only; unused yet).
     @sym_names = "".split(",")
+
+    # instance_eval block hoisting: parallel arrays indexed by synthetic
+    # function id N. Each lifted block becomes a file-scope static
+    # function `sp_ieval_<N>` that takes a typed `self` parameter.
+    @ieval_counter = 0
+    @ieval_class_idxs = []
+    @ieval_body_ids = []
   end
 
   # Backslash-n for C string literals - bootstrap-safe (avoids escape level issues)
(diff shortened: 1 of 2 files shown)


khasinski:
Pushed `e632f14` addressing all four review points:

1. **Scope (high)**: Replaced the global `@nd_count` sweep with a recursive descent that does not enter `Def`/`Lambda`/`Class`/`Module`/`Block` bodies. The top-level local map is no longer applied to identically-named locals inside methods. Verified with a test where a top-level `cfg : Config` shadows a method parameter `cfg : Other` — Spinel now correctly leaves the inner call alone instead of emitting a `Config*` cast over an `Other*`. Added flow-sensitive tracking too: a reassignment to a non-`Class.new` RHS poisons the entry, so subsequent `instance_eval` on that name is not rewritten.

2. **Yield (high)**: Now uses `body_has_yield(body_id)` instead of hardcoded `0`, so the synthetic method's C signature matches blocks that contain `yield` or `block_given?`.

3. **ConstantPath (medium)**: `Foo::Bar.new` now resolves through the leaf name, matching how `collect_class` records nested classes (Spinel's class registry is keyed on the leaf, not the full path).

4. **Block parameters (medium)**: `obj.instance_eval { |x| ... }` is now skipped at the rewrite step, so a parameterized block can never be lifted into a no-arg synthetic method.

Bootstrap still closes (`gen2.c == gen3.c`) and `make test` is unchanged (74 pass, 0 fail, 1 pre-existing error).

matz:
Thanks for the careful framing — the README limitation is intentional in spirit, but I'm open to lifting it as long as the implementation fits the rest of the compiler. Yes to direction, but I'd prefer a different shape than synthetic-method-lifting.

**On the approach.** What `instance_eval` actually needs is "compile the block body with `self` rebound to `recv`". That doesn't have to be done by adding a method to the class. We already have a precedent for compiling a block as a free function at file scope: `compile_proc_literal` was hoisted to file scope in #6 (commit `4593581`), and `compile_lambda_body_expr` already emits a static function with a captures struct. I'd like `instance_eval` to use the same mechanism rather than mutating the class's method list.

**Concretely**, I'd suggest emitting something like:

```c
static <ret> sp_ieval_<N>(sp_<Class> *self, /* capture struct, if any */) {
  /* block body — `self` is just a regular parameter typed to recv's class */
}
```

and rewriting the call site to `sp_ieval_<N>(recv, ...)`. This is the same code path the proc/lambda hoist takes, plus a typed `self` parameter. Compared to synthetic-method-lifting:

- No `__sp_ieval_<N>` entries in class metadata, no risk of colliding with user method names, no entries showing up in method-lookup or `respond_to?`-style dispatches we may add later.
- The receiver-class flow analysis you've already written stays — we still need to know recv's class to type the `self` parameter — but it's the only role that analysis plays. The class otherwise stays untouched.
- Closure capture (your v2 follow-up) drops onto the existing lambda capture-struct path rather than a new mechanism.
- `return` inside the block becomes a function-level return, which is a deviation from Ruby's "return from enclosing method" semantics. In practice the DSL pattern doesn't use `return`, but if/when we want it that's a separate, contained piece of work — same as it is for lambdas today.

**On the v1 limits**: yes to incremental follow-ups (wider receiver tracking, then block params, then closure capture). One PR per step, each landing only when bootstrap + bench + test are all green on master.

If you're up for re-spinning this on top of `master` (which now has CI on push, plus a bunch of perf and macOS portability changes), I'd merge a hoisted version. Don't worry about closing this draft — leave it open until the new approach lands, or convert in place, whichever is easier for you.

khasinski:
Pushed the re-spin per your feedback.

What changed:

- Synthetic-method-on-class lifting is gone. Each `recv.instance_eval do ... end` now hoists the block into a file-scope `static void sp_ieval_<N>(sp_<Class> *self)` and rewrites the call site to invoke that function directly. Class metadata is untouched.
- The receiver-class flow analysis stays. Its only role now is to type the `self` parameter.
- Added a sibling inference pass `infer_ieval_body_call_types` so bare self-calls inside the block (e.g. `app.instance_eval { get("/") }`) still propagate arg types into the receiver class's methods, the same way `infer_class_body_call_types` does for regular method bodies. Without this, `Routes#get(path)` lost its `path: const char*` inference once the block stopped being a registered method.

v1 scope is unchanged from the original draft (top-level local previously bound to `Class.new`, no block params, no closures, no yield). Block return value is discarded; lifted function is `void`. Expression-context calls evaluate to the receiver via a comma expression so `if obj.instance_eval { ... }` still type-checks.

Rebased onto current `master`. Bootstrap closes (`gen2.c == gen3.c`). `make test`: 104 pass, 0 fail, 0 error.

Wider receivers, block params, closures, and a real return value are the planned follow-ups, one PR per step.

matz:
Merged. Thanks for the careful v1 scoping and the question up front — both the synthetic-method-lifting direction and the staged-relaxation plan look right. Welcome follow-ups in the order you proposed: wider receiver tracking (ivars / params / method returns) first, then block params, then closure capture.

Note that #124 landed in parallel for the trampoline shape (`def m(&b); instance_eval(&b); end`). The two paths cover different call sites and don't overlap, but if either one's detection grows it's worth checking the other so the gates stay disjoint.

PR #21, opened 2026-04-26
fix(codegen): route Array#size to length for IntArray/SymArray/StrArray

khasinski:
## Summary

`[1, 2, 3].size` returned `0` even though `[1, 2, 3].length` returned `3`.
The IntArray/SymArray and StrArray dispatch in `compile_array_method_expr`
matched only `mname == "length"`, so `"size"` fell through to a default
that returned `0`. FloatArray and PtrArray already accept both names; this
brings the remaining three array types in line.

\`\`\`ruby
b = [1, 2, 3, 4, 5]
puts b.size            # was: 0    now: 5
puts [:a, :b].size     # was: 0    now: 2
puts [\"x\", \"y\"].size   # was: 0    now: 2
\`\`\`

## Implementation note

In the \`str_array\` branch, \`\"size\"\` is added as its own \`if\` arm rather
than \`if mname == \"length\" || mname == \"size\"\`. The compound form
destabilises the self-build: gen1 emits valid C, but bin1 then
mis-compiles unrelated \`==\` comparisons in \`spinel_codegen\` itself,
producing fragments like \`if (( && ...))\`. The duplicate-arm form
sidesteps it. Worth chasing the underlying inference quirk separately;
I hit the same pattern in two other unrelated fixes.

## Test plan

- [x] \`make test\` passes (74 pass; the one pre-existing \`gvar\` error is
  unaffected).
- [x] \`test/bm_array_lit.rb\` extended with \`.size\` on IntArray, SymArray,
  StrArray, and on an array post \`push\`/\`pop\`. CRuby and Spinel outputs
  match.

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -14008,6 +14008,12 @@ class Compiler
         end
         return "sp_IntArray_length(" + rc + ")"
       end
+      if mname == "size"
+        if @hoisted_strlen_var != "" && @hoisted_strlen_recv == rc
+          return @hoisted_strlen_var
+        end
+        return "sp_IntArray_length(" + rc + ")"
+      end
       if mname == "[]"
         return "sp_IntArray_get(" + rc + ", " + compile_arg0(nid) + ")"
       end
@@ -14225,6 +14231,12 @@ class Compiler
     end
     if recv_type == "str_array"
       if mname == "length"
+        if @hoisted_strlen_var != "" && @hoisted_strlen_recv == rc
+          return @hoisted_strlen_var
+        end
+        return "sp_StrArray_length(" + rc + ")"
+      end
+      if mname == "size"
         if @hoisted_strlen_var != "" && @hoisted_strlen_recv == rc
           return @hoisted_strlen_var
         end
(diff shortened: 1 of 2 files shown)


khasinski:
Thanks for the review. Pushed an update (`57b5cb6`) addressing both comments:

1. **str_array `size` arm now carries `@hoisted_strlen_var`** -- direct fix
   from your suggestion. `size` and `length` now have identical perf
   characteristics on str_array.

2. **Split `int_array`/`sym_array` into two arms for consistency** -- the
   bot's first comment misread the original diff (`if mname == "length" || mname == "size"`
   covered both branches; `size` did return `sp_IntArray_length(...)` in
   non-hoisted contexts, the test verifies this). But the underlying
   point about consistency with the str_array shape is fair, so all
   three array types now use the same two-arm pattern.

The `||` form does work for int_array/sym_array (both my original commit
and the test confirm it), but compounding it specifically on str_array
breaks the self-build. Documented in the commit message; worth a
separate look at the inference quirk.

Test extended with a hoisted-length-style usage (`while i < arr.size`)
for all three array types, so the `.size` perf path is exercised.

PR #22, opened 2026-04-26
feat(codegen): add Math.tan, atan, asin, acos, log10, hypot

khasinski:
## Summary

\`Math.sin\`, \`Math.cos\`, \`Math.sqrt\`, \`Math.log\`, \`Math.log2\`, \`Math.exp\`,
and \`Math.atan2\` are dispatched, but their close relatives are not.
Calls to the missing methods returned \`0\`:

\`\`\`ruby
Math.sin(0.5)    # 0.479  (works)
Math.cos(0.5)    # 0.878  (works)
Math.tan(0.5)    # was: 0       now: 0.546
Math.atan(1.0)   # was: 0       now: 0.785
Math.asin(1.0)   # was: 0       now: 1.5708
Math.acos(0.0)   # was: 0       now: 1.5708
Math.log10(100)  # was: 0       now: 2.0
Math.hypot(3, 4) # was: 0       now: 5.0
\`\`\`

Each method had no entry in \`infer_method_name_type\` (so \`Math.tan(x)\`
was inferred as \`int\` instead of \`float\`) and no dispatch in the \`Math\`
constant handler in \`compile_call_expr\`. Both gaps fixed; the dispatch
forwards directly to libm (\`tan\`, \`atan\`, \`asin\`, \`acos\`, \`log10\`, \`hypot\`).

## Test plan

- [x] \`make test\` passes (74 -> 75 with the new \`bm_math.rb\`; pre-existing
  \`gvar\` error unaffected).
- [x] New \`test/bm_math.rb\` exercises every Math method this PR touches
  (and the existing ones), with multiple inputs (0.0, 0.5, 1.0). All
  results compared against CRuby via the standard \`make test\` harness.

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -2500,13 +2500,28 @@ class Compiler
     if mname == "sin"
       return "float"
     end
+    if mname == "tan"
+      return "float"
+    end
+    if mname == "acos" || mname == "asin" || mname == "atan"
+      return "float"
+    end
     if mname == "log"
+      return "float"
+    end
+    if mname == "log2"
+      return "float"
+    end
+    if mname == "log10"
       return "float"
     end
     if mname == "exp"
       return "float"
     end
     if mname == "atan2"
+      return "float"
+    end
+    if mname == "hypot"
       return "float"
     end
     if mname == "freeze"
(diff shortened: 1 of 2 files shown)


khasinski:
Thanks for the review. Pushed an update (`a4e4e15`) addressing both:

1. **Math.log2 added to inference** -- valid catch. `log2` was already
   dispatched in codegen but missing from `infer_method_name_type`,
   so the result type was inferred as `int`. Pre-existing inconsistency,
   fixed here since we're already touching the inference table. Test
   added that verifies the float-typed result.

2. **Math.hypot 1-arg form** -- declined: CRuby's `Math.hypot` raises
   `ArgumentError` on 1-arg calls, it doesn't return `|x|`. Verified:

   ```
   $ ruby -e 'p Math.hypot(3)'
   wrong number of arguments (given 1, expected 2) (ArgumentError)
   ```

   Keeping it 2-arg-only matches CRuby. Adding a 1-arg fabs form would
   diverge from the language spec.

PR #24, opened 2026-04-26
fix(codegen): add Float#min/max/sum/first/last for FloatArray

khasinski:
## Summary

\`[1.5, 2.5, 0.5].min\` returned \`0\` because the FloatArray dispatch in
\`compile_array_method_expr\` only handled \`length/[]/push/pop/empty?/size\`.
Calls to \`min/max/sum/first/last\` fell through to a default that
returned \`0\` -- a silent wrong-answer bug.

```ruby
f = [1.5, 2.5, 0.5, 3.5]
puts f.min   # was: 0       now: 0.5
puts f.max   # was: 0       now: 3.5
puts f.sum   # was: 0       now: 8 (sum is exact here)
puts f.first # worked       now: 1.5
puts f.last  # worked       now: 3.5
```

Adds three runtime helpers (\`sp_FloatArray_min/_max/_sum\`) and five
dispatch lines, mirroring the existing IntArray section.

## Test plan

- [x] \`make test\` passes (one new test file \`bm_float_array.rb\`).
- [x] Test uses values with non-zero fractional parts so Spinel's
      float-puts (which strips trailing \`.0\`) and CRuby's match.


khasinski:
Thanks for the review. Pushed an update (`40697d7`):

1. **Empty-array safety in `sp_FloatArray_min`/`_max`** — added the
   `if (a->len == 0) return 0;` guard. Previously the functions
   read `a->data[0]` unconditionally. Note: `sp_IntArray_min`/`_max`
   have the same gap; left untouched here to keep this PR scoped to
   FloatArray, but worth a follow-up.

2. **`last` simplified to `sp_FloatArray_get(rc, -1)`** — same pattern
   IntArray uses for `last`; the runtime's `_get` handles negative
   indices already. Both forms have the same out-of-bounds behavior
   on empty arrays, but the negative-index form is the idiomatic one.

Issue #2, opened 2026-04-24
Eventually Get Rid of the GC


sirinath:
Maybe you can use concepts like in https://salt-lang.dev/ and https://higherorderco.com/ to get rid of GC and also increase speed.

matz:
Thanks for the pointers.

Spinel's source language is (a realistic subset of) Ruby, and I'd
like to keep Ruby's semantics unchanged. Both `Salt`'s arena-plus-
lifetime-proofs model and `HVM`'s graph-reduction model would require
either sacrificing dynamic features that idiomatic Ruby depends on
or forcing programmers to annotate / restructure their code. Neither
is something I want to ask of a Ruby programmer, so I can't adopt
them here.

sirinath:
Can the dynamic features be modeled through effect tracking and inference, at least to some extent, without changing the syntax?

matz:
Thanks for the suggestion. Effect systems are an interesting area of research, but I don't see them as a fit for Ruby — at least not for Spinel.

Even an inference-based effect system needs the effects to actually be inferable. In Ruby, things like `eval`, `define_method`, `send`, `method_missing`, open classes, and runtime monkey-patching mean that the set of effects a given call can produce is not generally knowable until the program runs. To make inference tractable, you eventually have to either restrict those features or fall back to a "may do anything" effect — and at that point the analysis stops paying for itself.

Spinel's goal is to keep Ruby *as Ruby*: same semantics, same dynamic feel, no annotations, no hidden restrictions on what programmers can write. GC is the price I'm willing to pay for that. Improving the GC (and the rest of the runtime) is in scope; reshaping the language around an effect discipline isn't.

Appreciate the thought, though — closing this line of discussion here.

PR #25, opened 2026-04-26
fix(codegen): respect start/length in Array#fill(val, start, len)

khasinski:
## Summary

The 2-arg and 3-arg forms of `Array#fill` silently filled the whole
array. The dispatch only used `compile_arg0` for the value and ignored
the remaining args:

```ruby
arr = [1, 2, 3, 4, 5]
arr.fill(0, 1, 3)
# was: [0, 0, 0, 0, 0]
# now: [1, 0, 0, 0, 5]   (matches CRuby)
```

Adds proper 2-arg (`val, start`) and 3-arg (`val, start, length`)
handling. Negative start indexes from the end, matching CRuby.

## Note on a related issue

`arr = []; arr.fill(0, 0, 3) rescue puts \"x\"` still fails to compile
in Spinel (`incompatible pointer to integer conversion assigning to
mrb_int from sp_IntArray*`). That's a separate inference bug where
empty `[]` infers as IntArray and `rescue` tries to unify the
expression's type. Out of scope here; this PR only fixes the dispatch
correctness for the 2/3-arg forms.

## Test plan

- [x] `make test` passes (new `bm_array_fill.rb` covering 1-, 2-,
      and 3-arg forms plus negative-start).

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -14041,9 +14041,43 @@ class Compiler
     end
     if mname == "fill"
       pfx = array_c_prefix(recv_type)
+      args_id_fill = @nd_arguments[nid]
       val = compile_arg0(nid)
+      start_expr = "0"
+      # Default end: current array length (matches CRuby's no-args
+      # form which fills the entire existing array).
+      end_expr = "sp_" + pfx + "_length(" + rc + ")"
+      if args_id_fill >= 0
+        aargs_fill = get_args(args_id_fill)
+        if aargs_fill.length >= 3
+          # arr.fill(value, start, length): negative start counts from
+          # the end; if still negative after that, clamp to 0
+          # (matches CRuby: `[1,2,3].fill(9, -5, 2) #=> [9, 9, 3]`).
+          # end_expr = start + length lets the array grow past its
+          # current length when start+length > length; sp_*_set
+          # auto-grows and zero-fills gaps (matches CRuby:
+          # `[1,2,3].fill(9, 5, 2) #=> [1, 2, 3, 0, 0, 9, 9]`).
+          start_tmp = new_temp
+          len_tmp = new_temp
+          emit("  mrb_int " + start_tmp + " = " + compile_expr(aargs_fill[1]) + ";")
+          emit("  mrb_int " + len_tmp + " = " + compile_expr(aargs_fill[2]) + ";")
+          emit("  if (" + start_tmp + " < 0) " + start_tmp + " += sp_" + pfx + "_length(" + rc + ");")
+          emit("  if (" + start_tmp + " < 0) " + start_tmp + " = 0;")
+          start_expr = start_tmp
+          end_expr = "(" + start_tmp + " + " + len_tmp + ")"
+        elsif aargs_fill.length == 2
+          # arr.fill(value, start): fills from start to end of EXISTING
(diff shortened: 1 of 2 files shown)


khasinski:
Pushed an update (`6af4d43`) addressing each
point with one minor pushback on #2.

1. **Negative-start clamp** — added `if (start_tmp < 0) start_tmp = 0;`
   after the wrap-from-end adjustment, on both the 2-arg and 3-arg
   paths. Matches CRuby:
   ```ruby
   [1,2,3].fill(9, -5, 2)  # => [9, 9, 3]   (start = -5 + 3 = -2; clamps to 0)
   ```

2. **2-arg form should grow** — declined; CRuby's 2-arg form does not
   grow, only 3-arg does. Verified:
   ```ruby
   [1,2,3].fill(9, 5)     # => [1, 2, 3]               (no growth)
   [1,2,3].fill(9, 5, 2)  # => [1, 2, 3, nil, nil, 9, 9] (grows)
   ```
   So end_expr stays at `length` for 2-arg; the loop simply doesn't
   execute when start >= length.

3. **Remove early-termination check** — agreed. `sp_*_set` already
   auto-grows for indexes past length (`set_slow` for IntArray;
   `set` itself for Float/Str/PtrArray). Dropping
   `&& itmp < sp_*_length(rc)` lets 3-arg fill grow as expected.
   For 2-arg, end_expr equals length so the condition was redundant
   anyway.

Test extended with: 3-arg-grows-past-length, 2-arg-beyond-length-no-op,
and very-negative-start clamping. One small caveat for IntArray: the
gap-fill semantics differ (CRuby fills with `nil`, Spinel's IntArray
can't hold nil so it uses 0). The test asserts on length + the
explicit fill values, not on the gap slots, to keep `make test`'s
CRuby diff happy.

khasinski:
Re-checking against the current diff:

**1. Negative-start clamp (3-arg, line 14064)** — already in place on the line right after the `+=`:
```
emit("  if (" + start_tmp + " < 0) " + start_tmp + " += sp_" + pfx + "_length(" + rc + ");")
emit("  if (" + start_tmp + " < 0) " + start_tmp + " = 0;")
```
That's the exact pattern your suggestion proposes; the second emit is the safety clamp. Verified against `[1,2].fill(0, -5, 1)`: works on Spinel and matches CRuby (`[0, 2]`).

**2. Negative-start clamp (2-arg, line 14076)** — same pattern at lines 14074–14075. The growth half of the suggestion does not match CRuby semantics for the 2-arg form: `[1,2,3].fill(9, 5)` returns `[1, 2, 3]` in MRI (no growth). The 3-arg form *does* grow, which the existing `end_expr = (start_tmp + len_tmp)` already handles via the runtime's auto-growing `sp_*_set`.

**3. Loop bound preventing growth** — the loop emitted is `for (... ; itmp < end_expr; ...)`, where `end_expr` is `(start + len)` for the 3-arg case. The `< sp_*_length(rc)` bound your snippet refers to is the no-arg default; it isn't applied in 3-arg. So growth already works: `[1,2,3].fill(9, 5, 2)` produces `[1, 2, 3, 0, 0, 9, 9]` on both Spinel and CRuby.

No code changes here, but happy to add a regression test for the negative-start case if useful. Let me know if you'd like a closer look at any specific input.

PR #28, opened 2026-04-26
fix(parser): dedupe require_relative inlines via canonical-path tracking

khasinski:
## Summary

If two files both `require_relative 'lib/types'`, the parser inlined
the file content twice. For files defining structs, this produced
duplicate `struct sp_X_s` declarations in the generated C and a
`redefinition of 'sp_X_s'` compile error. Ruby's load-once semantics
had no equivalent here.

```ruby
# main.rb
require_relative 'lib/types'   # defines DT_Vertex
require_relative 'lib/data'    # also requires lib/types

# Before: error: redefinition of 'sp_DT_Vertex_s'
# After:  compiles cleanly; lib/types inlined exactly once.
```

## Implementation

`spinel_parse.c` adds a small (256-entry) static array of canonical
paths already inlined. Subsequent `require_relative`s of the same
file are replaced with a comment instead of the file content. Paths
go through `realpath()` so relative-path / symlink / cwd variants
resolve to the same key.

The 256-entry limit is bounded but well above any realistic project
size (Spinel itself uses far fewer). Could be made dynamic if you'd
prefer; the static form keeps the change small and allocation-free.

## Test plan

- [x] `make test` passes (existing 81 tests unaffected).
- [x] New `test/bm_require_dedup/` with a `main.rb` that
      `require_relative`s `lib/types` both directly and transitively
      through `lib/data`. Compiles successfully where it previously
      failed with the redefinition error.

The dedup test lives under `test/bm_require_dedup/` rather than as a
top-level `test/*.rb` because the `make test` runner globs only the
top level (same convention as the existing `test/bm_require/`).

### DIFF
--- a/spinel_parse.c
+++ b/spinel_parse.c
@@ -938,13 +938,26 @@ static char *resolve_plain_requires(char *source, const char *exe_path) {
     snprintf(lib_path, sizeof(lib_path), "%s/%s", lib_dir, lib_name);
     if (!strstr(lib_path, ".rb")) strcat(lib_path, ".rb");
 
-    char *content = read_file(lib_path);
-    if (!content) {
-      content = strdup("# require not resolved");
+    /* Same dedup as resolve_requires: a file pulled in via plain `require`
+       must not be re-inlined if a previous `require` or `require_relative`
+       already pulled it. Otherwise mixing the two forms for the same lib
+       still produces struct-redefinition errors. */
+    char *canonical = sp_canonical_path(lib_path);
+    char *content;
+    if (sp_path_already_included(canonical)) {
+      content = strdup("# require skipped (already included)");
+      free(canonical);
     } else {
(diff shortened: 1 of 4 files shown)


khasinski:
Pushed an update (`e145473`) addressing all three:

1. **Dedup extended to `resolve_plain_requires`** — the same canonical-path
   table is now consulted there too, so mixing `require` and
   `require_relative` for the same library no longer re-inlines.

2. **Dynamic table** — `sp_included_paths` is now a `char**` grown via
   `realloc` (start cap 16, doubles on demand). No silent drop on large
   projects; no fixed limit.

3. **Cleanup at end of run** — added `sp_includes_free()` called from
   `main` to free both the path strings and the array. Process is
   short-lived so this matters mainly for embedders / leak checkers, but
   it's the right hygiene.

PR #29, opened 2026-04-26
fix(codegen): make Array#replace work for IntArray, StrArray, FloatArray

khasinski:
## Summary

`Array#replace` was a silent no-op on arrays. The dispatch in
`compile_array_op_stmt`'s `replace` branch only handled
`mutable_str` and `string` receivers; array calls fell through
with no emitted code, so the receiver was unchanged with no error:

```ruby
b = [1, 2, 3, 4, 5]
b.replace([10, 20, 30])
puts b.length   # was: 5    now: 3
puts b[0]       # was: 1    now: 10
```

## Implementation

Three runtime helpers added to `lib/sp_runtime.h`:
`sp_IntArray_replace`, `sp_StrArray_replace`, `sp_FloatArray_replace`.
Each: reset `len` (and `start` for IntArray's deque-style buffer),
grow capacity if the source is larger, then `memcpy` from src.

Codegen dispatches to the right helper based on receiver type
(int_array, sym_array, str_array, float_array). int_array and
sym_array share the helper since they share the underlying container.

## A small style note

The four type arms (`int_array`, `sym_array`, `str_array`,
`float_array`) are written as four separate `if` blocks rather than a
combined `||` for the int+sym pair. Compounding them with `||` here
destabilises the bootstrap (gen1 emits valid C, but bin1 then mis-compiles
unrelated declarations). Same pattern as in #21 for the
`length`/`size` arms; sticking with the duplicate-arm form for now.

## Test plan

- [x] `make test` passes (one new `test/bm_array_replace.rb` covering
      Int/Str/Float arrays plus the existing mutable-String path,
      with both shrinking and growing replacements).


khasinski:
Pushed an update (`2b062a3`) checking the
`realloc` return in all three new replace helpers
(`sp_IntArray_replace`, `sp_FloatArray_replace`, `sp_StrArray_replace`).
On failure: `perror("realloc"); exit(1);` — same fail-fast behaviour
as a hypothetical successful replace into a now-broken array would
have on the very next `memcpy`, just with a useful error message
attached.

Note: the existing `sp_*_push_grow` and similar helpers in the runtime
also realloc unchecked. Left those alone here to keep the PR scoped;
worth a follow-up to bring them into line.

PR #30, opened 2026-04-26
fix(codegen): support block form of Array.new(N) { |i| ... }

khasinski:
## Summary

`Array.new(5) { |i| i * 2 }` returned an empty array because the
`Array.new` dispatch handled only no-args, 1-arg (`n`), and 2-arg
(`n, val`) forms. The block form fell through to the no-args path
that emitted `sp_IntArray_new()` with no loop:

```ruby
arr = Array.new(5) { |i| i * 2 }
puts arr.length    # was: 0    now: 5
puts arr.sum       # was: 0    now: 20
```

## Implementation

Adds an IntArray block path to the `cname == \"Array\"` handler
in `compile_call_expr`: allocates an IntArray, loops `count` times
binding the block param to the index, pushes the block's last
expression each iteration. Mirrors the existing `N.times.map`
codegen.

## Scope: IntArray only

This PR limits the block path to IntArray. Introspecting the block
body via `infer_type` to pick a typed container (StrArray /
FloatArray) destabilises the bootstrap (calling `infer_type` from
this dispatch site causes bin1 to mis-compile unrelated comparisons
elsewhere — same issue I worked around in \#21 for `Array#size`).

For now, Float/String collectors must be built manually with
`[]` + `N.times { ... << }`. Expanding to other typed arrays is a
clean follow-up once the underlying inference fragility is sorted.

## Test plan

- [x] `make test` passes (one new `test/bm_array_new_block.rb`
      covering block-with-index, block-with-constant, and zero-length).

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -13201,11 +13201,15 @@ class Compiler
             arrnew_iv = new_temp
             @needs_int_array = 1
             emit("  sp_IntArray *" + arrnew_tmp + " = sp_IntArray_new();")
+            # Root the new array before running the block body — pushing
+            # poly/string values inside the loop can trigger a GC cycle
+            # that would otherwise sweep the local pointer.
+            emit("  SP_GC_ROOT(" + arrnew_tmp + ");")
             emit("  for (mrb_int " + arrnew_iv + " = 0; " + arrnew_iv + " < " + arrnew_count + "; " + arrnew_iv + "++) {")
-            if arrnew_bp != ""
-              emit("    lv_" + arrnew_bp + " = " + arrnew_iv + ";")
-            end
             @indent = @indent + 1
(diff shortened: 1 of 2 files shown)


khasinski:
Pushed an update (`c2f50e9`) addressing both:

1. **`SP_GC_ROOT` after the IntArray allocation** — agreed and applied.
   The block body can call `compile_expr`/`compile_stmt`, which in turn
   can emit allocations (string concats, hash lookups, etc.); rooting
   the new array prevents the freshly-allocated pointer from being
   swept mid-loop.

2. **Indentation tidy-up** — moved `@indent = @indent + 1` to *before*
   the block-param assignment and dropped the extra spaces in the
   `lv_` emit string. The block-param line is now in the same indent
   regime as the rest of the loop body.

Issue #12, opened 2026-04-25
`require_relative` silently skipped when the word appears earlier in a comment or string


rramsden:
*Output from Claude Opus 4.7 but testing locally and breaks. Feel free to close if irrelevant!*

### Summary

The text-level preprocessor in `resolve_requires()` (`spinel_parse.c`, lines 776–842) uses `strstr` to find `require_relative`, then checks for line-start position. When the first match is **not** at a line start (e.g., the word appears inside a comment or string literal), the loop `break`s — abandoning *all* remaining `require_relative` statements in the file. No warning, no error; the referenced files are silently not inlined.

### Reproducer (2 files)

**`lib.rb`**
```ruby
puts "hello from lib"
```

**`main.rb`**
```ruby
# uses require_relative to load helpers
require_relative 'lib'
puts "done"
```

### Expected output
```
hello from lib
done
```
(matches `ruby main.rb`)

### Actual output
```
done
```
`lib.rb` is silently not inlined. Only `done` is printed.

### Remove the word from the comment and it works:
```ruby
# loads helpers
require_relative 'lib'
puts "done"
```
Output: `hello from lib` / `done` as expected.

### Root cause

```c
// spinel_parse.c:778–783
while ((pos = strstr(result, "require_relative")) != NULL) {
  /* Check it's at start of line */
  if (pos != result && *(pos - 1) != '\n') {
    /* Not at line start, skip */
    break;   // ← quits the whole loop, not just this iteration
  }
  ...
}
```

The comment says "skip" but `break` exits the `while` entirely. Subsequent real `require_relative` statements are never examined. `strstr` also matches the literal substring anywhere — inside string literals, identifiers, comments — so this can trip on legitimate Ruby code where the word appears for unrelated reasons.

### Impact

- **Silent failure**: no diagnostic emitted. Symptom is a confusing downstream error (e.g., an unresolved constant) hundreds of lines later in the generated C, or a runtime logic error if the missing file only affected behavior.
- **Footgun for documentation**: any file with a doc comment describing its own `require_relative` usage triggers the bug.

### Proposed fix (for discussion)

The immediate bug — `break` where a "skip past this occurrence" was intended — is fixable in ~8 lines:

```diff
--- a/spinel_parse.c
+++ b/spinel_parse.c
@@ -775,11 +775,14 @@ static char *resolve_requires(const char *source, const char *source_path) {

   char *result = strdup(source);
   char *pos;
[...]

PR #31, opened 2026-04-26
Fix infinite recursion on circular require_relative

hobostay:
## Summary
- Both `spinel_parse.rb` and `spinel_parse.c` would infinite-loop (crashing with a stack overflow) when processing files with circular `require_relative` chains (e.g., file A requires B, B requires A).
- Adds a visited-files set that tracks which files have already been resolved in the current recursion chain, skipping any re-encounter.

## How to reproduce
Create two files:
```ruby
# a.rb
require_relative "b"
puts "a"
```
```ruby
# b.rb
require_relative "a"
puts "b"
```
Then run `spinel a.rb` — it crashes with `SystemStackError` (Ruby parser) or a segfault (C parser).

## Test plan
- [x] Verified the fix compiles and runs
- [ ] Existing test suite should still pass (`make test`)
- [ ] Manually test with circular require files

### DIFF
--- a/spinel_parse.c
+++ b/spinel_parse.c
@@ -762,8 +762,17 @@ static char *read_file(const char *path) {
 }
 
 /* Simple require_relative resolver: replace lines matching
-   require_relative "path" with the file content */
-static char *resolve_requires(const char *source, const char *source_path) {
+   require_relative "path" with the file content.
+   visited/visited_count track absolute paths already resolved to prevent
+   infinite recursion on circular require_relative chains. */
+static int resolve_visited_contains(char **visited, int count, const char *path) {
+  for (int i = 0; i < count; i++)
+    if (strcmp(visited[i], path) == 0) return 1;
+  return 0;
+}
+
+static char *resolve_requires_impl(const char *source, const char *source_path,
+                                    char **visited, int visited_count) {
   /* Get base directory */
   char *path_copy = strdup(source_path);
   char *dir = strdup(path_copy);
@@ -772,6 +781,13 @@ static char *resolve_requires(const char *source, const char *source_path) {
   if (slash) *slash = '\0';
   else { free(dir); dir = strdup("."); }
   free(path_copy);
+
+  /* Record this file as visited */
+  {
+    char abs[1024];
+    if (realpath(source_path, abs))
+      visited[visited_count++] = strdup(abs);
+  }
 
   char *result = strdup(source);
   char *pos;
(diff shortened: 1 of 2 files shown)


matz:
Thanks for the patch — circular `require_relative` was a real footgun. There's some overlap with PR #28 (merged after this PR was opened), so a small rebase + scope trim would land it cleanly.

## What changed since this PR was opened

PR #28 added a global `sp_included_paths` set in `spinel_parse.c` that records every file inlined by `resolve_requires` / `resolve_plain_requires`, with `realpath()` canonicalisation. As a side effect of the load-once dedup it also breaks circular cycles, because once `a` is marked the second hop on `b → a` short-circuits to a comment.

I verified on current master:

```sh
# a.rb requires b, b.rb requires a
$ ./spinel a.rb -o ./a && ./a
a
b
a
```

No infinite recursion, output matches `ruby a.rb`. So the `.c`-side circular logic this PR adds (the per-recursion `visited[]` cursor + `resolve_requires_impl`) is now redundant with what's on master.

## What's still needed

Two pieces of this PR are still useful and don't overlap with #28:

**1. The `spinel_parse.rb` Ruby fallback is still infinite-looping.** The bootstrap parser doesn't have an equivalent dedup, so:

```sh
$ ruby spinel_parse.rb a.rb /tmp/a.ast
SystemStackError: stack level too deep
        ... 8922 levels...
        from spinel_parse.rb:32:in 'block in resolve_requires'
```

The visited-set approach in your `spinel_parse.rb` change fixes this. Worth keeping as-is.

**2. The two extra `break` → `continue` fixes in `spinel_parse.c`.** Issue #12 explicitly called these out as out-of-scope for that fix:

```c
if (q1 && q1 < line_end && (!q2 || q1 < q2)) { ... }
else if (q2 && q2 < line_end) { ... }
- else break;                                /* unparseable quote → abort whole file */
+ else { scan_from = pos + 1; continue; }    /* skip this hit, keep scanning */

char *end = strchr(start, quote_char);
- if (!end || end > line_end) break;         /* unclosed quote → abort */
+ if (!end || end > line_end) { scan_from = pos + 1; continue; }
```

These two `break`s have the same silent-failure shape as the one fixed in commit `c1c1058` — a malformed `require_relative` line should skip past, not abandon the rest of the file. Keeping these is a real improvement.

## Suggested slim version

If you'd rebase onto current master and trim to:

- `spinel_parse.rb`: full circular detection (your patch unchanged)
- `spinel_parse.c`: only the two `break` → `continue` lines above; drop `resolve_requires_impl`, the `visited[]` array, and the wrapper.

…then I'll merge it on the spot. Happy to take it on this side instead if you'd rather not bother — just let me know.

matz:
Took the slim residue on this side per the offer above (no rebase reply for a few days, and the picture shifted since the original review):

- The `.c` circular-require detection from this PR — `resolve_requires_impl` + `visited[]` — has been redundant since #28 landed the global `sp_included_paths` dedup. Verified circular `a ↔ b` runs to completion on master.
- The `.rb` sibling fix is no longer needed; `spinel_parse.rb` was dropped entirely in 80bd08d (C-only parse).
- That left only the two `break` → `continue` lines in `resolve_requires`. Landed in de2db08, with you co-attributed.

Thanks for the original report and the patch — the malformed-line abort would still be there without it.

PR #38, opened 2026-04-27
perf(codegen): cache repeated String#split on class metadata

mattn:
Profiling showed `;`-split on class metadata fields was 33.6% of codegen
time with 99.9% reuse. Added `split_{semi,pipe,comma}` caches; the 23
mutation sites (type-refinement) keep raw `.split`. Bootstrap fixpoint
holds; no test regressions.

| Stage                  | Before  | After   |
| ---------------------- | ------- | ------- |
| codegen self-compile   | 18.5s   | 13.2s   |
| CRuby codegen          | 22.48s  | 13.58s  |

---
Implemented with assistance from Claude.

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -185,6 +185,22 @@ class Compiler
     @parse_id_cache = {}
     @parse_id_pool = [[0]]
 
+    # Caches for split_semi/split_pipe/split_comma. Class metadata fields
+    # like @cls_meth_names are read via .split(sep) hundreds of thousands
+    # of times per codegen run; profiling showed >99% reuse. Cached results
+    # are READ-ONLY — call sites that mutate (type-refinement code paths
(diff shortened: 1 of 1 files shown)


mattn:
Bootstrap regressed on origin/master: gen1.c (CRuby output) and gen2.c (bin1 output) diverge — 640 method signatures lose their non-self parameters in bin1's emission, hence read_text_ast etc. fail to compile. Closing for now and rethinking the caching approach.

mattn:
Closing — this didn't survive self-hosting.

Fresh bootstrap on `origin/master` fails at step 3: `gen1.c` (CRuby) is fine, but `gen2.c` (`bin1`) drops the non-`self` parameters on ~640 method signatures, so `bin1` can't reproduce its own input. CRuby ran it fine, which is why local checks missed it. The cache shape probably needs a rethink — will revisit.

Issue #36, opened 2026-04-26
expanding Spinel. (added eval, threading, mutex, dynamic file handling, byte matcher and BOP tracker.)


CufeHaco:
Im working on a few experiments with spinel.  So far ive integrated the builtin ByteMatcher and BOP tracker i made for Jruby 10.1 builtins.  

This has allowed me to add eval and threading to spinel.

A few things that have had me caught up, class nesting in modules, among others. The biggest one was it trying to pass IntArray for a StrArray.

Im currently cleaning up my test to see if it will run my Rubian shell,  which successfully compiled its bootloader in spinel.  

Im going to circle back and build a  rubian cli from scratch just for Spinel, until I can work on it more to actualy run Rubian and Kestówv(RedRoot) kernel on it.  Once I'm happy with the progress, I'll push a PR.

If anyone has had a simular issue with the arrays, let me know. I think i may have found the problem.   

I'll update you as I can with my findings.

CufeHaco:
@matz The BOP tracker I contributed to JRuby 10.1 tracks builtin hotness via a bitmask  no dynamic dispatch, just observational instrumentation. What I wanted to explore for Spinel is a compile-time analog: using byte array pattern matching to map foreign bytecode (e.g. Ruby YARV, JVM) onto Spinel's compiled C output.

The idea is that Spinel already produces deterministic native binaries. If you can fingerprint the byte patterns those binaries emit for common operations, you can build a matcher that recognizes equivalent patterns in foreign bytecode and cross-compiles them directly into Spinel's output without a VM in the middle.

The BOP tracker provides the hotness signal: which operations are worth matching. The bytematcher does the structural comparison via byte arrays. Neither requires runtime dynamic dispatch, both live at the AOT layer.

I'd like to know if that direction is worth pursuing for Spinel, and if so, what constraints I should design around.

CufeHaco:
Now that im rested, I think i can explain it a bit better. 

In the end of the day, it doesnt matter what we use as languages, compilers vs interpreters, etc.  They always become Binary.  Binary itself is an abstraction layer of the physical electrical current running in the hardware.  Things like Seg Faults are literal dead shorts of current, the current took the wrong circuit path in the hardware, creating the mismatch of the stored emf.

[...]

PR #41, opened 2026-04-27
experimental additions. added Bytematcher and BOP tracker just like J…

CufeHaco:
Experimental prototypes for Spinel.

Added:
-ByteMatcher and BOP tracker with flags.
-added Eval
-added Threads with Mutex

ByteMatcher and BOP tracker was rewritten from the original code I did for Jruby 10.1 builtins. it has been repurposed to support Spinel.

this isn't for a merge yet, but to allow for others and Matz to review what works, what doesn't, and what Matz wants for this project.


CufeHaco:
@matz thank you for the time you put into spinel.  i know my code needs work, but i wanted to get my ideas out so i can get feedback on what areas to work on.  i'll jump on the issues gemini has for me here shortly.

matz:
Thanks for the writeup and for being clear that this isn't for merging. That's the right framing for an experimental PR.

The features here (eval, Thread/Mutex, dynamic send, the small VM) all assume a runtime VM with dynamic dispatch. That's not a TODO in the README's limitations list; it's the design. Spinel's value is "Ruby-the-static-subset compiled to standalone C, no VM, no embedded interpreter." Once `eval` and `Thread.new` are first-class, the static type inference that makes the rest of the compiler tractable starts unraveling. JRuby's runtime model fits the JVM well, but it's the wrong donor for an AOT C compiler.

There are also concrete issues that would block merging regardless of direction: `spinel_parse.c` (1061 lines, the Prism AST serializer) is deleted with no replacement, breaking bootstrap; the new `src/*.c` files aren't wired into the Makefile; build artifacts (`spinel`, `spinel_parse`, `src/libspinel_dynamic.a`) are committed; `Thread.new` returns the literal pointer `(sp_Thread*)0x1`; `sp_eval` is a tiny recursive-descent integer-arithmetic evaluator (literals plus `+ - * /` and parentheses, identifiers and everything else silently return 0), so it doesn't share semantics with Ruby's `eval` despite the name; and the branch is `CONFLICTING` against master.

I'm going to close this. Let's continue the design discussion on your Issue #36. Pick one feature, lay out why it fits Spinel's static-AOT model and how it interacts with type inference and GC, and a focused PR built off current master with green test/bench is welcome. Closing is not a verdict on the work, just on the fit.

CufeHaco:
Thanks for the feedback.  I was going though and spotted the same thing.  Im still getting acquainted with git commands.  The VM wasn't my design either, that was the liberty of the llm I used to help me plug in my code in the right spots.  

The bytematcher and BOP tracker isnt supposed to be a VM (but this gives me another project to try with it!)  The idea is simular between ruby bytecode and the jvm bytecode.  My goal was to give spinel the ability to match other bytecode to its compiled binaries from byte arrays.

The other stuff like eval, was just to see if I could get rubian to compile.  I would like to go down that road eventually.  I would definitely like to hear your thoughts on how we can make the bytematcher and BOP tracker fit your needs.

PR #46, opened 2026-04-27
feat(codegen): add Array#shuffle and String#% with str_array

mattn:
Adds Fisher-Yates `shuffle` / `shuffle!` for `IntArray` and `StrArray` in `sp_runtime.h` and routes `Array#shuffle` / `shuffle!` through them, inferring the result type from the receiver. Also routes `String#%` with a `str_array` RHS to a new `sp_str_format_strarr` helper so `"%s and %s" % ["a", "b"]` works; the integer `%` path is unchanged for non-string receivers.

Whenever `rand`, `srand`, `sample`, `shuffle`, or `shuffle!` is referenced, the generated `main` now seeds with `srand((unsigned)time(NULL))` so calls do not start from a fixed sequence.

Note: index selection uses `rand() % (i+1)`, which carries the usual modulo bias. This matches how `sample` already picks an index in this codebase, so no regression there. A separate change can switch both call sites to an unbiased helper if desired.

`test/bm_shuffle.rb` exercises the new paths but only validates length/membership/sort, not that the result is actually permuted; manual runs confirm distinct orderings across invocations.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -210,6 +210,8 @@ static inline mrb_int sp_FloatArray_length(sp_FloatArray*a){return a->len;}
 static inline mrb_bool sp_FloatArray_empty(sp_FloatArray*a){return a->len==0;}
 static inline mrb_float sp_FloatArray_get(sp_FloatArray*a,mrb_int i){if(i<0)i+=a->len;return a->data[i];}
 static inline void sp_FloatArray_set(sp_FloatArray*a,mrb_int i,mrb_float v){if(i<0)i+=a->len;while(i>=a->cap){a->cap=a->cap*2+1;a->data=(mrb_float*)realloc(a->data,sizeof(mrb_float)*a->cap);}while(i>=a->len){a->data[a->len]=0.0;a->len++;}a->data[i]=v;}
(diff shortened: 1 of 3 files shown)


matz:
Landed on master via rebase + ff-merge so your two commits keep their authorship:

- 75c6315 (your initial feat commit, rebased onto current master to resolve the lib/sp_runtime.h conflict around the UTF-8 sp_str_reverse / FloatArray reverse_bang/sort_bang additions)
- f12be04 (your review-tightening commit, also rebased)
- d6301ae (test additions on top: float_array shuffle, empty/single-element edges, and a new bm_str_format_arr.rb covering the supported %s / %% / extras-ignored / no-placeholder cases)

Bootstrap fixed-point clean, 101 tests + 55 benches green.

Note for follow-ups if you want to take them: `sp_str_format_strarr` currently fills missing %s slots with the empty string instead of raising ArgumentError like CRuby; the test deliberately doesn't exercise that since matching CRuby would be a separate behaviour change. The auto-`srand((unsigned)time(NULL))` seeding is also a behaviour change (previously the C default fixed seed) and might be worth a CHANGELOG-style note.

Closing the PR since the substance is on master.

PR #57, opened 2026-04-27
fix(runtime): move cls_id out of sp_RbVal union to avoid pointer corruption

mame:
### Reproducer

```ruby
class C
  attr_accessor :n
  def initialize(n); @n = n; end
  def name; @n.to_s; end
end
class D
  def name; "other"; end
end
def show(o); o.name; end
puts show(C.new(42))
puts show(D.new)
```

### Expected

```
42
other
```

### Actual (master)

```
$ ./r3_bin
Segmentation fault (core dumped)
$ echo $?
139
```

### Why it crashes — `sp_RbVal` layout (before)

```c
typedef struct {
  int tag;
  union {
    mrb_int i;
    const char *s;
    mrb_float f;
    mrb_bool b;
    void *p;     // 8 bytes
    int cls_id;  // 4 bytes  ← shares storage with v.p
  } v;
} sp_RbVal;

static sp_RbVal sp_box_obj(void *p, int cls_id) {
  sp_RbVal r;
  r.tag = SP_TAG_OBJ;
  r.v.p = p;             // writes 8 bytes
  r.v.cls_id = cls_id;   // overwrites the first 4 of those 8 bytes
  return r;
}
```

The second store clobbers the **low 32 bits** of the pointer. Reading `cls_id` itself returns the correct value (so the dispatcher picks the right branch), but the moment the dispatched method dereferences `(sp_C *)recv.v.p`, the program SIGSEGV's. Existing tests masked this because every poly-dispatched method ignored its receiver.

### `sp_RbVal` layout (after)

```c
typedef struct {
  int tag;
  int cls_id;                                  // separate field — no longer shares
  union { ...; void *p; } v;
} sp_RbVal;

static sp_RbVal sp_box_obj(void *p, int cls_id) {
  sp_RbVal r;
  r.tag = SP_TAG_OBJ;
  r.cls_id = cls_id;
  r.v.p = p;             // independent storage, doesn't overlap
  return r;
}
```

The struct stays 16 bytes (the 4-byte slot was previously padding after `tag`). Each `sp_box_*` helper writes `r.cls_id = 0` (or the actual value), and the dispatch path reads `recv.cls_id` instead of `recv.v.cls_id`.

Adds `test/bm_poly_self_deref.rb` as a small reproducer.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -527,14 +527,14 @@ typedef uint64_t sp_RbValue;
 #define SP_TAG_NIL  4
 #define SP_TAG_OBJ  5
 #define SP_TAG_SYM  6
-typedef struct { int tag; union { mrb_int i; const char *s; mrb_float f; mrb_bool b; void *p; int cls_id; } v; } sp_RbVal;
-static sp_RbVal sp_box_int(mrb_int v) { sp_RbVal r; r.tag = SP_TAG_INT; r.v.i = v; return r; }
-static sp_RbVal sp_box_str(const char *v) { sp_RbVal r; r.tag = SP_TAG_STR; r.v.s = v; return r; }
-static sp_RbVal sp_box_float(mrb_float v) { sp_RbVal r; r.tag = SP_TAG_FLT; r.v.f = v; return r; }
-static sp_RbVal sp_box_bool(mrb_bool v) { sp_RbVal r; r.tag = SP_TAG_BOOL; r.v.b = v; return r; }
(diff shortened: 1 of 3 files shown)


matz:
Landed on master via rebase + ff-merge so your commit keeps the original author:

- b42ae56 (your fix, rebased onto current master to resolve a conflict in `compile_poly_method_call` against #56's `recv_tmp` / `arg_strs` / boxing changes — `recv.cls_id` reads now sit on top of those #56 additions)
- c7f6779 (rename `bm_poly_self_deref.rb` → `poly_self_deref.rb` so it matches the bm_-prefix-drop convention adopted in 9091afc just before your PR landed)

Bootstrap fixed-point clean; 107 tests + 56 benches green. Closing the PR.

PR #69, opened 2026-04-27
Fix boxed object class id storage

### LINKED ISSUE #59: sp_box_obj overwrites the boxed object pointer
### Summary

`sp_RbVal` stores `void *p` and `int cls_id` in the same union. `sp_box_obj` writes both fields, so assigning the class id overwrites the pointer.

### Repro

Create `/tmp/spinel_box_obj_repro.c`:

```c
#include "sp_runtime.h"

int main(void) {
  int obj = 123;
  sp_RbVal value = sp_box_obj(&obj, 42);
  return value.v.p == &obj ? 0 : 1;
}
```

From the Spinel checkout:

```sh
cc -O0 -Ilib /tmp/spinel_box_obj_repro.c lib/libspinel_rt.a -lm -o /tmp/spinel_box_obj_repro
/tmp/spinel_box_obj_repro
echo $?
```

### Expected

Exit status `0`; the boxed value should preserve both the object pointer and the class id.

### Actual

Exit status `1`; `value.v.p` no longer equals the original pointer.

### Notes

The runtime currently has:

```c
typedef struct { int tag; union { ... void *p; int cls_id; } v; } sp_RbVal;
static sp_RbVal sp_box_obj(void *p, int cls_id) {
  sp_RbVal r;
  r.tag = SP_TAG_OBJ;
  r.v.p = p;
  r.v.cls_id = cls_id;
  return r;
}
```

A simple fix is to store `cls_id` outside the union.

---
Reproduced against matz/spinel `7bfc13c` on 2026-04-27.

companygardener:
Fixes #59.

`sp_box_obj` stored the object pointer and class id in the same union slot, so writing `cls_id` overwrote `v.p`. This moves `cls_id` next to the tag and updates generated poly dispatch to read it from the separate field.

Checked for contribution docs/templates before filing; Spinel has no top-level contribution guide or PR template.

Verification:
- compiled and ran a direct `sp_box_obj` runtime repro; it now exits 0
- compiled and ran `test/bm_poly2.rb` through `spinel_parse` + `ruby spinel_codegen.rb` + `cc`

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -527,13 +527,13 @@ typedef uint64_t sp_RbValue;
 #define SP_TAG_NIL  4
 #define SP_TAG_OBJ  5
 #define SP_TAG_SYM  6
-typedef struct { int tag; union { mrb_int i; const char *s; mrb_float f; mrb_bool b; void *p; int cls_id; } v; } sp_RbVal;
+typedef struct { int tag; int cls_id; union { mrb_int i; const char *s; mrb_float f; mrb_bool b; void *p; } v; } sp_RbVal;
 static sp_RbVal sp_box_int(mrb_int v) { sp_RbVal r; r.tag = SP_TAG_INT; r.v.i = v; return r; }
 static sp_RbVal sp_box_str(const char *v) { sp_RbVal r; r.tag = SP_TAG_STR; r.v.s = v; return r; }
 static sp_RbVal sp_box_float(mrb_float v) { sp_RbVal r; r.tag = SP_TAG_FLT; r.v.f = v; return r; }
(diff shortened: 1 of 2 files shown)


matz:
Thanks for the clean diagnosis and the standalone runtime reproducer in #59. The exact same fix landed about an hour before this PR via #57 (b42ae56) by @mame, which is why GitHub now flags the branch as CONFLICTING.

Both patches lift `cls_id` out of the union so `sp_box_obj`'s second store no longer clobbers `v.p`. The version on master goes one step further and also writes `r.cls_id = 0` in `sp_box_int` / `sp_box_str` / `sp_box_float` / `sp_box_bool` / `sp_box_nil` / `sp_box_sym`, so the field is never an uninitialized read for non-obj values. Functionally the two PRs are equivalent for the bug you're reproducing.

Closing as superseded; #59 is closed in the same step. Sorry the timing didn't work out, and please keep the bug reports / fixes coming.

PR #71, opened 2026-04-27
fix(codegen): resolve nested ConstantPath reads with lexical/root scope and unify scoped constant collection

sacckey:
## Summary

Follow-up to #53 (already merged).

This PR improves constant resolution for nested constant paths by handling lexical scope vs root scope (`::`) more consistently, and it unifies scoped constant collection logic.

## Changes

- Added lexical-aware constant resolution for constant reads and constant paths.
  - Relative paths prefer lexical scope.
  - `::` paths resolve from root scope.
- Updated constructor/class-name resolution to use the new lexical-scope-aware resolve_const_ref_name.
- Unified scoped constant collection:
  - `collect_module_with_prefix` now uses `collect_scoped_constant` for `ConstantWriteNode`.
  - `collect_constant` now delegates to `collect_scoped_constant("", nid)`.
- Kept `Struct.new` constant handling in the unified collection path.

## Tests

Updated existing test:

- `test/bm_constant_path_nested_read.rb`

Added cases for:

- `A::B::C` and `::A::B::C`
- `M::C::X` and `::M::C::X`
- lexical-vs-root conflict case to verify:
  - relative path resolves lexically
  - `::` path resolves from root

## Notes

This PR focuses on lookup/collection behavior only.
Internal-name encoding ambiguity with `_` in constant names is a separate concern and can be handled in a follow-up PR.

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -743,7 +743,109 @@ class Compiler
     end
     rt = @nd_type[recv_nid]
     if rt == "ConstantReadNode" || rt == "ConstantPathNode"
-      return const_ref_flat_name(recv_nid)
+      return resolve_const_ref_name(recv_nid)
+    end
+    ""
+  end
+
+  def module_name_exists(name)
+    i = 0
+    while i < @module_names.length
+      if @module_names[i] == name
+        return 1
+      end
+      i = i + 1
+    end
+    0
+  end
+
+  def builtin_const_root_name(name)
+    if name == "Float" || name == "Integer" || name == "Math"
+      return 1
+    end
+    if name == "Array" || name == "Hash" || name == "Proc" || name == "StringIO" || name == "Fiber"
+      return 1
+    end
+    0
+  end
+
+  def const_namespace_exists(name)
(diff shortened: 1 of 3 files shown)


matz:
Landed on master via rebase + ff-merge so your three commits keep their authorship:

- d3171e3 (your `resolve_const_read_name` / `resolve_const_ref_name` / `collect_scoped_constant` unification)
- 2cda5c7 (lexical scope tracking for const initialization)
- 54308ae (lexical scope when inferring module ivar initializer types)

Resolved a small conflict against #70 (which I'd merged shortly before yours and which added a narrow `@current_class_idx`-only lookup in `compile_expr` / `infer_type` for `ConstantReadNode`). Your `resolve_const_read_name` walks the full lexical chain and is a strict superset, so the manual resolution dropped #70's specific block in favor of yours. The class-scoped constant test from #70 (`test/class_constant_sym_array.rb`) still passes under your resolver.

Bootstrap fixed-point clean; 108 tests + 56 benches green. Closing the PR.

PR #76, opened 2026-04-28
feat(codegen): support built-in pointer types in poly slots

mame:
A `poly`-typed parameter doesn't handle built-in pointer types (Array, etc.).

## Reproducer

```ruby
class Foo; def [](x); 42; end; end
def read0(a); a[0]; end
puts read0(Foo.new)
puts read0([100, 200, 300])
```

## Expected

```
42
100
```

## Actual

Doesn't compile:

```
/tmp/_t.c: In function ‘main’:
/tmp/_t.c:52:53: error: passing argument 1 of ‘sp_box_int’ makes integer from pointer without a cast [-Wint-conversion]
   52 |     printf("%lld\n", (long long)sp_read0(sp_box_int(_t3)));
      |                                                     ^~~
      |                                                     sp_IntArray *
note: expected ‘mrb_int’ {aka ‘long int’} but argument is of type ‘sp_IntArray *’
```

## Analysis and fix

The poly path has two cooperating gaps:

**Boxing.** `box_expr_to_poly` only knew how to box scalars and `obj_<UserClass>`; anything else fell through to `sp_box_int(val)`. For an `IntArray` (or any built-in pointer) `val` is a pointer, so `sp_box_int(IntArray*)` is a C type error.

Reserving one `SP_TAG_*` per built-in doesn't scale (`Array<Float>`, `Array<Array<int>>`, …). Instead, reserve negative `cls_id` values under `SP_TAG_OBJ` and add thin `sp_box_int_array(p)` etc. wrappers that delegate to `sp_box_obj(p, SP_BUILTIN_*)`. `box_expr_to_poly` picks the matching wrapper. Adding a new built-in is one constant + one wrapper.

**Dispatch.** With boxing fixed, the call compiles and the `IntArray` reaches `compile_poly_method_call` with `cls_id == SP_BUILTIN_INT_ARRAY` (negative). But that function only emitted `if (cls_id == N)` branches for user classes (`cls_id >= 0`) — no branch matches and the result temp keeps its default value, so the call silently returns `0` instead of `100`.

Split the dispatcher under `if (tag == SP_TAG_OBJ)` into two halves: user-class branches (unchanged) and built-in branches via a new `emit_poly_builtin_dispatch` helper. The return-type lookup is factored into `poly_method_return_type` so the built-in side can supply fallback types (`[]`, `length`, `size` → int) when no user class defines the method.

Adds `test/poly_box_int_array.rb` and `test/poly_dispatch_builtin.rb` as reproducers.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -528,6 +528,17 @@ typedef uint64_t sp_RbValue;
 #define SP_TAG_NIL  4
 #define SP_TAG_OBJ  5
 #define SP_TAG_SYM  6
+/* Negative cls_id values let SP_TAG_OBJ also carry built-in pointer
+   types (IntArray, FloatArray, ...) — avoids minting a new SP_TAG_*
+   per type. Non-negative cls_id stays an index into the user-class
+   table as before. The element-type tag and the array cls_id are
+   paired by `array_cls_id = -element_tag - 1`. */
+#define SP_BUILTIN_ARRAY_OF(tag) (-(tag) - 1)
+#define SP_BUILTIN_INT_ARRAY    SP_BUILTIN_ARRAY_OF(SP_TAG_INT)   /* -1 */
+#define SP_BUILTIN_STR_ARRAY    SP_BUILTIN_ARRAY_OF(SP_TAG_STR)   /* -2 */
+#define SP_BUILTIN_FLT_ARRAY    SP_BUILTIN_ARRAY_OF(SP_TAG_FLT)   /* -3 */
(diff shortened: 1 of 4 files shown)


matz:
Merged via manual rebase to master at c4697f3. Bootstrap fixed-point clean; 120 tests + 56 benchmarks pass.

Filing a follow-up issue to track the partial-coverage gaps Gemini flagged: `emit_poly_builtin_dispatch` only emits IntArray branches today, and `sp_poly_puts`/`sp_poly_to_s` don't yet handle negative cls_id values. The architecture (negative `cls_id` under SP_TAG_OBJ + thin box wrappers) is in place — extending to FloatArray/StrArray/SymArray/PtrArray is straightforward additive work.

PR #80, opened 2026-04-28
fix(codegen): declare synthesized index for paramless Integer#step blocks

mame:
`Integer#step` with a do-block that omits its parameter referenced an undeclared `_i` index.

## Reproducer

```ruby
sum = 0
1.step(10, 1) do
  sum += 1
end
puts sum
```

## Expected

```
10
```

## Actual

```
/tmp/_step.c: In function ‘main’:
/tmp/_step.c:14:10: error: ‘lv__i’ undeclared (first use in this function)
   14 |     for (lv__i = 1; lv__i <= 10; lv__i += 1) {
      |          ^~~~~
```

## Analysis and fix

`compile_block_iteration_stmt` synthesizes `_i` as the index name when the block omits its parameter, but never declares it. Two paramless `step` blocks in the same function would hit `redefinition of 'lv__i'` for the same reason if the declaration *were* added without a scope.

Wrap the loop in a `{ ... }` block and emit `mrb_int lv__i = 0;` inside it when the block param is synthesized — the block scope makes the declaration local and side-steps the redefinition case at the same time.

Adds `test/range_step_no_block_param.rb`.

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -18390,14 +18390,27 @@ class Compiler
           end
         end
         bp1 = get_block_param(nid, 0)
+        synth = 0
         if bp1 == ""
           bp1 = "_i"
+          synth = 1
+        end
+        # When the block omits its parameter, the synthesized `_i` isn't
+        # declared anywhere — wrap the loop in a block scope and declare
+        # it locally. This also avoids redefinition errors when multiple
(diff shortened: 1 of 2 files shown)


matz:
Merged via fast-forward to master. Bootstrap fixed-point clean; 127 tests + 56 benchmarks pass.

PR #82, opened 2026-04-28
fix(codegen): poly_array follow-ups — shuffle, +, concat, each_with_object, flat_map

mame:
Several Array methods didn't pick up `poly_array` (heterogeneous arrays) when that type was introduced. The dispatch sites kept their original `int_array | str_array | float_array | sym_array` lists and either fell through silently or compiled to the wrong runtime call.

## Reproducers

```ruby
# 1. shuffle (sym_array also affected; bin output is empty on master)
puts [1, "a", :s, 2.0].shuffle.length

# 2. +
a = [1, "x"]; b = [2, "y"]
puts (a + b).length

# 3. concat
a = [1, "x"]; a.concat([2, "y"])
puts a.length

# 4. each_with_object
n = 0
[1, "a", :sym].each_with_object("") {|_e, _a| n += 1 }
puts n

# 5. flat_map (block returns a heterogeneous array)
puts [1, 2].flat_map { |x| [x, x.to_s] }.length
```

## Expected

```
4
4
4
3
4
```

## Actual

```
# 1. shuffle
(empty — ran but produced no output; lv_arr ended up `0` because the
dispatch fell through and the result temp was never assigned)

# 2. +
/tmp/_t.c:24:18: error: invalid operands to binary +
(have 'sp_PolyArray *' and 'sp_PolyArray *')

# 3. concat
2     # the loop was never emitted; receiver kept its original length

# 4. each_with_object
0     # body never ran

# 5. flat_map
/tmp/_t.c:28:26: error: initialization of 'sp_IntArray *' from
incompatible pointer type 'sp_PolyArray *'
```

## Analysis and fix

Each of these sites checked the receiver (or block return) type against an explicit list. The list pre-dated the poly_array type, so the new type just wasn't in it — the check returned false, and the dispatcher either fell through or used the IntArray fallback in `array_c_prefix`. They're all the same kind of miss; commits split per method:

1. `array_c_prefix`: map `poly_array → "PolyArray"` so callers don't need a per-call ternary.
2. `Array#shuffle / shuffle!`: extend the type list to include `sym_array` (uses the IntArray runtime internally) and `poly_array` (new `sp_PolyArray_dup` / `_shuffle{_bang}` runtime helpers).
3. `Array#+`: extend both `infer_call_type` (so the result type is poly_array) and `compile_operator_call_expr`.
4. `Array#concat`: extend the receiver-type check.
5. `Array#each_with_object`: extend the receiver-type check; let `elem_type_of_array` map poly_array → "poly" so the block param is typed `sp_RbVal`.
6. `Array#flat_map`: extend the block-return recognition (in `infer_call_type`) and the accumulator typing (in `compile_flat_map_expr`).
7. After the above, eight sites independently spelled out the same `int_array | str_array | float_array | sym_array | poly_array` predicate. Factor it into `is_array_type` and use it at those sites. *_ptr_array stays out of the helper — `sp_PtrArray` lacks several helpers (`_dup`, `_shuffle`, …) and the dispatchers don't route ptr_array correctly even on master, so that's a separate concern.

Each fix has its own reproducer test in `test/`.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -594,6 +594,9 @@ static sp_PolyArray *sp_PolyArray_new(void) { sp_PolyArray *a = (sp_PolyArray *)
 static void sp_PolyArray_push(sp_PolyArray *a, sp_RbVal v) { if (a->len >= a->cap) { a->cap = a->cap * 2 + 1; a->data = (sp_RbVal *)realloc(a->data, sizeof(sp_RbVal) * a->cap); } a->data[a->len++] = v; }
 static mrb_int sp_PolyArray_length(sp_PolyArray *a) { return a->len; }
 static sp_RbVal sp_PolyArray_get(sp_PolyArray *a, mrb_int i) { if (i < 0) i += a->len; return a->data[i]; }
+static sp_PolyArray *sp_PolyArray_dup(sp_PolyArray *a) { sp_PolyArray *b = sp_PolyArray_new(); for (mrb_int i = 0; i < a->len; i++) sp_PolyArray_push(b, a->data[i]); return b; }
+static void sp_PolyArray_shuffle_bang(sp_PolyArray *a) { for (mrb_int i = a->len - 1; i > 0; i--) { mrb_int j = (mrb_int)(rand() % (i + 1)); sp_RbVal t = a->data[i]; a->data[i] = a->data[j]; a->data[j] = t; } }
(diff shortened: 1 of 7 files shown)


matz:
Merged via fast-forward to master. Bootstrap fixed-point clean; 136 tests + 56 benchmarks pass.

Re Gemini's two notes:
- `sp_PolyArray_dup` memcpy optimization is a reasonable follow-up but not blocking; can be addressed separately if profiling shows it.
- `lib/sp_runtime.h` is hand-maintained, not generated, so the manual edits there are correct. (Gemini false positive.)

Issue #67, opened 2026-04-27
case obj when SomeClass emits undeclared C identifiers


companygardener:
### Summary

Class-pattern `case` conditions are lowered as comparisons against undeclared C identifiers instead of class/type checks.

### Repro

```ruby
class Var; end
class List; end

node = Var.new
case node
when Var
  puts "var"
when List
  puts "list"
end
```

Commands:

```sh
./spinel_parse /tmp/case_class_pattern.rb /tmp/case_class_pattern.ast
ruby spinel_codegen.rb /tmp/case_class_pattern.ast /tmp/case_class_pattern.c
cc -O0 -Werror -ffunction-sections -fdata-sections -Ilib /tmp/case_class_pattern.c lib/libspinel_rt.a -lm -o /tmp/case_class_pattern
```

### Expected

The program compiles and prints:

```text
var
```

### Actual

Compilation fails:

```text
use of undeclared identifier 'Var'
use of undeclared identifier 'List'
```

The generated C also initializes an `mrb_int` temporary from an object pointer.

---
Reproduced against matz/spinel `7bfc13c` on 2026-04-27.

PR #86, opened 2026-04-28
feat(case/when): support class matching and full poly when coverage

v420v:
## 概要
- `case/when` の不具合（issue #67）を修正
- poly 変数に対する全値型・配列型での `when` マッチングに対応

### 対応 `when` 値型
| 型 | 
|----|
| クラス述語 (`when SomeClass`) | 
| クラスの継承  |
| int / float / string | 
| bool / nil / symbol | 
| int range (`1..3`) | 
| array (`["a","b"]`) | 

## テスト
`test/case_when_class.rb` テスト追加


matz:
Hi @v420v, thanks for the substantial work here. I rebased onto master locally and ran the bootstrap + benchmarks. The mechanical rebase against PR #76 (built-in array tagging) was straightforward (rename `SP_CLS_*_BOX` → `SP_BUILTIN_*_ARRAY`, add `SP_BUILTIN_POLY_ARRAY`), but two issues block merge that I think need the original author's eyes:

### 1. Nested `if` at the variable-assign site breaks nullable narrowing

In `compile_lvar_write` (currently around line 17569 in this branch):

```ruby
if rhs_t != "nil" || is_nullable_type(vt) == 0
  if is_nullable_type(vt) == 0           # <- new
    set_var_type(lname, rhs_t)
  end
end
```

The inner gate prevents every `T?` variable from ever being narrowed to `T`, including the existing `var = nil; var = LNode.new(...)` pattern that `scan_locals` already supports. After this change, `bm_linked_list` and `bm_rbtree` regress: `lv_head` stays at `obj_LNode?` and the codegen emits `lv_head = sp_box_obj(sp_list_push(...), 0);` (boxing a `sp_LNode *` argument-mismatch). Same shape applies to any `Foo? + Foo` flow.

The original guard `rhs_t != "nil" || is_nullable_type(vt) == 0` is what was preserving the narrowing path; the inner gate negates it. I think the new gate isn't load-bearing for the case/when work, but I want to flag it so it doesn't get lost.

### 2. `scan_poly_assigns` fall-through over-marks `T? + T` as poly

```ruby
elsif old == at
  ...
elsif old == "nil" && is_nullable_pointer_type(at) == 1
  ...
else
  types[idx] = "poly"     # <- catches T? + T, T + T?, T? + T?
```

For `obj_LNode? + obj_LNode` (extremely common: nullable head pointer reassigned to a non-null node), this hits the `else` branch and forces the variable to poly, which then fights with both the existing nullable inference and item (1) above.

The two changes are synchronized: (1) blocks narrowing at the per-assign site, (2) demotes the variable to poly anyway. I tried adding elif branches for `base_type(old) == at`, `base_type(at) == old`, and `base_type(old) == base_type(at)`, which recovers `bm_linked_list` and `bm_rbtree`, but breaks `test/case_when_class.rb` because the test relies on the aggressive poly-detection. The two designs collide: the PR wants "any reassignment with mismatched type → poly", the existing inference wants "prefer T? when one side is nil/T-compatible". They can't both be right at the same site.

### 3. Gemini concern about descendant-class match

The static-type case branch at line ~18327 (`is_obj_type(pred_type) == 1 && ConstantReadNode`) does:

```ruby
if is_class_or_ancestor(pred_cname, cname) == 1
  result = result + "1"
end
```

This is correct only when `pred_type` is statically the most-derived class. If `pred_type` is `obj_Animal` and a runtime instance is `Dog < Animal`, `case animal when Dog` should match but emits `0` because `is_class_or_ancestor(Animal, Dog)` is false. A runtime `cls_id` check is needed for the descendant case (the poly branch already does this via `collect_matching_cls_cond`, which could be reused).

### Suggested split

This PR is doing three things that I'd ask to be separated so each can land cleanly:

1. **issue #67 monomorphic fix only**: just emit the type-correct predicate temporary (`sp_Foo *tmp` for `obj_Foo` predicate) and the static-class condition, with the descendant-class fix in (3). Small, contained, easy to review.
2. **`case`/`when` over poly variables**: `poly_literal_cond`, range/scalar-on-poly handling, the `pred_type == "poly"` arm.
3. **The new `scan_poly_assigns` whole-method pass + nested-if narrowing change**: needs to be reconciled with the existing nullable inference. Likely the cleanest answer is to leave T?/T narrowing alone and only mark "poly" when the two types have no common base (e.g. `obj_Var` then `int`), but that wants discussion before code.

I'll leave issue #67 open and this PR open. Happy to review (1) on its own as a quick follow-up if you'd like to split it out.

matz:
（上記コメントの日本語要約）

@v420v さん、まずは大規模な実装ありがとうございます。手元で master に rebase してビルド・ベンチを通したところ、PR #76 との衝突は機械的に解消できる範囲でしたが、設計上の問題が 2 点ありマージは保留しています。

### 1. 変数代入箇所の nested `if` が nullable narrowing を壊す

`compile_lvar_write` (このブランチの 17569 行付近):

```ruby
if rhs_t != "nil" || is_nullable_type(vt) == 0
  if is_nullable_type(vt) == 0           # <- 新規追加
    set_var_type(lname, rhs_t)
  end
end
```

内側の gate により `T?` 型の変数が `T` に narrow されなくなり、既存の `var = nil; var = LNode.new(...)` パターンが壊れます。`bm_linked_list` と `bm_rbtree` がリグレッションし、`lv_head = sp_box_obj(sp_list_push(...), 0);` のような型不整合な C コードが出力されます。元の `rhs_t != "nil" || is_nullable_type(vt) == 0` ガードが narrowing 経路を保っていたものを、内側の追加 gate が打ち消している形です。これは case/when の本筋ではないように見えるため、確認をお願いしたいところです。

### 2. `scan_poly_assigns` の else が `T? + T` を poly に降格させる

```ruby
elsif old == at
  ...
elsif old == "nil" && is_nullable_pointer_type(at) == 1
  ...
else
  types[idx] = "poly"     # <- T? + T, T + T?, T? + T? を全部拾ってしまう
```

`obj_LNode? + obj_LNode` のような nullable 先頭ポインタの再代入で `else` に落ちて poly になり、(1) と合わさって既存の nullable 推論と衝突します。`base_type(old) == at` などの elif を足すと `bm_linked_list`/`bm_rbtree` は復活しますが、今度は `test/case_when_class.rb` が落ちます。PR の方針「型違いの再代入はすべて poly」と既存の方針「片方が nil/T 互換なら T? を優先」が同じ箇所で衝突しているため、片方を直すと他方が壊れる構造です。

### 3. Gemini が指摘した派生クラスマッチ

静的型 case 分岐 (~18327 行):

```ruby
if is_class_or_ancestor(pred_cname, cname) == 1
  result = result + "1"
end
```

`pred_type` が静的に最派生クラスである場合のみ正しく、たとえば `pred_type` が `obj_Animal` で実体が `Dog < Animal` だと `case animal when Dog` はマッチしてほしいのに `0` を出力します。派生方向のマッチは runtime の `cls_id` チェックが必要で、poly 側ですでに使っている `collect_matching_cls_cond` を再利用できそうです。

### 分割の提案

以下の 3 つに分けて頂けると個別にレビュー・マージしやすいです:

1. **issue #67 の monomorphic 修正のみ**: predicate 一時変数の型を正す (`sp_Foo *tmp`) + 静的クラス条件 + (3) の派生クラス対応。小さく独立。
2. **poly 変数に対する case/when**: `poly_literal_cond`、scalar/range の poly 対応、`pred_type == "poly"` 分岐。
3. **`scan_poly_assigns` 全メソッドパス + narrowing 変更**: 既存の nullable 推論との整合が必要。共通基底のない型 (`obj_Var` から `int` 等) のみ poly 化する方針が綺麗そうですが、コード前に方針議論をお願いしたいところです。

issue #67 と本 PR は open のままにしておきます。(1) を切り出して頂ければ follow-up としてすぐにレビューできます。

v420v:
@matz 
Thanks for the review. I'll extract the issue #67 fix into a new PR and close this one.

PR #110, opened 2026-04-29
Cleanup up dead code

rramsden:
### Context

I noticed while working on https://github.com/matz/spinel/pull/73 there is a lot of dead code inside `spinel_codegen.rb`. 

Running `make clean && make && make test` doesn't produce any errors.

```
make test
Tests: 160 pass, 0 fail, 0 error
```

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -7111,10 +7111,7 @@ class Compiler
       @needs_gc = 1
       @needs_str_array = 1
     end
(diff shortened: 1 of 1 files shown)


rramsden:
I believe these findings from Gemini are false positives as the test suite runs correctly.

Gemini's review did flag some additional @needs_* variables which I removed in https://github.com/matz/spinel/pull/110/commits/7e33d42bb386ee3af10a93ce1b30135a6ea994dc

PR #173, opened 2026-05-01
Fix keyword call-site type inference

### LINKED ISSUE #170: Keyword argument with nil default called with string emits incompatible C
## Summary

A keyword argument defaulting to `nil` can still be inferred as `mrb_int` after a call site passes a string. The generated C then passes a `char *` to an `mrb_int` parameter and fails with `-Werror`.

## Reproduction

Save this as `/tmp/kw_nil_default_string_call.rb`:

```ruby
def show(x: nil)
  puts x
end

show(x: "hi")
show(x: nil)
```

Then run:

```sh
make bootstrap
./spinel_parse /tmp/kw_nil_default_string_call.rb /tmp/kw_nil_default_string_call.ast
./spinel_codegen /tmp/kw_nil_default_string_call.ast /tmp/kw_nil_default_string_call.c
cc -O2 -Werror -ffunction-sections -fdata-sections -Ilib /tmp/kw_nil_default_string_call.c lib/libspinel_rt.a -lm -Wl,-dead_strip -o /tmp/kw_nil_default_string_call
```

## Actual

```text
error: incompatible pointer to integer conversion passing 'char *' to parameter of type 'mrb_int'
```

## Expected

The generated program should compile and print:

```text
hi
```

followed by the output for `nil`.

### LINKED ISSUE #192: Windows MSYS2 make test output comparisons fail for platform-specific output
## Summary

`make test` can fail on Windows/MSYS2 even when the generated program output is logically equivalent to Ruby output.

Observed in the `windows-mingw` CI job:

```text
FAIL: fileio
FAIL: fileopen
FAIL: system
```

After normalizing CRLF differences, `fileio` and `fileopen` pass. `system` can still fail because Ruby captured output may place child-process output from `system("echo ...")` before earlier parent-process `puts` output, while the generated C flushes stdout before calling `system()` and prints in source order.

## Reproduction

On Windows/MSYS2 MINGW64:

```sh
make deps
make -j$(nproc) all
make test
```

## Expected

The test harness should compare logical stdout consistently across platforms, and the `system` fixture should not depend on parent/child stdout buffering order.

## Notes

The fix is to normalize captured CRLF output in the harness and avoid the buffering-sensitive ordering in `test/system.rb`.

companygardener:
## Summary

Fix call-site type propagation for keyword/default arguments so calls like `show(x: "hi")` can refine a parameter that defaults to `nil`.

Changes:
- use the existing call-type unification path for keyword argument inference
- keep function/class body call inference inside the fixed-point loop
- track explicit rest parameters instead of inferring them from trailing int_array parameters
- tighten GC-rooting and nullable expected-type handling from review feedback
- normalize captured test output before comparison so CRLF output does not fail Windows/MSYS2 jobs
- print expected/actual output when a harness comparison fails
- stabilize the `system` fixture so it does not depend on parent/child stdout buffering order
- add a regression test for a keyword parameter with a `nil` default called with a string

Fixes #170.
Closes #192.

## Verification

```sh
make -B bootstrap
make test
# Tests: 232 pass, 0 fail, 0 error
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,10 +193,14 @@ test: spinel_parse$(EXE) $(SP_RT_LIB)
 	      expected=$$($(TIMEOUT10) ruby "$$f" 2>/dev/null); \
 	    fi; \
 	    actual=$$($(TIMEOUT10) /tmp/_sp_t_bin$(EXE) 2>/dev/null); \
+	    expected=$$(printf "%s" "$$expected" | tr -d '\r'); \
+	    actual=$$(printf "%s" "$$actual" | tr -d '\r'); \
 	    if [ "$$expected" = "$$actual" ]; then \
 	      pass=$$((pass+1)); \
 	    else \
-	      echo "FAIL: $$bn"; fail=$$((fail+1)); \
+	      echo "FAIL: $$bn"; \
+	      printf '%s\n%s\n%s\n%s\n' "--- expected ---" "$$expected" "--- actual ---" "$$actual"; \
+	      fail=$$((fail+1)); \
 	    fi; \
 	  else \
 	    echo "ERR:  $$bn"; err=$$((err+1)); \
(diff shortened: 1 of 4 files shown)


companygardener:
Addressed the review feedback in 9a7ac4a:

- constructor argument rooting now protects pointer args when their own evaluation or a later argument can trigger GC
- parent ivar type updates now guard against self-recursive parent links
- rest-parameter handling now tracks explicit rest slots instead of inferring from trailing int_array
- nullable expected array/string types normalize through base_type, and poly_array is treated as a pointer type for GC-safety checks

Verification:

```sh
make -B bootstrap
make test
# Tests: 232 pass, 0 fail, 0 error
```

companygardener:
The windows job is slow and failing for an unknown reason. Adding diagnostics to the harness in order to better see what's going on.

PR #174, opened 2026-05-01
Add poly runtime operators

### LINKED ISSUE #171: poly equality emits invalid C for mixed call sites
## Summary

When a method parameter is inferred as `poly`, equality currently emits a raw C `==` between `sp_RbVal` values. That produces invalid C for mixed call sites.

## Reproduction

Save this as `/tmp/poly_eq.rb`:

```ruby
def same(a, b)
  a == b
end

puts same(1, 1)
puts same("x", "x")
```

Then run:

```sh
make bootstrap
./spinel_parse /tmp/poly_eq.rb /tmp/poly_eq.ast
./spinel_codegen /tmp/poly_eq.ast /tmp/poly_eq.c
cc -O2 -Werror -ffunction-sections -fdata-sections -Ilib /tmp/poly_eq.c lib/libspinel_rt.a -lm -Wl,-dead_strip -o /tmp/poly_eq
```

## Actual

```text
error: invalid operands to binary expression ('sp_RbVal' and 'sp_RbVal')
```

## Expected

The generated program should compile and print:

```text
true
true
```

### LINKED ISSUE #172: poly || emits C boolean instead of selected value
## Summary

For non-boolean expressions, Ruby's `||` returns one of the original operand values. When the inferred type is not a C boolean, Spinel can still emit C `||`, producing an `int` instead of the selected Ruby value.

## Reproduction

Save this as `/tmp/poly_or.rb`:

```ruby
def pick(a, b)
  a || b
end

puts pick(nil, "fallback")
puts pick("actual", "fallback")
```

Then run:

```sh
make bootstrap
./spinel_parse /tmp/poly_or.rb /tmp/poly_or.ast
./spinel_codegen /tmp/poly_or.ast /tmp/poly_or.c
cc -O2 -Werror -ffunction-sections -fdata-sections -Ilib /tmp/poly_or.c lib/libspinel_rt.a -lm -Wl,-dead_strip -o /tmp/poly_or
```

## Actual

```text
error: incompatible integer to pointer conversion returning 'int' from a function with result type 'const char *'
```

## Expected

The generated program should compile and print:

```text
fallback
actual
```

companygardener:
## Summary

Add runtime/codegen support for `poly` equality and value-returning truthiness.

Changes:
- compare `sp_RbVal` values through runtime helpers instead of raw C operators
- handle cross-type numeric equality/comparison and non-comparable poly comparisons
- compile non-boolean `||` as a value-select expression with coherent branch typing/boxing
- preserve nullable pointer semantics when boxing into `poly`, so NULL becomes `nil`
- handle NULL strings in poly equality/comparison without normalizing NULL to an empty string
- normalize captured test output before comparison and stabilize the `system` fixture for Windows/MSYS2
- add `sp_RbVal` truthiness and supporting poly operator helpers
- add regression tests for mixed-call-site equality, poly comparisons, and value-returning `||`

Fixes #171.
Fixes #172.

## Verification

```sh
make -B bootstrap
make test
# Tests: 233 pass, 0 fail, 0 error
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -193,10 +193,14 @@ test: spinel_parse$(EXE) $(SP_RT_LIB)
 	      expected=$$($(TIMEOUT10) ruby "$$f" 2>/dev/null); \
 	    fi; \
 	    actual=$$($(TIMEOUT10) /tmp/_sp_t_bin$(EXE) 2>/dev/null); \
+	    expected=$$(printf "%s" "$$expected" | tr -d '\r'); \
+	    actual=$$(printf "%s" "$$actual" | tr -d '\r'); \
 	    if [ "$$expected" = "$$actual" ]; then \
 	      pass=$$((pass+1)); \
 	    else \
-	      echo "FAIL: $$bn"; fail=$$((fail+1)); \
+	      echo "FAIL: $$bn"; \
+	      printf '%s\n%s\n%s\n%s\n' "--- expected ---" "$$expected" "--- actual ---" "$$actual"; \
+	      fail=$$((fail+1)); \
 	    fi; \
 	  else \
 	    echo "ERR:  $$bn"; err=$$((err+1)); \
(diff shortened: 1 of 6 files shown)


companygardener:
Addressed the review feedback in 04dad06:

- poly equality now handles cross-type numeric equality
- poly comparisons now carry a comparable flag, return false for unsupported mixed comparisons, and avoid undefined pointer ordering by comparing uintptr_t values
- integer/non-boolean `||` now preserves value-returning semantics, including `0 || fallback`
- `||` branch results now share a coherent result type, boxing to poly when needed
- poly equality paths now set `@needs_rb_value`
- integer exponentiation with a negative exponent now returns a float value instead of truncating to int

Verification:

```sh
make -B bootstrap
make test
# Tests: 233 pass, 0 fail, 0 error
```

PR #215, opened 2026-05-02
feat(codegen): Method object

mame:
## What

`method(:foo)` on an instance method used to be a pure compile-time
alias — the variable held a placeholder int and only direct
in-scope calls worked. Storing the method in an ivar / passing it
across a method-call boundary lost the binding.

## Approach

Register `Method` as a regular user class with a synthetic
`initialize(self_obj, fn_ptr)`. The existing class machinery emits
its typedef, struct, GC scan, and constructor.

Two `method(:foo)` forms are handled:

- **No receiver, inside a class body** — bind to `self`.
- **Obj-typed receiver** (e.g. `@worker.method(:shout)`) — bind to
  the inferred receiver class.

`m.call(args)` / `m[args]` lower to a function-pointer cast and
an indirect call. `detect_method_taken_classes` excludes captured
receivers from the value-type optimization so the captured pointer
can't dangle.

## Commits

This PR ships as four atomic commits:

1. **`feat(codegen): method(:foo) via synthetic user class (Method)`** —
   the core feature plus a critical detail: `register_builtin_classes`
   runs at the *end* of `collect_all`, not the start. Inserting `Method`
   at the head would shift every user class's `cls_idx` up by one
   and break the `sp_box_obj(p, 0)` fallback in
   `emit_poly_builtin_dispatch`'s BUILTIN_PTR_ARRAY branch (which
   assumes the first user class is `cls_id 0`). Verified with the
   existing `test/poly_dispatch_ptr_array.rb`.

2. **`fix(codegen): emit null fn_ptr when method's name is unknown on receiver`** —
   Optcarrot has registration sites like `@pulse_0.method(:poke_2)`
   where the receiver class (or its inheritance chain) defines no
   `poke_2`. Pre-fix, codegen emitted `&sp_Pulse_poke_2` and the
   linker rejected the unresolved symbol. Walk the inheritance chain;
   if no ancestor defines the symbol, fall back to
   `sp_Method_new(..., (mrb_int)0)`. The Method instance is still safe
   to capture / pass / GC; an actual `m.call(...)` invocation crashes,
   matching Ruby's NoMethodError semantics roughly enough.

3. **`fix(codegen): use defining class for inherited method(:foo) symbols`** —
   when the receiver's class inherits `foo` from a parent, the generated
   C function lives under the *defining* class. Emitting
   `&sp_<RecvCls>_foo` left an unresolved symbol; emit
   `&sp_<DefiningCls>_foo` instead.

4. **`fix(codegen): use defining class for no-recv method(:foo) symbols`** —
   same inheritance-walk logic for the no-receiver path
   (`def foo; bar = method(:baz); end`). Without this,
   `method(:peek_6000)` inside `MMC1#reset` emitted
   `&sp_Optcarrot_MMC1_peek_6000` even though `peek_6000` lives on
   the parent `ROM` class as `sp_Optcarrot_ROM_peek_6000`.

## Tests

- `test/bound_method_basic.rb` — store `method(:foo)` in an ivar,
  call via `.call(x)` and `[x]`.
- `test/bound_method_array.rb` — array of Methods.
- `test/bound_method_gc.rb` — GC scan picks up the captured
  receiver.
- `test/bound_method_obj_recv.rb` — bind on a receiver other than
  `self` (local + ivar across class boundaries).

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -16147,21 +16147,20 @@ class Compiler
               mref = @nd_name[arg_ids[0]]
             end
             recv_cname = recv_t[4, recv_t.length - 4]
-            # Verify the method actually exists on the receiver's class
-            # (or any ancestor). Optcarrot has registration sites like
-            # `@pulse_0.method(:poke_2)` where Pulse — the inferred
-            # type — doesn't define poke_2; in Ruby the resulting
-            # NoMethodError surfaces only when that BM is invoked,
-            # which (for an address-mapping table) may never happen.
-            # Emitting `&sp_Pulse_poke_2` regardless leaves an
-            # unresolved symbol the linker rejects. Fall back to a
-            # null fn_ptr — safe as long as the BM isn't called.
-            recv_ci = find_class_idx(recv_cname)
-            method_exists = 0
-            walker_ci = recv_ci
-            while walker_ci >= 0 && method_exists == 0
+            # Walk the inheritance chain to find which class actually
+            # defines `mref`. The C function symbol for an inherited
+            # method lives under the *defining* class (`sp_Parent_foo`),
+            # not the receiver's own name (`sp_Child_foo`). Without
+            # this walk a `Noise.method(:poke_0)` where Noise inherits
+            # poke_0 from Oscillator emitted `&sp_Noise_poke_0` and
+            # the linker rejected the unresolved symbol. If no
+            # ancestor defines the method, fall back to a null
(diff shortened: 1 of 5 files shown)


mame:
While I believe this fix is necessary to get optcarrot working, I've realized that I haven't properly verified whether the current implementation actually contributes to running optcarrot. I'll conduct some further testing, and once I'm confident, I'll either reopen this issue or submit a new PR. Sorry for the noise.

PR #230, opened 2026-05-03
build: split test/bench compile+link to enable compiler caching

ryanseys:
## Summary
- Split the test rule (`Makefile:232`) into a separate compile step (`-c source.c -o source.o`) and link step (`source.o $(SP_RT_LIB) ... -o test_bin`).
- Same split applied to the bench rule (`Makefile:294`) for consistency, plus added `/tmp/_sp_b.c.o` to the bench cleanup list.
- No behavior change for users running `make test` or `make bench`; only the shape of the cc invocations changes.

## Why
Compiler caches like sccache and ccache only cache **pure compile invocations** — `cc -c source.c -o out.o`. Combined compile+link calls (no `-c`, output is an executable, links static libs and `-lm`) are classified as non-cacheable.

On the recent sccache CI experiments, this showed up clearly: of 307 sccache invocations on `ubuntu-gcc`, only **25 were cached** (the prism + regexp + bigint build-phase compiles). The 278 per-test calls were all classified as `non-compilation calls` and skipped.

Splitting unlocks per-test caching. The warm-cache test phase should drop from ~3min to ~1min for PRs that don't touch tests.

## Test plan
- [x] `make -j$(sysctl -n hw.ncpu) LTO=0 all` — clean build succeeds.
- [x] `make -j$(sysctl -n hw.ncpu) test OPT=-O0` — **282/282 pass** in 1m02s locally (macOS, clang).
- [x] CI confirms all 4 matrix jobs still green.

## Trade-offs
- One extra `exec()` per test (the link step) adds maybe 1ms × 282 = ~280ms of overhead on a clean run. Negligible vs. the ~120s saved on warm cache.
- The intermediate `.o` lives in the same `\$\$tmpdir` as the other test artifacts, so cleanup at line 258 already handles it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -229,7 +229,8 @@ build/test-results/%.ok: test/%.rb spinel_parse$(EXE) $(SP_RT_LIB) spinel_codege
 	rm -f "$@.diff"; \
 	./spinel_parse$(EXE) "$<" "$$ast" 2>/dev/null && \
 	./spinel_codegen$(EXE) "$$ast" "$$cfile" 2>/dev/null && \
-	$(CC) $(CFLAGS) -Werror $(SEC_FLAGS) -Ilib "$$cfile" $(SP_RT_LIB) $(LDFLAGS) -lm $(GC_FLAGS) -o "$$bin" 2>/dev/null; \
+	$(CC) $(CFLAGS) -Werror $(SEC_FLAGS) -Ilib -c "$$cfile" -o "$$cfile.o" 2>/dev/null && \
+	$(CC) $(CFLAGS) "$$cfile.o" $(SP_RT_LIB) $(LDFLAGS) -lm $(GC_FLAGS) -o "$$bin" 2>/dev/null; \
 	if [ $$? -eq 0 ]; then \
 	  if [ -f "$<.expected" ]; then \
 	    LC_ALL=C sed 's/\r$$//' "$<.expected" >"$$exp.n"; \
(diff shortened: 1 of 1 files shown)


matz:
Direction is good — the sccache/ccache classification of "compile+link" as non-cacheable is exactly why per-test caching wasn't kicking in, and splitting unlocks it. Numbers match what I'd expect.

On the Gemini feedback, partial agreement:

- **`$(CFLAGS)` at the link step — valid.** The link step uses the compiler driver, which interprets ABI-affecting flags like `-m32`/`-m64` and especially `-fsanitize=...` to pick the right runtime libraries (`libubsan`, `libasan`, etc.) and select the correct multilib path. With the current change, `make test CFLAGS=-fsanitize=undefined` would compile sanitized object files but link them against the non-sanitized runtime, producing a confusing link error or a binary missing the sanitizer hooks. Worth carrying `$(CFLAGS)` through to the link step on both rules.
- **`-Werror` at the link step — not needed.** Linker diagnostics aren't warnings the C front-end controls, and `-Werror` is a no-op at link. Skipping it there is correct.

So the suggested form would be:

```
$(CC) $(CFLAGS) "$$cfile.o" $(SP_RT_LIB) $(LDFLAGS) -lm $(GC_FLAGS) -o "$$bin"
```

(and the symmetric change to the bench rule).

If you want to keep this minimal, this is a one-token fix. Once that's in and you flip out of draft, happy to merge.

PR #234, opened 2026-05-03
fix(codegen): chained `@a = @b = expr` evaluates RHS once via poly

mame:
## Reproduction

```ruby
# Two ivars whose slots end up with disagreeing concrete types
# (here: a string slot and an int slot), assigned via a chain.
class C
  def initialize
    @a = "hello"
    @b = 42
  end

  def reset
    @a = @b = nil
  end
end

C.new.reset
```

## Expected behavior

The program type-checks under spinel's codegen and runs to completion, matching CRuby's behavior of clearing both slots.

## Actual behavior

`spinel_codegen` lowers the chain as a single nested expression:

```c
static inline mrb_int sp_C_reset(sp_C *self) {
    self->iv_a = (self->iv_b = 0);
```

The outer assignment inherits the inner write's type, so a string slot ends up receiving an `mrb_int`. gcc rejects the emission:

```
spinel_out.c: In function 'sp_C_reset':
spinel_out.c:40:16: error: assignment to 'const char *' from 'mrb_int' {aka 'long int'} makes pointer from integer without a cast [-Wint-conversion]
   40 |     self->iv_a = (self->iv_b = 0);
      |                ^
```

The same shape with a side-effecting RHS — `@a = @b = foo` — fails identically:

```c
self->iv_a = (self->iv_b = sp_foo());
```

```
spinel_out.c:51:16: error: assignment to 'const char *' from 'mrb_int' {aka 'long int'} makes pointer from integer without a cast [-Wint-conversion]
```

## Analysis

The parser nests `@a = @b = expr` as

```
InstanceVariableWriteNode(@a, expr=InstanceVariableWriteNode(@b, expr=<bottom>))
```

Two things go wrong against this shape:

1. **`scan_writer_calls` does not widen through the chain.** When it visits the outer `@a` write, it asks `infer_type` for the type of the inner `@b` write. `infer_type` has no `InstanceVariableWriteNode` case and falls through to its `"int"` default, which the surrounding guard `if at != "int" && at != "nil"` then drops. The chain head (`@a`) therefore never sees the bottom RHS type, and its slot is never widened to absorb a value of a disagreeing type.

2. **Lowering is structurally a single nested expression.** Even when scan widens the slots correctly, lowering `iv_a = (iv_b = bottom_expr)` ties the outer assignment's RHS type to the inner ivar's slot type. With disagreeing slot types, the C compile fails.

A naive workaround would be to unroll the chain by recursively lowering each ivar write with the bottom RHS substituted in. That re-evaluates the RHS once per ivar — a `CallNode` like `@a = @b = foo` would emit `sp_foo()` twice and double-fire the call's side effects.

## Fix

Two changes in `spinel_codegen.rb`:

### `scan_writer_calls`: drill through chains

Walk through inner `InstanceVariableWriteNode` expressions to the bottom RHS, and widen the chain head against that type. Chain heads bypass the `int`/`nil` guard because `update_ivar_type`'s widening logic still needs to fire when a previously-concrete slot is reused for a chain whose bottom RHS disagrees.

```ruby
bottom = expr_id
is_chain_head = 0
while bottom >= 0 && @nd_type[bottom] == "InstanceVariableWriteNode"
  bottom = @nd_expression[bottom]
  is_chain_head = 1
end
if is_empty_hash_literal(bottom) == 0 && is_empty_array_literal(bottom) == 0
  at = infer_type(bottom)
  if is_chain_head == 1
    update_ivar_type(@current_class_idx, iname, at)
  elsif at != "int" && at != "nil"
    update_ivar_type(@current_class_idx, iname, at)
  end
end
```

In the reproduction, this lifts `@a` from `"string"` to `"poly"` once it sees the chain's bottom RHS (`nil` or `int`).

### `compile_stmt`: route chains through a dedicated helper

A new `compile_chained_ivar_writes` evaluates the RHS once into a typed temp and emits one store per slot. Slots widened to `poly` get `box_value_to_poly`; slots whose type matches the RHS get the temp directly.

```ruby
def compile_chained_ivar_writes(nid)
  ivars = []
  cur = nid
  while @nd_type[cur] == "InstanceVariableWriteNode"
    ivars.push(@nd_name[cur])
    cur = @nd_expression[cur]
  end
  rhs_id = cur
  rhs_type = infer_type(rhs_id)
  # NilNode rhs is special-cased: see below.
  if @nd_type[rhs_id] == "NilNode"
    # ... per-slot recurse ...
    return
  end
  rhs_expr = compile_expr(rhs_id)
  tmp = new_temp
  emit("  " + c_type(rhs_type) + " " + tmp + " = " + rhs_expr + ";")
  # per-slot: poly → box, otherwise temp directly
  ...
end
```

The native-temp-plus-box-for-poly shape (rather than "always a poly temp + unbox per slot") is deliberate: it avoids any poly→non-poly unbox. Cross-tag union reads (e.g. `.v.s` on an int-tagged box) are UB in C, so boxing is only injected for slots that have already been widened to poly.

The constructor body's emit path (in `compile_class_def`) has its own `emit_raw`-based ivar-write handling. For chained writes it now defers to `compile_stmt`, which routes through the same helper.

#### NilNode RHS

`@a = @b = nil` with disagreeing concrete slots can't go through the typed-temp strategy: `mrb_int _t = 0; iv_a = _t;` is rejected by C because an `mrb_int` variable doesn't implicitly convert to `const char *`. The literal `0` from `compile_expr(NilNode)`, however, *is* a null pointer constant and assigns cleanly to any slot. Since `NilNode` has no side effects, the helper recurses per slot for nil-RHS chains and produces `iv_x = 0;` for each — type-correct for every slot type.

## Emit after fix

`@a = @b = nil`:

```c
static inline mrb_int sp_C_reset(sp_C *self) {
    SP_GC_SAVE();
    SP_GC_ROOT(self);
    self->iv_a = sp_box_nil();
    self->iv_b = 0;
```

`@a = @b = foo`:

```c
static inline mrb_int sp_C_reset(sp_C *self) {
    SP_GC_SAVE();
    SP_GC_ROOT(self);
    mrb_int _t1 = sp_foo();
    self->iv_a = sp_box_int(_t1);
    self->iv_b = _t1;
```

`@a` is widened to poly and stores the boxed value; `@b` keeps its `mrb_int` slot and gets the temp directly. The RHS is evaluated exactly once.

## Test plan

- [x] `make -j4 bootstrap` green (`gen2.c == gen3.c`)
- [x] `make test` — 288 pass, 0 fail, 0 error
- [x] `test/chained_ivar_write_split.rb` — chained literal nil with disagreeing slot types
- [x] `test/chained_ivar_write_call_rhs.rb` — side-effecting RHS evaluated exactly once (`$count == 1` after `@a = @b = foo`)


matz:
Merging — this is a real improvement over master, the once-only RHS evaluation strategy is right, and the test coverage for the nil and call-RHS shapes lands cleanly. CI 4/4 green.

Filing the two `high`-priority Gemini concerns as separate issues for follow-up rather than blocking on them here, since neither is exercised by the current test suite and the on-master behavior they cover is "fail to compile" — not "wrong result":

1. **Tail-node widening gap** (#235). When the bottom RHS is `int`/`nil` and the chain *tail* slot was previously a pointer type (e.g. `string`), `scan_writer_calls`'s `is_chain_head == 0` branch keeps the gate `at != "int" && at != "nil"` and the tail stays unwidened. `compile_chained_ivar_writes` then assigns an `mrb_int` temp directly to a `const char *` slot — same C error shape this PR was fixing for the head.

2. **Empty literal in chain not specialized** (#236). `@a = @b = {}` with a slot that's been promoted to a typed-container (e.g. `str_str_hash`) should emit `sp_StrStrHash_new()` per the standard `InstanceVariableWriteNode` path; the chain helper's `compile_expr(rhs_id)` returns the default `sp_StrIntHash_new()` and mismatches the promoted slot.

Issue #235, opened 2026-05-03
Chained ivar write: tail-node slot type not widened when bottom RHS is int/nil


matz:
Follow-up to #234 (chained `@a = @b = expr` evaluation), filing a `high`-priority Gemini concern that wasn't exercised by the merged test suite.

## Summary

`scan_writer_calls`'s post-#234 chain-drilling logic widens the chain *head* through the inner-write nodes down to the bottom RHS, but the chain *tail* is visited as an ordinary `InstanceVariableWriteNode` (its `expr` is the bottom expression, not a nested write). That visit hits the existing gate `at != \"int\" && at != \"nil\"` and the tail's slot is not widened. If the tail was previously a pointer type, `compile_chained_ivar_writes` then assigns an `mrb_int` temp directly to it.

## Reproduction

```ruby
class C
  def initialize
    @a = \"hello\"   # string slot
    @b = \"world\"   # string slot
  end

  def reset
    @a = @b = 42    # int RHS; both slots ought to widen to poly
  end
end

C.new.reset
```

Pre-#234 master fails differently (single nested expression). Post-#234 the head (`@a`) widens to poly via the new `is_chain_head == 1` branch. The tail (`@b`) keeps its `string` recording. `compile_chained_ivar_writes` emits:

```c
mrb_int _t1 = 42;
self->iv_a = sp_box_int(_t1);   /* OK: @a is poly */
self->iv_b = _t1;               /* C error: const char * = mrb_int */
```

## Fix sketch

Inside `scan_writer_calls`, when the current write's expression is *not* itself an `InstanceVariableWriteNode` but the surrounding chain has one (i.e., this write is a tail of some chain visited above), the `int`/`nil` gate should be relaxed the same way the chain-head branch relaxes it. Cleanest is to detect tail membership during the same drill-down pass that handles the head and widen all participants in one sweep.

Alternative shape: `compile_chained_ivar_writes` could box the temp on a per-slot mismatch instead of relying on type-inference to have widened. That keeps the type model honest at the cost of a mid-codegen branch. Either is acceptable; widening at scan time is the more idiomatic spinel pattern and matches what the head-widening already does.

## Test fixture

A regression test that didn't ship with #234:

```ruby
class C
  def initialize
    @a = \"hello\"
    @b = \"world\"
  end
  def reset; @a = @b = 42; end
end
C.new.reset                 # must compile + run
```

Mentioning briefly per the cadence agreed in #83.

Refs #234

PR #238, opened 2026-05-03
fix(codegen): nil chain head must not bypass scan_writer_calls guard

mame:
## Reproduction

```ruby
class Base
  def reset
    @a = @b = nil
  end
end

class Sub < Base
  def initialize
    @a = "hello"
    @b = 42
  end
  attr_reader :a, :b
end

s = Sub.new
puts s.a
puts s.b
```

A parent class declares its ivar slots via a chained `nil` assignment. A subclass writes typed values (string + int) into the same slots from its own `initialize`.

## Expected behavior

CRuby:

```
$ ruby repro.rb
hello
42
```

Spinel should emit C that compiles cleanly and prints the same.

## Actual behavior

`spinel_codegen` widens both `Sub#iv_a` and `Base#iv_a` to `sp_RbVal` (poly), then emits the subclass's `@a = "hello"` initializer as a raw `const char *` store into the poly slot:

```c
struct sp_Base_s {
  sp_RbVal iv_a;
  mrb_int iv_b;
};

struct sp_Sub_s {
  sp_RbVal iv_a;
  mrb_int iv_b;
  ...
};

static inline sp_Sub *sp_Sub_new(void) {
  ...
  self->iv_a = (&("\xff" "hello")[1]);   // RHS is const char *, slot is sp_RbVal
  self->iv_b = 42;
  ...
}
```

gcc rejects the emission:

```
spinel_out.c: In function 'sp_Sub_new':
spinel_out.c:64:16: error: incompatible types when assigning to type 'sp_RbVal' from type 'char *'
   64 |   self->iv_a = (&("\xff" "hello")[1]);
      |                ^
spinel_out.c: In function 'sp_Sub_initialize':
spinel_out.c:71:16: error: incompatible types when assigning to type 'sp_RbVal' from type 'char *'
spinel_out.c: In function 'main':
spinel_out.c:85:5: error: cannot convert to a pointer type
```

(`@b` happens to stay `mrb_int` because the subclass write `@b = 42` and the parent chain bottom both type-infer as `int`, so neither side disagrees with the other.)

## Analysis

`scan_writer_calls` was taught to drill through chained `InstanceVariableWriteNode` expressions so the chain head could be widened against the bottom RHS — necessary for `@string_slot = @int_slot = expr_returning_int`, where the head's existing slot type would otherwise stay at `string` and conflict with the typed-temp store at emit time.

That bypass forced `update_ivar_type` even when the bottom RHS was `nil`, which wasn't required (nil chains are handled at emit time by `compile_chained_ivar_writes`' per-slot recurse path — `compile_expr(NilNode)` produces the literal `0`, a null pointer constant valid for any slot type, without needing the slot to be widened).

Forcing nil into a slot then interacts badly with `update_ivar_type`'s parent-cascade and the iterative inference fixpoint:

1. **Iter 1** — `Base#reset`'s chain head bypass calls `update_ivar_type(Base, @a, "nil")`. `Base.@a` was `"int"` (the chain placeholder default); the `old == "int"` branch overwrites it to `"nil"`. Then `Sub#initialize`'s `@a = "hello"` calls `update_ivar_type(Sub, @a, "string")`, which cascades up to `Base`. `Base.@a` is `"nil"`, so `old == "nil"` overwrites it to `"string"`.
2. **Iter 2** — chain head bypass fires again, calling `update_ivar_type(Base, @a, "nil")` against the now-string slot. `string` is a nullable pointer type, so `update_ivar_type` widens to `"string?"`. `Sub`'s cascade then tries `update_ivar_type(Base, @a, "string")` against `"string?"` — neither side recognizes `T?` ↔ `T` as compatible, so it falls through to the catch-all `types[k] = "poly"`.
3. The poly cascades down to `Sub.@a` on the next iteration, and the fixpoint stabilizes with both slots as `sp_RbVal`. The subclass's `@a = "hello"` then emits as the raw `const char *` literal into a poly slot.

## Fix

Restrict the chain-head bypass to non-nil bottoms — the only case where it was load-bearing — and let nil chains fall through the existing `int`/`nil` guard.

```diff
-          if is_chain_head == 1
+          if is_chain_head == 1 && at == "int"
             update_ivar_type(@current_class_idx, iname, at)
           elsif at != "int" && at != "nil"
             update_ivar_type(@current_class_idx, iname, at)
           end
```

After the fix, the slots stay typed and the emit is clean:

```c
struct sp_Sub_s {
  const char * iv_a;
  mrb_int iv_b;
  ...
};

static inline mrb_int sp_Base_reset(sp_Base *self) {
    self->iv_a = 0;       // null pointer constant — accepted for const char *
    self->iv_b = 0;
```

## Test plan

- [x] `make -j4 bootstrap` green (`gen2.c == gen3.c`)
- [x] `make test` — 289 pass, 0 fail, 0 error (with the new `chained_ivar_write_subclass.rb`)
- [x] Reproduction repro compiles and prints `hello` / `42`
- [x] Existing chained-ivar tests (`chained_ivar_write_split.rb`, `chained_ivar_write_call_rhs.rb`) still pass — the bypass remains in force for `int` bottoms, which is what they needed


matz:
Auto-closed by 3f1e588 `Closes #238` trailer. Your branch was based off master before #235's chain_inames-based widening, so the diff didn't apply cleanly — I ported your gate (`is_chain_head == 1 && at == "int"`) to the new structure (`chain_inames.length > 1 && at == "int"`). Test fixture is verbatim from this PR. 291 pass, bootstrap clean. Thanks for the quick regression catch + analysis.

PR #257, opened 2026-05-04
fix(codegen,runtime): box Range values in poly-hash literals

### LINKED ISSUE #249: Hash literal with mixed `Integer` and `Range` values fails to compile — `Range` boxed via `sp_box_int`
## Summary

A hash literal whose values mix `Integer` and `Range` fails to compile. Spinel synthesizes a `sp_SymPolyHash` for the polymorphic value type, but every value insertion uses `sp_box_int(...)` — including when the value is an `sp_Range`. The C compiler rejects it: `passing 'sp_Range' to parameter of incompatible type 'mrb_int'`.

## Reproduction

```ruby
H = {
  success: 200..299,
  ok:      200,
}
```

```sh
spinel repro.rb -c -o /tmp/repro.c
cc -fsyntax-only -I ~/git/spinel/lib /tmp/repro.c
```

## Actual behavior

```
/tmp/repro.c:25:53: error: passing 'sp_Range' to parameter of incompatible type 'mrb_int' (aka 'long long')
   25 |     sp_SymPolyHash_set(_t1, SPS_success, sp_box_int(sp_range_new(200, 299)));
      |                                                     ^~~~~~~~~~~~~~~~~~~~~~
note: 'sp_box_int' declared here
   static sp_RbVal sp_box_int(mrb_int v) { ... }
```

The hash element insertion goes through `sp_box_int` for *every* value, but `sp_range_new(200, 299)` returns `sp_Range`, not `mrb_int`. The boxing function selection is wrong for non-int values.

## Expected behavior

A poly hash with heterogeneous value types should use the right boxing function per value — `sp_box_range(...)` for ranges, `sp_box_int(...)` for ints, etc. Same as how a method that returns a poly value selects the boxer based on the inferred type of each return path.

## Why this matters in practice

Roundhouse's emitted test_helper synthesizes a status-code lookup table:

```ruby
STATUS_SYMBOLS = {
  success:              200..299,
  redirect:             300..399,
  missing:              404,
  not_found:            404,
  error:                500..599,
  ok:                   200,
  created:              201,
  no_content:           204,
  # ...
}.freeze
```

This is the canonical Rails `assert_response :success` lookup shape. Three of the entries (`success`, `redirect`, `error`) use `Range` values; the rest are `Integer`. The hash trips this bug at every Range entry — three errors in real-blog's emitted tests today.

## Roundhouse-side workaround

If a fix is far off, Roundhouse can split the lookup into two homogeneous hashes (one of integers, one of `(min, max)` tuples) and dispatch by symbol presence, or compose a flat dispatch in code. The `Range`-keyed shape is more idiomatic Ruby but not load-bearing.

Mentioning briefly per the cadence agreed in matz/spinel#83.

Halvanhelv:
Hash literals with mixed Integer and Range values failed to compile because `box_non_nullable_value_to_poly` had no branch for `"range"` and fell through to `sp_box_int`, which rejects `sp_Range` structs.

Added `SP_BUILTIN_RANGE` and `sp_box_range()` to heap-allocate the Range and box via `sp_box_obj`, following the same pattern as `SP_BUILTIN_PROC`.

Test in `test/hash_poly_range.rb` covers mixed poly-hash access and verifies Range presence doesn't break compilation.

Closes #249


Halvanhelv:
Closing — the same fix (sp_box_range via sp_gc_alloc + SP_BUILTIN_RANGE) was already landed on master by the maintainer, along with sp_Range_inspect and SP_BUILTIN_TIME.

PR #288, opened 2026-05-05
feat(parse): support MissingNode (parse-error placeholder; defensive handler)

ryanseys:
Adds a parser case for `PM_MISSING_NODE` that exits with a clear internal-error message. Prism emits `MissingNode` as an error-recovery placeholder.

## Why no test fixture

The parser's main loop bails on parse errors before `flatten()` is called (`spinel_parse.c:1201`), so reaching this case indicates Prism produced a `MissingNode` without flagging an error in `error_list` — a contract violation we surface clearly rather than route through the generic `UnsupportedNode` path. The handler is defense-in-depth in case Prism's contract changes.

I tried to construct an exercising fixture from both valid and invalid Ruby; the `error_list` bail-out catches every case I could come up with. Happy to add one if a reviewer can think of input that reaches this arm.

## Why `exit(1)` and not `mrb_raise`

Matches the existing `UnsupportedNode`-style `fprintf(stderr, ...); exit(1)` precedent already in `flatten()`. Loud-failure over silent miscompile.

## Verification

- `make test` — 326 pass, 0 fail.

### DIFF
--- a/spinel_parse.c
+++ b/spinel_parse.c
@@ -794,6 +794,16 @@ static int flatten(pm_node_t *node) {
     node_counter--;
     return flatten(n->value);
   }
+  case PM_MISSING_NODE: {
+    /* Prism emits MissingNode as an error-recovery placeholder. main()
+       bails out before flatten() runs when parser.error_list is
+       non-empty, so reaching this case means Prism produced a
+       MissingNode without flagging an error -- a contract violation we
+       surface clearly rather than silently miscompile. */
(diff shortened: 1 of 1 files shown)


ryanseys:
Both addressed in 86cb99f:

- Dropped the stale `(line 1201)` reference; now describes the bail-out as "main() bails out before flatten() runs when parser.error_list is non-empty".
- Switched the byte-offset format to `%td` and removed the `(long)` cast so the ptrdiff_t prints correctly on LLP64 platforms.

Re-verified: `make test` (326 pass) + `make bootstrap` (gen2.c == gen3.c).

PR #291, opened 2026-05-05
feat(parse,codegen): support Call{Operator,And,Or}WriteNode (3 nodes)

ryanseys:
Bundle of three `Call*` compound-write forms:
- `CallOperatorWriteNode` — `obj.attr += val` (and `-=`, `*=`, `/=`, `%=`, `<<=`, `>>=`, `&=`, `|=`, `^=`).
- `CallAndWriteNode` — `obj.attr &&= val`.
- `CallOrWriteNode` — `obj.attr ||= val`.

Together with the `Index{Operator,And,Or,Target}WriteNode` set and the existing `IndexOperatorWriteNode`, this closes out the compound-write family across all variable kinds (locals, ivars, globals, cvars, indexed slots, and now method-style accessors).

## Single-receiver-evaluation semantics

Critical for correctness: `obj.bar += val` is shorthand for `tmp = obj; tmp.bar = tmp.bar + val`, NOT `obj.bar = obj.bar + val`. Single-evaluation matters when the receiver expression has side effects. The test exercises `get_cfg(c).name &&= "second"` and verifies the helper fires exactly once via a global counter. `compile_call_assign_typed` emits a temp-receiver block:

```c
{ sp_<Cls> *_t = recv; _t->iv_<bname> = _t->iv_<bname> <op> (val); }
```

so the read and write share one temp.

## Restriction: attr_accessor-backed typed instance receivers only

The codegen checks `rt[0,4] == "obj_"` to detect a typed instance, then `cls_has_attr_writer(ci, bname)` to confirm the attr exists. For non-attr receivers (untyped, polymorphic, or method-without-backing-ivar), we `exit(1)` with a precise message rather than emit silently-wrong C — matching the loud-failure spirit even though the parser succeeded.

`+=` on a string-valued attr routes to `sp_str_concat` (mirroring `LocalVariableOperatorWriteNode`).

## CallTargetNode intentionally not included

The 4th compound form (`obj.x, p.y = 1, 2` for multi-assign LHS) landed upstream via mame's #216, so it is already present.

## Test fixtures

- `test/call_op_write.rb` — `Counter#n` exercising all 10 numeric operators; `Greeting#msg += " world"` for the string-concat path.
- `test/call_or_write.rb` — don't-fire-on-truthy + receiver-once via side-effecting helper.
- `test/call_and_write.rb` — fire-on-truthy + receiver-once verification.

The `||=` "fire on nil" branch for string slots is documented as a Spinel-specific gap inline (nil-typed string slots get sentineled to `""` by the constructor, and `""` is C-truthy).

## Verification

- `make test` — all pass.


ryanseys:
All three addressed in 9d7ecf1: each of `PM_CALL_OPERATOR_WRITE_NODE`, `PM_CALL_AND_WRITE_NODE`, and `PM_CALL_OR_WRITE_NODE` now mirrors the `PM_CALL_NODE` pattern — when `PM_CALL_NODE_FLAGS_SAFE_NAVIGATION` is set, emit `call_operator: &.` so the codegen sees the safe-nav marker on `obj&.attr <op>= val` forms.

Re-verified: `make test` (324 pass) + `make bootstrap` (gen2.c == gen3.c).

PR #292, opened 2026-05-05
feat(parse,codegen,runtime): support BackReferenceReadNode ($&, $~, $', $`)

ryanseys:
Adds the four symbolic regex back-references that CRuby populates on every match alongside the numbered captures (\$1..\$9, already supported via `NumberedReferenceReadNode`).

Spinel's regex runtime already calls `sp_re_set_captures` after every successful match; this commit extends that helper to also populate the three new globals:

- `sp_re_match_str` — \$& (the whole matched substring)
- `sp_re_match_pre` — \$\` (substring before the match)
- `sp_re_match_post` — \$' (substring after the match)

## Memory lifecycle

Each is heap-allocated via `sp_str_alloc_raw` and **overwritten** (not appended) on every `sp_re_set_captures` call. The runtime's existing free path handles the previous string.

## Read-time null guard

The codegen accessor null-guards the read so an untouched-since-no-match access returns `""` instead of `NULL` — matches CRuby's "all-empty after no match" semantics.

## $~ aliasing

\$~ is the `MatchData` object in CRuby; Spinel has no `MatchData` wrapper, so the codegen aliases \$~ to \$&. Most user code that touches \$~ ends up calling `.to_s` on it anyway, which agrees. Implementing `MatchData` properly would be a larger separate change.

## Test (test/back_ref.rb)

- A first match with \$&, \$\`, \$' all populated.
- A second match overwrites the captures (verifies the runtime resets on each `sp_re_set_captures` call).

## Verification

- `make test` — all pass.


ryanseys:
Both addressed in ddec333:

**GC root registration (high)** — added `sp_re_mark_globals()` that calls `sp_mark_string` on the five static-lifetime regex globals (`sp_re_last_str`, `sp_re_captures[10]`, `sp_re_match_str`, `sp_re_match_pre`, `sp_re_match_post`). Wired in via forward-declaration before `sp_gc_mark_all` and called there right after the root loop, so the str-heap sweep no longer reaps live back-reference / capture buffers. `sp_mark_string` is null-safe and no-ops on string literals (the `""` default of `sp_re_last_str`), so unconditional calls are safe.

**`$~` comment (medium)** — corrected to `the MatchData object (aliased to $& in Spinel; no MatchData wrapper)`.

Re-verified: `make test` (322 pass) + `make bootstrap` (gen2.c == gen3.c).

PR #373, opened 2026-05-07
fix(codegen): typed-array RHS in multi-write picks matching C type and getter

mame:
## Reproduction

`test/multi_write_float_array_destructure.rb` (committed alongside the fix):

~~~ruby
TBL = [[-0.12, 0.40], [0.00, 0.68], [0.31, 1.00], [0.72, 1.00]]

a, b = TBL[0]
puts a
puts b

c, d = TBL[3]
puts c
puts d

LABELS = [["lo", "low"], ["hi", "high"], ["mid", "middle"]]

s, t = LABELS[1]
puts s
puts t
~~~

## Expected behavior

~~~
$ ruby test/multi_write_float_array_destructure.rb
-0.12
0.4
0.72
1.0
hi
high
~~~

## Actual behavior on master (`0d10645`)

The destructure temp is declared `sp_IntArray *` but
`compile_expr(val_id)` emits a typed cast that produces a
different pointer type — gcc rejects the resulting line under
its default `-Wincompatible-pointer-types` / `-Wint-conversion`:

~~~
$ ./spinel test/multi_write_float_array_destructure.rb -o /tmp/r
/tmp/spinel_out.Ogvjoo.c: In function ‘main’:
/tmp/spinel_out.Ogvjoo.c:88:25: error: initialization of ‘sp_IntArray *’ from incompatible pointer type ‘sp_FloatArray *’ [-Wincompatible-pointer-types]
   88 |     sp_IntArray *_t10 = ((sp_FloatArray *)sp_PtrArray_get(cst_TBL, 0));
      |                         ^
/tmp/spinel_out.Ogvjoo.c:100:25: error: initialization of ‘sp_IntArray *’ from incompatible pointer type ‘sp_StrArray *’ [-Wincompatible-pointer-types]
  100 |     sp_IntArray *_t12 = ((sp_StrArray *)sp_PtrArray_get(cst_LABELS, 1));
      |                         ^
/tmp/spinel_out.Ogvjoo.c:102:10: error: assignment to ‘const char *’ from ‘mrb_int’ {aka ‘long int’} makes pointer from integer without a cast [-Wint-conversion]
  102 |     lv_s = sp_IntArray_get(_t12, 0);
      |          ^
spinel: C compilation failed
~~~

The runtime damage is the same shape as the C error: even when
gcc is coaxed into accepting the cross-type assignment, the
subsequent `sp_IntArray_get(_t10, 0)` reads float bytes through
the int accessor and returns garbage.

## Analysis & fix

`compile_multi_write` already handles the array-literal RHS
(`a, b = [1, 2]`) and tuple-call RHS branches with proper typed
dispatch. The third branch — RHS is a method/expression call
returning a typed array — only checked for `poly_array`, falling
through to `sp_IntArray *` / `sp_IntArray_get` for everything
else:

~~~ruby
if val_t_local == "poly_array"
  emit("  sp_PolyArray *" + tmp + " = " + compile_expr(val_id) + ";")
else
  emit("  sp_IntArray *" + tmp + " = " + compile_expr(val_id) + ";")
end
~~~

`compile_multi_write_splat` (the splat-assign equivalent) already
handles `float_array` / `str_array` / `poly_array` / `<X>_ptr_array`
in its `get_fn` selection. Mirror that here: pick the temp's C
type and the per-slot getter from `val_t_local`, and drive
`emit_multi_write_target` with the matching element type. For
`<X>_ptr_array` RHSes, additionally cast the void* return to the
element type.

The `CallTargetNode` branch (used when the LHS is `obj.attr =`)
hard-coded `sp_IntArray_get(...)` for its slot expression; that
now reuses the same typed `rhs` so an attr-writer target gets
the typed slot value too.

## Diff stat

~~~
 spinel_codegen.rb                                  | 53 +++++++++++++++++-----
 test/multi_write_float_array_destructure.rb        | 27 +++++++++++
 ...multi_write_float_array_destructure.rb.expected |  6 +++
 3 files changed, 74 insertions(+), 12 deletions(-)
~~~

## Test plan

- [x] `make -j4 bootstrap` green; gen2 == gen3 fixpoint OK.
- [x] `make test` 291 pass / 0 fail / 0 error
      (290 baseline + 1 new test).
- [x] Manual reproduction: master fails with the
      `-Wincompatible-pointer-types` / `-Wint-conversion`
      cluster above; this branch produces the expected output.


matz:
This now conflicts with master after #372 (`fix(codegen): multi-write destructure dispatches by RHS array type`) merged a few minutes ago, since both touch the same multi-write destructure path in `spinel_codegen.rb`. CI didn't re-run yet on the pre-conflict state.

Could you rebase against current master? The two PRs look complementary (372 dispatches by RHS array type, this one picks the matching C type / getter for the destructure temp), so the rebase should mostly be a clean integration — but a fresh look from you to confirm the right ordering of the two patches is more reliable than me force-pushing on your behalf.

Happy to merge once mergeable + CI green.

mame:
It looks like #372 was the same fix! My local optcarrot worked the same way without #373, so I'll close this.

FWIW, (modified) optcarrot is basically running on my end, and I'm at the final cleanup stage now.

PR #384, opened 2026-05-08
fix(codegen): collapse stale unqualified obj-name observations against qualified peers

mame:
## Reproduction

`test/forward_call_class_method_obj.rb` (committed alongside the fix):

~~~ruby
# `Class.new(obj_X)` shape: caller passes a user-class instance to a
# forward-defined Class.new.

class Holder
  def initialize(name)
    @name = name
  end
  def label
    "h:#{@name}"
  end
end

class Caller
  def make
    @h = Holder.new("alpha")
    @t = Target.new(@h)
  end
  def show
    @t.held_label
  end
end

class Target
  def initialize(h)
    @holder = h
  end
  def held_label
    @holder.label
  end
end

c = Caller.new
c.make
puts c.show
~~~

Companion tests cover an `IntArray` arg (`forward_call_class_method_int_array.rb`) and a nested `Outer.new(Inner.new)` shape (`forward_call_class_method_nested.rb`).

## Expected behavior

~~~
$ ruby test/forward_call_class_method_obj.rb
h:alpha
~~~

The simple two-class repro happens to work on master because the
constructor-arg path resolves the forward ref through a single
in-flight scan. The bug shows up when a multi-class single-file
program reorders sibling classes inside a module so that a class
lexically *earlier* than its sibling references the sibling's
constant in an ivar init — `@b = B.new(@conf)` inside
`class A` while `class B` comes later in the same `module Mod`.
Pass-1 scan stamps the unqualified `obj_<bare>` form into the
slot, a later pass with all classes registered records the
qualified `obj_<scope>_<bare>` form, and the codegen sees the two
strings as different types and widens to poly all the way through
the type fixpoint.

## Actual behavior on master (`b19b722`)

A multi-class lib-order single-file build that puts a caller class
ahead of its sibling callees produces poly-typed callee params:

~~~
$ ./spinel /tmp/single.rb -c -o /tmp/out.c
$ grep "sp_Mod_B_initialize" /tmp/out.c | head -1
static inline void sp_Mod_B_initialize(sp_Mod_B *self, sp_RbVal lv_conf, sp_RbVal lv_x, sp_IntArray * lv_arr);
~~~

`B#initialize`'s `conf` and `x` params are typed `sp_RbVal` (poly)
even though every call site in the caller class passes a concrete
`obj_Mod_Config` and `obj_Mod_X`. The `@x = x` body assignment
then widens the callee's `@x` ivar to poly, which makes
`@x.<method>` dispatch through every user class that defines the
method — *including the receiver's own class* — producing
infinite recursion at runtime (`cls_id == B → B#update`):

~~~c
static mrb_int sp_Mod_B_update(sp_Mod_B *self, mrb_int lv_data) {
  sp_RbVal _t450 = self->iv_x;
  mrb_int _t451 = 0;
  if (_t450.tag == SP_TAG_OBJ) {
    if (_t450.cls_id == 5) _t451 = sp_Mod_X_update(...);
    if (_t450.cls_id == 6) _t451 = sp_Mod_Y_update(...);
    if (_t450.cls_id == 15) _t451 = sp_Mod_B_update(...);  /* self-recurse */
  }
  return sp_Mod_B_finalize(self, (lv_data + _t451));
}
~~~

After the fix the same call site emits the concrete signature:

~~~c
static inline void sp_Mod_B_initialize(sp_Mod_B *self,
    sp_Mod_Config * lv_conf, sp_Mod_X * lv_x,
    sp_IntArray * lv_arr);
~~~

## Analysis & fix

`scan_ivars` registers ivar types during Pass 1 (class enumeration)
in source order. A class lexically *earlier* than its sibling
— for example `Mod::A` ahead of `Mod::B` — has its
`@b = B.new(@conf)` write scanned while `B` itself isn't yet in
`@cls_names`. `resolve_const_read_name`'s lexical-scope fallback
then misses `Mod_B` and stamps the unqualified `obj_B` into
`@cls_ivar_types[A]`.

A subsequent pass with all classes registered records the
qualified `obj_Mod_B` through `scan_writer_calls`'s `infer_type`
path. The two strings refer to the same class, but the codegen
has three sites that compare them as strings:

1. `update_ivar_type` — `old="obj_B"` / `new="obj_Mod_B"` falls
   into the *genuinely incompatible* branch and widens the slot
   to poly.
2. `unify_call_types` — same disagreement on `Class.new(@b, …)`'s
   widen-from-args flow collapses every callee's ptype to poly.
3. `finalize_ivar_heterogeneity` — observation list contains both
   `obj_B` and `obj_Mod_B` (distinct count = 2), so the slot is
   locked at poly even after rounds of fixpoint refinement.

Once the caller's `@b` ivar is poly, the cascade reaches every
`Class.new(@b, …)` call site as a poly arg, the callee's
`#initialize` ptype gets widened to poly, the callee's body's
`@b = b` makes its own `@b` ivar poly, and `@b.<method>` dispatches
through the poly arm — including the receiver's own class.

The fix adds a stale-unqualified-obj normalization at all three
sites. When old / new are both `obj_X` and `obj_Y_X`, *only* one
of them resolves through `find_class_idx`, and the registered name
ends with `_<unregistered_bare>`, treat them as the same class —
keep the qualified form. The unregistered side is the Pass-1
fallback; the qualified side is the post-registration result.

A new helper `drop_stale_unqualified_obj_obs` does the same
filtering on the observation list before
`finalize_ivar_heterogeneity`'s distinct-count widening decision.
The helper is type-primed with empty-string locals so the
self-bootstrap path infers the right `const char *` C types
through `sp_str_sub_range` / `sp_str_concat`.

## Diff stat

~~~
$ git diff --stat origin/master
 spinel_codegen.rb                                  | 129 ++++++++++++++++++++-
 test/forward_call_class_method_int_array.rb        |  26 +++++
 ...forward_call_class_method_int_array.rb.expected |   1 +
 test/forward_call_class_method_nested.rb           |  35 ++++++
 test/forward_call_class_method_nested.rb.expected  |   1 +
 test/forward_call_class_method_obj.rb              |  36 ++++++
 test/forward_call_class_method_obj.rb.expected     |   1 +
 7 files changed, 224 insertions(+), 5 deletions(-)
~~~

## Test plan

- [x] `make -j4 bootstrap` — fixpoint clean (`gen2.c == gen3.c`).
- [x] `make test` — `316 pass, 0 fail, 0 error` (313 baseline + 3 new).
- [x] `ruby test/forward_call_class_method_*.rb` matches each
      spinel-compiled binary's output.


mame:
Sorry, I created a wrong PR.

PR #390, opened 2026-05-08
Two small codegen additions: ENV.fetch(k,d) + module-scope @@cvars

OriPekelman:
## Two small additions found while building tep on master

Both surfaced while iterating on a Sinatra clone. Neither is a hard
miscompile in spinel itself — they're holes in the supported Ruby
surface that user code reasonably reaches for.

### 1. `ENV.fetch(key, default)`

`ENV[k]` works today; `ENV.fetch(k, dflt)` doesn't — the codegen
emits `cannot resolve call to 'fetch' on int (emitting 0)` and the
program segfaults at runtime when the result is touched.

CRuby semantics for the two-arg form:

```ruby
ENV.fetch("HOME", "/tmp")          # => "/usr/me"   (or "/tmp" if unset)
```

Patch adds the type-inference and codegen arms:

```ruby
ENV.fetch(key, default)
# expands to:
({ const char *_t = getenv(key); _t ? sp_str_dup_external(_t) : (default); })
```

The block form (`ENV.fetch("X") { computed }`) is not handled —
keeps the patch small. KeyError-on-missing for the single-arg form
isn't modeled either; that already gracefully returns the
`getenv()` result (NULL → empty string), matching `ENV[]`.

Repro: `test/env_fetch.rb` (reads three configurations: missing
with literal default, set with explicit value, missing with
non-literal default).

### 2. Module-scope `@@cvars`

`@@var = ...` works inside `class Foo; ...; end`. Inside a module
body, or at top level, the codegen emits references to
`cvar_Toplevel_X` but the `collect_cvars` walker only descended
into `ClassNode` subtrees, so the corresponding `static <type>
cvar_Toplevel_X = ...;` declaration was never written and the C
compile failed with `use of undeclared identifier`.

```ruby
module Tep
  @@session_secret = ""           # collected? -> no, before fix
  def self.session_secret;     @@session_secret;     end
  def self.session_secret=(v); @@session_secret = v; end
end
```

Patch flips `collect_cvars` to start from the program root with
`class_idx = -1` (the Toplevel namespace) and lets `collect_cvars_in`
re-enter the class/module distinction itself: nested `ClassNode`
switches the namespace to the class's index, `ModuleNode` keeps
Toplevel, everything else recurses. Module-scope, top-level, and
class-body cvars all flow through the same registration path now.

Repro: `test/module_cvars.rb` (covers module-scope `@@var`,
read+write through `def self.x` accessors, and a bare top-level
`@@plain` for completeness).

### Why these came up

Building `tep` (a Sinatra clone targeting spinel-AOT) hits both
patterns naturally:

  - `Tep.session_secret = ENV.fetch("TEP_SESSION_SECRET")` is the
    canonical app-startup line.
  - `module Tep; @@session_secret = ""; ...` is the most idiomatic
    place to keep that secret.

We had workarounds (an instance var on a singleton, plus
`ENV[k] || default`), but figured shipping the spinel-side fixes
was cheaper than maintaining workarounds long-term.

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -4802,6 +4802,9 @@ class Compiler
           if mname == "[]"
             return "string"
           end
+          if mname == "fetch"
+            return "string"
+          end
         end
         if rcname == "Dir"
           if mname == "home"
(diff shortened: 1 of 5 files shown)


OriPekelman:
Addressed @gemini-code-assist's point on the set-path:

`test/env_fetch.rb` now also probes the env-set branch via `HOME` (always exported by POSIX `make` and in CI). The assertion is host-agnostic — "non-empty result that isn't the fallback string" — so the actual home value can differ per runner. A regression that inverted the getenv ternary would print `set=fallback-not-used` instead of `set=ok`.

A symmetric `unset=...` line catches the inverse mistake (always returning the env value regardless).

Pushed as 92594cd. Spinel `make test` 337/0/0.

PR #406, opened 2026-05-09
fix(codegen): String#index(sub, start) honours start offset

OriPekelman:
## `String#index(sub, start)` -- honour the start offset

The 2-arg form was silently dropping `start`: the codegen branch
emitted `sp_str_index(rc, compile_arg0(nid))` whether the call had
one or two args, so

```ruby
"a.b.c".index(".", 2)
```

returned `1` (the first dot) instead of `3`. The natural
"walk all matches" loop never advanced.

### Fix

  - New runtime helper `sp_str_index_from(s, sub, start)` in
    `sp_runtime.h`. Honours codepoint indexing (matches the 1-arg
    `sp_str_index`'s contract), supports negative start (clamps to
    0 when too negative, returns -1 when too positive).

  - Codegen branch peels off the 2-arg case before falling through
    to the existing 1-arg path:

    ```ruby
    if mname == "index"
      args_id_idx = @nd_arguments[nid]
      if args_id_idx >= 0
        a_idx = get_args(args_id_idx)
        if a_idx.length >= 2
          return "sp_str_index_from(" + rc + ", " + compile_expr(a_idx[0]) + ", " + compile_expr(a_idx[1]) + ")"
        end
      end
      return "sp_str_index(" + rc + ", " + compile_arg0(nid) + ")"
    end
    ```

### Coverage

`test/string_index_from.rb` exercises:
  - 1-arg baseline.
  - 2-arg with `start = 0`, on the dot itself, just past it,
    in/out of range.
  - Negative start (back from end), oversize-negative (clamp to 0).
  - Out-of-range positive start.
  - Multibyte string (codepoint, not byte indexing).
  - The "walk all dots" loop pattern -- builds the comma-joined
    list of all match positions in `header.payload.sig`.

### Where this came up

JWT parsing in tep needed to find both `.` separators in a token:

```ruby
dot1 = token.index(".")
dot2 = token.index(".", dot1 + 1)   # always returned dot1 -- bug
```

Pre-fix the verifier's `signing_input` was always just the header
(no payload), the HMAC didn't match, and every token verified as
"bad signature". The hand-rolled `find_dots` workaround in
`Tep::Jwt` can drop after this lands.


OriPekelman:
Addressed both points from @gemini-code-assist's review:

1. **Codegen claim**: the codegen change IS in the diff (`spinel_codegen.rb` lines 26028-26039 -- the new 2-arg branch that emits `sp_str_index_from`). The bot's per-line comment was anchored to `lib/sp_runtime.h:550` and it looks like it didn't pick up the spinel_codegen.rb hunk. No change needed.

2. **Optimisation**: real win, applied in 6b59bca. The codepoint counter now starts from `start` and walks from `s + boff` instead of recounting from 0; the "walk all matches" loop becomes O(N) total instead of O(N^2). Same regression test still passes.

Spinel suite still 351/0/0 with the optimisation.

PR #410, opened 2026-05-09
fix(codegen): Array#delete_at on user-class ptr_array

OriPekelman:
## `Array#delete_at(i)` on a `[Foo.new, ...]` array

Pre-fix the codegen had `delete_at` arms for `IntArray` and
`StrArray` but not for user-class `PtrArray`. A user-class
array could grow via `push` but never shrink, and the while-
drain pattern (`while arr.length > 0; arr.delete_at(0); end`)
hung the program because:

1. The `delete_at` call hit the unresolved-call warning and
   emitted 0 (the `arr.delete_at(0)` line was effectively a
   no-op).
2. Even if it did the right thing, `body_mutates_var?` -- the
   helper that decides whether `arr.length` can hoist outside
   a `while` -- knew about `push` / `pop` / `delete` / `clear`
   etc. but not `delete_at`, so `length` got hoisted and the
   loop spun forever.

### Fix (four small pieces)

1. **Runtime helper** (`lib/sp_runtime.h`):
   `sp_PtrArray_delete_at(a, i)` peer to the existing
   IntArray / StrArray versions. Returns the popped pointer
   or NULL on out-of-range. Negative indices count from end.

2. **Codegen dispatch**: ptr_array arm in
   `compile_poly_method_call` gains the `delete_at` case,
   mirroring the existing `push` / `[]` / `first` / `last`
   arms. Result cast to the element class so
   `v = arr.delete_at(0)` lowers cleanly.

3. **Return-type inference** (`infer_method_name_type`):
   `delete_at` already covered `str_array` → string,
   `float_array` → float; add `is_ptr_array_type(rt)` →
   `ptr_array_elem_type(rt)`. Without this the assignment
   target defaults to int and trips a
   `incompatible pointer to integer conversion` C-compile
   error.

4. **Loop-invariant analysis** (`body_mutates_var?`):
   `delete_at` joins `push` / `pop` / `shift` / etc. in the
   mutating-method set so `while arr.length > 0` no longer
   hoists `length` past the body's `delete_at`.

### Coverage

`test/ptr_array_delete_at.rb` (11 lines of expected output):
  - push / delete_at round trip
  - delete_at returns the popped instance + decrements length
  - middle-index delete (shifts later elements down)
  - negative index counts from end
  - out-of-range index returns nil-equivalent + array unchanged
  - **drain loop** (`while a.length > 0; a.delete_at(0); end`)
    must terminate -- pre-fix this hung due to the length hoist

### Where this came up

Building Tep::Scheduler (fiber-based cooperative scheduler for
the tep web framework: https://github.com/OriPekelman/tep) --
needed a slot-list of FiberSlot wrappers, and the natural
"clear" path is a `while length > 0; delete_at(0); end` drain.
With this fix the natural shape just works.


OriPekelman:
@gemini-code-assist the codegen changes ARE in this PR -- the diff stat shows `spinel_codegen.rb | 20 ++++++++++++++--` and `git diff master..fix/ptr-array-delete-at spinel_codegen.rb` shows the three hunks the description lists:

1. **Return-type inference** at line 4042-4049 (the `infer_method_name_type` `delete_at` branch gains the `is_ptr_array_type(rt)` arm).

2. **Codegen dispatch** at line 27760-27767 (new `delete_at` case in the ptr_array arm of `compile_poly_method_call`, mirroring `push` / `[]` / `first` / `last`).

3. **Loop-invariant analysis** at line 33464-33466 (`body_mutates_var?` adds `delete_at` to the mutating-method set so `while arr.length > 0` doesn't hoist `length` past a `delete_at` body).

The bot's per-line review was anchored to `lib/sp_runtime.h:372` and didn't surface the `spinel_codegen.rb` hunks. Same shape as the false-positive on #406. No change needed -- regression test passes, full spinel suite 354/0/0, downstream tep suite 159/0/4.

PR #458, opened 2026-05-13
fix(analyze,codegen): bridge no-recv IO; poly_array of procs; poly .call

ryanseys:
## Summary

Bundles three related fixes that the test fixture surfaces together:

1. **`puts` / `print` / `printf` in expression context** — last stmt of a `proc { ... }` body, a consumed `if`-arm, a ternary branch. `compile_no_recv_call_expr` had no symmetric path for these no-recv IO methods, so the call fell through to `warn_unresolved_call` and emitted `0` *without* firing the side-effect. Fix: bridge through `compile_io_call_stmt(nid, mname, -1)` for the whitelist (all three return nil in MRI), passing `"0"` up as the C expression value. Also extracts `compile_printf(nid)` next to `compile_print` so both dispatchers share the emitter.

2. **Heterogeneous array literals of procs** (`[proc { ... }, ...]`) were inferred as `int_array` and construction emitted `sp_IntArray_push(_, sp_proc_new(...))` — int-from-pointer C-compile error. Add a `proc`/`lambda` arm to `infer_array_elem_type_from_ids` in both `spinel_analyze.rb` and `spinel_codegen.rb` so such literals lower to `poly_array` and each element is boxed via `sp_box_proc`. Mirrors the existing `is_hash_type(et)` arm.

3. **Poly `.call(...)` dispatch.** When the per-element block param of `each` iterating a poly_array of procs binds with static type `poly`, `p.call` lands in `compile_poly_method_call`. Add a `mname == "call"` arm inside `emit_poly_builtin_dispatch` that fires for `cls_id == SP_BUILTIN_PROC`: unbox via `(sp_Proc *)recv.v.p` and invoke `sp_proc_call`. Lives **after** the user-class loop, so classes with their own `.call` method (e.g. functor objects, `test/poly_method_args.rb`) still resolve first.

## Test plan

- [x] `test/io_call_in_proc_body.rb` exercises all three together: `[proc { puts "a" }, proc { print "b\n" }, proc { printf("c=%d\n", 3) }].each { |p| p.call }`

## Out of scope

- **Symbol#to_proc (`&:call`)** — still TODO in spinel (`find_block_arg` returns `-1` for the SymbolNode shape, per the comment at `spinel_codegen.rb:1067`). The explicit-block form `{ |p| p.call }` is the closest equivalent that works today; the `&:sym` lowering is a separate concern.


ryanseys:
Thanks for the review. Both suggestions addressed:

**1. Test refactor to iterate procs.** Adopted in spirit — the test is now:

```ruby
[
  proc { puts "a" },
  proc { print "b\n" },
  proc { printf("c=%d\n", 3) }
].each { |p| p.call }
```

Used the explicit-block form rather than `&:call` because `Symbol#to_proc` (`&:sym`) is explicitly TODO in spinel — see the comment at `spinel_codegen.rb:1067`: "`&:sym` (SymbolNode) ... would need symbol-to-proc / nil-as-no-block lowering." Implementing `&:sym` is a separate concern; the explicit-block form is the closest equivalent that works today. Enabling it also required two adjacent fixes in this PR:

- Heterogeneous-array literals of procs → `poly_array` (analyzer + codegen), since `[proc, proc, proc]` previously inferred as `int_array` and `sp_proc_new` returned a pointer that didn't fit `mrb_int`.
- Poly `.call` dispatch — when the block param iterates a poly_array of procs, the recv's static type is `poly`; added a `SP_BUILTIN_PROC` arm in `emit_poly_builtin_dispatch` that unboxes via `(sp_Proc *)recv.v.p` and invokes `sp_proc_call`.

**2. Trailing newline on `.expected`.** The file already ends with `\n` — confirmed via `od -c`: `a\nb\nc=3\n` (10 bytes). The output of `printf("c=%d\n", 3)` is exactly `c=3\n` (one newline). Adding another newline would diverge from program output and the test would fail. Test passes as-is.

matz:
Manually merged in 9e645e2 (master) after local rebase to resolve conflict with the poly-recv .dup arm at the same emit_poly_builtin_dispatch site. Both arms coexist post-merge; CI was green on the original commit, archive + tests verified locally (424/0/0).

PR #459, opened 2026-05-13
fix(codegen): resolve qualified-constant arg in is_a?/kind_of?/instance_of?

ryanseys:
## Summary

`obj.is_a?(Mod::Klass)` previously fell through to `compile_expr(arg)` and emitted a bare C identifier (`Mod_Klass`) with no declaration — "use of undeclared identifier" at C compile time. The dynamic-arg guard in `compile_introspection_expr` only excluded the `ConstantReadNode` shape, so any qualified constant (`ConstantPathNode`) fell off the static-resolution path entirely. Compounding it: `compile_introspection_expr` only matched `is_a?` — neither `kind_of?` (alias) nor `instance_of?` (exact-class) were routed through it.

Fix:
- Extract `is_static_const_ref` and `resolve_introspection_arg_name` helpers.
- Widen the `is_a?` branch to also match `kind_of?` (shared ancestor-walk semantics in MRI).
- Add an `instance_of?` branch using exact-class comparison.

## Test plan

- [x] `test/introspection_constant_path.rb` covers: same-class match, unrelated-sibling, ancestor case (via `Mod::Klass < Mod::Base`), deeper constant path (`Mod::Sub::Deep`)

- Class-as-local (`k = Mod::Klass; obj.is_a?(k)`) — separate ConstantPathNode-to-local assignment bug.
- Built-in ancestor (`obj.is_a?(Object)`) — spinel's class registry doesn't materialize `Object` as a user-class ancestor.

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -17126,9 +17126,29 @@ class Compiler
     ""
   end
 
+  def is_static_const_ref(aid)
+    if @nd_type[aid] == "ConstantReadNode" || @nd_type[aid] == "ConstantPathNode"
+      return 1
+    end
+    0
+  end
+
+ # Resolve an is_a?/kind_of?/instance_of? argument to its registered
+ # class name. ConstantPathNode (`Foo::Bar`) walks the path via
+ # resolve_const_ref_name; ConstantReadNode (`Foo`) carries the leaf
+ # in @nd_name. Other shapes fall back to @nd_name so static lookup
+ # still has a name to try.
+  def resolve_introspection_arg_name(aid)
+    if @nd_type[aid] == "ConstantPathNode"
+      return resolve_const_ref_name(aid)
+    end
+    @nd_name[aid]
+  end
+
   def compile_introspection_expr(nid, mname, rc, recv_type)
- # is_a? - check class hierarchy
-    if mname == "is_a?"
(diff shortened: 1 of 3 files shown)


ryanseys:
Thanks for the review. The implementation is present in this PR — see `spinel_codegen.rb` (+51 / -9 in the file list). Specifically:

- `is_static_const_ref(aid)` @ ~17129 — new helper (1 if Constant{Read,Path}Node, else 0).
- `resolve_introspection_arg_name(aid)` @ ~17137 — new helper, walks ConstantPath via the existing `resolve_const_ref_name`; falls back to `@nd_name`.
- `compile_introspection_expr` @ ~17151 — widened the `is_a?` guard to `is_a? || kind_of?`; added an `instance_of?` branch using exact-class comparison; both routes through the new helpers.

`gh pr view 459 --repo matz/spinel --json files` confirms `spinel_codegen.rb` is in the diff.

The test-strengthening suggestion was also adopted in the amended commit: the fixture now covers inheritance (`Mod::Klass < Mod::Base`) and a deeper constant path (`Mod::Sub::Deep`). Two cases were documented as out-of-scope in the fixture header: class-as-local (`k = Mod::Klass; obj.is_a?(k)`) hits a separate ConstantPathNode-to-local assignment bug, and `obj.is_a?(Object)` returns false in spinel today because the class registry doesn't materialize `Object` as a user-class ancestor.

PR #460, opened 2026-05-13
fix(codegen): dispatch .call on anonymous proc receiver (incl. chained / ternary)

ryanseys:
## Summary

`proc { ... }.call` — where the receiver is the inline `proc { ... }` call rather than a local variable — fell through to `warn_unresolved_call` ("cannot resolve call to 'call' on proc"). `compile_dot_call_expr` only matched `LocalVariableReadNode` receivers.

Fix: after the LocalVariableReadNode special-case, fall through to a generic branch keyed on `infer_type(recv) == "proc"`. The check is node-shape-agnostic, so any receiver whose inferred type is `proc` routes through the same emitter.

## Test plan

- [x] `test/proc_anon_dot_call.rb` covers: anonymous proc literal, anonymous proc with args, method returning proc (factory pattern), ternary returning proc.

- `lambda { ... }.call` — separate `compile_lambda_call_expr` handler.
- `@ivar = proc { ... }; @ivar.call` — pre-existing ivar/proc codegen gap.
- `$global = proc { ... }; $global.call` — unverified inference.
- `[procs].each(&:call)` — Symbol#to_proc unsupported; addressed by rs-no-recv-io-expr-bridge with the explicit-block form.

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -13011,6 +13011,20 @@ class Compiler
           return "sp_proc_call(lv_" + rname + ", (mrb_int[]){" + ca + "})"
         end
       end
+
+ # Anonymous-proc-receiver dispatch — `proc { ... }.call`, or any
+ # expression chain whose inferred type is `proc` (e.g.
+ # `factory().call`). The LocalVariableReadNode branch above handles
+ # the named-proc case directly; this generic branch compiles the
+ # receiver expression and feeds it to sp_proc_call, using the same
+ # arg-array packing convention.
+      if infer_type(recv) == "proc"
+        ca_anon = compile_call_args(nid)
+        if ca_anon == ""
+          ca_anon = "0"
+        end
+        return "sp_proc_call(" + compile_expr(recv) + ", (mrb_int[]){" + ca_anon + "})"
+      end
     end
     ""
   end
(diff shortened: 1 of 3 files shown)


ryanseys:
Thanks for the review. The amended commit extends the test fixture to cover three additional receiver shapes that all funnel through my `infer_type(recv) == "proc"` branch — proving the fix is genuinely generic, not just for inline-anonymous procs:

- **CallNode receiver returning a proc** (factory pattern): `def factory; proc { show("factory") }; end; factory.call`
- **Ternary returning a proc**: `(cond ? proc { ... } : proc { ... }).call` (IfNode receiver with proc-unified arms)
- (already covered) anonymous proc with args: `proc { |n| show((n * 2).to_s) }.call(7)`

Documented as **out of scope** in the fixture header:

- **`lambda { ... }.call`** — lambdas have a separate `compile_lambda_call_expr` handler (~line 11941 in master); not exercised by this PR's `compile_dot_call_expr` fix.
- **`@ivar = proc { ... }; @ivar.call`** — pre-existing ivar/proc codegen gap (sp_Proc* doesn't fit the ivar slot today).
- **`$global = proc { ... }; $global.call`** — unverified inference for `$global` procs.
- **`[proc1, proc2].each(&:call)`** — Symbol#to_proc unsupported in master; addressed instead by rs-no-recv-io-expr-bridge with the explicit-block form `[procs].each { |p| p.call }`.

Each demoted case was verified empirically — attempting them triggers separate codegen issues that warrant their own PRs.

PR #461, opened 2026-05-13
fix(codegen): case/when on value-type class and qualified-constant arm

ryanseys:
## Summary

`case obj when ClassName` had two C-compile-blocking bugs when the predicate's class was a value-type instance (per-class free-list pool layouts where `lv_x` is `sp_X` rather than `sp_X *`):

1. **Pointer/value tmp mismatch.** `compile_case_stmt` and the case-as-expression branch in `compile_expr` always emitted the predicate tmp as `sp_<X> *`, regardless of whether the predicate's actual storage was a value type. The generated C failed with "initializing 'sp_X *' with an expression of incompatible type 'sp_X'; take the address with &".

2. **ConstantPathNode when-arm.** The static-resolution branch in `compile_when_conds` only matched the bare `ConstantReadNode` arg shape, so a qualified `when Mod::Klass` (`ConstantPathNode`) fell through to the poly-recv path and emitted `<undeclared identifier> Mod_Klass` against a non-poly predicate.

Fix:
- Route both tmp declarations through `c_type(pred_type)` so value-type classes get `sp_<X>` (no pointer) and standard heap classes get `sp_<X> *`. Assignment then matches the predicate's actual storage shape.
- Extend the const-shape guard in `compile_when_conds` to also match `ConstantPathNode`, resolving the qualified name via the existing `resolve_const_ref_name` helper.

## Test plan

- [x] `test/case_when_qualified_class.rb` exercises both fixes via a value-type module-namespaced class and a value-type plain class

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -10625,12 +10625,11 @@ class Compiler
         if pred_type == "string"
           emit("  const char *" + ptmp + " = " + pred_val + ";")
         elsif is_obj_type(pred_type) == 1
- # See compile_case_stmt — the temp must be the right pointer
- # type so compile_when_conds can resolve `when ClassName` to
- # a static match. .
-          bt = base_type(pred_type)
-          obj_cname = bt[4, bt.length - 4]
-          emit("  sp_" + obj_cname + " *" + ptmp + " = " + pred_val + ";")
+ # See compile_case_stmt — route the temp declaration through
+ # c_type so value-type classes get `sp_<X>` (no pointer) and
+ # standard heap classes get `sp_<X> *`. compile_when_conds then
+ # resolves `when ClassName` against the right shape.
+          emit("  " + c_type(pred_type) + " " + ptmp + " = " + pred_val + ";")
         else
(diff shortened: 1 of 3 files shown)


ryanseys:
The implementation is present in this PR — see `spinel_codegen.rb` (+21 / -24 in the file list). Specifically:

- `compile_case_stmt` @ ~22425-22431 — replaced hardcoded `sp_<X> *` tmp declaration with `c_type(pred_type)`, so value-type classes (per-class free-list pool layouts) get `sp_<X>` (no pointer) and standard heap classes keep `sp_<X> *`.
- `compile_expr` CaseNode-as-expression branch @ ~10627-10633 — same `c_type(pred_type)` fix applied to the case-as-expression call site.
- `compile_when_conds` @ ~22513-22524 — extended the const-shape guard from `ConstantReadNode` to also accept `ConstantPathNode`, routing the qualified name through the existing `resolve_const_ref_name` helper.

`gh pr view 461 --repo matz/spinel --json files` confirms `spinel_codegen.rb` is in the diff.

PR #462, opened 2026-05-13
fix(codegen): emit side-effect stmts in consumed if-expr branches

ryanseys:
## Summary

When an `if` is consumed for its value (RHS of assignment, return slot, ternary arm, etc.), `compile_if_expr` previously kept only the last statement of each branch — emitting `(cond ? lastexpr : ...)`. Any leading side-effect statements (`puts`, ivar writes, method calls, etc.) in the branch body were silently dropped.

  ```ruby
  x = if true
        puts "a"   # this was being dropped
        1
      end
  ```

Fix: detect multi-stmt branches and switch from the ternary form to a real `if (cond) { ... tmp = lastexpr; } else { ... tmp = lastexpr; }` that compiles each non-final statement with `compile_stmt`, assigns the final statement's expression to a temp, and returns the temp. Single-stmt branches stay on the ternary fast path. Elsif chains recurse; poly-unified branches keep the existing `box_value_to_poly` wrapping.

## Test plan

- [x] `test/if_expr_multi_stmt_body.rb` covers: multi-stmt then- and else-branches, elsif chain recursion, nested if-expr as method argument, mixed branch lengths, empty/missing else.

- **Poly type-unification of mixed concrete branches** (int/string). Spinel's inference picks one concrete type as the unified type rather than `poly`, so `tmp = 1LL;` is emitted against a `const char *` slot. The bug isn't in this PR's multi-stmt restructuring — the ternary fast path has the same inference issue. Separate fix.

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -20202,12 +20202,90 @@ class Compiler
     "sp_" + owner + "_" + sanitize_name(op) + "(" + cast + recv_c + ", " + rhs_c + ")"
   end
 
+  def if_branch_has_multi_stmt(body)
+    if body < 0
+      return 0
+    end
+    if get_stmts(body).length > 1
+      return 1
+    end
+    0
+  end
+
+ # Assign the value of an if-arm body to a temp, executing any
+ # leading side-effect statements before the assignment. Body
+ # convention: stmts[0..n-2] run for their side-effects;
+ # stmts[n-1] supplies the arm's value. Empty/missing body → "0".
+  def compile_if_arm_to_tmp(body, tmp, unified_t)
+    if body < 0
+      emit("  " + tmp + " = 0;")
+      return
+    end
+    stmts = get_stmts(body)
+    if stmts.length == 0
+      emit("  " + tmp + " = 0;")
+      return
+    end
+    i = 0
+    while i < stmts.length - 1
+      compile_stmt(stmts[i])
+      i = i + 1
+    end
+    last_id = stmts.last
+    val = compile_expr(last_id)
+    if unified_t == "poly" && infer_type(last_id) != "poly"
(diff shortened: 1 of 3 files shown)


ryanseys:
Thanks for the review. The implementation is present in this PR — see `spinel_codegen.rb` (+79 / -2 in the file list). Specifically:

- `if_branch_has_multi_stmt(body)` @ ~20205 — new helper (1 if body has >1 stmt, else 0).
- `compile_if_arm_to_tmp(body, tmp, unified_t)` @ ~20217 — new helper that compiles leading stmts via `compile_stmt` and assigns the final expr to `tmp`.
- `compile_if_expr` @ ~20245-20305 — added multi-stmt detection branch that emits a real `if (...) { } else { }` with a typed temp, returning the temp; the existing ternary fast path remains unchanged for single-stmt branches.

`gh pr view 462 --repo matz/spinel --json files` confirms `spinel_codegen.rb` is in the diff.

The test-strengthening suggestions are also adopted in the amended commit:

- **elsif chain recursion** — outer `if` uses multi-stmt path, inner `elsif` (recursive `compile_if_expr`) has multi-stmt branches.
- **Nested if-expression as method argument** — `puts(if true; puts "nested"; 10; end)` exercises `emit()` ordering (inner if-stmt prologue must land before outer call site).
- **Mixed branch lengths** — single-stmt then-branch, multi-stmt else.
- **Empty/missing else** — implicit nil → `tmp = 0;`.

**Poly type-unification with mixed concrete branches** was attempted but documented as out-of-scope: branches with int and string concrete types make spinel pick one concrete type as the unified type rather than `poly`, so `tmp = 1LL;` is emitted against a `const char *` slot — the same inference bug exists on the ternary fast path. Not introduced by this PR.

PR #466, opened 2026-05-13
Proc codegen fixes

DmitryBochkarev:
# Proc/Codegen Fixes

This branch fixes every code path needed for the `defer.rb` pattern to compile
and execute correctly — a deferred execution utility that stores procs in an
array and invokes them in LIFO order.

## The Motivating Example

```ruby
class Defer
  def initialize
    @blocks = []
  end

  def add(&blk)
    @blocks.push(blk)
    nil
  end

  def call
    i = @blocks.length - 1
    while i >= 0
      @blocks[i].call
      i -= 1
    end
    nil
  end
end

class Runner
  def self.with_deferred(&outer)
    d = Defer.new
    outer.call(d)
    d.call
    nil
  end
end

Runner.with_deferred do |d|
  i = 1
  d.add { puts "first #{i}" }
  i += 1
  d.add { puts "second #{i}" }
  puts "mid"
  i += 1
  d.add { puts "third #{i}" }
end

# Expected output:
#   mid
#   third 3
#   second 3
#   first 3
```

This 30-line example hits every broken code path. Running it through the
compiler without the fixes in this branch would produce one of: a C compile
failure, a silent no-op, or garbage output.

## What Was Broken

Spinel's proc pipeline had gaps at every layer — analysis, IR serialization, and
codegen — that silently dropped behavior without warnings or errors.

**Analysis.** Block parameters (`&block`) carried no type information from call
sites to proc bodies. A parameter `d` inside `do |d| ... end` that received a
`Defer` instance was typed `mrb_int` by default, making `d.add` unresolvable.
Empty arrays (`[]`) promoted to hold object types but not `proc`, so
`@blocks.push(blk)` left `@blocks` as `int_array` — elements typed `mrb_int`,
`.call` silently skipped.

**IR.** The block param analysis pass had no serialization path, so even correct
analysis results were lost between the analyzer and codegen phases.

**Codegen.** `.call` on a Proc was a silent no-op in both dispatch paths: the
poly dispatch table had no `SP_BUILTIN_PROC` arm, and direct dispatch only fired
for local-variable receivers (not `@ivar.proc.call` or `method_return.call`).
Class method dispatch never forwarded literal blocks, so `Runner.with_deferred
{ ... }` never received its block argument. Pointer arguments through
`sp_proc_call`'s compound literal failed C compilation because `Defer *` could
not implicitly convert to `mrb_int`. Free variable scanning missed variables
inside interpolated strings (`"first #{i}"`), nested proc capture state leaked
between sibling closures, and duplicate heap cells were allocated when inner
procs could reuse their enclosing proc's capture.


matz:
Merged locally to master in 28a8536 with one small fix on top (72744c2): removed a duplicate `SP_BUILTIN_PROC` arm in `emit_poly_builtin_dispatch` that conflicted with #458 (commit d9bf5e7, merged earlier today). Without that fix, every `.call` on a poly receiver was emitted twice and three tests failed with duplicated output (`io_call_in_proc_body`, `proc_array_push`, `proc_body_nested_block`).

All 431 tests pass and bootstrap is clean. Thanks for the careful work across analysis, IR, and codegen for the defer.rb pattern; the five new tests cover the regressions well.

Closing in favor of the merge commit.

PR #468, opened 2026-05-13
fix(codegen): Array#sum honours its init argument

o8n:
### Summary

`Array#sum(init)` silently dropped its `init` argument on the
IntArray / FloatArray dispatch paths. This PR threads `init`
through the runtime helpers, unboxes poly-typed `init` safely,
and widens the block-form accumulator to `mrb_float` for
FloatArray receivers so the sub-second precision (literally) is
not lost.

### Reproduction

```ruby
puts [1, 2, 3].sum(10)
puts [].sum(7)
puts [1, 2, 3].sum(10) { |x| x * 2 }
puts [1.5, 2.5].sum(0.5) { |fx| fx }
```

### Expected output

```
16
7
22
4.5
```

### Actual behavior (before this PR)

```
6
0
12
3.0
```

No warning. No compile error. The compiled binary just returns
the wrong number. A separate variant — passing a poly-typed
local (`p = [1, "x"][0]; [10].sum(p)`) — used to compile (init
ignored) but fails the moment `init` is forwarded into the typed
runtime helper, so the naive forwarding alone would have turned
a silent miscompile into a hard C build error.

### What was wrong

| Site | Pre-PR emission |
|---|---|
| `int_array.sum` | `sp_IntArray_sum(rc)` — runtime helper took no `init`, started accumulator at `0` |
| `float_array.sum` | `sp_FloatArray_sum(rc)` — same shape |
| `array.sum { ... }` (block form) | `compile_array_sum_block` emitted `mrb_int t = 0;` for every receiver, dropping init and truncating float receivers |

In all three sites the call's first argument was parsed but
never forwarded. Once we start forwarding it, two further
constraints surface:

- The static type of `init` may be `poly` (a heterogeneous
  array's element, an `ARGV` read, etc.). `compile_arg0`
  returns the raw `sp_RbVal` expression, which the typed
  helpers reject.
- A poly carrying an INT-tagged value cannot be unboxed
  through `.v.f` (the union slot read is bit-for-bit, not a
  coercion). The existing `sp_poly_to_f` runtime helper does
  the tag-dispatched coercion the rest of the codebase
  already uses for poly->float arithmetic.

### Fix

| File | Change |
|---|---|
| `lib/sp_runtime.h` | `sp_IntArray_sum` / `sp_FloatArray_sum` take `init` and seed the accumulator with it |
| `spinel_codegen.rb` | New helper `compile_arg0_as_float` — symmetric to the existing `compile_arg0_as_int`, but routes poly through `sp_poly_to_f` for tag-aware coercion (poly INT-tagged -> `(mrb_float)v.i`, poly FLT-tagged -> `v.f`, etc.) |
| `spinel_codegen.rb` | IntArray sum dispatch forwards `compile_arg0_as_int(nid)`; FloatArray sum dispatch forwards `compile_arg0_as_float(nid)`. No-arg call sites still get the literal `"0"` from the helpers' default, so `[1,2,3].sum` emits the same C as before |
| `spinel_codegen.rb` | `compile_array_sum_block` picks `mrb_float` + `compile_arg0_as_float` when `recv_type == "float_array"`, and keeps `mrb_int` + `compile_arg0_as_int` otherwise — this matches `infer_type`'s existing `float_array.sum -> "float"` answer end-to-end |

#### Runtime diff

```c
// before
static mrb_int sp_IntArray_sum(sp_IntArray*a){
  mrb_int s=0;
  for(mrb_int i=0;i<a->len;i++) s+=a->data[a->start+i];
  return s;
}
// after
static mrb_int sp_IntArray_sum(sp_IntArray*a, mrb_int init){
  mrb_int s=init;
  for(mrb_int i=0;i<a->len;i++) s+=a->data[a->start+i];
  return s;
}
```

`sp_FloatArray_sum` takes the symmetric `mrb_float init` change.

#### Codegen diff (final shape)

```ruby
# IntArray sum dispatch
- return "sp_IntArray_sum(" + rc + ")"
+ return "sp_IntArray_sum(" + rc + ", " + compile_arg0_as_int(nid) + ")"

# FloatArray sum dispatch
- return "sp_FloatArray_sum(" + rc + ")"
+ return "sp_FloatArray_sum(" + rc + ", " + compile_arg0_as_float(nid) + ")"

# compile_array_sum_block
- emit("  mrb_int " + tmp_sum + " = 0;")
+ if recv_type == "float_array"
+   emit("  mrb_float " + tmp_sum + " = " + compile_arg0_as_float(nid) + ";")
+ else
+   emit("  mrb_int " + tmp_sum + " = " + compile_arg0_as_int(nid) + ";")
+ end
```

### Test plan

`test/array_sum_init.rb` (17 assertions + `done`) covers:

| Group | Assertions |
|---|---|
| IntArray no-block, with init | `[1,2,3].sum(10) == 16` |
| IntArray no-block, no init regression | `[1,2,3].sum == 6` |
| IntArray no-block, empty + init | `[].sum(7) == 7`, `[].sum == 0` |
| IntArray no-block, negative init | `[1,2].sum(-5) == -2` |
| IntArray block, with init | `[1,2,3].sum(10) { |x| x*2 } == 22` |
| IntArray block, no init regression | `[1,2,3].sum { |x| x*2 } == 12` |
| IntArray block, empty + init | `[].sum(7) { |x| x*2 } == 7` |
| FloatArray no-block, with float init | `[1.5,2.5].sum(0.5) == 4.5` |
| FloatArray no-block, no init regression | `[1.5,2.5].sum == 4.0` |
| FloatArray no-block, int init implicit widening | `[1.5,2.5].sum(1) == 5.0` |
| FloatArray no-block, negative float init | `[1.5,2.5].sum(-0.5) == 3.5` |
| Poly int init on IntArray | `[1,2].sum([10,"x"][0]) == 13` |
| Poly float init on FloatArray | `[1.0,2.0].sum([1.5,"x"][0]) == 4.5` |
| Tag-mixed: poly-INT init on FloatArray | `[1.0,2.0].sum([1,"x"][0]) == 4.0` |
| FloatArray block + float init | `[1.5,2.5].sum(0.5) { |fx| fx } == 4.5` |
| FloatArray block + float-returning body | `[1.0,2.0].sum { |fy| fy * 1.5 } == 4.5` |

The FloatArray block cases use distinct param names (`fx` /
`fy`) so they don't share the function-scope local that earlier
`|x|` blocks declared — spinel hoists block params to a single
function-scope variable, an orthogonal pre-existing limitation
outside this PR's scope.

### Out of scope

| Item | Note |
|---|---|
| `int_array.sum(float_init)` returning float, with or without a block | Pre-existing miscompile. `[1, 2].sum(0.5)` returns `3` (CRuby: `3.5`); the same answer comes back from `[1, 2].sum(0.5) { |x| x }` because the block-form accumulator is also `mrb_int` for non-float receivers. `int_array.sum`'s return type is inferred as `int` and a float init is unboxed via `compile_arg0_as_int`. Promoting the result type when init is float (or when a poly init carries a FLT-tagged value) requires changes in `infer_type` and is left as a follow-up. |
| `float_array.sum { block }` returning Integer in CRuby | This PR widens the accumulator to `mrb_float` for FloatArray receivers, so the common case where the block returns a float (or where init is a float) now works. The corner case where the block returns Integer (e.g. `[1.0, 2.0].sum { _1.to_i }` -> CRuby `3` vs spinel `3.0`) gets a value that prints the same digits but is tagged float; choosing the result type from the block's return type rather than the receiver type requires a parallel change in `infer_type` and is left as a follow-up. |
| Tag-aware unbox for `compile_arg0_as_int` | The existing `compile_arg0_as_int` keeps its `.v.i` direct-access convention (matches every other callsite in the file). Switching it to `sp_poly_to_i` would widen the diff across already-merged dispatchers and is an orthogonal cleanup. |
| `String#sum` | Different method (bit-width arg, not init). Unchanged here. |

### Run results

- [x] `make` reaches the bootstrap fixpoint (`gen2.c == gen3.c`,
  `analyze.rb` and `codegen.rb` both IR + C OK)
- [x] `make test` — 385 pass, 0 fail, 0 error (the new
  `test/array_sum_init.rb` is included)
- [x] `make bench` — 49 pass, 1 fail (`bm_io_wordcount`),
  2 error (`bm_csv_process`, `bm_wordfreq`). The fail + 2 errors
  reproduce on master with no diff, unrelated to this PR.


matz:
Merged to master in f909396. Rebased locally on master (since the branch was conflicting); the only conflict was the comment indent style on the helper that lives next to the new `compile_arg0_as_float` — resolved by normalizing the new function's comment style to the surrounding HEAD convention. All 432 tests pass, bootstrap 4-way fixpoint OK. Thanks for the thorough writeup and the runtime/codegen split.

Issue #463, opened 2026-05-13
`Hash#fetch(key, {})` (hash default) on int-leaf hash emits mismatched-type ternary (sequel to #454)


rubys:
## Summary

`5e8269d` (#454) fixed `Hash#fetch(key, default)` with a **string** default on an int-leaf hash. The variant with a **hash** default (`fetch "k", {}`) still emits a type-mismatched ternary and fails C compile.

## Reproduction

```ruby
class P
  def self.lookup(params)
    params.fetch "article", {}
  end
end

P.lookup({ "x" => 1 })
```

```sh
$ spinel repro.rb -c -o repro.c
$ cc -fsyntax-only -I$SPINEL/lib repro.c
repro.c:N:N: error: incompatible pointer to integer conversion returning
  'sp_StrIntHash *' from a function with result type 'mrb_int' (aka 'long long')
```

## Actual

```c
static mrb_int sp_P_cls_lookup(sp_StrIntHash * lv_params) {
    return (sp_StrIntHash_has_key(lv_params, (&("\xff" "article")[1]))
            ? sp_StrIntHash_get(lv_params, (&("\xff" "article")[1]))   // mrb_int
            : sp_StrIntHash_new());                                     // sp_StrIntHash *
}
```

The ternary's two arms: `mrb_int` (from `sp_StrIntHash_get`) and `sp_StrIntHash *` (the empty-hash default). Function return type pinned to `mrb_int` (from the get arm). The hash-default arm fails the cast.

## Expected

Same shape as the string-default fix in `5e8269d`: detect the type mismatch at the fetch site and either widen the receiver to a variant that accommodates both (`StrPolyHash` here) or box both branches to `sp_RbVal` for unification.

## Passing variants

- **String default** (closed by #454 / `5e8269d`): `params.fetch "k", ""` on `Hash[String, Int]` — compiles cleanly.
- **Matched int default**: `params.fetch "k", 0` on `Hash[String, Int]` — compiles cleanly (no type mismatch).

The current break is symmetric to the string-default case but the fix path apparently didn't cover the `{}` literal default. Real-blog evidence: the string-default sites `params.fetch "id", "0"` (4 errors each in ArticlesController + CommentsController) all closed in `5e8269d`; the hash-default sites `params.fetch "article", {}` (1 each in ArticleParams.from_raw + CommentParams.from_raw, 2 errors total at lines 3478/3714) remain.

## Roundhouse-side note

[...]

Issue #456, opened 2026-05-13
Chained `hash[k1][k2] = v` on poly-valued outer hash: inner `[]=` emits Array arms only, Hash arms missing


rubys:
## Summary

Chained `outer_hash[k1][k2] = v` where `outer_hash` is a `Hash[String, untyped]` (StrPolyHash) emits **polymorphic `[]=` dispatch that enumerates only Array container types** (`POLY_ARRAY` / `PTR_ARRAY` / `INT_ARRAY`) for the inner setter. Hash container arms are absent. The String key passed as the index then fails C compile against the Array setters' `mrb_int` index parameter. Even at runtime, if the inner value is a Hash (the common case for nested-form params), the setter is a silent no-op — none of the emitted arms match.

`is_a?(Hash)` narrowing on the previous line doesn't fire (sibling to closed #438 / #448): both with and without an `is_a?(Hash)` guard, the same Array-only dispatch is emitted.

## Reproduction

```ruby
def assign(into)
  into["x"]["y"] = "z"
end

assign({"x" => {"y" => "old"}})
```

```sh
$ spinel repro.rb -c -o repro.c
$ cc -fsyntax-only -I$SPINEL/lib repro.c
repro.c:N:N: error: incompatible pointer to integer conversion passing
  'char *' to parameter of type 'mrb_int' (aka 'long long')
```

Three errors at three Array setter call sites.

## Actual

```c
static inline mrb_int sp_assign(sp_StrPolyHash * lv_into) {
    sp_RbVal _t1 = sp_box_str((&("\xff" "z")[1]));
    if (sp_StrPolyHash_get(lv_into, (&("\xff" "x")[1])).cls_id == SP_BUILTIN_POLY_ARRAY)
        sp_PolyArray_set((sp_PolyArray *)sp_StrPolyHash_get(lv_into, (&("\xff" "x")[1])).v.p,
                         (&("\xff" "y")[1]), _t1);          // ← String index → mrb_int slot
    else if (... == SP_BUILTIN_PTR_ARRAY)
        sp_PtrArray_set(..., (&("\xff" "y")[1]), _t1.v.p);  // ← same
    else if (... == SP_BUILTIN_INT_ARRAY)
        sp_IntArray_set(..., (&("\xff" "y")[1]), _t1.v.i);  // ← same
    return 0;
}
```

[...]

PR #472, opened 2026-05-13
fix(analyze,codegen): emit Hash arms in poly-recv .fetch

### LINKED ISSUE #470: Poly-recv `.fetch` dispatch on `sp_RbVal` emits only user-class arms, missing built-in Hash variants (sibling to #456)
## Summary

Polymorphic `.fetch` dispatch on an `sp_RbVal` receiver enumerates only user-defined classes that define a `fetch` method, missing built-in Hash variants (`StrIntHash`, `StrStrHash`, `StrPolyHash`, `SymPolyHash`, etc.). At the source level the receiver is always a Hash, but the emitted dispatch can never match a Hash `cls_id` at runtime — so `.fetch` silently returns nil/empty regardless of the actual hash contents. Inverse of #456 (which added Hash arms to poly-recv `[]=` but not to other Hash methods).

Surfaces post-#463 in real-blog: after the hash-default fix boxed `lv_raw_sub` to `sp_RbVal`, the downstream `sub.fetch "title", ""` site dispatches over `sp_RbVal` cls_ids and only finds user-class arms (`ActionDispatch::Flash#fetch`, `ActionDispatch::Session#fetch`) — both of which were defined for sym-keyed hash-like semantics, not as general Hash#fetch substitutes. The mismatched parameter types (Flash/Session take sym keys; the dispatch passes a String) cascade into 16 C compile errors in real-blog.

## Reproduction — silent-wrong runtime

```ruby
class FlashLike
  def fetch(key, default)
    "from-flash"
  end
end

# Keep FlashLike#fetch alive.
puts FlashLike.new.fetch("x", "y")

def get_raw(h)
  h["x"]
end

def lookup(h)
  sub = get_raw(h)             # sub: sp_RbVal (StrPolyHash lookup)
  sub.fetch("title", "DEFAULT")
end

# At runtime, sub is the inner Hash {"title" => "real"}.
puts lookup({ "x" => { "title" => "real" } })
```

```sh
$ ruby repro.rb
from-flash
real

$ spinel repro.rb -o repro && ./repro
from-flash

```

CRuby prints `real`; spinel prints an empty line. The Hash lookup silently returns nil because the dispatch matched no arm.

## Actual emit

```c
static inline const char * sp_lookup(sp_StrPolyHash * lv_h) {
    sp_RbVal lv_sub = sp_get_raw(lv_h);
    sp_RbVal _t2 = lv_sub;
    sp_RbVal _t3 = sp_box_nil();                  // ← initial empty
    if (_t2.tag == SP_TAG_OBJ) {
        if (_t2.cls_id == 21)                      // FlashLike only
            _t3 = sp_box_str(sp_FlashLike_fetch(..., "title", ""));
    }
    // No Hash arms. At runtime sub is a StrStrHash, no arm matches,
    // _t3 stays sp_box_nil(). Returned as ""/empty.
    return _t3;
}
```

The dispatch enumerates user-class cls_ids that define `fetch`, but the canonical Hash variants (whose `sp_<Hash>_get` / `sp_<Hash>_fetch_or` already implement the semantics) are absent.

## Expected

Hash arms should appear in the dispatch alongside any user-class arms. Roughly:

```c
if (_t2.tag == SP_TAG_OBJ) {
    if (_t2.cls_id == SP_BUILTIN_STR_INT_HASH)
        _t3 = sp_StrIntHash_has_key(...) ? sp_box_int(sp_StrIntHash_get(...)) : sp_box_str(default);
    else if (_t2.cls_id == SP_BUILTIN_STR_STR_HASH)
        _t3 = sp_StrStrHash_has_key(...) ? sp_box_str(sp_StrStrHash_get(...)) : sp_box_str(default);
    else if (_t2.cls_id == SP_BUILTIN_STR_POLY_HASH)
        _t3 = sp_StrPolyHash_has_key(...) ? sp_StrPolyHash_get(...) : sp_box_str(default);
    // ... and so on for SYM_*_HASH variants
    else if (_t2.cls_id == 21)
        _t3 = ... user-class arms
}
```

Parallel to `4859894` (#456) which did this for poly-recv `[]=` with statically-String key.

## Real-blog cc error trace

```
/tmp/blog.c:3486: error: passing 'char *' to parameter of type 'mrb_int' (Flash#fetch key slot is sym)
/tmp/blog.c:3487: error: passing 'char *' to parameter of type 'mrb_int' (Session#fetch key slot is sym)
/tmp/blog.c:3489: error: assigning to 'const char *' from incompatible type 'sp_RbVal' (lv_raw_title was pinned to const char *)
/tmp/blog.c:3490: error: initializing 'sp_RbVal' with an expression of incompatible type 'const char *'
```

Per-site, 4 errors. Same pattern repeated for the `body` key and for the Comment side: 16 errors total in real-blog at HEAD `2f820bf`.

## Roundhouse-side note

Surfaces in `app/models/article_params.rb` (and `comment_params.rb`):

```ruby
def self.from_raw(params)
  raw_sub = params.fetch "article", {}    # #463 boxed result to sp_RbVal
  sub = if raw_sub.is_a?(Hash)
    raw_sub                                # still RbVal
  else
    {}
  end
  raw_title = sub.fetch "title", ""       # ← this issue
  ...
end
```

The shape was suppressed pre-#463 (the outer fetch failed C compile first). After #463's hash-default fix, the outer fetch now compiles, and the inner fetch's missing-Hash-arms surfaces.

Cluster root for 16 of 16 real-blog spinel-AOT errors at HEAD `2f820bf` (the only remaining bucket).

## Related dispatch gaps (warning-only, same family)

Spinel currently warns "cannot resolve" for several Hash methods on poly variants:

```
warning: cannot resolve call to 'delete' on str_poly_hash (emitting 0)
warning: cannot resolve call to 'delete' on sym_poly_hash (emitting 0)
warning: cannot resolve call to 'dup'    on sym_poly_hash (emitting 0)
warning: cannot resolve call to 'fetch'  on sym_poly_hash (emitting 0)
warning: cannot resolve call to 'merge'  on sym_poly_hash (emitting 0)
warning: cannot resolve call to 'merge'  on sym_str_hash  (emitting 0)
```

These are the same family — Hash-method codegen incomplete for poly-valued / mixed-value variants. Mentioning here as related; happy to file separately if matz prefers per-method tickets.

rubys:
## Summary
- Closes #470. Polymorphic `.fetch` dispatch on an `sp_RbVal` receiver enumerated only user-class arms (`def fetch`) and missed every built-in Hash variant (`StrIntHash` / `StrStrHash` / `StrPolyHash`). At runtime a Hash receiver matched no arm and the result temp stayed at its nil default — silent miscompile, not a C compile failure.
- Sibling to 4859894 (#456), which added Hash arms to poly-recv `[]` / `[]=` when the key is statically a String. Same shape, same gating rationale.
- Surfaces in real-blog: after #463 boxed `lv_raw_sub` to `sp_RbVal`, the downstream `sub.fetch "title", ""` site dispatched over `sp_RbVal` cls_ids and only found user-class arms (`ActionDispatch::Flash#fetch`, `ActionDispatch::Session#fetch`, both sym-keyed), cascading to 16 C compile errors.

## Changes
- **analyze**: `fetch` on a poly recv now widens to `poly`, matching the `[]` poly-recv widening. Without this the result temp's static C type comes from the first user-class `fetch` return (e.g. `const char *`) and the int-valued Hash arm's `sp_box_int(...)` rhs is a C type mismatch.
- **codegen**: `emit_poly_builtin_dispatch` gains a `fetch` block parallel to the existing `[]` Str-Hash arms. Each arm tests `_has_key` first, returns the boxed value on hit, or the boxed default on miss. Default is boxed via `box_value_to_poly` when its source type is concrete. Sym-keyed variants are deferred for the same `@needs`-gating reason as `[]`.
- **test**: `test/poly_recv_fetch_hash_arms.rb` mirrors the issue's repro (FlashLike user class + outer StrPolyHash carrying an inner StrStrHash).

Before:
```c
if (_t2.tag == SP_TAG_OBJ) {
  if (_t2.cls_id == 21) _t3 = sp_FlashLike_fetch(...);   // FlashLike only
}
// Runtime cls_id == SP_BUILTIN_STR_STR_HASH: no match, _t3 stays nil.
```

After:
```c
if (_t2.tag == SP_TAG_OBJ) {
  if (_t2.cls_id == 21) _t3 = sp_box_str(sp_FlashLike_fetch(...));
  if (_t2.cls_id == SP_BUILTIN_STR_INT_HASH) _t3 = sp_StrIntHash_has_key(...) ? sp_box_int(sp_StrIntHash_get(...)) : sp_box_str("DEFAULT");
  if (_t2.cls_id == SP_BUILTIN_STR_STR_HASH) _t3 = sp_StrStrHash_has_key(...) ? sp_box_str(sp_StrStrHash_get(...)) : sp_box_str("DEFAULT");
  if (_t2.cls_id == SP_BUILTIN_STR_POLY_HASH) _t3 = sp_StrPolyHash_has_key(...) ? sp_StrPolyHash_get(...) : sp_box_str("DEFAULT");
}
```

## Test plan
- [x] Issue's repro now prints `from-flash\nreal` (matching CRuby), was `from-flash\n`.
- [x] `make test`: 435 pass / 0 fail / 0 error (added 1 new regression test, 434 pre-existing pass).
- [x] New emit verified by hand against issue's "Expected" sketch.


matz:
Merged to master in 260f36c. Rebased locally since the branch needed a small touch-up over #469's parent-walking owner_idx in the narrowed-int gating arm (the conflict was a simple replacement of `i` -> `owner_idx`; the new arm_incompat check is preserved). All 439 tests pass and bench 55/55. Thanks!

PR #495, opened 2026-05-14
fix(analyze): preserve nil state for scalar ivars

ydah:
### Summary

Instance variables initialized with `nil` and later assigned numeric values can be generated as plain numeric fields, making `nil?` checks invalid.

### Repro

```ruby
@hit_material_index = nil

# later
@hit_material_index = obj.material_index

return background if @hit_material_index.nil?
```

### Actual

The ivar was generated as an integer field initialized to `0`, and the `nil?` control flow did not behave like Ruby.

### Expected

Spinel should either support nilable ivar types or preserve correct `nil?` semantics for mixed nil/numeric ivars.

### DIFF
--- a/spinel_analyze.rb
+++ b/spinel_analyze.rb
@@ -7560,6 +7560,15 @@ class Compiler
     end
   end
 
+  def scalar_ivar_type?(t)
+    t == "int" || t == "float" || t == "bool" || t == "symbol"
+  end
+
+  def nil_scalar_ivar_mix?(old_type, new_type)
+    (old_type == "nil" && scalar_ivar_type?(new_type)) ||
+      (new_type == "nil" && scalar_ivar_type?(old_type))
+  end
+
   def update_ivar_type(ci, iname, new_type)
     names = @cls_ivar_names[ci].split(";")
     types = @cls_ivar_types[ci].split(";")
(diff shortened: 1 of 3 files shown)


matz:
Heads-up: I had to revert this merge at e45e735 to get CI green. The optcarrot integration job (ubuntu/gcc, added on 2026-05-10 in 0f66aaf) fails since this PR with `-Werror=int-conversion` on `self->iv_step += 1`, `self->iv_timer += self->iv_freq` and similar APU::Pulse / Triangle / Noise ivar arithmetic.

**Root cause**

The new `scan_writer_calls` branch routes plain `@ivar = nil` literal writes through `update_ivar_type` when the current slot is scalar. Optcarrot's `@wave_length` follows the pattern

```ruby
def initialize(apu)
  @wave_length = nil          # init
end
def reset
  @wave_length = 0 if @wave_length    # later
end
```

Pre-this-PR, `scan_writer_calls` had `elsif at != "nil" && ...` and silently skipped the nil branch, so `@wave_length` settled at `int`. With this PR, the nil write reaches `update_ivar_type`, `nil_scalar_ivar_mix?` widens `@wave_length` to `poly`, and downstream ivars like `@freq = (@wave_length + 1) * 2 * @fixed` cascade to poly, breaking the int arithmetic emit at every `iv_freq` / `iv_timer` / `iv_step` site.

**The bug this PR fixed is still real**

The `@hit_material_index = nil` then `@hit_material_index = obj.method` then `@hit_material_index.nil?` pattern from the original description does need poly tracking to compile correctly. The local `make test` suite passed because none of the test fixtures exercised the optcarrot-style "ivar declared nil but body is pure int arithmetic" path. I learned today that `make optcarrot` isn't part of `make test` and didn't run it locally before merging.

**Proposed narrower trigger**

Widen scalar-and-nil mixes to poly only when the ivar is actually read via `.nil?` somewhere in the program (or `== nil` / pattern-matched against nil). That gates `path 4`'s new nil branch on `cls_ivar_nil_checked?(ci, iname)`. Quick AST scan over CallNodes with `nil?` as the method name and an InstanceVariableReadNode receiver should be enough.

That keeps the test cases (`HitState`, `ResetState`, `ParamResetState`) working — they all use `.nil?` explicitly — while leaving optcarrot's `@wave_length` (used only via truthiness `if @wave_length` and as int arithmetic) at the cheaper `int` storage.

Happy to take this on as a follow-up if you'd rather not redo the PR — but flagging in case you want to chase it (probably faster from your end given the context).

PR #499, opened 2026-05-14
fix(codegen): emit (NULL, NULL) for block-ABI trailing args when call site has no block

ryanseys:
## Summary
Instance methods whose body uses `yield` get a C signature with two trailing slots — `void (*_block)(..., void*)` and `void *_benv`. When such a method also declares `&block`, calling it without a literal block at the call site dropped the trailing `_block, _benv` pair, tripping clang's `too few arguments to function call`.

Threads a `yargs` parameter through `build_call_tail` and adds `cls_call_yargs(ci, midx)` which returns `, NULL, NULL` when `cls_method_has_yield` is set. Both class-instance method dispatch paths (no-recv bare-method-in-class and `recv.method`) now append the yield-ABI suffix when the callee's body uses `yield`. The literal-block path (`compile_yield_call_stmt`) bypasses the C call entirely by inlining the body, so the suffix is only relevant when no literal block is present at the call site — `NULL, NULL` is always safe in that case.

## Test plan
- [x] New regression test `test/block_optional_unused.rb` (`def m(text = nil, &block); ...; yield if block; end` called without a block) compiles and matches the reference Ruby output
- [x] `make bootstrap` stays at fixpoint
- [x] Full test suite passes

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -9224,10 +9224,17 @@ class Compiler
   end
 
  # Builds the trailing portion of a call-args list — each non-empty
- # piece prefixed with ", ", empties skipped. Returns "" when both
+ # piece prefixed with ", ", empties skipped. Returns "" when all
  # are empty. Mirrors build_params_str: callers concatenate the
  # result onto a self/recv prefix to form the full arg list.
-  def build_call_tail(ca, bp)
+ #
+ # `yargs` carries the trailing yield ABI args (`_block, _benv`) when
+ # the callee declares `yield` but the call site doesn't pass a
+ # literal block. Without it, methods with both `&block` and `yield`
+ # in their body get a 3-arg call against a 5-arg signature and clang
+ # fails the build. Mirrors the `, NULL, NULL` shape that the
+ # top-level no-recv path appends in compile_no_recv_call_expr.
+  def build_call_tail(ca, bp, yargs = "")
     result = ""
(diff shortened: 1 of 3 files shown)


ryanseys:
Both concerns addressed in 4a4d854:

**1. No-recv (internal) call path coverage** — added `Greeter#call_internal` which makes a bare `greet("internal")` call inside the class body. This routes through `compile_no_recv_call_expr` (line 12956) and confirms `cls_call_yargs` is wired into both dispatch paths.

**2. Block forwarding via `&block_arg`** — you were right to flag this; investigation showed it would have silently no-op'd. `compile_yield_stmt` previously emitted `if (_block) _block(args..., _benv)` unconditionally, so when a call site forwards via `&block` the captured proc landed in `lv_block` but `yield` only consulted the (NULL) yield-ABI slots and dropped the block.

Extended `compile_yield_stmt` to fall back to `sp_proc_call(lv_<block_param>, ...)` when the enclosing method declares `&block`. This matches the existing precedent in `block_given?` lowering (which already prefers `lv_block`) and reflects Ruby's semantics that `yield` and `block.call` drive the same proc.

Test now covers all three cases (`Greeter.new.greet("hi")`, `Greeter.new.call_internal`, `forwarder { ... }`), and `make bootstrap` reaches the 4-way C/IR fixpoint with all 468 tests passing.

PR #568, opened 2026-05-17
fix(runtime): raise Errno::EISDIR for directory file I/O

ydah:
## Summary

Make `File.read`, `File.binread`, and `File.write` raise `Errno::EISDIR` when the target path is a directory.

## Expected

Directory paths should fail with `Errno::EISDIR`, matching CRuby behavior.

For example:

```ruby
begin
  File.read(".")
rescue => e
  puts e.class
end

begin
  File.binread(".").bytes
rescue => e
  puts e.class
end

begin
  File.write(".", "hello")
rescue => e
  puts e.class
end
```

should produce:

```text
Errno::EISDIR
Errno::EISDIR
Errno::EISDIR
```

## Actual

The runtime did not check whether the path was a directory before file I/O.

`File.read` and `File.binread(...).bytes` could treat a directory as empty content, making a directory path indistinguishable from an empty file.

`File.write` ignored the failed open and returned as if the write succeeded, so writing to a directory could appear successful.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -1743,8 +1743,35 @@ static mrb_bool sp_file_file(const char *path) {
    and `\r\n` becomes `\r\r\n`. fread's actual byte count drives
    null-termination because text mode shrinks the byte count below
    ftell's raw-file size. */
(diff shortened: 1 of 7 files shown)


matz:
#569 merged first and brought in `sp_file_directory` (with Windows wide-char support via `GetFileAttributesW`) plus the `<sys/stat.h>` include and `S_ISDIR` macro fallback. This PR now conflicts on those lines.

Could you rebase to drop the duplicate helper / include / macro from this PR and keep only the `Errno::EISDIR` raise checks inside `sp_file_read`, `sp_file_write`, and `sp_file_binread_bytes`? The use-sites already call `sp_file_directory(path)` which is now defined upstream. Tests should be unchanged.

Thanks for the fix.

ydah:
Rebased `fix-file-directory-io-errors` onto current `upstream/master` in f6a59a7.

The duplicate `<sys/stat.h>` / `S_ISDIR` / `sp_file_directory` additions from this PR are gone now; the branch uses the `sp_file_directory(path)` implementation merged via #569 and keeps only the `Errno::EISDIR` checks in:

- `sp_file_read`
- `sp_file_write`
- `sp_file_binread_bytes`

PR #569, opened 2026-05-17
feat(runtime): add File.directory? and File.file? predicates

ydah:
## Summary

Add `File.directory?` and `File.file?` so compiled programs can reliably distinguish directories from regular files.

## Expected

Applications should be able to detect directory and regular-file paths before doing file I/O, matching CRuby predicate behavior.

For example:

```ruby
path = "spinel_file_predicate_test.txt"
File.write(path, "x")

puts File.directory?(".")
puts File.directory?(path)
puts File.file?(path)
puts File.file?(".")
puts File.directory?("spinel_missing_predicate_path")
puts File.file?("spinel_missing_predicate_path")

File.delete(path)
```

should produce:

```text
true
false
true
false
false
false
```

## Actual

Spinel did not lower `File.directory?` or `File.file?` as supported File class methods.

Calls to these predicates fell through to the unresolved class-method path and produced falsey `0` values, so compiled applications could not reliably distinguish a directory like `.` from other unsupported or missing paths.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -1699,8 +1699,42 @@ static mrb_int sp_catch_val[SP_CATCH_STACK_MAX];
 static volatile int sp_catch_top = 0;
 static void sp_throw(const char *tag, mrb_int val) { int i = sp_catch_top - 1; while (i >= 0) { if (strcmp(sp_catch_tag[i], tag) == 0) { sp_catch_val[i] = val; sp_catch_top = i + 1; longjmp(sp_catch_stack[i], 1); } i--; } fprintf(stderr, "uncaught throw: %s\n", tag); exit(1); }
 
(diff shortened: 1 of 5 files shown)


ydah:
Addressed the review comments in f06d433.

- Added a Windows-specific predicate path using `GetFileAttributesW` with UTF-8 to UTF-16 conversion, avoiding `stat()` limitations for UTF-8 paths and directory paths with trailing slashes.
- Renamed `sp_file_regular` to `sp_file_file` and updated the `File.file?` codegen mapping accordingly.
- Added `File.directory?("./")` coverage for the trailing-slash directory case.

Verified:
- `make build/test-results/file_directory_predicates.ok build/test-results/file_class_methods_int_recv.ok`
- `make bootstrap`

Issue #636, opened 2026-05-20
Tep::Server::Scheduled segfaults under sustained concurrent HTTP/1.1 keep-alive burst (SIGSEGV in recv+dispatch path)


OriPekelman:
A `Tep::Server::Scheduled`-hosted app segfaults under sustained
concurrent HTTP/1.1 keep-alive bursts. Stand-alone "hello world"
route -- no PG, no other batteries. The default prefork-blocking
server (`Tep::Server`) handles the same workload fine; the bug
is specific to the cooperative scheduler path.

## Minimal repro

`repro.rb`:

```ruby
require_relative "../lib/tep"
set :scheduler, :scheduled

get '/hi' do
  "hello"
end
```

```sh
bin/tep build repro.rb -o /tmp/repro
/tmp/repro -p 4985 &
ruby -e '
  require "net/http"; require "uri"
  uri = URI("http://127.0.0.1:4985/hi")
  2.times.map { Thread.new {
    Net::HTTP.start(uri.host, uri.port) { |h|
      100.times { h.request(Net::HTTP::Get.new(uri.path)) }
    }
  } }.each(&:join)
'
```

## Observed

Server processes ~60-80 requests successfully across the two
persistent-keep-alive TCP connections, then dies with SIGSEGV.
Exit code 139 (SEGV_ACCERR).

Strace tail (the segfault happens right after a normal-looking
recvfrom; the next expected syscall is the response write):

```
recvfrom(5, "GET /hi HTTP/1.1\r\nAccept-Encodin"..., 4096, 0, NULL, NULL) = 131
sendto(5, "HTTP/1.1 200 OK\r\nContent-Type: t"..., 102, 0, NULL, 0) = 102
sendto(5, "hello", 5, 0, NULL, 0) = 5
rt_sigprocmask(SIG_BLOCK, NULL, [], 8) = 0
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
... ~50 more clean recv+send pairs ...
recvfrom(5, "GET /hi HTTP/1.1\r\nAccept-Encodin"..., 4096, 0, NULL, NULL) = 131
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_ACCERR, si_addr=0xb37bc3df54e4} ---
+++ killed by SIGSEGV (core dumped) +++
```

The fault address `0xb37bc3df54e4` is a high heap-shaped pointer
that's been corrupted by a few bytes -- typical of either a UAF
[...]

PR #641, opened 2026-05-21
fix(runtime): walk sp_fiber_root.saved_roots in suspended-fiber GC (#636 follow-up)

OriPekelman:
The post-#636 fix (`c7d247b`) moved per-fiber GC roots into a save/restore model on every `Fiber#resume` but missed walking `sp_fiber_root`'s saved region. When a `Fiber.new`'d fiber is current and a GC pass fires, main's roots — stashed in `sp_fiber_root.saved_roots` by the resume side — are never marked. Main's live references get freed, including the `Fiber*` objects themselves (held only via main's locals). The next walk of `sp_fiber_list_head` dereferences a stale `sp_Fiber*` and segfaults at a high-pointer address: the same shape #636 was meant to fix.

## Empirical bisect

The bug is reachable via tep's Scheduled-server burst load. The original repro from #636 (2 fibers × 100 keepalive requests) now runs clean post-fix, but a slightly different shape (1 fiber × 200 keepalive requests, plus higher-concurrency variants) SEGVs consistently after 60–90 requests. Bisect against [`OriPekelman/tep@test/spinel_scheduled_burst_segv_repro.rb`](https://github.com/OriPekelman/tep/blob/main/test/spinel_scheduled_burst_segv_repro.rb):

| spinel | tep case A (1 conn × 200 keepalive) |
|---|---|
| `6513d2d` (pre-fix) | ok=200/200 |
| `1bbf3d7` (commit before #636) | ok=200/200 |
| `c7d247b` (#636 fix) | **ok=87, SEGV** |
| `c7d247b` + this patch | ok=200/200 |

Higher-load variants (4×50, 8×100, 200 fresh non-keepalive, 1000 short-conn churn) all crash on `c7d247b` and all run clean with the patch.

## Fix

`sp_mark_suspended_fibers` walks `sp_fiber_root.saved_roots` unconditionally (when `sp_fiber_root != sp_fiber_current`), in addition to iterating `sp_fiber_list_head`. Mirrored in both the Windows and the Linux/ucontext variants.

## Note on the test

I tried to build a self-contained Ruby regression test that reproduces the bug, but spinel's 2-generation GC (full sweep every 8 cycles) shelters the small synthetic test cases — promoted survivors aren't swept on partial cycles. The tep integration repro is the most reliable trigger I could land on. Happy to iterate on a tighter standalone test if there's a preferred shape.

Refs #636.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -2297,7 +2297,13 @@ static void sp_fiber_save_roots(sp_Fiber*f){if(f->saved_roots_cap<sp_gc_nroots){
 static void sp_fiber_restore_roots(sp_Fiber*f){if(f->saved_nroots>0)memcpy(sp_gc_roots,f->saved_roots,sizeof(void**)*f->saved_nroots);sp_gc_nroots=f->saved_nroots;}
 static void sp_fiber_list_add(sp_Fiber*f){f->fiber_prev=NULL;f->fiber_next=sp_fiber_list_head;if(sp_fiber_list_head)sp_fiber_list_head->fiber_prev=f;sp_fiber_list_head=f;}
 static void sp_fiber_list_remove(sp_Fiber*f){if(f->fiber_prev)f->fiber_prev->fiber_next=f->fiber_next;else if(sp_fiber_list_head==f)sp_fiber_list_head=f->fiber_next;if(f->fiber_next)f->fiber_next->fiber_prev=f->fiber_prev;f->fiber_prev=NULL;f->fiber_next=NULL;}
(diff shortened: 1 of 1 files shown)


OriPekelman:
Thanks — extracted `sp_mark_fiber_roots` as a static helper in both platform variants, per the suggestion. Cross-platform-block duplication (most of the Fiber section is duplicated identically between the Windows and Linux `#ifdef` branches) is an existing structural issue I deliberately kept out of scope here. Happy to do that as a separate PR if it's wanted. Re-ran the A–F probe against the refactored runtime: A=200/200, C=200/200, D=200/200, E=800/800, F=1000/1000 — semantically identical.

Issue #630, opened 2026-05-20
Destructuring (parallel assignment) a [UserClass, UserClass] return value segfaults


OriPekelman:
## Repro

```ruby
class Mat
  attr_accessor :nrows, :ncols
  def initialize(nrows, ncols)
    @nrows = nrows
    @ncols = ncols
  end
  def matmul(other)
    Mat.new(@nrows, other.ncols)
  end
end

def fwd(w, x)
  y = w.matmul(x)
  z = w.matmul(x)
  [y, z]
end

w = Mat.new(3, 4)
x = Mat.new(4, 1)
y, z = fwd(w, x)             # ← segfaults at runtime
puts y.nrows
```

Compiles cleanly. Segfaults at runtime (exit 139).

Replacing the destructure with explicit indexing works fine:

```ruby
result = fwd(w, x)
puts result[0].nrows         # 3 — fine
puts result[1].nrows         # 3 — fine
```

So Spinel emits good code for `Array<UserClass>` returns and index access, but the parallel-assignment destructure path is the broken one.

## Why I tripped over this

Building a small LoRA training step where a method naturally returned two intermediate `Mat`s (forward output + a cached activation needed for backward). Worked around by exposing two methods and recomputing.

Spinel HEAD `1e2394d`.

PR #648, opened 2026-05-22
fix(codegen): defer top-level CONST init when RHS reads a main local (#647)

### LINKED ISSUE #647: Top-level `CONST = recv.method` hoists init above local-var assignment — reads zero-struct, returns 0 (or SIGSEGV when call chain dereferences via FFI)
Sibling of #630 (multi-write fix in 9ac50a6), same shape, single target.

When a top-level `CONST = recv.method` is preceded by `recv = ...`,
codegen emits the const init *before* the local-var assignment runs.
The const init reads from a zero-initialized struct, so the constant
ends up wrong (silently 0 for value chains), or the program segfaults
if the call chain reaches FFI / C code that dereferences a pointer
field of the zero-init struct.

Confirmed on Spinel `8ff3a4e` (HEAD of matz/spinel master at time of
report). macOS arm64, Apple clang from Xcode CLT 26.x.

## Pure-Ruby repro (no deps)

```ruby
class Cfg
  attr_reader :nested
  def initialize(n); @nested = n; end
end

class Inner
  attr_reader :v
  def initialize; @v = 42; end
end

inner = Inner.new
cfg = Cfg.new(inner)
CFG_VOCAB = cfg.nested.v
puts "vocab=" + CFG_VOCAB.to_s
```

Expected: `vocab=42`
Actual: `vocab=0`

## Generated C confirms the hoist

`spinel repro.rb -S`:

```c
static mrb_int cst_CFG_VOCAB = 0;

int main(int argc,char**argv){
    sp_Inner lv_inner = {0};
    sp_Cfg lv_cfg = {0};
    sp_Inner _t1 = lv_cfg.iv_nested;   // ← reads from zero-init lv_cfg
    cst_CFG_VOCAB = _t1.iv_v;          // ← stores 0
    lv_inner = sp_Inner_new();         // ← real init lands here
    lv_cfg = sp_Cfg_new(lv_inner);
    /* ... puts ... */
}
```

The pre-body const-init loop plants `cst_CFG_VOCAB = lv_cfg.iv_nested.iv_v;`
at the top of `main`, before either local has been assigned. With the
structs still zero-filled, the chain yields zeros.

## Crash variant — same shape via FFI

When the const init's RHS reaches FFI / C code that dereferences a
pointer field of the zero-init struct, the program SIGSEGVs at startup
instead of just emitting a bogus value.

Hit in `OriPekelman/toy` while wiring `examples/04_serve_http.rb`:

```ruby
require_relative "../lib/toy_smollm2_loader"  # provides SmolLM2ConfigLoader

cfg   = SmolLM2ConfigLoader.read(GGUF)
gguf  = TinyNN.tnn_gguf_load(GGUF)
KV    = SmolLM2KVFFICache.new
KV.realize_for_mmap(gguf, cfg, 512, flags.untied, flags.qkv_bias)
CFG_VOCAB = cfg.vocab           # ← triggers the hoist
puts "loaded " + GGUF + " — vocab=" + CFG_VOCAB.to_s
```

Crash: `EXC_BAD_ACCESS (code=1, address=0x8)` at `ldr x8, [x8, #0x8]`
in `main + 4460`. Exit 139, no output produced — the hoisted const init
runs before the very first `puts`.

## Workaround

Assign through a local, not directly to a constant. Either of these
works:

```ruby
cfg_vocab = cfg.vocab           # local var — no hoist
CFG_VOCAB = cfg_vocab           # RHS is now a local read, not a method call
```

```ruby
# or, equivalently, inline at use site
puts "loaded " + GGUF + " — vocab=" + cfg.vocab.to_s
```

## Likely fix region

The same shape as #630 (multi-write fix in 9ac50a6) — an
`@multi_const_inits` push happened unconditionally, and codegen planted
the init before locals were assigned. The single-target analogue of
that pre-body loop needs the same guard: skip when the RHS reads from
a local that hasn't been assigned yet, and let the real assignment in
source order land the value.

OriPekelman:
## What

Defer the emit for top-level `CONST = recv.method` initializers whose
RHS reads a main-scope local. Without this, codegen's const-init
pre-loop plants `cst_X = lv_recv.iv_z;` at the top of `main`, before
`lv_recv = sp_Recv_new()` runs — so the chain either silently returns
0 (plain attr_reader chain) or SIGSEGVs when the call reaches FFI.

Closes #647.

## Repro

```ruby
class Cfg; attr_reader :nested; def initialize(n); @nested = n; end; end
class Inner; attr_reader :v; def initialize; @v = 42; end; end
inner = Inner.new
cfg = Cfg.new(inner)
CFG_VOCAB = cfg.nested.v
puts CFG_VOCAB.to_s
```

Before: `vocab=0`
After:  `vocab=42`

Generated C before:

```c
int main(int argc,char**argv){
    sp_Inner lv_inner = {0};
    sp_Cfg lv_cfg = {0};
    sp_Inner _t1 = lv_cfg.iv_nested;   // ← reads zero-init lv_cfg
    cst_CFG_VOCAB = _t1.iv_v;          // ← stores 0
    lv_inner = sp_Inner_new();
    lv_cfg = sp_Cfg_new(lv_inner);
    ...
}
```

Generated C after:

```c
int main(int argc,char**argv){
    sp_Inner lv_inner = {0};
    sp_Cfg lv_cfg = {0};
    lv_inner = sp_Inner_new();
    lv_cfg = sp_Cfg_new(lv_inner);
    sp_Inner _t1 = lv_cfg.iv_nested;
    cst_CFG_VOCAB = _t1.iv_v;          // ← in source order now
    ...
}
```

## Approach

Sibling of #630 (multi-write fix in 9ac50a6). There, the push into
`@multi_const_inits` was skipped entirely when no target was a
constant. For single-target writes the const still has to be declared
(it's a global), so we keep the registration but split the emit:

1. New helper `expr_reads_main_local(nid, lnames)` in `spinel_codegen.rb`
   — a recursive AST walker mirroring `scan_for_argv`; returns 1 if
   the expression rooted at `nid` reads any `LocalVariableReadNode`
   whose name is one of `main`'s declared locals.

2. `emit_main` (right before the const-init pre-loop) builds a
   `defer_const_init[]` flag list parallel to `@const_names`. A
   top-level const (scope `""`) is marked deferred when its RHS reads
   any main local.

3. The const-init pre-loop skips flagged entries.

4. The body-statement loop normally skips `ConstantWriteNode`
   (because the pre-loop handled it). When the target is flagged, we
   compile the RHS in place and emit `cst_X = <val>;` (+ `SP_GC_ROOT`
   for pointer types) at the in-source position — by then the local
   it depends on has been assigned.

## Scope

Only top-level constants whose RHS actually reads a main local change
emit ordering. Class constants, constants over literal RHS, constants
over method-call-on-constant RHS, and the existing multi-write /
self-reference handling all keep their current paths.

## Test

* `make test` — **625 pass / 0 fail / 0 error**, same as master.
* Hand repro (above) prints `vocab=42` after, `vocab=0` before.
* Surfaced while wiring an HTTP serving example in
  [OriPekelman/toy](https://github.com/OriPekelman/toy) (the FFI
  crash variant from #647); that binary now boots and handles
  requests after the fix.

## Diff

```
spinel_codegen.rb | 105 ++++++++++++++++++++++++++++++++++++++++++++++++++++--
1 file changed, 103 insertions(+), 2 deletions(-)
```

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -10669,6 +10669,44 @@ class Compiler
     end
   end
 
+ # Return 1 if the AST rooted at `nid` reads any LocalVariableReadNode
+ # whose name appears in `main_lnames` (main()'s declared locals).
+ #
+ # Used by emit_main to detect single-target top-level constants whose
+ # initializer depends on a local that hasn't been assigned yet at the
+ # const-init pre-loop. Without this guard, codegen plants
+ #   cst_X = lv_y.iv_z;
+ # at the top of main, before `lv_y = sp_Y_new()` runs — so the chain
+ # reads from a zero-init struct (silently returns 0, or SIGSEGVs when
+ # the call reaches FFI). matz/spinel#647; sibling of #630 (multi-write
+ # fix in 9ac50a6).
+  def expr_reads_main_local(nid, main_lnames)
+    if nid < 0
+      return 0
+    end
+    if @nd_type[nid] == "LocalVariableReadNode"
+      nm = @nd_name[nid]
+      li = 0
+      while li < main_lnames.length
+        if main_lnames[li] == nm
+          return 1
+        end
+        li = li + 1
+      end
+      return 0
+    end
+    cs = []
+    push_child_ids(nid, cs)
+    k = 0
+    while k < cs.length
+      if expr_reads_main_local(cs[k], main_lnames) == 1
+        return 1
+      end
+      k = k + 1
+    end
+    0
+  end
+
  # Walk the AST looking for class-hierarchy usage patterns
  # (`<klass>.superclass`, `.ancestors`, `<`, `<=`, dynamic
  # is_a? / kind_of? / instance_of?, case-when on a Class const).
@@ -11088,13 +11126,42 @@ class Compiler
       j = j + 1
     end
 
+ # Compute deferred const inits: a top-level `CONST = recv.method`
+ # whose RHS reads from a main local must wait for the body to run,
+ # otherwise the pre-emit below plants `cst_X = lv_y.iv_z;` before
+ # `lv_y = sp_Y_new()` and reads from a zero-init struct
+ # (matz/spinel#647). The single-target sibling of #630 (9ac50a6
+ # multi-write fix): there the push was skipped entirely; here the
+ # const must still be declared but the init emit moves to source
+ # order, handled in the body-stmt loop below.
+    defer_const_init = "".split(",")
+    dci = 0
+    while dci < @const_names.length
+      defer_const_init.push("0")
+      dci = dci + 1
+    end
+    dci = 0
+    while dci < @const_names.length
+      scope_dci = ""
+      if dci < @const_scope_names.length
+        scope_dci = @const_scope_names[dci]
+      end
+      if scope_dci == "" && dci < @const_expr_ids.length
+        eid_dci = @const_expr_ids[dci]
+        if eid_dci >= 0 && expr_reads_main_local(eid_dci, lnames) == 1
+          defer_const_init[dci] = "1"
+        end
+      end
+      dci = dci + 1
+    end
+
  # Constants (initialize global declarations)
     i = 0
     while i < @const_names.length
  # expr_id == -1 means this constant came from a `A, B, C = expr`
  # multi-write: its value is computed by the deferred init block
  # below, not from a single per-target expression.
-      if i < @const_expr_ids.length && @const_expr_ids[i] >= 0
+      if i < @const_expr_ids.length && @const_expr_ids[i] >= 0 && defer_const_init[i] != "1"
         old_scope = @current_lexical_scope
         if i < @const_scope_names.length
           @current_lexical_scope = @const_scope_names[i]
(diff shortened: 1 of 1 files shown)


matz (commented):
LGTM. Reviewed approach against the related #630 multi-write fix (9ac50a6) — single-target sibling looks consistent.

**Logic correctness**:
- `expr_reads_main_local` recursively walks the AST via `push_child_ids`, matches `LocalVariableReadNode` names against main's declared locals. Standard pattern, mirrors `scan_for_argv`. ✅
- `defer_const_init[]` is parallel to `@const_names`; only top-level consts (scope `""`) with main-local RHS get flagged. ✅
- Pre-loop skips flagged entries; body loop emits them in source order with the matching `SP_GC_ROOT` for pointer slots. Pre-loop still runs first for everything else, so any cross-const dependencies in non-deferred consts stay correct. ✅

**Edge cases checked**:
- `A = lv.x` (deferred) → `B = A` (non-deferred) chain: B sees `cst_A = 0` at its pre-loop emit point. Not handled here, but out of #647 scope — same shape as the existing pre-loop semantics. Could be a separate follow-up.
- Class const + same-named top-level const: body loop's `cwn_idx` lookup gates on `scope == ""`, picks the top-level entry correctly. ✅

**One nit**: no regression test in `test/` for the repro. The repro from the PR body would slot in cleanly as `test/const_init_defers_for_main_local.rb` (with `.rb.expected` `vocab=42`). Happy to add it post-merge if easier.

**CI**: ubuntu gcc/clang and macos clang green. windows-mingw failure was the `file_write_poly_arg` test's `/tmp/...` path tripping on MinGW — already fixed on master in `d36bd93`. I've triggered `update-branch` so the PR's CI re-runs against current master; should go green.

Merge once CI is green.

Issue #646, opened 2026-05-21
Reading top-level constant from within its own assignment's RHS segfaults instead of raising NameError


OriPekelman:
When a top-level constant `APP = SomeClass.new` is being assigned, and `SomeClass#initialize` (transitively) reads `APP` to dispatch a method on it, the read should either raise `NameError` (MRI semantics — the constant isn't bound yet) or return `nil`. Spinel does neither: depending on the dispatch shape it either emits a warning + uses `0`, or it returns the partially-initialized object pointer and the caller deref'd it, segfaulting.

## Real-world hit

Filed because tep tripped this while wiring Battery 2's PG-backed broadcast (OriPekelman/tep#24). The chain:

```ruby
module Tep
  class App
    def initialize
      # ... other init ...
      @broadcast_pg_conn = PG::Connection.new("")  # ← segfaults here
      # ... rest of init ...
    end
  end
  APP = App.new   # ← APP being constructed; APP not bound yet
end

module PG
  class Connection
    def initialize(opts)
      # ...
      if opts.is_a?(String)
        if Tep::Scheduler.scheduled_context?
          # ...
        else
          # ...
        end
      end
    end
  end
end

module Tep
  module Scheduler
    def self.scheduled_context?
      Tep::APP.sched_current >= 0   # ← Tep::APP being constructed
    end
  end
end
```

A minimal `tep` app (just `require "lib/tep"` + a single `get '/' do "ok" end`) produces:

```
Segmentation fault (core dumped) /tmp/probe_app -p 4985
```

at module load, before the server binds. Removing the `@broadcast_pg_conn = PG::Connection.new("")` line (and seeding via a top-level setter call AFTER `APP = App.new` completes) eliminates the crash.

## What I think is happening

Spinel allocates the `APP` slot up-front, zero-initializes it, then runs the initializer in-place on that slot. While `App#initialize` is running, `APP` points to the partially-initialized struct. The `Tep::APP.sched_current` access reads `@sched_current` at its expected offset — which is currently zero (initializer hasn't gotten there yet). `Scheduler.scheduled_context?` returns `0 >= 0` → true. The branch picks the async path. async_connect then walks further into Tep::APP's not-yet-initialized fiber/scheduler arrays, dereferencing zero or partially-initialized PtrArray slots → segv.

## Synthetic repro doesn't crash

[...]

PR #687, opened 2026-05-24
[codex] Remove obsolete issue 646 plan

ledsun:
## Summary

Remove the issue #646 implementation plan from docs. The tracked direct-shape case (`CONST = Class.new` with reads during initialization) appears to be covered by the landed guard and regression test, while the remaining implementation context is documented in code comments and the test.

I am not certain whether removing this is the preferred project policy, and I am not trying to claim that every possible generalized const-initialization shape is handled. It looked like this planning note may simply have been left behind after the tracked fix landed, so I opened this PR as a cautious cleanup proposal. If the project intentionally keeps issue-specific planning documents around, or wants this plan kept for follow-up edge cases, please feel free to close it without concern.

## Validation

Docs-only change; no tests run.


ledsun:
Closing this since the plan still appears useful for documenting follow-up const-initialization edge cases that are not fully covered by the direct-shape fix.

PR #747, opened 2026-05-25
feat(gem): add Spinelfile + path-gem loadpath resolution (Phase 0)

neidiom:
spinel-gem reads a Spinelfile (path: gems only) and writes a Spinelfile.loadpaths sidecar. spinel_parse learns --loadpaths=FILE and searches those dirs before <exe_dir>/lib. The driver walks up from the source file to discover the sidecar, so `require "foo"` in any nested .rb finds the gem.

Also fixes a latent bug where a top-of-file plain `require` was silently dropped after the first splice rebuilt the buffer -- the bootstrap clause only fired when result == source, then never again.

Tests: 11 unit + 6 E2E + 1 parser regression; new make test-gems.


OriPekelman:
Personally, I believe anything like this would like an issue first? 

Not very clear to me what a Spinelfile is. It looks like a Gemfile ? If it is why not use `Gemfile` and with the existing `ruby` directive? `ruby '3.x', :engine => 'spinel', :engine_version => '0.0.x'` This way, for the time being, no design choices need to be made. Bundler will fail loudly, as it should. At any rate this seems a tiny-bit premature ?  For the time being almost any gem not built specifically to support Spinel would not work out of the box - and the "ecosystem" of Spinel libs is still very early stage as far as I can see.

matz:
Reverted in 9e5101a.

The gem-discovery / loadpath strategy is still under discussion in #925 — Phase 0 was merged before that conversation concluded, exposing the `Spinelfile` DSL, `spinel-gem` CLI, and `Spinelfile.loadpaths` sidecar format publicly. Unresolved design questions that motivated the revert:

- Whether gem loadpaths should shadow stdlib (current PR placed them first)
- Whether `Spinelfile.loadpaths` is a committed lockfile or generator-only artifact (no `.gitignore` guidance shipped)
- The full Spinelfile DSL surface (`spinel`, `source`, `gem`, `group`) becomes a compatibility commitment once landed

Apologies for the churn — please feel free to revisit once #925 settles on a direction.

The standalone `spinel_parse.c` top-of-file `require` bug fix that was bundled here is also reverted; if it's still hitting users it can be re-landed independently as a small patch with just the parser fix + regression test.

PR #1001, opened 2026-05-27
fix: normalize negative start in slice_range before computing length

neidiom:
Array and string slicing with a negative start index computed the slice length from the raw (negative) start value before it was normalized by _slice(), producing too many elements:

  a = [0,1,2,3,4]; a[-4..-2]  =>  [1,2,3,4]  (wrong)
  CRuby:                        =>  [1,2,3]    (correct)

The _slice_range helpers already normalized negative end_ against len but left start negative, so n = end_ - start + 1 overflowed. Fix all five variants (IntArray, FloatArray, StrArray, str_sub_range_r, str_sub_range_len_r) to normalize start the same way end_ is normalized before computing n.


matz:
Merged via rebase onto master (ee28ce9) — CI failures were the at_exit_lifo/warn_kernel cases fixed in 115fb80. Rebased and merged. Thanks!

matz:
Gemini's review flagged a double-adjustment, but applying the suggested clamp (`if (start < 0) start = 0;` after `start += a->len;`) would introduce a real bug.

Traced through the merged code with `a = [0,1,2,3,4]; a[-6..-2]`:

- end_ = -2 + 5 = 3, start = -6 + 5 = -1 (still negative)
- n = 3 - (-1) + 1 = 5, but the `if (n<0 || start<0) n=0` guard fires → n becomes 0
- sp_IntArray_slice(a, -1, 0) → start is re-adjusted to 4 inside, but len=0 immediately returns the empty array

Result: `[]` for an out-of-range negative start. CRuby returns `nil` here; the subset returns the empty array, but the index semantics are correct.

Gemini's suggested clamp would change the result to `[0, 1, 2, 3]` (clamping a -6 start to 0), which is plainly wrong vs CRuby. The existing `n=0` guard already handles the over-negative case correctly, so the PR is fine as merged — no update needed.

Verified locally:
```
[0,1,2,3,4][-4..-2]   => [1, 2, 3]   (CRuby match)
[0,1,2,3,4][-6..-2]   => []          (CRuby: nil; subset empty)
[0,1,2,3,4][-100..-2] => []          (CRuby: nil; subset empty)
```

PR #1004, opened 2026-05-27
fix: defined?() returns nil for never-written globals

neidiom:
CRuby returns nil for defined?() when  has never been assigned, but Spinel always returned "global-variable" since the analyzer registered every gvar read (including bare references in defined? checks) into @gvar_names with no write tracking.

Add @gvar_written flag array that records whether each global variable has a write node in the program. The analyzer sets the flag on write and exports it through the IR; the codegen's defined? handler checks it before returning "global-variable".

### DIFF
--- a/spinel_analyze.rb
+++ b/spinel_analyze.rb
@@ -581,6 +581,7 @@ class Compiler
  # Global variables ($x)
     @gvar_names = "".split(",", -1)
     @gvar_types = "".split(",", -1)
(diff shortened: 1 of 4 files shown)


matz:
Merged via rebase onto master (9a3bcaa) — CI failures were the at_exit_lifo/warn_kernel cases fixed in 115fb80. Rebased on top of master and merged. Thanks!

PR #1026, opened 2026-05-28
runtime: add frozen-array guard to FloatArray#pop/#shift and StrArray…

neidiom:
…#pop

IntArray#pop and IntArray#shift already call sp_raise_frozen_array() before mutating a frozen array. FloatArray and StrArray variants were missing the check, silently allowing mutation on frozen arrays where MRI raises FrozenError.


matz:
Superseded by #1029, which adds frozen-array guards to all mutating methods (IntArray/FloatArray/StrArray reverse!/rotate!/sort!/shuffle!/pop/shift/delete/delete_at/insert), a superset of this PR's FloatArray#pop/#shift + StrArray#pop. Thanks for the fix - merging the broader one to avoid overlapping edits on the same lines.

PR #1029, opened 2026-05-28
runtime: add frozen-array guard to all mutating methods

neidiom:
IntArray#pop/#shift already had the sp_raise_frozen_array() check, but many other mutating methods across all array types were missing it. MRI raises FrozenError when calling any mutation method on a frozen Array. Added the guard to:

IntArray: reverse!, rotate!, sort!, shuffle!
FloatArray: reverse!, rotate!, sort!, shuffle!, pop, shift StrArray: reverse!, rotate!, sort!, shuffle!, pop, delete_at,
          delete, insert
PolyArray: shuffle!, rotate!


matz:
Thanks! Merged. Addressed the two review points from the bot in follow-up 424a1e1: added the missing `slice!` frozen guard across Int/Float/Str/Poly/Ptr arrays, and flattened the `if(!a||a->frozen){if(a&&a->frozen)...;return;}` guards into `if(!a)return;if(a->frozen){...;return;}`. Regression test test/array_frozen_slice_bang.rb added.

Issue #1027, opened 2026-05-28
Codegen: string interpolation of `sp_String *`-returning method emits `%lld` (pointer-as-integer) instead of `%s`


rubys:
## Summary

A `#{...}` interpolation whose expression is a method call returning `sp_String *` (the mutable-string type backing `s = String.new; s << ...; s`) emits the pointer cast to `long long` and formats it with `%lld`. The interpolated chunk renders as a decimal pointer address instead of the string's contents.

Direct `<<` on the same return value works correctly via `sp_String_append`. Only the string-interpolation path is broken.

## Reproduction

```ruby
module M
  def self.make
    s = String.new
    s << "hello"
    s
  end
end

x = "[#{M.make}]"
puts x
```

```
$ spinel -E repro.rb
[4331330144]
```

Expected `[hello]`.

## Generated C

```c
static inline sp_String * sp_M_cls_make(void) {
    SP_GC_SAVE();
    sp_String * lv_s = NULL;
    SP_GC_ROOT(lv_s);
    lv_s = sp_String_new();
    sp_String_append(lv_s, "hello");
    SP_GC_RESTORE();
    return lv_s;
}
```

Function signature is correct (`sp_String *` return). The interpolation site:

```c
lv_x = sp_sprintf("[%lld]", (long long)sp_M_cls_make());
```

Should emit something along the lines of:

```c
lv_x = sp_sprintf("[%s]", sp_String_cstr(sp_M_cls_make()));
```

(`sp_String_cstr` already exists at `lib/sp_runtime.h:1804`.)

## Symmetric working case (no interpolation)

```ruby
io = String.new
io << "["
io << M.make
io << "]"
puts io
```

This emits `sp_String_append(lv_io, sp_M_cls_make())` directly — works correctly, prints `[hello]`. So the type information *is* available; the gap is specifically in the StringInterp format-spec dispatch not recognizing `sp_String *` alongside `const char *` (and the various boxed string variants).

## Context

[...]

PR #1030, opened 2026-05-29
runtime: stop sp_sprintf truncating long interpolations at 4095 bytes

rubys:
## Problem

Every string interpolation (`"...#{x}..."`) compiles to `sp_sprintf` (`compile_interpolated` → `sp_sprintf("fmt", …)`). `sp_sprintf` formats into a fixed `char _sp_tmp[4096]` stack buffer and then **clamps** the length:

```c
int _sp_n = vsnprintf(_sp_tmp, sizeof(_sp_tmp), fmt, ap);
if (_sp_n >= (int)sizeof(_sp_tmp)) _sp_n = (int)sizeof(_sp_tmp) - 1;  // silent truncation
char *b = sp_str_alloc(_sp_n); memcpy(b, _sp_tmp, _sp_n);
```

So any interpolation that produces ≥ 4096 bytes is silently cut to 4095 — the head survives, everything past 4095 (including the closing of the string) is dropped. Direct `String#<<` concatenation is unaffected (it doesn't route through `sp_sprintf`); only interpolation hits the cap.

```ruby
big = "A" * 5000
puts "[#{big}]".length   # => 4095 (expected 5002)
```

## Fix

`vsnprintf` already returns the *would-be* length, so re-render into a correctly-sized heap buffer (via `va_copy`) for the overflow case. The common `< 4096` path keeps the stack-temp + `memcpy` fast path, so typical interpolations do no extra formatting work. `sp_str_alloc(n)` provides `n` usable bytes + NUL, so `vsnprintf(b, n+1, …)` fits exactly.

## Test

`test/sprintf_long_interp.rb` checks the length and both endpoints of a 5002-byte interpolation (a mid-string clamp keeps the head but drops the closing bracket). Boundary-verified locally at 4095 → 4097, 4096 → 4098, and 100000 → 100002.

## Provenance

Surfaced in [roundhouse](https://github.com/rubys/roundhouse) view emit: a lowering that coalesces `io << a; io << b; io << c` append runs into a single `io << "...#{}..."` produces interpolations larger than 4096 bytes, which truncated rendered HTML pages mid-content. Complements #1027/#1028 (the other string-interpolation arms).


rubys:
Good catch — adopted in ac3af79. Re-arming the `va_list` with `va_start` in the slow path (instead of `va_copy` on every call) keeps the common <4096-byte fast path free of any `va_copy` cost and drops the extra `ap2`. Re-running `va_start`/`va_end` after the first matched cycle is standard C (C11 §7.16.1) and portable. Repro test + 4095/4096/4097/100000-byte boundaries still pass.

PR #1051, opened 2026-05-29
codegen: GC scan walks full ancestor chain, not just immediate parent (#1050)

### LINKED ISSUE #1050: GC scan function only unions ivars one inheritance level — grandparent ivars dropped, causing use-after-free
## Summary

A class's generated GC scan function only marks ivars from **the class itself and its
immediate superclass** — not the transitive set over all ancestors. When a passthrough
intermediate class sits in the chain, a grandchild's scan function **drops ivars
introduced two-or-more levels up**. Because instances are allocated with the leaf
class's scan function, those grandparent-introduced ivars are never marked, so the GC
frees the objects they hold while the instance is still live → use-after-free.

This is distinct from #1031 (fiber capture-struct NULL scan hook, fixed in `08eaa4b`);
that fix is present in my toolchain and does not cover this.

## Minimal repro (deterministic via `GC.start`, no fibers/threads/IO)

```ruby
class A
  def payload=(v); @payload = v; end   # @payload introduced HERE (grandparent)
  def payload; @payload; end
end
class B < A          # passthrough middle class, introduces no ivars of its own
end
class C < B          # leaf; introduces its own @own
  def own=(v); @own = v; end
  def own; @own; end
end

def churn(n)
  s = ""; i = 0
  while i < n; s = s + i.to_s + ","; i += 1; end
  s.length
end

obj = C.new
obj.payload = "secret_" + 42.to_s        # heap string, only reference is the inherited ivar
obj.own     = "kept_" + 7.to_s
GC.start                                  # @payload collected: C's scan fn doesn't mark it
churn(8000)                               # reuse the freed slot
GC.start
puts "own="     + obj.own                 # => kept_7      (C's own ivar — scanned, survives)
puts "payload=" + obj.payload             # => (empty!)    (grandparent ivar — dropped, freed)
```

**Expected:** `own=kept_7` / `payload=secret_42`
**Actual:** `own=kept_7` / `payload=` — the grandparent ivar's value was collected.

### Confirmation it's the scan function

`spinel repro.rb -c` shows the generated scan functions:

```c
static void sp_A_gc_scan(void *p){ sp_A *self=(sp_A*)p; if(self->iv_payload) sp_gc_mark((void*)self->iv_payload); }
static void sp_B_gc_scan(void *p){ sp_B *self=(sp_B*)p; if(self->iv_payload) sp_gc_mark((void*)self->iv_payload); }
static void sp_C_gc_scan(void *p){ sp_C *self=(sp_C*)p; if(self->iv_own) sp_gc_mark((void*)self->iv_own); }   // <-- iv_payload dropped
```

`A` (defines it) and `B` (one hop) mark `iv_payload`; `C` (two hops, via empty `B`)
does not. The `sp_C` struct still *declares* `iv_payload` — only the scan is missing.

### Controls / scope

- **Negative control:** comment out both `GC.start` calls → prints `payload=secret_42`.
  So it's premature collection, not a setter/codegen-other bug.
- **Two levels does NOT reproduce:** with just `class Leaf < Base`, `sp_Leaf_gc_scan`
  correctly marks the inherited ivar. The bug needs **≥3 levels with a passthrough
  middle class** (one that introduces no ivars of its own). This suggests the scan-set
  is computed as `own_ivars ∪ immediate_parent.own_ivars` rather than the union over
  the full ancestor chain — an empty middle class breaks the chain.
- Object-typed grandparent ivars are dropped the same way; they may land in a benign
  window (freed slot not yet reused → reads stale-but-intact) and not crash immediately.

## Real-world impact

This is what crashes a roundhouse-transpiled Rails app under concurrent keep-alive load.
`ArticlesController < ApplicationController < ActionController::Base`: `@params` is
introduced in `ActionController::Base`, so `sp_ApplicationController_gc_scan` marks it
but `sp_ArticlesController_gc_scan` drops it. `@params` holds a `str_poly_hash`; the GC
frees it mid-request, and the next `@params.fetch("id","0")` →
`sp_StrPolyHash_has_key((sp_StrPolyHash*)iv_params.v.p, …)` dereferences a freed/NULL
pointer. Symbolicated crash (post-`08eaa4b`, release + `-g` rebuild + ASan-style atos):

```
sp_StrPolyHash_has_key      sp_runtime.h     <- SIGSEGV, read at 0x0
sp_ArticlesController_show   (@params.fetch)
sp_ArticlesController_process_action
sp_Main_cls_dispatch / _fiber_body_3 / sp_fiber_trampoline
```

`iv_params` reads as `{tag: SP_TAG_OBJ, cls_id: STR_POLY_HASH, v.p: NULL}` — a valid
discriminant with a freed pointer, so the tag-guarded dispatch in the emitted fetch
passes straight through to the NULL deref.

(Minor, separate observation: `sp_StrPolyHash_has_key` lacks the `if(!h) return` guard
that its siblings `sp_StrPolyHash_get`/`_keys` have — adding it would soften this
particular crash into a silent miss, but the real fix is the scan function.)

## Environment

- spinel master (built today; includes `08eaa4b` capture-struct fix and `GC.start` #1021)
- macOS arm64 (POSIX `ucontext` backend); mechanism is platform-independent.

---

Found while debugging #1031 (compare-spinel still segfaulted after the capture-struct
fix). cc: @OriPekelman

rubys:
Fixes #1050.

## Problem

A class's generated `_gc_scan` function unioned ivars from the class itself plus
its **immediate** parent's own-ivar list — not the transitive set over all
ancestors. Each class records only its *own* ivars (step 1 deliberately skips
inherited names so they're marked under the ancestor's recorded type). So when a
**passthrough class** that introduces no ivars of its own sits in the chain, a
grandchild leaf **drops the grandparent's ivars** from its scan. Instances are
allocated with the leaf class's scan fn, so those ivars are never marked and the
collector frees objects still reachable through them → use-after-free.

Concretely: `ActionController::Base` introduces `@params`; `ApplicationController`
(direct child) keeps it; `ArticlesController < ApplicationController` drops it.

## Fix

Walk the full ancestor chain via `@cls_parents` instead of stopping at the
immediate parent. Each ivar lives in exactly one ancestor's own-list, so there's
no double-marking, and the existing per-ancestor type handling is preserved.

## Test

`test/i1050.rb` — a 3-level hierarchy (`A` introduces `@payload`, empty
passthrough `B`, leaf `C` with its own `@own`). The grandparent ivar `@payload`
is collected across `GC.start` before the fix (prints empty) and survives after.
Full suite: 650 pass / 0 fail / 0 error.

## Notes / for review

- Found while debugging the roundhouse Rails-blog crash referenced in #1050.
  This fix is **necessary but not sufficient** for that app — there is a separate,
  independent bug (a poly `sp_RbVal` local assigned from a method return is not
  `SP_GC_ROOT`'d) that I'm filing as its own issue.
- The chain walk uses `@cls_parents` (superclass only). If ivars can be introduced
  via included **modules**, those would need handling elsewhere — flagging since
  the cases I can exercise are single inheritance.

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -9976,9 +9976,16 @@ class Compiler
           end
           j = j + 1
         end
- # Also scan parent fields
-        if @cls_parents[i] != ""
-          pi = find_class_idx(@cls_parents[i])
+ # Also scan inherited fields. Walk the FULL ancestor chain, not just
+ # the immediate parent: each class records only its OWN ivars (step 1
+ # skips inherited names so they're marked under the ancestor's type),
+ # so a passthrough parent that introduces no ivars of its own would
+ # otherwise drop a grandparent's ivars from this class's scan -- the
+ # collector then frees objects still reachable through those fields
+ # (issue #1050).
+        anc = @cls_parents[i]
+        while anc != ""
+          pi = find_class_idx(anc)
           if pi >= 0
             pnames = @cls_ivar_names[pi].split(";", -1)
             ptypes = @cls_ivar_types[pi].split(";", -1)
@@ -9991,6 +9998,9 @@ class Compiler
               end
               pj = pj + 1
             end
+            anc = @cls_parents[pi]
+          else
+            anc = ""
           end
         end
         emit_raw("}")
--- a/test/i1050.rb
+++ b/test/i1050.rb
@@ -1,0 +1,41 @@
+# A grandchild's GC scan function must mark ivars introduced two or more
+# levels up the inheritance chain. When a passthrough class (B) sits
+# between the class that introduces an ivar (A) and the leaf (C), the
+# leaf's scan must still walk PAST the empty middle class to mark the
+# grandparent's ivar -- otherwise a GC fired while the instance is live
+# frees the object that ivar holds (use-after-free / lost data). The
+# object is reachable ONLY through the inherited ivar, and GC.start
+# forces the collection that surfaces it (issue #1050).
+class A
+  def payload=(v); @payload = v; end
+  def payload; @payload; end
+end
+
+class B < A   # passthrough: introduces no ivars of its own
+end
+
+class C < B
+  def own=(v); @own = v; end
+  def own; @own; end
+end
+
+def churn(n)
+  i = 0
+  s = ""
+  while i < n
+    s = s + i.to_s + ","
+    i += 1
+  end
+  s.length
+end
+
+obj = C.new
+obj.payload = "secret_" + 42.to_s   # grandparent ivar; only live reference
+obj.own     = "kept_" + 7.to_s      # leaf's own ivar (already scanned)
+
+GC.start
+churn(8000)                          # reuse any freed slot
+GC.start
+
+puts obj.own
+puts obj.payload
--- a/test/i1050.rb.expected
+++ b/test/i1050.rb.expected
@@ -1,0 +1,2 @@
+kept_7
+secret_42


rubys:
Matz beat me to it: https://github.com/matz/spinel/commit/9402a7e203f61ddbe265817c4fcc036f329f1388

Issue #1054, opened 2026-05-29
Proposal: sp_net — POSIX TCP/poll/process primitives in libspinel_rt.a (+ where TLS lives)


OriPekelman:
Following the `sp_crypto` precedent (#514): tep's `lib/tep/sphttp.c` has a ~260-LoC core of **generic POSIX runtime** (TCP / poll / prefork / shell) that any HTTP-shaped spinel program wants — roundhouse being the obvious peer. Proposing to extract it into `libspinel_rt.a` as `sp_net.{c,h}` so frameworks build their HTTP layer on top instead of fork-and-modifying. Full surface + transition plan in **OriPekelman/tep#12** (the long-standing tep tracking issue; cc was flagged there). sphttp.c is now 822 LoC, so the split is overdue.

Candidate `sp_net` surface (plain C, no spinel-runtime dep, static return buffers — same shape as sp_crypto):
```
sp_net_listen/accept/accept_nb/connect/close, set_nonblock
sp_net_recv_some/recv_all, write_str/write_bytes
sp_net_poll_reset/add/run/ready
sp_net_fork/exit/getpid/wait_any
sp_net_shell_capture
```
(HTTP framing + WebSocket accessors stay framework-side.)

## The decision that needs your call: where does TLS live?

tep just added **outbound TLS** (libssl): a fd→`SSL*` registry so `connect_tls` returns an fd whose `recv`/`write`/`close` transparently route through `SSL_*`. The catch: that weaves `SSL_*` branches **into** the very recv/write/close that `sp_net` would own. So extracting `sp_net` forces a choice:

- **(A) `sp_net` stays plaintext + optional `sp_net_tls`** — keep libspinel_rt.a OpenSSL-free (matching sp_crypto's deliberate "no OpenSSL/libsodium" stance). The fd→SSL* registry + handshake + TLS read/write live in a separate `sp_net_tls` unit linked only when an app needs it (`-lssl -lcrypto`). Requires the plaintext recv/write to expose a small hook (or sp_net_tls provides its own read/write the framework selects) rather than the registry-inside-IO shape tep uses today. **My lean** — keeps the core dependency-light; TLS is opt-in.
- **(B) `sp_net` includes TLS** — simplest for consumers (HTTPS "just works"), but libspinel_rt.a gains a hard OpenSSL dependency, which sp_crypto pointedly avoided.

[...]

PR #1055, opened 2026-05-29
sp_net: POSIX TCP/poll/process/shell primitives in libspinel_rt.a

OriPekelman:
Concrete proposal for **matz/spinel#1054** (and the long-standing OriPekelman/tep#12). Adds `lib/sp_net.{c,h}` — a vendored, audit-sized POSIX-runtime helper archived into `libspinel_rt.a` alongside `sp_crypto`, same precedent + spirit as #514.

## Why
HTTP-shaped Spinel frameworks (tep, roundhouse) each carry the same ~250-LoC POSIX core (TCP / poll(2) / prefork / shell). This lifts it into the runtime so they build their HTTP layer on top instead of fork-and-modifying. Extracted from tep's `lib/tep/sphttp.c` (stable for months); HTTP framing + WebSocket accessors stay framework-side.

## This implements option (A) from #1054 — TLS stays out
`sp_net` is **plaintext, OpenSSL-free**, keeping `libspinel_rt.a` dependency-light (matching sp_crypto's deliberate "no OpenSSL/libsodium" stance). An optional `sp_net_tls` unit can layer over these fds later for consumers that need HTTPS — not in this PR.

## Surface (`sp_net_` prefix)
```
install_term_handlers / shutdown_requested   # prefork graceful shutdown; accept() honors the flag
listen / accept / accept_nb / connect / close / set_nonblock
recv_some / recv_all / write_str / write_bytes
poll_reset / poll_add / poll_run / poll_ready
fork / exit / getpid / wait_any
shell_capture
```
Pure C, no spinel-runtime dep, static return buffers, single-threaded server model. Consumers declare the FFI surface their side (as tep does for `sphttp_*` today); auto-linked from the archive like sp_crypto, so no `ffi_lib` needed.

## Build + test
- `Makefile`: `build/sp_net.o` rule + added to the `libspinel_rt.a` archive + `install lib/sp_net.h`.
- `test/sp_net_basic.rb` (+ `.expected`): deterministic smoke — `shell_capture` + a 127.0.0.1 TCP loopback (listen/set_nonblock/connect/accept_nb/write_str/recv_some/close) + getpid. Auto-discovered by `make test` (`wildcard test/*.rb`). Verified passing locally (Linux aarch64). poll/fork/wait_any are covered by tep's consumer suite (they need multi-process/timing setups that don't make a stable `.expected`).

## Open questions for you / @rubys
1. **Naming** — `sp_net` vs `sp_posix_net` vs a separate `libspinel_net.a`?
2. **term-signal trio** — does `install_term_handlers`/`shutdown_requested` (+ accept honoring the flag) belong in `sp_net`, or should graceful-shutdown stay framework-side and `accept` be signal-agnostic?
3. **TLS placement** confirmation (option A here) — good, or do you want TLS in (B) / fully framework-side?

tep would follow up by slimming `sphttp.c` to its HTTP/WS-specific surface and consuming `sp_net` from the archive (separate tep PR, gated on this landing + a pin bump).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -401,6 +401,12 @@ TESTS := $(wildcard test/*.rb)
 ifeq ($(SPINEL_INT_OVERFLOW),promote)
 TESTS := $(filter-out test/int_overflow_raises.rb,$(TESTS))
 endif
+# sp_net is POSIX-only; on Windows the TU compiles to stubs, so the
+# sp_net smoke's output diverges. Skip it there (the POSIX surface is
+# exercised by consumer suites on POSIX targets).
+ifeq ($(OS),Windows_NT)
+TESTS := $(filter-out test/sp_net_basic.rb,$(TESTS))
+endif
 TEST_TARGETS := $(patsubst test/%.rb,build/test-results/%.ok,$(TESTS))
 
 # `make test` is incremental via mtime tracking on .ok files;
--- a/Makefile
+++ b/Makefile
@@ -288,9 +288,13 @@ build/sp_core.o: lib/sp_core.c lib/sp_core.h
 	@mkdir -p build
 	$(CC) -c -O2 -Wno-all $(SEC_FLAGS) -Ilib lib/sp_core.c -o build/sp_core.o
 
+build/sp_net.o: lib/sp_net.c lib/sp_net.h
+	@mkdir -p build
+	$(CC) -c -O2 -Wno-all $(SEC_FLAGS) -Ilib lib/sp_net.c -o build/sp_net.o
+
 SP_RT_LIB = lib/libspinel_rt.a
 
-$(SP_RT_LIB): $(RE_OBJ) build/sp_bigint.o build/sp_crypto.o build/sp_pack.o build/sp_strscan.o build/sp_time.o build/sp_core.o
+$(SP_RT_LIB): $(RE_OBJ) build/sp_bigint.o build/sp_crypto.o build/sp_pack.o build/sp_strscan.o build/sp_time.o build/sp_core.o build/sp_net.o
 	ar rcs $@ $^
 
 regexp: $(SP_RT_LIB)
@@ -594,6 +598,7 @@ install: all
 	install -m 644 lib/sp_runtime.h      $(SPNLDIR)/lib/
 	install -m 644 lib/sp_core.h         $(SPNLDIR)/lib/
 	install -m 644 lib/sp_time.h         $(SPNLDIR)/lib/
+	install -m 644 lib/sp_net.h          $(SPNLDIR)/lib/
 	install -m 644 lib/*.rb              $(SPNLDIR)/lib/
 	install -d $(PREFIX)/bin
 	ln -sf $(SPNLDIR)/spinel $(PREFIX)/bin/spinel
--- a/lib/sp_net.c
+++ b/lib/sp_net.c
@@ -4,8 +4,48 @@
  * Extracted from tep's lib/tep/sphttp.c (the POSIX-runtime core that
  * is generic across HTTP-shaped Spinel programs), per matz/spinel#1054
  * and OriPekelman/tep#12. HTTP framing + WebSocket accessors + TLS stay
- * framework-side. */
+ * framework-side.
+ *
+ * Platform split: sp_net is a POSIX prefork/poll runtime (sockets,
+ * poll(2), fork, sigaction). Those APIs don't exist under MinGW, so on
+ * Windows the whole TU compiles to thin error-sentinel stubs -- the
+ * functionality is simply absent there, but libspinel_rt.a still
+ * builds everywhere. Mirrors sp_runtime.h's per-platform fiber backend
+ * (Win32 Fiber vs POSIX ucontext). */
 #include "sp_net.h"
+
+#if defined(_WIN32)
+
+/* ---------- Windows (MinGW): no POSIX net/process surface ---------- */
+
+int sp_net_install_term_handlers(void) { return -1; }
+int sp_net_shutdown_requested(void)    { return 0; }
+
+int sp_net_listen(int port, int reuseport) { (void)port; (void)reuseport; return -1; }
+int sp_net_accept(int sfd)                 { (void)sfd; return -1; }
+int sp_net_accept_nb(int sfd)              { (void)sfd; return -1; }
+int sp_net_connect(const char *host, int port) { (void)host; (void)port; return -1; }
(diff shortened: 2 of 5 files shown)


matz:
Thanks for turning #1054 into a concrete PR so quickly, Ori — the extraction is the right move and the surface looks good.

One blocker before this can land: CI is red on two targets.

**1. Windows (build failure).** `sp_net.c` pulls in `<poll.h>` / `fork` / `sigaction` / BSD sockets, none of which exist under MinGW, so `build/sp_net.o` fails and takes the whole `libspinel_rt.a` build down with it (`lib/sp_net.c:16: fatal error: poll.h: No such file or directory`). We'd like to keep `sp_net` in the core archive (no separate `libspinel_net.a` — a prefork/`poll` server is inherently POSIX, and that's fine; the functionality just differs per platform). So please platform-guard the TU: wrap the body in `#if defined(_WIN32)` → thin stubs returning error sentinels, `#else` → the real POSIX implementation. That keeps the archive building everywhere and `sp_net` simply absent/no-op on Windows. This mirrors what `sp_runtime.h` already does for the fiber backend (Win32 Fiber API vs POSIX `ucontext`), so it's consistent with the existing precedent.

**2. macOS (test failure).** The build is fine there, but `test/sp_net_basic.rb` fails — the loopback / non-blocking-`accept` smoke isn't deterministic off Linux (you noted it was Linux-aarch64-verified). Could you either make that portion deterministic or split it so the `shell_capture` / `getpid` parts run everywhere and the TCP-loopback part is gated to where it's reliable? `make test` runs on all three CI targets, so the `.expected` has to hold cross-platform.

On your open questions: keeping it in `libspinel_rt.a` answers (1) — `sp_net` as-is, in the core archive, is good. TLS placement option (A) is the right call. The `install_term_handlers`/`shutdown_requested` trio is fine to keep in `sp_net` for now. Once CI is green on all four targets we'll merge.

rubys:
From the roundhouse (second-consumer) side — we're on board with this, and aligned with matz's answers to the three open questions:

- **Naming / core archive (Q1):** keeping `sp_net` in `libspinel_rt.a` is right for us — roundhouse FFI-declares the `sp_net_*` surface its side, exactly as it does `sphttp_*` today.
- **Signal trio (Q2):** keep it in `sp_net`. Roundhouse's scheduled server already calls `install_term_handlers` and checks `shutdown_requested` before parking on `io_wait`, so having it in the shared layer suits us — we'd just call `sp_net_*` instead of our vendored `sphttp_*`.
- **TLS placement (Q3):** option A (plaintext core, optional `sp_net_tls` layer later) is the right call.

I checked the proposed surface against roundhouse's vendored server: the socket/poll/process core (`listen / accept{,_nb} / connect / close / set_nonblock / recv_some / write_str / poll_{reset,add,run,ready} / fork / exit / getpid / wait_any` + the term-signal trio) maps cleanly to what `server_scheduled.rb` + the scheduler actually use. The HTTP-framing bits we also use (`drain_body`, `read_request`, `request_buf`, `write_chunk{,_end}`) correctly stay framework-side per your scope.

**One gap, which we'll take as a follow-on — not a blocker here.** Roundhouse serves static assets via zero-copy `sendfile(2)` + `fstat` (`Sock.sphttp_filesize` / `Sock.sphttp_sendfile`, live call sites in both our prefork and scheduled servers). `sp_net` doesn't cover those, and tep almost certainly serves its own assets the same way — so by the same DRY rationale that motivates this PR, `sp_net_sendfile` / `sp_net_filesize` probably belong in the archive eventually. But they're separable (file I/O, not socket/poll/process) and carry their own platform-divergence (Linux vs macOS `sendfile` signatures, Windows `TransmitFile`/stub) — exactly the kind of second cross-platform problem that shouldn't ride into this PR.

So: land `sp_net` as scoped here, and **roundhouse will file a follow-on PR adding `sp_net_sendfile` / `sp_net_filesize`** (with the same per-platform guarding you're applying to the socket TU), then re-vendor to consume them. That fits the rollout model already described here (core lands → consumers follow up, like tep's `sphttp.c`-slimming PR). In the interim roundhouse keeps its framework-side sendfile, so nothing is blocked.

OriPekelman:
Pushed fixes for both CI blockers + the robustness items (7adf0f0):

**Windows build** — wrapped `sp_net.c` in `#if defined(_WIN32)` → thin error-sentinel stubs / `#else` → the POSIX impl, so `libspinel_rt.a` builds everywhere with `sp_net` simply absent on Windows (mirrors `sp_runtime.h`'s per-platform fiber backend, as you suggested). Verified the stub branch compiles with `-D_WIN32` (no `<poll.h>`/sockets pulled in).

**macOS test** — dropped the non-deterministic TCP-loopback from `test/sp_net_basic.rb`; it now exercises only `shell_capture` + `getpid`, deterministic on every POSIX target. Added a Windows filter (`OS=Windows_NT` → `filter-out`) since `sp_net` is stubbed there. The socket/poll/process surface stays covered by the POSIX consumer suites (tep).

**Robustness (gemini):**
- `poll_run`: check `sp_net_term_flag`, return -1 on `EINTR` if set (shutdown no longer retried into an indefinite block).
- `recv_some` / `recv_all`: retry on `EINTR` (was treated as EOF) while honoring `term_flag`.
- `connect`: `signal(SIGPIPE, SIG_IGN)` (a client that never listens needs it too).
- `listen` / `connect`: validate port ∈ [0, 65535].

Open questions: thanks for confirming — core archive (Q1), signal trio stays (Q2), TLS option A (Q3) all match what's here. And noted @rubys's `sendfile`/`filesize` follow-on (separate PR, per-platform-guarded) — agreed it shouldn't ride this one. CI should be green now; let me know if anything else surfaces.

matz:
Merged — thanks Ori. The platform guard (`#if defined(_WIN32)` → error-sentinel stubs / `#else` → POSIX) is exactly the shape we wanted, and CI is green on all four targets including windows-mingw and macOS. `sp_net` lands in `libspinel_rt.a`, auto-linked, absent on Windows.

Note for anyone reading the gemini-code-assist review above: those comments were on the original commit (`64686ee`); your `7adf0f0` already addressed all of them — `recv_some`/`recv_all`/`poll_run`/`accept` all check `sp_net_term_flag` before blocking and on `EINTR` (graceful shutdown preserved), and `connect`/`listen` validate the port range and ignore SIGPIPE. No follow-up needed.

Closes #1054 in spirit (the extraction proposal). Next step is yours: the tep PR slimming `sphttp.c` to consume `sp_net` from the archive, gated on a pin bump.

PR #1058, opened 2026-05-29
runtime,codegen: implement Integer#round(ndigits)

neidiom:
Integer#round without args already returned self, but calling round with a precision argument (e.g. 1234.round(-2)) fell through to the unresolved-call path and silently returned 0. CRuby rounds to the nearest 10^(-ndigits) for negative ndigits.

Added sp_int_round(v, nd) runtime helper using pow/round, and a codegen arm for Integer receivers with an argument.


matz:
Superseded by your #1059, which implements the same `Integer#round(ndigits)` plus `#ceil`/`#floor`/`#truncate` with ndigits in one change. Merged #1059 (CI 4/4 green, all values verified against CRuby 3.2). Thanks — closing this one as redundant; the round-with-precision fix is on master.

PR #1059, opened 2026-05-29
runtime,codegen: implement Integer#ceil/floor/round/truncate with ndi…

neidiom:
…gits

Integer#ceil, #floor, #round, and #truncate without an argument already returned self, but calling with a precision argument (e.g. 1234.ceil(-2)) fell through to the unresolved-call path and silently returned 0. CRuby rounds to the nearest 10^(-ndigits) for negative ndigits; positive ndigits are a no-op on integers.

Added sp_int_round, sp_int_ceil, sp_int_floor, sp_int_truncate runtime helpers using pow/ceil/floor/round, and a codegen arm for Integer receivers with an argument.


matz:
Follow-up: gemini-code-assist's review flagged a real issue — the double-based implementation loses precision for |v| > 2^53 and casting `pow(10,-nd)` is UB once 10^(-nd) > INT64_MAX. Confirmed: `1000000000000000005.round(-1)` gave `...000` (should be `...010`) and `5000000000000000000.round(-19)` returned INT64_MIN garbage.

Addressed on master in 1687982 — reimplemented round/ceil/floor/truncate with exact integer division/modulo (half-away-from-zero, sign-aware ceil/floor, overflow-guarded round-up, `-nd>=19`→0 since 10^19 doesn't fit mrb_int). All in-range results now match CRuby exactly; added a large-value regression test. Thanks @neidiom — the feature and your expected values were spot on, this just hardens the arithmetic.

PR #1060, opened 2026-05-29
runtime,codegen: fix String#oct prefix auto-detection

neidiom:
String#oct used strtoll(base=8) unconditionally, ignoring the 0x/0b/0o/0 prefixes that CRuby auto-detects. "0xff".oct returned 0 instead of 255. Added sp_str_oct runtime helper that checks for prefixes before falling back to base-8, matching CRuby behavior.


matz:
Thanks — the approach and the expected values are right (verified against CRuby 3.2: `[255, 5, 63, 63, 8]`). CI is red on all four targets because of the binary branch.

In `sp_str_oct`, the `0b` case calls `strtoll(p, NULL, 2)`, but `strtoll` only auto-skips a `0x` prefix for base 16 — it does **not** skip `0b` for base 2. So on `"0b101"` it parses the leading `0`, stops at `b`, and returns `0` (which is why `"0b101".oct` fails the test). The hex branch works precisely because `strtoll(p, …, 16)` handles `0x` itself.

One-line fix — skip the prefix explicitly, like you already do for `0o`:

```c
if(p[1]==b||p[1]==B)return(mrb_int)strtoll(p+2,NULL,2);
```

Confirmed locally: `strtoll("0b101", 0, 2)` → `0`, `strtoll("0b101"+2, 0, 2)` → `5`. With that change the suite should go green. (The `0x` branch can stay as `strtoll(p, …, 16)` — it is fine.)

neidiom:
@matz please check now.

PR #1061, opened 2026-05-29
Fix String#rindex with position argument

neidiom:
The 2-arg form s.rindex(sub, pos) silently dropped the position argument and always returned the last match, so "hello".rindex("l", 2) returned 3 instead of 2.  Add sp_str_rindex_from runtime helper that limits the search to codepoint index <= pos (with negative pos support), and route codegen to it when a second argument is present.


matz:
Follow-up: gemini-code-assist correctly flagged that the codepoint walk reset to the string start on each match — O(N²) for a short needle in a long repetitive haystack. Fixed on master (hoisted the walk cursor + counter so they advance monotonically with strstr's forward results → O(N)), plus a NULL-receiver guard. Output unchanged; the regression test still matches CRuby. Thanks @neidiom — the position logic and all 17 expected values were correct.

PR #1063, opened 2026-05-30
Fix Float#ceil/floor/round/truncate return type with negative ndigits

neidiom:
CRuby returns Integer when ndigits <= 0 (e.g. 1.9.ceil(-1) => 10) but Spinel was returning Float (10.0). The analyzer only recognized the literal 0 case; the codegen likewise only cast to mrb_int for ndigits==0. Now the analyzer checks ndigits <= 0 for the IntegerNode literal, and the codegen wraps the stmt-expr in (mrb_int) when the literal is <= 0. Non-literal ndigits route through new runtime helpers sp_flt_ceil_nd/floor_nd/round_nd/truncate_nd which check the sign at runtime.


matz (changes requested):
Thanks for this — the **literal** half is exactly right and fixes the reported case. With your patch `1.9.ceil(-1)` returns `10` (Integer) and `1.234.round(2)` returns `1.23` (Float), both matching CRuby; on master the negative-literal case wrongly returns Float. The analyzer + codegen handling for `IntegerNode` ndigits (`<= 0 → Integer`, `> 0 → Float`) is the correct call.

I can't merge as-is, though: the **non-literal** path (the new `sp_flt_*_nd` runtime helpers) has a value bug and a regression. The PR's own test only exercises literal ndigits, so CI is green while these slip through. Built locally and compared against CRuby and master:

| call (ndigits via a variable) | CRuby | master | this PR |
|---|---|---|---|
| `n = 2; 1.234.round(n)` | `1.23` (Float) | `1.23` (Float) | `1` (Integer) |
| `m = -1; 15.5.round(m)` | `20` (Integer) | `20.0` (Float) | `2` (Integer) |

Two separate problems:

1. **Value bug for `nd <= 0`.** `sp_flt_round_nd` (and the ceil/floor/truncate siblings) do `return (mrb_int)round(v/f)` with `f = pow(10, -nd)`, which divides by `10^(-nd)` but never multiplies back — so `15.5.round(-1)` becomes `round(1.55) = 2` instead of `20`. Your *literal* codegen already has the correct, sign-agnostic formula; the helper should just reuse it:

   ```c
   static mrb_int sp_flt_round_nd(mrb_float v, mrb_int nd){
     double f = pow(10, (double)nd);
     return (mrb_int)(round(v * f) / f);
   }
   ```
   (and the same shape for ceil/floor/trunc). That gives `15.5.round(-1) == 20` for the negative case.

2. **Truncation regression for `nd > 0`.** Because the helper always returns `mrb_int`, `1.234.round(n)` with a runtime `n = 2` truncates to `1`. master returns `1.23` (Float) here, so this is a regression on what's probably the more common non-literal shape (configurable precision). This one isn't really a bug in your code — it's the genuine limitation you flagged in the PR description: the return type of `x.round(n)` depends on the *runtime sign* of `n`, and a single C return type can't be Integer-or-Float. That's a design decision I'd rather make explicitly than ship silently-wrong-on-positive. Options, roughly:
   - keep the old non-literal path (return Float) — value-correct for `n > 0`, wrong type for `n < 0` (today's master behavior);
   - return a boxed poly (Integer when `nd <= 0`, Float when `nd > 0`) — fully CRuby-correct, but a larger change touching the analyzer's type for this call and codegen's poly handling;
   - accept truncation and document it.

   I lean toward the poly-return being the only fully-correct answer, but it's a bigger piece; happy to discuss which way you'd like to take it (or I can pick it up separately).

Could you (a) switch the helpers to the unified formula so the `nd <= 0` value is correct, (b) settle the `nd > 0` direction per above, and (c) add non-literal cases (a variable ndigits, both signs, with `.class` and value assertions) to the test so CI guards this path next time? The literal work is solid and I'd be glad to merge once the non-literal path is value-correct. Thanks again.

matz:
Thanks again, and sorry for the back-and-forth. I landed a different resolution on master in 5ba0470, so this PR is closed — but the fix is built on what you surfaced here and you're credited as co-author on the commit.

CRuby picks Integer vs Float from the **runtime sign** of `ndigits`, and spinel is statically typed: `x.round(n)` with a variable `n` (or `x.round(*args)` / `x.round(*[])`) has no single statically-knowable return class. The `sp_flt_*_nd` helpers forced everything to `mrb_int`, which truncated the float for runtime `n > 0` (`1.234.round(n)` -> `1`) and had a scaling bug for `n <= 0` (`15.5.round(-1)` -> `2` not `20` — the helper divided by `10^(-nd)` without multiplying back).

The resolution: decide the return type by argument **presence**, not value — no arg -> Integer, any `ndigits` arg -> Float. Fully static, never boxes, always numerically correct. The one deliberate divergence from CRuby is that `x.round(-1)`/`(0)` is now `Float`-typed (`10.0`) rather than `Integer` (`10`); values are equal, only `#class` / default string form differ. Details in [`docs/FLOAT-ROUNDING.md`](https://github.com/matz/spinel/blob/master/docs/FLOAT-ROUNDING.md).

Really appreciate the report and reproductions — it's a genuine compatibility corner. Hope you'll send more.

PR #1065, opened 2026-05-30
Fix String#squeeze with charset argument

neidiom:
squeeze("a") was silently dropping the argument and squeezing all consecutive characters (same as the no-arg form). Add sp_str_squeeze_chars runtime helper that checks each character against the charset before deciding to squeeze, with ^-negation support (squeeze("^a") squeezes everything except 'a'). Route codegen to the new helper when an argument is present.


matz:
Merged, thanks! This fixes the silent arg-drop and adds `^`-negation, which is more than `count`/`delete` even do today.

I followed up in a0cfe54 with one CRuby-compat fix: a **lone `^`** charset (`"x".squeeze("^")`) is the literal caret, not a negation, so it should squeeze only `^` runs (`"aaa^^^".squeeze("^")` -> `"aaa^"`). The original always treated a leading `^` as negation, which squeezed everything. Added a regression case for it.

Two remaining limitations, both **family-wide** (shared with `count`/`delete`/`tr`, not specific to this PR, so left as-is for consistency):
- **Multiple charset args** (`squeeze("a-z", "^aeiou")` = intersection) use only the first argument — the codegen passes `compile_arg0` like the rest of the family.
- **Backslash escapes** in the charset aren't processed (`sp_utf8_decode_charset` handles ranges but not `\`).

Neither affects the common single-charset usage. Appreciate the fix!

PR #1066, opened 2026-05-30
Fix String#delete and String#count with ^-negated charset

neidiom:
"hello".delete("^l") returned "heo" (deleting ^ and l literally) instead of "ll" (deleting everything except l). The charset decoder sp_utf8_decode_charset treats ^ as a literal codepoint; the ^- negation is a higher-level Ruby convention that each consumer must handle. Add ^-stripping and in_set inversion to both sp_str_delete and sp_str_count, matching the pattern already used in sp_str_squeeze_chars.


matz:
Merged, thanks — this closes the `^`-negation gap for `count`/`delete` that I'd noted as a family limitation on #1065.

Followed up in 168b231 with the same lone-`^` fix I applied to `squeeze`: a lone `"^"` charset is the literal caret, not a negation (`"a^b".count("^")` -> `1`, `"a^b".delete("^")` -> `"ab"`), so `^` only negates when followed by more of the set. Added regression cases.

Still family-wide (shared with `tr`/`squeeze`, left as-is): multiple charset args use only the first, and backslash escapes aren't processed. Neither affects normal single-charset usage.

PR #1090, opened 2026-05-30
Fix String#tr/tr_s with ^-negation in from-set

neidiom:
"abc".tr("^a", "x") returned "xbc" instead of "axx". The ^ at the start of the from-set inverts which characters get translated — all chars EXCEPT those listed. sp_str_tr and sp_str_tr_s didn't handle ^, treating it as a literal caret. Add ^-stripping and in_set inversion, using tcps[tn-1] as the replacement for negated matches (CRuby uses the last to-char for all negated matches), while preserving positional mapping (tcps[mi]) for non-negated tr.


matz:
Merged into master as de75220. Rebased onto current master and resolved the conflict with #1077 (empty replacement string) locally: the ^-negation version already subsumes the empty-to-set deletion, so both behaviors are preserved. Verified `tr`/`tr_s` negation and empty-replacement test cases against CRuby, full suite 686 pass. Thanks!

PR #1271, opened 2026-06-01
Specialize implicit allocation-site helper classes

fc-anjos:
## What changed

This adds implicit allocation-site specialization for direct no-argument helper class construction when the helper's state is inferred from empty-array ivars. Each eligible `Class.new` site is cloned to a synthetic internal class so sibling helper instances can keep independent typed array storage instead of widening through the shared template class.

Codegen now lowers constructor calls to the analyzer-selected synthetic class, resolves shared-body ivar/local/return cache entries against the active specialized class, and skips unused template helper method declarations/bodies when all real calls use synthetic specializations.

## Why

A single ordinary helper class can be used at multiple closed-world allocation sites with different record-like element types. Before this change, inference merged those sites through the base class and widened useful typed storage.

## Tests

- `git diff --check`
- `make build/test-results/implicit_alloc_site_store_specialization.ok build/test-results/implicit_alloc_site_cache_helper_specialization.ok build/test-results/bundle_classd_49.ok build/test-results/param_body_hash_inference.ok`


matz:
Thanks for this — the allocation-site specialization is a genuinely nice idea, and the implementation holds up well on the checks that matter most for a compiler-source change:

- **Self-host fixpoint is stable.** `make bootstrap` reaches byte-equality on both `analyze.rb` and `codegen.rb` (IR and C). The five new recursive shared tree-walk inference helpers don't destabilize stage-2, which is the failure mode I was most worried about for a change of this shape.
- **No optcarrot regression.** Checksum stays `59662`.
- Both new tests pass, and CI is green.

There is one soundness hole, though, that the current tests don't exercise: **a specialized instance that escapes into a unified / boxed receiver position.**

### Minimal repro

```ruby
class Store
  def initialize; @records = []; end
  def add(r); @records << r; r; end
  def count; @records.length; end
end

def use(s)        # s receives two *different* clones -> param unifies to base obj_Store
  s.add(1)
  s.count
end

a = Store.new
b = Store.new
a.add(10)
b.add(20)
puts use(a)
puts use(b)
```

```
undefined reference to `sp_Store_count'
undefined reference to `sp_Store_add'
collect2: error: ld returned 1 exit status
```

### Root cause

`implicit_specialization_template_class?` treats the base/template class as fully dead once any clone exists, so `emit_forward_decls` and `emit_class_methods` skip its instance-method bodies. But the base class is still reachable: when an instance flows into a position that unifies the two sites (here, the parameter `s`, typed `obj_Store`), codegen lowers the call through the base and the boxed `cls_id` dispatch:

```c
static inline mrb_int sp_use(sp_RbVal lv_s) {
  if (_t1.cls_id == 47) _t2 = sp_Store_add((sp_Store *)_t1.v.p, 1LL);   // 47 = base Store, body suppressed
  if (_t3.cls_id == 47) _t4 = sp_Store_count((sp_Store *)_t3.v.p);
}
```

So the base methods are *called but never defined* → link error. There's also a latent second bug hiding behind the first: even if those bodies were emitted, the boxed values carry the **clone** class ids (`49`/`51` = `Store__implicit_33/36`), so the `cls_id == 47` arm never matches and the call would be silently skipped — wrong result rather than a link failure. And the suppression is half-applied today: the base class still emits its constructor adapter, `gc_scan`, and `new`, only the instance methods are dropped, which is inconsistent.

The reason the suite stays green is that none of the current tests have a method taking the helper class as a parameter (or otherwise unifying two sites through the base), so this path never gets generated.

### Suggested directions

1. **Escape-aware specialization (the robust fix):** only clone sites whose instances can't reach a position that unifies them back to the base — parameter, return value, collection element, etc. If an instance can escape to a base-typed slot, fall back to the shared base class for *all* sites of that class. This is the classic allocation-site-specialization + escape interaction; it also avoids the deeper struct-layout problem (two clones with `sp_IntArray*` vs `sp_StrArray*` `@records` can't soundly share one `sp_Store *`).
2. **Minimal stop-gap:** don't suppress the template class — always emit its instance methods (and a base `cls_id` dispatch arm) so the boxed path links. This fixes the link error but not the mis-dispatch, and not the layout mismatch when clones differ in element type, so it's only safe as an interim step.

Happy to help with whichever direction you take. Nice work on getting the self-host and shared-body cache resolution right — that's the hard part.

fc-anjos:
Thanks for the detailed review, Matz!

I see the problem: the specialization needs to stop once an instance can flow into a shared receiver position. I added your example as a regression test, plus a layout-sensitive variant to cover the deeper shape/layout problem.

The fix now leaves that class unspecialized when a candidate escapes through method arguments, returns, or shared storage, while still allowing closed local helper cases to specialize.

PR #1276, opened 2026-06-02
Add `spinel --emit-rbs`: export whole-program inferred types as RBS

OriPekelman:
# Add `spinel --emit-rbs`: export whole-program inferred types as RBS

## What

A new analyze-only flag that dumps Spinel's whole-program type inference as an
RBS file:

```sh
spinel app.rb --emit-rbs            # writes app.rbs, no binary
spinel app.rb --emit-rbs -o app.rbs
```

It stops after the analyzer (no codegen, no compile), walks the inferred
per-method/ivar types, and emits structured RBS — proper `class … ; def
self.foo: (String) -> Integer` — with a `# spinel: widened to untyped (slow
path)` marker on any slot that degraded to the boxed poly path.

## Why

The analyzer already computes a concrete type for (almost) every node; today
that knowledge is internal. Exposing it as RBS makes it consumable by the wider
Ruby toolchain:

- **Type checkers** — drop the export into `sig/` and run Steep/Sorbet against
  it; the output is `rbs validate`-clean.
- **Editors / the inference** — the same data backs a ruby-lsp hover addon.
- **The silent-miscompile signal** — the `untyped` markers point exactly at the
  spots where inference fell back to the boxed slow path, which is where
  surprising runtime behavior tends to live.

## Design / safety

- **Opt-in and analyze-only.** Without the flag nothing changes; with it, the
  compiler stops before codegen and writes only the RBS. Zero effect on the
  generated C or the compiled binary of a normal build.
- **Self-contained.** Touches two files: the `spinel` wrapper (the flag +
  `SPINEL_EMIT_RBS` env + the analyze-only early-exit) and `spinel_analyze.rb`
  (the RBS formatter + a single `ENV["SPINEL_EMIT_RBS"]`-gated emit hook at the
  end of analysis). No parser or codegen changes.

## Validation

Built and tested on aarch64-Linux against current `master`:

- `make test`: **778 pass / 0 fail / 0 err** (unchanged — the flag is additive).
- Exercised on **tep** (a ~40-file Sinatra-style framework, the largest
  real-world pure-Ruby Spinel app): the inferred signatures were compared against
  the project's **47 hand-authored `.rbs` files**. Of the 524 authored methods,
  **74% were reachable** from a single entrypoint, and **71% of those matched the
  human signatures exactly** (278/391); the differences
  are systematic and defensible (Spinel reports the *compiled* return type — e.g.
  a setter returning `Integer` 0 — and infers param nilability the author left
  implicit), plus the expected `untyped` widenings at polymorphic/heterogeneous
  boundaries, which the export self-marks.

Sample (from the tep run):

```rbs
class Tep::Url
  def self.escape: (String) -> String
  def self.split_url: (String) -> Hash[String, String]
  def self.parse_query: (String) -> Hash[String, String]
end
```

## Notes

This is the first of a small series of opt-in *observability* surfaces (inferred
types, `#line` debug stepping, native backtraces) developed against real apps;
each is gated and leaves release output untouched. Sending `--emit-rbs` first
since it's the most self-contained and immediately useful.

### DIFF
--- a/spinel
+++ b/spinel
@@ -43,6 +43,7 @@ esac
 SOURCE=""
 OUTPUT=""
 C_ONLY=0
+EMIT_RBS=0
 STDOUT_MODE=0
 RUN_MODE=0
 OPT_LEVEL="2"
@@ -107,6 +108,8 @@ $1"
                 EVAL_USED=1
                 shift ;;
     -c)         C_ONLY=1; shift ;;
+    --emit-rbs) # Analyze-only: dump inferred signatures as RBS, no binary.
+                EMIT_RBS=1; shift ;;
     -S)         STDOUT_MODE=1; shift ;;
     -E)         RUN_MODE=1; shift ;;
     *.rb)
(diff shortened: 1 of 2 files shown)


OriPekelman:
Thanks @gemini-code-assist — good catch. Added the conflicting-mode guards so `--emit-rbs` errors (exit 2) instead of silently ignoring `-E` / `-c` / `-S`. I placed them up front alongside the existing `-E` mode-conflict checks (rather than in the RBS block) so it fails fast before any parse/analyze work:

```
$ spinel app.rb --emit-rbs -c
spinel: --emit-rbs cannot be combined with -c (C source only)   # exit 2
```

Pushed as an amend to the single commit.

PR #1292, opened 2026-06-02
Add `spinel --debug` / `-g`: #line-mapped builds for native-debugger stepping

OriPekelman:
# Add `spinel --debug` / `-g`: `#line`-mapped builds for native-debugger stepping

## What

A debug build mode that makes a Spinel binary steppable in gdb/lldb against the
**original Ruby source**:

```sh
spinel app.rb --debug      # #line directives + non-inlined methods + -g -O0
spinel app.rb -g           # add #line + -g at the current -O
gdb ./app                  # break app.rb:42 ; step ; bt   -> Ruby lines
```

The codegen emits `#line <rubyline> "<rubyfile>"` directives mapping the
generated C back to the `.rb`, keeps methods non-inlined (so each Ruby method is
a real C frame), and the wrapper adds `-g -O0`. Multi-file: the `#line` map
follows `require_relative`'d sources, so stepping crosses file boundaries.

## Why

Today a compiled Spinel program is a black box to native tooling — a crash or a
breakpoint lands in generated C with no path back to the Ruby the developer
wrote. With `#line`, the line table points at the `.rb`, so breakpoints by
`app.rb:line`, single-stepping, and `bt` all speak Ruby. It's also the
foundation the inference-position export (`--emit-types`) and native
`Exception#backtrace` build on.

## Design / safety

- **Opt-in.** Without `--debug`/`-g`, codegen and the binary are byte-for-byte
  unchanged. `--debug` forces `-O0` and non-inlined methods for faithful
  stepping; `-g` is the lighter "add line info at the current `-O`" variant.
- The parser stamps each node with its `(file, line)`; codegen emits the
  matching `#line` directives and suppresses inlining under debug.

## Known limitation (please read)

`#line` directives perturb the C compiler's **DWARF variable-location** info: a
debugger then reads a function's **locals** from the wrong stack slot (their
zero-init) when that function has a GC-rooted (heap) local. **Source-line
stepping and the line table are correct** — breakpoints, stepping, and `bt` work;
only *local-value inspection* (`p lv_x`) is unreliable in heap-local functions.

This reproduces **cross-toolchain** (clang/lldb on macOS and gcc/gdb on Linux),
so it reads as a fundamental `#line`-vs-DWARF interaction rather than a one-compiler
bug. It's disclosed up front because it's load-bearing: a value-differential
tool we built around this sidesteps it by tracing a `#line`-free build and mapping
C lines back via the directives. Happy to discuss whether to narrow it (e.g.
emit `#line` only at statement boundaries, or gate per-function) in review.

## Validation

- `make test`: **787 pass / 0 fail / 0 err** (unchanged — debug is opt-in).
- A `--debug` build carries `#line app.rb:<n>` directives; gdb/lldb break on
  `.rb` lines and `bt` shows Ruby frames.

First of the debug-surface series; `--emit-types` (position-keyed inferred types)
and native `Exception#backtrace` build directly on this.

### DIFF
--- a/compiler_helpers.rb
+++ b/compiler_helpers.rb
@@ -466,6 +466,8 @@ class Compiler
     @nd_type.push("")
     @nd_name.push("")
     @nd_value.push(0)
+    @nd_line.push(0)   # keep the source-map arrays parallel with the rest, so a
+    @nd_file.push(0)   # dynamically-allocated node can't drift them out of sync
     @nd_content.push("")
     @nd_flags.push(0)
     @nd_operator.push("")
@@ -529,6 +531,7 @@ class Compiler
     @nd_name = Array.new(n, "")
     @nd_value = Array.new(n, 0)
     @nd_line = Array.new(n, 0)
+    @nd_file = Array.new(n, 0)
     @nd_content = Array.new(n, "")
     @nd_flags = Array.new(n, 0)
     @nd_operator = Array.new(n, "")
@@ -673,6 +676,21 @@ class Compiler
     if field == "node_line"
       @nd_line[nid] = val
     end
+ # Debug multi-file map: id into the FILE table (0 = toplevel source).
+    if field == "node_file"
+      @nd_file[nid] = val
+    end
+  end
+
+ # Debug multi-file map: record a FILE <id> <path> table entry emitted by
+ # the parser. Grown on demand so ids can arrive in any order.
+  def set_file_entry(id, path)
+    @file_table ||= []   # self-contained: the analyze/codegen consumers init it,
+                         # but don't rely on that from this shared helper
+    while @file_table.length <= id
+      @file_table.push("")
+    end
+    @file_table[id] = path
   end
 
   def set_ref_field(nid, field, ref_id)
--- a/compiler_helpers.rb
+++ b/compiler_helpers.rb
@@ -528,6 +528,7 @@ class Compiler
     @nd_type = Array.new(n, "")
     @nd_name = Array.new(n, "")
     @nd_value = Array.new(n, 0)
+    @nd_line = Array.new(n, 0)
     @nd_content = Array.new(n, "")
     @nd_flags = Array.new(n, 0)
     @nd_operator = Array.new(n, "")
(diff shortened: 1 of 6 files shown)


OriPekelman:
Thanks @gemini-code-assist — addressed all five (pushed as an amend; `make test` still 787 pass / 0 fail):

**Applied — genuine robustness in the new code:**
- **OOM NULL-checks** on `strdup(path)` in `sp_intern_file`, the `premap` `strdup`, and the line-map `calloc`/`malloc` — now fail with a clear message instead of a NULL deref. Consistent with the existing `realloc` check in the same function.

**Applied — defensive, though not a live bug:**
- **`alloc_node`** now pushes `@nd_line`/`@nd_file` too, keeping the parallel-node arrays in sync. (For the record: `alloc_node` has no dynamic callers today — all node creation goes through the bulk `alloc_nodes`, which already sizes both arrays — so there's no actual drift; I added it to preserve the invariant if that ever changes.)
- **`@file_table ||= []`** in `set_file_entry`. Worth noting the `NoMethodError` can't actually occur: both consumers initialize it (`spinel_analyze.rb` / `spinel_codegen.rb` `@file_table = []`). Added the guard anyway so the shared helper is self-contained.

matz:
Merged, thanks! The `--debug` / `-g` line-mapped builds are a great addition.

I applied two small maintainer follow-ups while merging:

- Dropped the defensive `@file_table ||= []` in `compiler_helpers.rb#set_file_entry`. Under the self-host inference the bare `[]` literal lowered to an `IntArray` and collided with the `StrArray`-typed slot, which is what broke the `windows-mingw` build on the second CI run (`sp_PolyArray * = sp_IntArray_new()`). Both consumers already initialize `@file_table = []`, so the shared helper can just assume the slot exists.
- Resolved a merge conflict in `spinel_parse.c` against a recent require-inlining change, so the trailing same-line modifier handling (`require 'x' rescue nil`, CRLF-safe) runs before `sp_wrap_included` marker-wraps the inlined content.

Verified locally: full test suite (800 pass), self-host bootstrap fixpoint, and optcarrot (checksum 59662) all green, plus a `--debug` smoke test across `require_relative` and modifier-`rescue` requires.

PR #1298, opened 2026-06-03
Add `spinel --emit-types`: position-keyed inferred types + degrade diagnostics (JSON)

OriPekelman:
# Add `spinel --emit-types`: position-keyed inferred types + degrade diagnostics (JSON)

> Builds on the now-merged #1292 (`--debug`): the per-node source positions come
> from that PR's parser stamps, and the wrapper turns on `SPINEL_DEBUG` for
> `--emit-types`.

## What

An analyze-only flag that dumps the whole-program inference as position-keyed
JSON:

```sh
spinel app.rb --emit-types            # writes app.types.json, no binary
```

The output is a single object with two arrays:

```json
{
  "types":       [ {"file":"app.rb","line":2,"col":4,"type":"Integer","rbs":"Integer"}, ... ],
  "diagnostics": [ {"file":"app.rb","line":7,"col":2,"severity":"warning","message":"...widened to untyped..."}, ... ]
}
```

`types` is every node with a concrete inferred type, keyed by source position
(1-based line / 0-based col, matching Prism's `node.location`), with both the
internal type and its RBS form. `diagnostics` flags where a value fell to the
boxed `untyped` / poly slow path, positioned at the enclosing `def`.

## Why

This is the machine-readable companion to `--emit-rbs` (#1276). Keyed by source
position, it's what an editor integration consumes directly: a ruby-lsp addon can
show the compiler's inferred type on hover and underline the degrade sites (the
silent-miscompile signal) without re-deriving anything. Same inference, exposed
per-position instead of per-signature.

## Design / safety

- **Opt-in and analyze-only.** Stops before codegen; writes only the JSON. No
  effect on a normal build's output.
- Touches the wrapper (the flag + `SPINEL_EMIT_TYPES` env + analyze-only exit),
  the analyzer (`emit_types_json`), `spinel_parse.c` (emit `node_col` alongside
  the `node_line`/`node_file` from #1292 — Prism's column helper, concatenation-
  stable across require splicing), and a slot to carry the column.

## A note on the schema

The **JSON shape is the least-settled** part of this series — happy to adjust the
keys / structure (e.g. how degrades are represented, whether to nest by file) to
whatever you'd want a stable consumer to depend on. Flagging it so we lock the
shape deliberately rather than by accident.

## Validation

- `make test`: 802 pass / 0 fail / 0 err on current master (unchanged
  — analyze-only and opt-in).
- `spinel app.rb --emit-types` writes a JSON of `{file,line,col,type,rbs}`
  entries + degrade diagnostics.

### DIFF
--- a/compiler_helpers.rb
+++ b/compiler_helpers.rb
@@ -468,6 +468,7 @@ class Compiler
     @nd_value.push(0)
     @nd_line.push(0)   # keep the source-map arrays parallel with the rest, so a
     @nd_file.push(0)   # dynamically-allocated node can't drift them out of sync
+    @nd_col.push(0)    # (and the self-host inferencer needs the literal push here)
     @nd_content.push("")
     @nd_flags.push(0)
     @nd_operator.push("")
@@ -532,6 +533,7 @@ class Compiler
     @nd_value = Array.new(n, 0)
     @nd_line = Array.new(n, 0)
     @nd_file = Array.new(n, 0)
+    @nd_col = Array.new(n, 0)
     @nd_content = Array.new(n, "")
     @nd_flags = Array.new(n, 0)
     @nd_operator = Array.new(n, "")
@@ -679,6 +681,10 @@ class Compiler
  # Debug multi-file map: id into the FILE table (0 = toplevel source).
     if field == "node_file"
       @nd_file[nid] = val
+    end
+ # Debug/types export: 0-based source column for hover position keys.
+    if field == "node_col"
+      @nd_col[nid] = val
     end
   end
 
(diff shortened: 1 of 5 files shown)


OriPekelman:
Thanks @gemini-code-assist — both applied (amended; `make test` 802 pass / 0 fail, conflict checks verified):

- **`@nd_col` in `alloc_node`** — added `@nd_col.push(0)` alongside the `@nd_line`/`@nd_file` pushes, keeping the parallel-node arrays in sync (and giving the self-host inferencer the literal push it types off).
- **`--emit-types` conflict checks** — it now errors (exit 2) on `-E`/`-c`/`-S` instead of silently ignoring them, mirroring `--emit-rbs`, placed with the other up-front mode-conflict guards.

Issue #1259, opened 2026-06-01
Runtime segfaults on process exit after a SIGTERM-handled sp_net server loop


OriPekelman:
## Summary

A Spinel-compiled server that installs `sp_net` term handlers, runs an `sp_net_accept` loop, and exits after catching SIGTERM **segfaults on process exit (139)** instead of exiting cleanly. The `sp_net` handler is correct — the crash is in the runtime's exit/teardown path.

## Repro (via tep, the smallest available)

```sh
git clone https://github.com/OriPekelman/tep && cd tep && make all
./bin/tep build examples/hello.rb -o /tmp/h     # any Tep.run!(port,1,false) server
/tmp/h -p 18091 &
sleep 3
kill -TERM %1
# → "Segmentation fault (core dumped)", exit 139
```

The server is a plain `sp_net_listen` → `sp_net_accept` loop; on SIGTERM `sp_net_accept` returns -1 (the handler set `sp_net_term_flag`), the loop breaks, and the program falls off the end of `main` → exit. The segfault is on that exit.

## Localization

- The `sp_net` handler (`lib/sp_net.c` `sp_net_term_signal`) is async-signal-safe (`sp_net_term_flag = 1`) with `SA_RESETHAND`. Correct.
- Bypassing everything after the accept loop with an explicit `sp_net_exit(0)` changes the exit code **139 → 143** — so the SIGSEGV is in the runtime's normal **process-exit teardown** (the path that runs when `main` returns or `exit()` is called), reached only after a SIGTERM was handled.
- **Build/timing-sensitive**: the same source compiled with a couple of extra `File.write` calls exits **143 (killed before the loop returns)** instead of segfaulting **139 (after)**. The outcome shifts with codegen/memory layout — consistent with the exit-time teardown (at-exit GC sweep? fiber/ucontext cleanup?) tripping on state the signal path left behind, rather than a deterministic bug.

## Impact

No clean shutdown for any Spinel program that catches SIGTERM and then exits — the process can't return from its main loop to run app-level teardown (flush buffers, emit a final record, close fds) without segfaulting. Surfaced in tep (`Tep.run!`) and toy (`toy serve`); see OriPekelman/tep#175 for the downstream thread.

## Notes
Happy to help reduce this to a pure-`sp_net` minimal repro (no tep) if useful — the build-sensitivity makes a stable minimal case fiddly, which is itself a clue. aarch64-Linux; can retest on other targets.

matz:
[...]

PR #1300, opened 2026-06-03
Add native `Exception#backtrace` under `--debug` (macOS + Linux)

OriPekelman:
# Add native `Exception#backtrace` under `--debug` (macOS + Linux)

> Builds on the merged #1292 (`--debug`): a `-g`, non-inlined debug build keeps
> the `sp_<method>` symbols around, so the runtime can snapshot the live C stack
> at `raise` and format it back to Ruby — no per-method shadow frames.

## What

Under `--debug`, a Spinel exception now carries the real call chain instead of an
empty array:

```ruby
begin
  Calc.run
rescue => e
  e.backtrace   # app.rb:in `Calc.run' / app.rb:in `Calc.helper' / app.rb:in `<main>'
end
```

The runtime captures `backtrace()` at the raise site and de-mangles each
`sp_<Class>_<method>` frame back to Ruby: instance methods render `Class#method`
(nested modules `A::B#m`), singleton methods `Class.method`, top-level methods
bare. Runtime-helper frames are filtered out.

Three pieces:
- **base capture + formatting** — `sp_raise_cls` snapshots the stack; the
  formatter parses `backtrace_symbols` output (macOS *and* glibc forms).
- **Linux linkage** — glibc's `backtrace_symbols` only resolves symbols in the
  dynamic symbol table, which `static` functions never reach. Under `--debug`,
  user (and class) methods get *external* linkage and the binary links with
  `-rdynamic` (non-Darwin), so frames actually resolve. Release stays `static`.
- **a debug-only null-receiver guard** — a method call on a nil heap receiver
  raises `NoMethodError` (CRuby parity) instead of a silent NULL deref deep in
  the callee (the matz/spinel#1259 class).

## Design / safety

Opt-in and `--debug`-gated throughout; a release build's codegen, linkage, and
output are byte-for-byte unchanged. The external-linkage + `-rdynamic` and the
null guard only appear in debug builds.

## Known limitation

Frames currently attribute to the toplevel `.rb` (`sp_bt_srcfile` is one path),
not the per-frame file. The function names identify frames correctly; per-frame
file attribution would need codegen to emit a symbol→file table — a follow-up.

## Validation

- `make test`: **803 pass / 0 fail / 0 err** (unchanged — debug-gated).
- Verified `Calc.run → Calc.helper → <main>` for a class-method raise and
  `Foo#bar` for instance methods; ZeroDivisionError drops the `sp_idiv` runtime
  frame; a null receiver raises `undefined method '...' for nil` under `--debug`.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -3378,14 +3378,16 @@ static const char *sp_bt_symbol(const char *line) {
     size_t len = (size_t)(end - p);
     if (len == 0 || len > 250) return 0;          /* unresolved (static/stripped) */
     memcpy(sym, p, len); sym[len] = 0;
-  } else {                                        /* macOS space-delimited form */
-    const char *p = strstr(line, "0x");           /* the address */
-    if (!p) return 0;
-    while (*p && *p != ' ') p++;                   /* skip the address */
-    while (*p == ' ') p++;                          /* to the symbol */
-    if (!*p) return 0;
-    const char *end = p;
-    while (*end && *end != ' ') end++;             /* symbol ends at space/" +" */
+  } else {                                        /* macOS: "<idx> <image> <addr> <symbol> + <off>" */
+    /* The symbol is the token just before the " + <off>" delimiter. Parse
+       backward from the last " + " rather than forward from "0x" — an image
+       path containing "0x" (e.g. /path/0x_proj/bin) would otherwise misparse. */
+    const char *plus = 0, *q = line;
+    while ((q = strstr(q, " + ")) != 0) { plus = q; q += 3; }
+    if (!plus) return 0;
+    const char *end = plus;
+    const char *p = end;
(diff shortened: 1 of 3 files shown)


OriPekelman:
Thanks @gemini-code-assist — all three applied (make test 803 pass; the Linux backtrace still names frames):

- **Leak in `sp_bt_format`** — `sp_bt_symbol` always returns a `strdup`'d string; now `free`d after it's copied into the formatted frame.
- **macOS parse robustness** — the symbol is now parsed *backward* from the last `" + "` offset delimiter rather than forward from `"0x"`, so an image path containing `0x` (e.g. `/path/0x_proj/bin`) no longer misparses the frame.
- **`-rdynamic` on Windows** — restricted to ELF targets via a `case` that skips `Darwin` *and* `MINGW*/MSYS*/CYGWIN*` (Windows has no native backtrace anyway, `SP_BT_AVAILABLE=0`).

PR #1301, opened 2026-06-03
Implement `Array#delete_at` and `Array#join` for float arrays

OriPekelman:
# Implement `Array#delete_at` and `Array#join` for float arrays

`Array#delete_at` and `Array#join` were implemented for `IntArray`, `StrArray`,
and `PtrArray` but not for `FloatArray`, so on a float-typed array they reached
codegen as unresolved calls and **raised `NoMethodError` at runtime**:

```ruby
[1.0, 2.0, 3.0].delete_at(1)   # NoMethodError: undefined method 'delete_at' for float_array
[1.0, 2.0].join(",")           # NoMethodError: undefined method 'join' for float_array
```

Surfaced by a real app (tep's OpenAI-server embeddings path does
`[0.0].delete_at(0)`).

## Fix

- **`lib/sp_runtime.h`** — add `sp_FloatArray_delete_at` (0-based, like the other
  `FloatArray` ops; returns `0.0` out-of-range, matching `delete_at`'s `nil`
  there) and `sp_FloatArray_join` (each element via the Ruby-faithful
  `sp_float_to_s`, placed where `sp_float_to_s` / `sp_String` are in scope).
- **`spinel_codegen.rb`** — dispatch both in the `float_array` branch
  (`join`'s no-arg default mirrors the `IntArray` peer).

This is a plain method-implementation fix — not gated, no effect on anything
else.

## Validation

Matches CRuby:

```
$ ruby   -e 'a=[1.0,2.0,3.0]; r=a.delete_at(1); puts "del=#{r} join=#{a.join(",")}"'
del=2.0 join=1.0,3.0
$ spinel -E -e 'a=[1.0,2.0,3.0]; r=a.delete_at(1); puts "del=" + r.to_s + " join=" + a.join(",")'
del=2.0 join=1.0,3.0
```

`make test`: full suite green.


OriPekelman:
Good catch @gemini-code-assist — fixed. `sp_FloatArray_join` now mirrors `sp_IntArray_join` exactly: it builds into a `malloc` buffer and returns an `sp_str_alloc`'d copy, instead of returning `->data` from an `sp_String` whose owner isn't GC-rooted across the return. (`sp_float_to_s`'s result is copied immediately, before the next call can reuse its buffer.) make test 803 pass; `[1.0,2.5,3.0].join("|")` → `1.0|2.5|3.0`.

PR #1304, opened 2026-06-03
Add SPINEL_GC_VERIFY: opt-in mark-path object validation

rubys:
## What

`SPINEL_GC_VERIFY=1` makes the collector validate, before it invokes an object's `scan` hook, that the object is a currently-registered heap allocation. If a raw/aliased pointer — e.g. one into a string or builder buffer — has reached the mark path, it **aborts with a diagnostic at that point** instead of jumping through a garbage scan-function pointer (which otherwise faults at an arbitrary address and is near-impossible to attribute post-mortem).

Implementation is small and contained in `lib/sp_runtime.h`:
- a per-mark sorted snapshot of `sp_gc_heap` + `sp_gc_old_heap` (taken at the start of `sp_gc_mark_all`, before the sweep, when every reachable object is still linked);
- an `O(log n)` membership check at the two `h->scan(obj)` call sites in `sp_gc_mark` / `sp_gc_mark_all`;
- a diagnostic + `abort()` on a miss.

## Why

This is in the spirit of the "expose what the compiler knows" discussion in OriPekelman/spinel-dev#5. The GC is the one place where a corrupt/aliased pointer turns into a *silent* crash: the collector calls through `h->scan`, and if `h` isn't a real header the program jumps to a garbage address. A timing-dependent version of that is extremely hard to localize — neither `gdb` nor ASan catch it cleanly (the reclaim path bypasses libc `free()` via the per-class object pools, and the fault is type-confusion on *live* memory, not a malloc use-after-free).

## Opt-in / output-neutral

With `SPINEL_GC_VERIFY` unset, behavior and codegen are unchanged — the only cost on the normal path is one branch on a zero static. Nothing is compiled differently.

## Validated

On a transpiled Rails app under concurrent fiber load, this deterministically converted a non-reproducible `SIGBUS` (`pc=0x80`) into:

```
  obj    = 0x9a3417700
  header = 0x9a34176d0
  ->scan = 0x80
  ->size = 43172101472
```

i.e. a raw pointer into a buffer (`header = obj − sizeof(sp_gc_hdr)` lands in buffer interior → bogus `scan`/`size`) on the mark path. A minimal repro + bug report will follow.

## Note

I also prototyped a `SPINEL_GC_STRESS` (force a collection per allocation) but **excluded it**: every-allocation collection segfaults at boot by prematurely collecting normally-safe, transiently-unrooted temporaries — noise rather than signal. Happy to revisit a safer form (e.g. delayed/every-N) if it'd be useful.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -324,6 +324,19 @@ static const char *sp_rational_inspect(sp_Rational r) {
    (pool push) or free it. Used by class-instance free-list pools. */
 static sp_gc_hdr *sp_gc_heap = NULL; static size_t sp_gc_bytes = 0; static size_t sp_gc_threshold = 256*1024;
 
+/* ---- GC verify: opt-in mark-path validation (release-neutral) ------------
+ * SPINEL_GC_VERIFY=1  before the collector invokes an object's scan hook,
+ *   check that the object is a currently-registered heap allocation. If a
+ *   raw/aliased pointer (e.g. into a string or builder buffer) has been put
+ *   on the mark path, abort with a diagnostic at that point instead of
+ *   calling through a bogus scan-function pointer (which otherwise faults at
+ *   a garbage address and is near-impossible to attribute).
+ * Default OFF; with it unset, behavior and codegen are unchanged. */
+static int sp_gc_verify = 0;
+static void sp_gc_verify_snapshot(void);
+static int  sp_gc_obj_registered(sp_gc_hdr *h);
+static void sp_gc_verify_fail(void *obj, sp_gc_hdr *h);
+
 /* ---- String GC ---- */
 static sp_str_hdr *sp_str_heap = NULL;
 #define SPL(s) (&("\xff" s)[1])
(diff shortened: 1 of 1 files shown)


rubys:
Thanks for the review — addressed in ded7e7a:

- **realloc failure (`sp_gc_vsnap_push`)** — now `sp_oom_die()` instead of silently dropping a header. Good catch: a dropped header would make `sp_gc_obj_registered` miss a live object and abort with a *false positive* under memory pressure, which would undermine the whole point of the check.
- **UB pointer comparison (`sp_gc_vsnap_cmp`, `sp_gc_obj_registered`)** — now compared via `uintptr_t`.
- **deref of a likely-bad header in `sp_gc_verify_fail`** — split the diagnostic: `obj`/`header` (pointer values only) are printed and `fflush`ed *before* touching the header for `->scan`/`->size`, so a possibly-unmapped header can't secondary-fault inside `fprintf` and swallow the message. Kept `->scan`/`->size` (they were the conclusive signal in the real case — `scan=0x80`, a buffer-interior "size"); a missing second line now itself indicates an unmapped header.
- **check before dereferencing header fields (`sp_gc_mark`)** — moved the registration check (which does *not* dereference `h`) to right after `h` is computed, before reading `h->marked`/`h->scan`, so a non-heap pointer is caught before any field access — not only at the scan call. Removed the now-redundant scan-site/drain checks.

One I did **not** change: the `__attribute__((constructor))` for env init. This file already uses the same attribute for the per-class object pools (`SP_POOL_DEFINE`), the generated app is a single TU, and spinel's toolchain is cc/clang/gcc — so the per-TU/MSVC concerns don't apply here, and matching the existing pattern seemed preferable to a per-call init branch. Happy to switch to lazy-init if you'd rather.

Rebuilt and re-verified: release build (env unset) boots unchanged; the verify build still traps the aliased pointer on the mark path.

PR #1371, opened 2026-06-07
Extend implicit holder specialization to pointer-array payloads

fc-anjos:
## Summary

This extends the holder specialization added in #1299 to constructor payloads that are concrete object pointer arrays.

The supported constructor payload shapes are now:

- direct object payloads, such as `obj_Row`
- concrete object pointer arrays, such as `obj_Row_ptr_array`

It also fixes shared method-body typing so expressions like `@record` and `@records[index]` are resolved from the active synthetic holder class before codegen consults shared node caches.

## Why

A holder class can be allocated at multiple closed-world sites with different record payloads. The existing holder specialization handled direct object payloads, but concrete record arrays did not participate in the allocation-site signature key.

That left shared holder methods vulnerable to stale cached expression types when the same AST node was emitted for multiple synthetic classes. The active synthetic holder class should determine the expression type.

## Changes

- Classify constructor arguments through one specialization-payload helper.
- Include concrete object pointer arrays in holder allocation-site signature keys.
- Keep direct object payloads on the same classifier path.
- Extend the self-reference guard to pointer-array payloads.
- Resolve active-specialization expression types before shared codegen node caches.
- Add a regression covering direct record payloads and concrete record-array payloads across two holder allocation sites.

## Test

Added:

```txt
test/implicit_alloc_site_holder_ptr_array_records.rb
```

Validation:

```txt
make -C .vendor/spinel LTO=0 codegen
make -C .vendor/spinel LTO=0 REF_RUBY='mise exec ruby -- ruby -EUTF-8' retest
```

Full retest summary:

```txt
RBS extractor tests: 13 pass
Analyze-fail tests: 3 pass
Tests: 823 pass, 0 fail, 0 error
```

### DIFF
--- a/spinel_analyze.rb
+++ b/spinel_analyze.rb
@@ -10488,15 +10488,38 @@ class Compiler
     end
   end
 
-  def implicit_new_site_has_obj_arg?(nid)
+  def implicit_new_holder_specialization_payload_type(t)
+    bt = base_type(t)
+    if is_ptr_array_type(bt) == 1
+      return bt
+    end
+    if is_obj_type(bt) == 1 && is_array_type(bt) == 0
+      return bt
+    end
+    ""
+  end
+
+  def implicit_new_site_has_specialization_payload?(nid)
     arg_types = implicit_new_site_arg_types(nid)
     k = 0
     while k < arg_types.length
-      bt = base_type(arg_types[k])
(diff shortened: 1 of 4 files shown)


matz:
Thanks! The C reimplementation (now `master`) already specializes implicit holders with pointer-array payloads — verified with a two-site pointer-array holder reproducer (compiles and runs correctly). Closing as already covered in C.

PR #1376, opened 2026-06-08
analyze,codegen: allow return/break/next inside direct instance_exec

ryanseys:
Lets a literal-block `recv.instance_exec { ... }` carry non-local control flow, since the spliced body runs in the caller's frame: `return` exits the enclosing method, `break` / `next` set the call's value.

## What's here
- Analyze gate (`iexec_rewrite_call`) stops rejecting `return` / `break` / `next`.
- Codegen wraps the spliced body in `do { } while (0)`: `break` → C `break`, `next` → `continue`, value captured into the result temp; `return` → C `return` from the enclosing function. Nested loops bind their own `break` / `next` first.

## Tests
`instance_exec_nonlocal_return`, `_break`, `_next` (expression-position values + nested-loop binding), vs ruby 4.0.5.

## Verification
`make bootstrap` fixpoint passes; `make retest` green (869 tests).


matz:
Implemented on master (commit 56b3ee7). `break`/`next` that bind to a direct `instance_exec` splice now work: the body is wrapped in `do {} while (0)` so `break` exits with its value (via the existing break-var capture) and `next` continues out with its value; a break/next inside a nested `each {}` still binds to that loop. `return` already worked. Ported from your test files.

PR #1377, opened 2026-06-08
analyze: receiver and context coverage for direct instance_exec/eval

ryanseys:
Extends where a direct `instance_exec` / `instance_eval` lift resolves its receiver, building on the inline-splice lowering.

## What's here
- **Implicit self** — receiverless `instance_exec { }` / `instance_eval { }` inside a class method resolves to `self` (the method's `@current_class_idx`).
- **Constant receiver** — `CONST.instance_exec { }` where `CONST = Foo.new`; `recv_class_idx_for_rebind` strips the `obj_<Class>` prefix, the same shape the local/ivar arms use.
- **Top-level method bodies** — `ieval_walk_toplevel_methods` walks file-scope `def`s (a method-local receiver resolves via the method's own param/local scope).
- **Nested lifts** — `descend_lifted_block` re-walks a lifted block's detached body so an inner `instance_exec` / `instance_eval` also lifts; codegen splices it within the outer inline with self rebound per level.

## Tests
`instance_exec_implicit_self`, `_const_receiver`, `_toplevel_method`, `_nested`, vs ruby 4.0.5.

## Verification
`make bootstrap` fixpoint passes; `make retest` green (870 tests).


matz:
Implemented on master (commit 4942d1b). The const-receiver, nested, and top-level-method forms already worked on the C compiler; the remaining gap — receiverless `instance_exec { ... }` / `instance_eval { ... }` resolving to self inside an instance method — now works (literal-block only, so a trampoline `instance_exec(args, &b)` body is unaffected). Thanks for the PR; the C compiler is a from-scratch reimplementation so the patch didn't apply directly, but your test files were the spec.

PR #1385, opened 2026-06-08
analyze: back-propagate a callee's param type to an uncalled instance method's param

OriPekelman:
## Problem

An instance method that is **never called** has no call site to constrain its params. If it merely *forwards* a param into another instance method, the param keeps its `int` default and then mismatches the callee's concrete C param at the (dead-but-still-emitted) call:

```ruby
class Cfg;    def initialize(v); @v=v; end; def v; @v; end; end
class Engine;  def realize(cfg); cfg.v + 1; end; end          # shared
class Called;   def go(cfg); Engine.new.realize(cfg); end; end  # called -> cfg pinned obj_Cfg by its call site
class Uncalled; def go(cfg); Engine.new.realize(cfg); end; end  # NEVER called -> cfg stays int
c = Cfg.new(5)
puts Called.new.go(c).to_s
```
```
poly.rb: error: incompatible type for argument 2 of 'sp_Engine_realize'   (int vs sp_Cfg)
spinel: C compilation failed
```

`Engine#realize`'s param is correctly `obj_Cfg` (and the emitted signature `sp_Engine_realize(sp_Engine*, sp_Cfg)` is correct). `Called#go`'s `cfg` is pinned `obj_Cfg` by its **call site**. `Uncalled#go` has no call site, so the only thing that could pin its `cfg` is back-propagation from the callee — and the existing pass, `infer_param_type_from_callee_slot`, covers only top-level / class-method callees (`callee_slot_type` resolves those), **not a receiver-dispatched instance method**. So forwarding from one instance method to another falls in the hole, and the dead method's body won't compile.

## Fix

`infer_instance_param_from_callee_slot` (the receiver-aware instance-method counterpart): for a class instance method's `int`-default param, walk the body, resolve each forwarding receiver to its class (`infer_type` → `obj_X`), look up `X#meth`'s param slot, and pin the param to the agreed obj type (bail on disagreement). It's a **pure narrowing** — more precise typing of the forwarding method — so it can't widen a shared method or cause a miscompile.

## Result

- The repro compiles and prints `6`.
- **Selectivity:** a genuinely-`int` param forwarded to an `int` callee is left untouched — `sp_Wrap_go(sp_Wrap*, mrb_int lv_x)` stays `mrb_int`. Only params forwarded to a concrete-obj callee slot are pinned.
- Self-host bootstrap converges (`make all`); full suite **845 pass / 0 fail / 0 error**. Regression test added (`test/uncalled_method_param_callee_backprop.rb`).

This is the analyzer blocker for a "full-API require" in [OriPekelman/toy#42](https://github.com/OriPekelman/toy/issues/42): loading an unused recipe whose `cfg` forwards into the shared engine broke the compile for every consumer. Diagnosed in OriPekelman/spinel-dev#11.


matz:
Thanks! On the C reimplementation (now `master`) this already works: an uncalled forwarding instance method's param is pinned from the callee's instance-method slot. Verified with the PR's `Uncalled` reproducer — it compiles and prints `6` with no type mismatch. Closing as already covered in C.

Issue #1342, opened 2026-06-05
Heterogeneous multi-value return (return s,s vs return nil,nil) emits a concrete sp_Tuple_* where sp_RbVal expected


OriPekelman:
## Summary

A method with two `return a, b` paths whose element types differ — e.g. `return s, s` (String, String) on one path and `return nil, nil` on another — emits a concrete tuple type (`sp_Tuple_nil_nil *` / `sp_Tuple_string_string *`) in a return slot that expects the generic `sp_RbVal`, so `cc` fails with `incompatible types when returning`. A single-shape multi-return (`return s, s` only) compiles fine — it's the **heterogeneous** pair of tuple shapes that can't unify.

Found via the [spinelgems](https://spinelgems.org) harness (gem **yrb**, whose `key_and_value_from_line` does `return $1, $2.strip` or `return nil, nil`). Auto-reduced from the gem's smoke by `spinel-reduce` (97 → 24 lines), then hand-cleaned.

## Minimal reproducer

```ruby
def kv(s)
  return nil, nil if s.empty?
  return s, s
end
a, b = kv("x")
p [a, b]
c, d = kv("")
p [c, d]
```

```
$ spinel kv.rb
out.c: error: incompatible types when returning type 'sp_Tuple_nil_nil *' but 'sp_RbVal' was expected
spinel: C compilation failed
```

CRuby prints `["x", "x"]` then `[nil, nil]`.

## Narrowing

| method body | result |
|---|---|
| `return nil, nil if …; return s, s` (nil-tuple + string-tuple) | **`sp_Tuple_nil_nil *` vs `sp_RbVal`** |
| `return s, s` then `return nil, nil` (same, dead 2nd path) | same failure |
| `return s, s.strip` (single tuple shape) | ✅ ok |

So two `return a, b` paths produce two *concrete* tuple types (`Tuple_string_string`, `Tuple_nil_nil`); the method's unified return type stays the concrete tuple of one path while the call site / return ABI expects `sp_RbVal`, and the mismatch reaches `cc`. A single tuple shape unifies fine.

## Expected

A method returning multiple values should have a return type that unifies across all `return a, b` paths (boxing to `sp_RbVal` / a poly tuple when the element types differ), so heterogeneous multi-value returns compile — `return s, s` and `return nil, nil` from the same method is ordinary Ruby.

## Environment

- Spinel `master` (`9c0a5f0`), `aarch64-linux-gnu`, built from source.

---
[...]

Issue #626, opened 2026-05-20
Three blockers found building a Ruby tokenizer: type widening, RBS advisory-only, :str FFI alias semantics


OriPekelman:
# Spinel issue: three blockers found while building a Ruby tokenizer

**Target repo:** `matz/spinel`
**Type:** Bug report (three findings, escalating severity)
**Reproducer commit:** [oripekelman/toy_ruby_neural_network@HEAD](https://github.com/OriPekelman/toy)
**Spinel HEAD tested:** `1e2394d` (also `7beeb54`, same behaviour)

## Context

Building a standalone Ruby BPE tokenizer (`lib/tokenizer.rb`) for
the project's LLM inference path. Three independent issues block
the full implementation. Reporting together because they're all
discovered by the same scaffold and may share a root cause.

### Symptom

Compiling a file that requires both `lib/transformer.rb` (which
defines `Mat`) and `lib/tokenizer.rb` (which has its own attr_readers)
produces generated C that fails to compile:

```
out.c:XXXX:5: error: aggregate value used where an integer was expected
  lv_ta = tnn_input_2d_f32(((void *)lv_sess), ((int)(lv_a->iv_nrows)), ...);
                                                ^
spinel: C compilation failed
```

### Root cause

`Mat`'s ivars are declared as `mrb_int iv_nrows; mrb_int iv_ncols`
when only `lib/transformer.rb` is loaded. But adding
`require_relative "../lib/tokenizer"` causes the analyzer to widen
them to `sp_RbVal iv_nrows; sp_RbVal iv_ncols`. The `(int)(...)`
cast then tries to cast a `sp_RbVal` struct to int, which is illegal.

### Reproducer (minimal)

```ruby
# tinynn/_repro_min.rb
require_relative "../lib/transformer"
require_relative "../lib/tokenizer"
t = Tokenizer.from_gguf("data/smollm2-135m-tok.gguf")
puts t.vocab_size
```

Compile:
```
spinel tinynn/_repro_min.rb -o /tmp/_repro
# fails with "aggregate value used where an integer was expected"
```

`lib/tokenizer.rb` shape:
```ruby
class Tokenizer
  attr_reader :vocab_size, :bos_id, :eos_id, :pad_id, :unk_id, :present
  def initialize(vocab, bos_id, eos_id, pad_id, unk_id)
    @vocab = vocab; @vocab_size = vocab.length
    @bos_id = bos_id; ...
  end
[...]

Issue #582, opened 2026-05-18
Array element type (PtrArray vs PolyArray) doesn't widen at call boundary


rubys:
## Summary

When a method is called with two different array shapes — \`sp_PtrArray *\` (homogeneous typed-pointer array, e.g. \`[A.new, A.new]\`) at one site and \`sp_PolyArray *\` (heterogeneous boxed array, e.g. \`[A.new, \"x\"]\`) at another — spinel picks one declared parameter type (PolyArray in this repro) but doesn't insert a widening / boxing step at the PtrArray call site. C emits an incompatible-pointer-types warning, and the pointer is delivered unboxed across the call boundary — which would manifest as field misreads / corruption at runtime.

Currently \`-Wincompatible-pointer-types\` so the build proceeds past the warning, but the underlying type-soundness gap is real: \`sp_PtrArray\` and \`sp_PolyArray\` have different element layouts and the callee expects \`sp_RbVal\` slots.

## Repro (10 lines)

\`\`\`ruby
class A
end

def take(arr)
  arr.first
end

a_arr = [A.new, A.new]
mixed = [A.new, \"x\"]
take(a_arr)
take(mixed)
\`\`\`

\`\`\`
$ spinel repro.rb -o /tmp/x
out.c:67: warning: incompatible pointer types passing 'sp_PtrArray *' to parameter of type 'sp_PolyArray *' [-Wincompatible-pointer-types]
out.c:42: note: passing argument to parameter 'lv_arr' here
\`\`\`

Emitted dispatch:

\`\`\`c
static inline sp_RbVal sp_take(sp_PolyArray * lv_arr) { ... }
...
sp_take(lv_a_arr);    // ← lv_a_arr is sp_PtrArray *, function expects sp_PolyArray *
sp_take(lv_mixed);    // ← matches
\`\`\`

The PtrArray needs either coercion to PolyArray (box each element) at the call site, or the parameter type needs to be PtrArray-of-(union) and the second call site narrowed.

## CRuby

\`\`\`
$ ruby repro.rb
\`\`\`
(no output, both calls succeed.)

## Context

Surfaced in roundhouse real-blog after #573 unblocked the broader emit. Sites:

[...]

Issue #474, opened 2026-05-13
FFI: add :float_array / :int_array type specs for zero-copy bulk transfer


OriPekelman:
## Summary

Propose two new FFI type specs that let C functions consume Spinel's contiguous Array storage as plain C pointers, with the same call-duration lifetime semantics as `:str`:

| spec | C type | Spinel arg | Notes |
|---|---|---|---|
| `:float_array` | `const double *` | `Array<Float>` | Pointer to the underlying `FloatArray` `.data` |
| `:int_array` | `const int64_t *` | `Array<Int>` | Pointer to the underlying `IntArray` `.data` |

Length comes through as a separate `:size_t` (or `:int`) arg — symmetric to `:str` + `strlen`. This is strictly additive; existing FFI signatures aren't affected.

## Motivation: a real-world bottleneck

In [`OriPekelman/toy_ruby_neural_network`](https://github.com/OriPekelman/toy_ruby_neural_network) (the Spinel transformer LM), the new `lib/tinynn.rb` bridges `Mat` (row-major `Array<Float>` storage) to ggml's CUDA/CPU backend. The current FFI surface forces per-element scratch upload before each ggml call:

```ruby
i = 0
n = a.nrows * a.ncols
while i < n
  TinyNN.tnn_scratch_set(sess, i, a.flat[i])   # one FFI call per scalar
  i = i + 1
end
TinyNN.tnn_upload(sess, ta)                    # one bulk cudaMemcpy
```

For the model's largest tensor (vocab × d_model = 5000 × 32 = 160K elements), that's 160 000 FFI calls per upload, all to write a single `double` into a session-owned scratch buffer that's already C-allocated and contiguous. The per-element cost dominates the actual GPU work.

With `:float_array`, the entire upload collapses to one call:

```ruby
TinyNN.tnn_upload_from_float_array(sess, ta, a.flat, a.flat.length)
```

backed by a C function that does a single memcpy (or directly hands the pointer to `ggml_backend_tensor_set`, which itself is a single `cudaMemcpy`):

```c
int tnn_upload_from_float_array(void *sess,
                                 void *tensor,
                                 const double *data,
                                 size_t n)
{
    tnn_session *s = (tnn_session *)sess;
    /* downcast to f32 once, on the C side */
    for (size_t i = 0; i < n; ++i) s->scratch[i] = (float)data[i];
    ggml_backend_tensor_set((struct ggml_tensor *)tensor,
                            s->scratch, 0, n * sizeof(float));
[...]

PR #1389, opened 2026-06-10
codegen: convert poly_array at concrete int_array/float_array boundaries

OriPekelman:
## Regression from #1342 (d6756fa) — silent pointer pun at concrete array boundaries

Since d6756fa, an array literal whose element widened to poly emits `sp_PolyArray`. When that value reaches a slot the analyzer pinned concrete — e.g. a callee param back-propagated from an FFI `:int_array` spec, with the literal's annotation cached before the element went poly — codegen emitted a silent C pointer pun:

```c
sp_PolyArray *_t993 = sp_PolyArray_new();
sp_PolyArray_push(_t993, lv_token_id);
sp_IntArray * _t994 = _t993;   /* len offsets align; ->data is sp_RbVal[], read as int64[] */
```

A bulk-FFI consumer (#474 zero-copy `->data` hand-off) then reads boxed tags as data. Real-world hit: a ggml-backed LLM eval runner fed garbage token ids into `ggml_backend_tensor_set` and died in its shape assert (OriPekelman/spinel-dev#13; bisected to d6756fa).

## Fix

Convert at both boundaries instead of punning — the reverse of #582's ptr_array→poly_array convert:

- `compile_expr_for_expected_type`: expected `int_array`/`float_array` with a poly_array value unboxes element-wise into a fresh concrete array (`.v.i`/`.v.f`, the same unbox contract as the scalar poly FFI args, #626);
- the direct `:int_array`/`:float_array` FFI lowering routes a poly_array arg through new `sp_PolyArray_ffi_int_data/_float_data` runtime bridges instead of handing out `->data` (call-duration lifetime, unchanged contract).

## Verification

- The original regression target (toy's eval runner, 272-tensor GGUF model) compiled with this fix runs green and byte-equals its recorded logprob baseline; at master it aborts in ggml.
- Full suite: 846 pass / 0 fail. Regression test included: heterogeneous multi-return destructure → poly slot → `[poly]` literal → live `:int_array` FFI call (via `ftok`, exported by both glibc and macOS libSystem and declared in no header the runtime includes).


matz:
Thanks! Spinel is now reimplemented in C on `master`, so this patch doesn't apply directly, but the underlying pointer-pun was present in the C version too and I've fixed it there (commit b60fbd7). The :int_array / :float_array FFI lowering now hands off element data instead of casting the array struct pointer: concrete arrays via zero-copy accessors (honoring IntArray's start offset), poly_arrays via new sp_PolyArray_ffi_int_data / _float_data bridges that unbox element-wise into a fresh call-duration buffer (sp_gc_alloc_nogc, so a sibling array arg can't be swept mid-build). Verified a poly_array and a concrete int_array reaching an :int_array call emit the bridges (not a pun) and run end-to-end via ftok; full suite + optcarrot + bench green. Closing as ported to C.

PR #1390, opened 2026-06-10
codegen: anchor sym_json_escape input as string for the self-host build

OriPekelman:
The **self-hosted** `spinel_codegen` binary emitted `--emit-symbol-map` (#1345) with every string field empty:

```json
{"c":"","ruby":"","kind":"imeth","file":"","line":2}
```

CRuby-run codegen was correct, so the bug only showed in shipped toolchains. Mechanism: under self-host inference the `@file_table` element read in `symbol_file_path` lands on poly (the table is filled in `compiler_helpers` across the require boundary), poisoning its return — and through it `sym_json_escape`'s param — to poly with no concrete arm; `s.length`/`s[i]` emit 0 (the build's own `cannot resolve call to 'length' on poly ... (emitting 0)` warnings were the fingerprint) and the escape loop returns "".

Fix: `.to_s` anchors at the table read and the escape input. After it, the self-host build is warning-free at `sym_json_escape` and the binary emits populated fields — including `file`/`line`, which were also casualties (the poisoned table read never yielded the path):

```json
{"c":"sp_Foo_save_p","ruby":"Foo#save?","kind":"imeth","file":"symmap_demo.rb","line":2}
```

Suite 845 pass / 0 fail. Found wiring perf tools to consume the map (OriPekelman/spinel-dev#16).

### DIFF
--- a/spinel_codegen.rb
+++ b/spinel_codegen.rb
@@ -9606,7 +9606,14 @@ class Compiler
 
   def symbol_file_path(fid)
     if fid != nil && fid >= 0 && fid < @file_table.length
-      p = @file_table[fid]
+ # .to_s anchor: under self-host the @file_table element read can
+ # land on poly (the table is filled in compiler_helpers across the
+ # require boundary), which poisons this return — and with it
+ # sym_json_escape's param — to poly with no concrete arm, so the
+ # escape loop emitted 0 and every --emit-symbol-map string field
+ # came out "" from the self-hosted binary (CRuby codegen was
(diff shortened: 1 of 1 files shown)


matz:
Thanks! This addressed a self-host inference artifact specific to the Ruby implementation (`@file_table` reads going poly across the require boundary, poisoning `sym_json_escape`'s param). The C version (now `master`) has no self-host inference pass, and `--emit-symbol-map` already emits fully-populated fields. Closing as no longer applicable.

PR #1391, opened 2026-06-10
runtime: raise Errno::ENOENT on File.read/write/binread open failure

OriPekelman:
`File.read` on a missing path silently returned `""` (CRuby: raises `Errno::ENOENT` at the call site). `File.write` to an un-openable path silently dropped the write; `File.binread` returned an empty byte array.

The silent empty value is close to the worst debugging experience for a one-line user error: it flows deep into the program until something dereferences garbage — in the wild this surfaced as a zero-output SIGSEGV (exit 139) in an ML data pipeline whose only actual problem was a missing fixture file (OriPekelman/spinel-dev#17).

Fix: raise on a failed `fopen` in all three entry points, using the exact shape `File.mtime` already uses — `Errno::ENOENT` (or `EACCES` / `RuntimeError` per errno) with the `strerror` text and the `rb_sysopen`-style location CRuby prints. `Errno::EISDIR` guards and success paths unchanged; `File.exist?` keeps probing with fopen.

Suite 846 pass / 0 fail (845 + a regression test covering the raise on read and write plus the surviving round-trip).

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -3774,7 +3774,14 @@ static const char *sp_file_read(const char *path) {
     sp_raise_cls("Errno::EISDIR", sp_sprintf("Is a directory @ io_fread - %s", path));
   }
   FILE *f = fopen(path, "r");
-  if (!f) return &("\xff" "")[1];
+  /* CRuby contract: a failed open raises (Errno::ENOENT for a missing
+     path), it does not yield "". The silent empty string let a missing
+     input flow deep into programs until something dereferenced garbage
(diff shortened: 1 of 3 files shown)


matz:
Thanks! Same situation as the other open PRs — Spinel is now reimplemented in C on `master`, so this won't apply directly, but the missing behavior was present in the C runtime as well. Fixed there: `File.read` / `File.write` / `File.binread` now raise `Errno::ENOENT` / `Errno::EACCES` (or `RuntimeError`) on open failure instead of silently returning an empty string/array. Closing as ported to C.

PR #1393, opened 2026-06-10
Root fiber and proc capture cells during closure construction

fc-anjos:
## The bug

`Fiber.new` and proc/lambda creation with captured locals lower to a run of back-to-back `sp_gc_alloc` calls: one heap cell per captured variable (plus a self cell), then the capture struct, all before the closure object itself is allocated and rooted. None of those intermediates is rooted, so a collection triggered by any allocation inside that window sweeps the already-allocated cells. The assembled closure then reads and writes freed memory on first resume/call.

The failure is allocation-pressure dependent, so small programs never hit it: a compiled HTTP server creating one fiber per connection died deterministically after ~297 fresh connections (SIGSEGV at `-O2`; `undefined method ... for nil` at `-O0` once the freed cell was recycled by a zeroing allocation).

## The fix

Emit `SP_GC_ROOT` for each capture heap cell, self cell, and capture struct immediately after its `sp_gc_alloc`, at all seven emission sites (fiber, lambda, and proc paths). This is the same rooting discipline #670 applied to value-type fields.

## Tests

Two new regression tests, `test/fiber_capture_cells_gc_during_creation.rb` and `test/proc_capture_cells_gc_during_creation.rb`: each round captures five small objects into a closure and almost nothing else allocates, so nearly every GC trigger lands inside the construction window. Both segfault before the fix (checked-in `.expected` files, so no reference-Ruby dependency) and print the deterministic `total=6250375000` after.

## Validation

- Both new tests fail (SIGSEGV) on unpatched master, pass with the fix
- Full suite: 847 pass / 0 fail / 0 error; RBS 13 pass; analyze-fail 6 pass
- The originating HTTP server, rebuilt with the fixed compiler, survives 1000 fresh connections (died at ~297 before)

## Follow up?

The seven roots are intentionally mechanical: each is locally verifiable next to its alloc. Two cleaner shapes are possible if you'd prefer: (a) allocate the capture struct first, root only it, and store each cell into its slot as it's allocated, so cells are reachable from a rooted parent the moment they exist; or (b) extract the three duplicated capture-construction sequences into one helper that owns the alloc+root pairing so the invariant can't be missed at a future call site. Happy to do either as a follow-up or rework this PR.


matz:
Thanks — this was a genuine GC bug in the C version too. The proc capture struct (`_capv`) was allocated and handed to `sp_proc_new_meta` (which itself allocates and can trigger a collection) without a root, so it could be swept mid-construction and `.call` would dereference a freed pointer. Fixed in the C codegen by emitting `SP_GC_ROOT` for the capture struct right after allocation. Closing as ported to C.

PR #1395, opened 2026-06-12
runtime: add sp_PolyArray pool with manual release for NO_GC interpreters

companygardener:
## Summary

Adds `sp_PolyArray_pool_acquire()` and `sp_PolyArray_pool_release()` to `lib/sp_runtime.h` — a fixed-cap LIFO pool for short-lived `sp_PolyArray`s. Existing `sp_PolyArray_new` and the rest of the PolyArray API are unchanged; this is strictly additive.

## Problem

Self-hosted compilers and tree-walking interpreters compiled by Spinel typically have a hot path where every method call constructs a fresh PolyArray to hold the evaluated argument list, passes it to a callee that consumes the values via a `bind_params`-style copy, and then drops the array. Under `-DSPINEL_NO_GC` (the recommended mode for bootstrap throughput, since the GC sweep is otherwise the dominant cost in a non-conservative tree-walker), every one of those allocations is a permanent ~448 B leak (16 sp_RbVal cap = 384 B + `sp_gc_hdr` 40 B + `sp_PolyArray` header 24 B).

For a Tungsten self-host stage 1 pass (the spinel-built interpreter running compile() against the compiler's own source, ~300 K method calls), the resulting dirty-page traffic is what macOS jetsam reads as memory pressure. The process is SIGKILL-ed at ~5 min CPU on a 128 GB box even though peak RSS stays under 90 GB, because jetsam's heuristic tracks dirty-page write rate, not just resident bytes. Verified by lldb: stop reason \`SIGKILL\` inside \`libsystem_malloc.dylib_xzm_xzone_thread_cache_fill_and_malloc\`; no diagnostic report, indicating an external kill rather than OOM.

## API

\`\`\`c
static inline sp_PolyArray *sp_PolyArray_pool_acquire(void);
static inline void          sp_PolyArray_pool_release(sp_PolyArray *a);
\`\`\`

These do not require GC and operate independently of \`sp_gc_bytes\`. The intended use is bounded-lifetime arrays whose escape is locally provable (the canonical case: arguments arrays consumed inside the call frame that built them). The caller is responsible for never releasing an array that retains live references — the API gives no safety net here; this is a perf primitive, not a managed handle.

Backed by a fixed-cap LIFO (default 4096, override via \`-DSP_POLY_ARRAY_POOL_CAP=...\`).

## Measured impact

Tungsten stage 1 — the spinel-built tungsten-stage0 interpreter binary running the Tungsten compiler against the compiler's own source:

| metric                     | before pool   | after pool    |
|----------------------------|---------------|---------------|
| RSS at 3:00 elapsed        | ~80 GB        | ~37 GB        |
| Survival past jetsam kill  | no (~5 min)   | yes (>15 min) |
| \`sp_PolyArray_new\` calls   | ~1 per method call | ~0 after warmup |

The pool lifts the jetsam ceiling that was killing every long-running interpreter pass.

## Test plan

- [x] Existing test suite passes: 998 pass / 0 fail / 0 error against current master.
- [x] Verified the helpers work in a downstream stress test (Tungsten stage 1 — see metrics above).
- [ ] Could add a runtime-level unit test exercising acquire/release/reacquire round-trips if preferred.

## Notes

Downstream Tungsten currently wires the helpers via a post-codegen C-rewrite script (because exposing them from user Ruby code requires codegen integration). A follow-up issue could discuss whether a codegen-recognized intrinsic (`__sp_pool_release_poly_array(x)` or a `force_class_method_param_pool_release` override) would make sense, so user Ruby code can request pool wiring without a rewrite step.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -2771,6 +2771,52 @@ typedef struct { sp_RbVal *data; mrb_int len; mrb_int cap; mrb_int frozen; } sp_
 static void sp_PolyArray_scan(void *p) { sp_PolyArray *a = (sp_PolyArray *)p; for (mrb_int i = 0; i < a->len; i++) sp_mark_rbval(a->data[i]); }
 static void sp_PolyArray_fin(void *p) { sp_PolyArray *a = (sp_PolyArray *)p; sp_gc_hdr *h = (sp_gc_hdr *)((char *)a - sizeof(sp_gc_hdr)); sp_gc_bytes -= sizeof(sp_RbVal) * a->cap; h->size -= sizeof(sp_RbVal) * a->cap; free(a->data); }
 static sp_PolyArray *sp_PolyArray_new(void) { sp_PolyArray *a = (sp_PolyArray *)sp_gc_alloc(sizeof(sp_PolyArray), sp_PolyArray_fin, sp_PolyArray_scan); a->cap = 16; a->data = (sp_RbVal *)malloc(sizeof(sp_RbVal) * a->cap); if (!a->data) sp_oom_die(); a->len = 0; { sp_gc_hdr *h = (sp_gc_hdr *)((char *)a - sizeof(sp_gc_hdr)); h->size += sizeof(sp_RbVal) * a->cap; sp_gc_bytes += sizeof(sp_RbVal) * a->cap; } return a; }
+
+/* PolyArray free-list pool (manual recycle, GC-independent).
+
+   sp_PolyArray_pool_acquire() pops a pre-cleared PolyArray off a
+   process-global LIFO; if the pool is empty it falls back to
+   sp_PolyArray_new(). sp_PolyArray_pool_release(a) clears `a` and
+   pushes it back (capped by SP_POLY_ARRAY_POOL_CAP, default 4096).
+
+   Intended use: short-lived arrays whose lifetime is bounded by the
+   enclosing call frame. The canonical case is tree-walking
+   interpreters that evaluate per-call argument lists (e.g.
+   `evaluate_args(node)` returning a fresh PolyArray that the called
+   method consumes via a `bind_params` step and never retains). In
+   that pattern the pool eliminates one ~448 B alloc (16 sp_RbVal
+   cap + sp_gc_hdr + sp_PolyArray header) per call site, which under
+   SPINEL_NO_GC=1 is otherwise an unrecoverable leak driving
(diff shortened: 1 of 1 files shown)


companygardener:
Thanks @copilot and @gemini-code-assist — addressed the load-bearing items in e4e9f96.

**Fixed**
- **GC UAF** (critical, both reviewers): added `sp_PolyArray_pool_scan_roots()` called from `sp_re_mark_globals` so pool entries are marked during GC.
- **`a->frozen` reset** (Copilot): cleared in both acquire and release.

**Noted, not changed**
- **Per-TU static state**: `sp_runtime.h` is included by exactly one generated TU. The other runtime helpers (`lib/sp_*.c`) re-declare their slice of the ABI rather than re-including the header, so there's no fan-out. Calling that out in case the architecture changes.
- **Thread safety**: legitimate concern, but the pool is scoped to single-threaded use as the doc states. Happy to thread-local it or add a lock in a follow-up if you'd prefer the API admit multi-threaded callers.

matz:
Thanks for the detailed writeup and metrics. After evaluating this against the current tree, I'm going to decline it as-is — not because the pooling idea is wrong, but because it doesn't have a home in the compiler today:

- **No in-tree caller.** Spinel's codegen never emits `sp_PolyArray_pool_acquire/release`; as you note, Tungsten wires them via a post-codegen C-rewrite script. Landing the helpers in `lib/sp_runtime.h` adds a public API that nothing in this repo calls — dead code that a `static` definition would also surface as an unused-function warning under our `-Werror` test builds.
- **The motivating workload isn't on the C compiler's path.** The self-hosted-under-`-DSPINEL_NO_GC` hot loop (every method call leaking a ~448 B arg array) was the old Ruby self-host's shape. The compiler is now a from-scratch C reimplementation and doesn't run that way, so the jetsam ceiling this lifts isn't one we hit in-tree. (We've also seen size-class pooling on `sp_str_alloc` regress throughput ~15%, so we're cautious about pools without a measured in-tree win.)

The right path is the one you already flagged: a **codegen-recognized intrinsic** (e.g. recognizing the `def m(*args); ...; end` arg-array pattern and emitting acquire/release around it) so user Ruby gets pool wiring without a rewrite step. That makes the helpers a live, measured part of the pipeline rather than an uninvoked API. If you'd like to pursue that, happy to discuss the shape on a fresh issue — and if a concrete in-tree consumer lands, the pool itself is easy to add then. Closing for now; thanks again.

Issue #282, opened 2026-05-04
Language implemented as a Ruby interpreter is 30x slower


companygardener:
This is more of a discussion than a specific issue. I'm working on a programming language that bootstraps via a stage-1 Ruby interpreter, and I've been landing changes upstream in order to get spinel to compile it. With those patches, plus a mix of builtins and source rewrites, I now have the interpreter compiling end-to-end.

The catch: the spinel-compiled tree-walking interpreter runs ~30x slower than CRuby running the same code. My hypothesis is that the workload is dominated by AST dispatch and env lookups - every variable reference walks a string-keyed hash and does strcmp; every node visit recurses through full polymorphic dispatch with GC root tracking - whereas spinel's example benchmarks tend to be tighter numeric/algorithmic code that compiles to clean C.

CRuby wins here because YARV is a byte code VM with interned-symbol environments, inline caches, and stack-allocated frames; an AOT-compiled tree walker doesn't get any of those wins for free.

Going to keep iterating and landing patches upstream - just wanted to share the data point in case it's useful or sparks ideas.

matz:
Thanks for the data point — useful even as a discussion. The 30× gap roughly matches what we'd expect from the architectural mismatch:

- Spinel's wins come from static type specialization — typed arrays, attr_accessors that lower to struct fields, scoped locals as C variables, monomorphic dispatch when class identity is statically resolvable. A tree-walking interpreter is the inverse: untyped `Node` hierarchy with virtual `eval`, and an env that's a `String → Object` hash. None of those compile to anything tighter than the obvious `sp_PolyArray_get` + tag-check + GC root push per access.
- CRuby wins on this kind of workload not because of CRuby per se but because of YARV's `getlocal` slot indexing + inline caches. The factor between an MRI 1.8 tree-walking interpreter and a YARV bytecode VM was already ~2× before AOT compilation entered the picture, and AOT-compiling a tree walker doesn't claw that back — the tree-walker overhead is the algorithmic shape, not the dispatch overhead.

A few changes you can make on your side that should move the needle:

- Replace the string-keyed env with a name-resolution pass that turns each variable reference into a slot index, then store the env as an `IntArray` (or a typed array of whatever your value representation is). `arr[k]` lowers cleanly in spinel; `hash[name_string]` doesn't.
[...]

PR #1396, opened 2026-06-12
codegen: pointer-eq pre-check before strcmp in case-on-string dispatch (lever 4 from #282)

companygardener:
## Summary

A starter implementation of one of the four perf levers we discussed in #282 — the "integer dispatch for hot AST node types" lever, applied to case-on-string dispatch arms.

## What changed

**Runtime** (lib/sp_runtime.h): adds \`sp_str_intern(s)\` — a linear-probing string-intern pool keyed by content. First call for each unique content allocates a canonical copy; subsequent calls return the same pointer. Capped at \`SP_STR_INTERN_CAP\` (default 1024), monotonic-growing (no eviction). Intended for short, long-lived strings like case-arm keys; not for per-call temporaries.

**Codegen** (spinel_codegen.rb, \`compile_when_conds\`): wraps the strcmp emission for both poly-string and direct-string predicate arms as:

\`\`\`c
(tmp == lit || strcmp(tmp, lit) == 0)
\`\`\`

instead of just \`strcmp(tmp, lit) == 0\`.

## Why this helps

With clang's \`-fmerge-all-constants\` (default at -O2+), two source-level string literals with identical content collapse to the same C string pointer at link time. When the predicate value originated from a literal — which is the case for many AST-walking patterns where the parser populates \`node["type"] = "method_def"\` from a string literal that's also the case-arm key — the pointer comparison decides in one cycle. \`strcmp\` only runs when the literals genuinely differ.

The pre-check adds one branch on the slow path (when pointer-eq fails); the strcmp fallback preserves correctness for any non-interned input.

## Test plan

- [x] Existing test suite: 998 pass / 0 fail / 0 error.
- [x] Smoke test (case k when "method_def" then 1 / when "class_def" then 2 / when "fn_def" then 3):
  \`\`\`c
  if ((_t1 == (&("\\xff" "method_def")[1]) ||
       strcmp(_t1, (&("\\xff" "method_def")[1])) == 0)) {
    return 1LL;
  }
  \`\`\`
  Output 2/1/3/0 matches expected.
- [x] Verified emission across a real spinel-compiled program (the Tungsten stage 0 interpreter binary, which is ~30 KLOC of Ruby compiled to ~1.5 MB of C): ~82 of 84 strcmp sites in case-on-string dispatches now have the pointer-eq pre-check.

## Notes on #282

This is the first concrete spinel-side implementation of one of the four levers from #282 ("symbol interning / inline cache / stack-promote / integer dispatch"). The other three (slot env, per-call IC, stack-promote env) are larger and remain to be tackled separately — happy to coordinate on the order if useful.

The interning pool's API is intentionally minimal. It's enough for the AST-walker pattern (a small key alphabet that lives for the duration of the compile) but not production-grade for general use. If you'd prefer a different shape (hash table, per-class pool, or codegen-recognized intrinsic instead of runtime call), I'm happy to iterate.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -1282,6 +1282,13 @@ static const char *sp_str_intern(const char *s) {
   if (sp_str_intern_count >= SP_STR_INTERN_CAP) return s;
   char *copy = sp_str_alloc(sl);
   memcpy(copy, s, sl);
+  /* sp_str_alloc returns a buffer with a marker byte at [-1] and the
+     content slot; we explicitly null-terminate at [sl] to make the
+     C-string contract independent of any future sp_str_alloc
+     internal change. Otherwise sp_str_intern's returned pointer
+     could feed into strcmp / strlen with undefined trailing bytes
+     if sp_str_alloc's invariant ever loosens. */
+  copy[sl] = '\0';
   sp_str_intern_pool[sp_str_intern_count++] = copy;
   return copy;
 }
@@ -2086,6 +2093,14 @@ static void sp_re_mark_globals(void) {
   sp_mark_string(sp_re_match_pre);
   sp_mark_string(sp_re_match_post);
   for (mrb_int i = 0; i < sp_argv.len; i++) sp_mark_string(sp_argv.data[i]);
+  /* Strings live in sp_str_intern_pool are owned by the pool, not
+     reachable through any user-visible root. sp_str_sweep would
+     otherwise free them on the next GC cycle, leaving dangling
+     pointers in the pool. Mark each pool entry so the str heap
+     keeps them alive while the pool holds a reference. */
+  for (int i = 0; i < sp_str_intern_count; i++) {
+    sp_mark_string(sp_str_intern_pool[i]);
+  }
   sp_mark_in_flight_exceptions();
   sp_mark_fiber_root_storage();
 }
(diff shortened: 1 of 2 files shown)


companygardener:
Thanks @copilot and @gemini-code-assist — addressed the critical items in f7746dd.

**Fixed**
- **GC UAF** (gemini, critical): `sp_re_mark_globals` now marks each `sp_str_intern_pool` entry via `sp_mark_string`. Without this, `sp_str_sweep` would free interned strings and leave dangling pointers in the pool.
- **Null terminator** (Copilot): explicitly write `copy[sl] = '\\0'` after memcpy. Removes the implicit dependency on `sp_str_alloc`'s allocation-and-termination invariant.
- **Double evaluation** (Copilot): hoisted `compile_expr(cid)` into a C statement-expression temp (`{ const char *_sl = LIT; (... == _sl || strcmp(..., _sl) == 0); }`). Today's case-arms are always literals so this is a no-op behaviorally, but it's future-proof if `compile_expr` ever expands to something with side effects.

**Noted, not changed**
- **Per-TU static state**: same as #1395 — `sp_runtime.h` is single-TU by design.
- **Thread safety**: legit but out of scope for this PR; the pool is single-threaded as documented.

matz:
Thanks — the case-on-string dispatch lever is now on master (193e974), implemented as a pointer-equality fast path inside the runtime `sp_str_eq` itself:
```c
static inline int sp_str_eq(const char*a,const char*b){if(a==b)return 1;if(!a||!b)return 0;return strcmp(a,b)==0;}
```
`case x when "foo"` lowers to `sp_str_eq(x, "foo")`, so under `-fmerge-all-constants` a literal-flow subject short-circuits exactly as your pre-check did — and every `sp_str_eq` caller benefits, not just case arms, with no per-site codegen change.

I left the `sp_str_intern` pool out for now: it needs codegen to intern subjects to pay off, and the literal-merge path already covers the common case. The spinel compiler here is a from-scratch C reimplementation, so the `spinel_codegen.rb` patch didn't apply directly — but your diagnosis and the lever are what landed. Closing this PR as implemented; #282 stays open for the other three levers.

Issue #1398, opened 2026-06-15
Regression (f6d5eef→b60fbd7): if/else with mismatched branch value-types (File.open{block} vs $stderr.puts) emits malformed C — int vs sp_RbVal


OriPekelman:
## Summary

A method whose value is an `if/else` where the two branches have **different value-types** — one branch `File.open(path, mode) { |f| ... }`, the other `$stderr.puts(...)` — now emits **malformed C** on `master` (`b60fbd7`). It compiled and ran correctly at `f6d5eef`. This is a return/value-position **type-unification regression** in the 763-commit window `f6d5eef..b60fbd7`.

It blocks bumping tep's spinel pin past `f6d5eef`: `SpinelKit::Json`/`SpinelKit::Log` (vendored into tep) use this exact pattern, so the whole lib fails to compile on master. (Filing per the cross-team workflow — tep is verified-green at `f6d5eef`; this is one of the regressions standing between the pin and master, now that #1369 and #1382 are closed.)

## Minimal repro (10 lines)

```ruby
def w(path, line)
  if path.length > 0
    File.open(path, "a") { |f| f.puts(line) }
  else
    $stderr.puts(line)
  end
end
w("/tmp/vmin.out", "hi")
puts "done"
```

- **`f6d5eef`**: compiles, runs, prints `done`. ✅
- **`b60fbd7` (master)**: ❌
  ```
  vmin.rb:5:35: error: expected ‘)’ before ‘;’ token
  vmin.rb:5:13: error: incompatible types when returning type ‘int’ but ‘sp_RbVal’ was expected
  spinel: C compilation failed
  ```
  (Line 5 = the `$stderr.puts(line)` else-branch; the emitted C for the branch can't satisfy the method's unified `sp_RbVal` return.)

## Isolation

- `File.open(path,"a"){|f| f.puts(line)}` **alone** (no if/else) → compiles+runs fine on master.
- `if/else` where **both** branches are `File.open{block}` (matching types) → fine on master.
- Only the **mismatched-branch-type** `if/else` (typed branch vs void/nil `$stderr.puts`) regresses. So this is value-position type unification across `if/else` when one arm yields a void/nil-ish value.

## Likely-same-family second site (from the real tep build)

Compiling tep against master also fails at a void-FFI call used in value position:
```ruby
if attempt < policy.max_attempts
  backoff = policy.backoff_for(attempt - 1)
  if backoff > 0
[...]

PR #1402, opened 2026-06-15
codegen: emit $stdout/$stderr.puts/print as one expression (valid in value position)

OriPekelman:
## The bug

The `$stdout`/`$stderr` `puts`/`print` lowering in `src/codegen_call.c` joins the per-arg `fputs(...)` and the trailing `fputc('\n', fd)` with a **`;`** — a statement sequence. That's fine in statement position, but the call can land in **value/return position** (e.g. the `else` arm of an `if/else` whose other arm yields a boxed value), where the surrounding emitter wraps it as `(<lowering>, sp_box_nil())`. The `;` then makes invalid C:

```c
return (fputs(lv_line, stderr); fputc('\n', stderr), sp_box_nil());
//                            ^ error: expected ')' before ';'
//        error: incompatible types when returning 'int' but 'sp_RbVal' was expected
```

### Minimal repro (`test/stdout_puts_value_position.rb`, included)

```ruby
class Out
  def initialize(p) ; @p = p ; end
  def write(line)
    if @p.length > 0
      File.open(@p, "a") { |f| f.puts(line) }   # boxed-value branch...
    else
      $stdout.puts(line)                         # ...forces this into value position
    end
  end
end
Out.new("").write("hello")
$stdout.print("done")
$stdout.puts("")
```

Each piece alone (bare `$stdout.puts`, the `File.open` block, the `if/else` of two plain `puts`) compiles; the bug needs `puts`/`print` in a value position where the result type is boxed. Surfaced on [tep](https://github.com/OriPekelman/tep)'s `Logger#log`, which blocked compiling a real downstream app's server.

## The fix

Emit a **single comma-operator expression** — join the args and the trailing-newline `fputc` with `,` instead of `;`. Valid in both statement and value position, and it also supplies the arg separator that was previously missing between multiple `puts` args (`fputs(a, fd)fputs(b, fd)` → `fputs(a, fd), fputs(b, fd)`). A bare `$stdout.print` (no args) emits `(void)0`.

## Verification

- New test fails pre-fix with the exact `expected ')' before ';'` + `int` vs `sp_RbVal` errors; passes after.
- Full suite: **849 pass / 0 fail / 0 error** (aarch64).

### DIFF
--- a/src/codegen_call.c
+++ b/src/codegen_call.c
@@ -7935,14 +7935,22 @@ else {
       const char *fd = is_err ? "stderr" : "stdout";
       if (!strcmp(name, "puts") || !strcmp(name, "print")) {
         int want_nl = !strcmp(name, "puts");
+        /* Emit a SINGLE C expression, joining the pieces with the comma
+           operator (not ';'). $stderr.puts/print can appear in value/return
+           position, where the surrounding emitter wraps this as
+           `(<here>, sp_box_nil())`; a ';' produced invalid C
+           (`(fputs(..); fputc(..), sp_box_nil())`). The comma operator also
+           gives the missing separator between multiple args. */
+        int emitted = 0;
         for (int k = 0; k < argc; k++) {
+          if (emitted++) buf_puts(b, ", ");
           TyKind at = comp_ntype(c, argv[k]);
           if (at == TY_STRING) { buf_printf(b, "fputs("); emit_expr(c, argv[k], b); buf_printf(b, ", %s)", fd); }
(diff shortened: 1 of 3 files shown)


matz:
Already fixed on master by cb4dafe (filed as #1398). The `$stdout`/`$stderr` `puts`/`print` lowering now joins its parts with the comma operator so it is a single valid C expression in value/return position (and multi-arg output is fixed). Your repro compiles and runs correctly on master:
```
$ spinel repro.rb -E
hello
done
```
Thanks for the precise diagnosis — it matched the fix exactly. Closing as already-resolved.

PR #1407, opened 2026-06-15
codegen: coerce a poly Array.new size operand to int

OriPekelman:
## The bug

`Array.new(size, fill)` (and `Array.new(size)` / `Array.new(size) { ... }`) lower the size into an `mrb_int` temp via plain `emit_expr`. When the size expression is **poly** — e.g. `nrows * ncols` where a factor widened to poly, so it emits `sp_poly_mul(...)` (returns `sp_RbVal`) — the result is assigned straight into the `mrb_int` temp:

```c
mrb_int _t2 = sp_poly_mul(sp_box_int(lv_nrows), lv_ncols);   // sp_RbVal -> mrb_int
// error: incompatible types when initializing type 'mrb_int' {aka 'long int'} using type 'sp_RbVal'
```

The array **element** type is inferred correctly (float, from the `0.0` fill — the field is emitted `sp_FloatArray *`); only the **size operand** is uncoerced.

### Minimal repro (`test/array_new_poly_size.rb`, included)

```ruby
class Mat
  def initialize(nrows, ncols)
    @flat = Array.new(nrows * ncols, 0.0)
  end
  def len ; @flat.length ; end
end
class Other
  def go
    Mat.new(1, @d)     # @d untyped -> poly; widens Mat#initialize's ncols to poly
  end
end
m = Mat.new(2, 3)
puts m.len             # => 6
```

`Other#go` is never called, but param inference scans its call site, so `ncols` widens to poly; combined with the concrete `Mat.new(2, 3)`, `nrows * ncols` becomes an `sp_poly_mul`.

## The fix

Emit the size via `emit_int_expr`, which wraps a poly value in `sp_poly_to_i` — the same coercion already used at other array-bound sites (indices, range bounds). Applied to all three `Array.new` size emissions (1-arg, 2-arg, block).

## Verification

- New test fails pre-fix with the exact `'mrb_int'`/`'sp_RbVal'` error; passes after.
- Full suite **864 pass / 0 fail** (aarch64).
- Real-world: this is the blocker that kept [toy](https://github.com/OriPekelman/toy)'s `Mat` substrate (`@flat = Array.new(nrows * ncols, 0.0)`) — and thus its entire eval/serve/infer/train compute surface — from compiling on master. With the fix, `toy eval` compiles clean.


OriPekelman:
For provenance: the 12-line repro wasn't hand-written — it was **mechanically reduced** from toy's full failing surface, which is how I'm confident it's the minimal trigger (and not an incidental shape).

The bug only reproduced on the full multi-file compilation unit (an isolated `Array.new(n, 0.0)` types `float_array` correctly — the poly-size widening needs the cross-file call graph). So single-file ddmin couldn't touch it. I used a multi-file reducer that does `git`-style delta-debugging across the real `require_relative` graph, with the project's own build (`spinel --rbs sig <entry>`) as the oracle and the **error pinned by a regex** (`transformer.rb:\d+ … incompatible types … 'mrb_int' … 'sp_RbVal'`, line wildcarded so the reduction stays faithful to the site as code shifts around it):

```
15 files / ~thousands of lines
  → pass 1 (drop whole files):      7 files
  → pass 2 (drop def/class blocks):  7 files / ~120 lines
  → pass 3 (drop remaining lines):   the 12-line standalone in this PR
```

That reduction is what corrected my initial guesses (I'd suspected an element-type / `@gamma` issue and a dead anchor-block) and pinned it to the **size operand** — `sp_poly_mul(...)` → `sp_RbVal` assigned into an `mrb_int` temp — which made the one-line `emit_int_expr` fix obvious.

The reducer (and the migrate/probe/bisect tooling around it) lives in [OriPekelman/spinel-dev](https://github.com/OriPekelman/spinel-dev) (`tools/reduce/spinel-reduce-project.rb`); happy to fold the repro-derivation into the test header if useful.

PR #1412, opened 2026-06-15
feat(proc): return a Float from a proc/lambda through `.call`

ryanseys:
## What

A proc/lambda whose body evaluates to a `Float` now carries that float back through `.call`. Previously spinel rejected it at compile time (`unsupported proc with float/range/time return (later slice)`).

```ruby
f = proc { 1.5 }
puts f.call       # was: compile error; now: 1.5
puts f.call + 0.5 # 2.0
```

## How

Float returns ride the **existing** poly return slot (`_sp_proc_poly_ret`) rather than a new ABI slot:

- **`src/codegen.c`** (`emit_proc_literal`): a `TY_FLOAT` return is stored into the slot via the poly-result path (boxed with `sp_box_float`), instead of deferring.
- **`src/codegen_call.c`** (`<proc>.call` / `.()` / `[]`): when the call's result type is `TY_FLOAT`, read the slot back and unbox with `sp_poly_to_f`.

`range`/`time` returns still defer (no fixture exercises them); the deferral message is narrowed accordingly.

## Scope / known limitation

Covers the **direct-literal**, **proc-variable**, and **method-tail** `.call` forms. A proc stored in a *poly container* and called there (`[proc { 3.5 }][0].call`) goes through a different poly-receiver `.call` path that reads the raw slot — that path is **already** lossy for non-int returns on master (a stored *string* proc mis-prints there too), so this change leaves it consistent rather than introducing a new behavior. Unifying that path is separate work.

The single-slot `_sp_proc_poly_ret` hazard (two poly/float-returning proc calls in one expression) is unchanged and out of scope; the fixture keeps one such call per expression.

## Tests

`test/proc_float_return.rb` — direct literal, proc-var, lambda-var, float expression in the body, float arithmetic at the call site, and a method whose tail is a float-returning proc call. Expected output generated from reference Ruby.

## Verification

- `make test` → 864 pass, 0 fail (no regressions).
- Pure `src/` change (no `lib/` ABI touch), so the bootstrap fixpoint is unaffected.


matz:
Thanks @ryanseys! Merged. Reviewed the poly-slot reuse (`sp_box_float` store via `g_result_poly`, `sp_poly_to_f` read-back) and confirmed the no-arg / proc-var / method-tail forms against CRuby. Full test/bench/optcarrot suite green. Your scoping is accurate: the float-lambda-*with-args* case hits the pre-existing `mrb_int[16]` proc-arg ABI limit (poly params / float args bad-C identically on master, independent of float returns) -- that's separate work, correctly left out here.

PR #1414, opened 2026-06-16
Support positional arguments in sprintf/% formatting

mattn:
Add support for positional argument references (%N$conv) in sprintf-style formatting.


matz:
Thanks @mattn! Merged. Verified the positional path covers every `%` formatting route (typed arrays and single args all go through `sp_str_format_polyarr`), confirmed against CRuby including width-not-positional (`%05d`), repeats (`%1$s%1$s`), `%%` interleaving, and out-of-range indices. Full test/bench suite green.

PR #1448, opened 2026-06-17
Ractor Milestones 3–6: spawn args, Hash/object messages, isolation, Ractor::Port

OriPekelman:
Follows up the merged minimal Ractor (#1333) through the roadmap. The deep-copy message codec now covers the full common value set, plus a shared channel type.

- **M3a — spawn arguments**: `Ractor.new(a,b){|x,y|}` deep-copies args into the block params.
- **M3b — Hash messages**: string/symbol/poly-keyed (+ typed variants), rebuilt as a poly hash in the receiver.
- **M4 — user-object/struct messages**: deep-copied by `cls_id` (a stable, Ractor-shared class index) + recursive ivars; codegen emits per-class `nivars`/`getivar`/`alloc`/`setivar` hooks the runtime codec drives. Non-copyable-ivar classes / exceptions / reopened builtins report `nivars -1` and raise rather than miscompile.
- **M5 — `@@cvar`/`$gvar` isolation**: accessing a class variable or user global inside a Ractor is rejected at compile time with `Ractor::IsolationError` (mirrors CRuby; thread-local/read-only specials like `$~`/`$stdout` are allowed).
- **M6 — `Ractor::Port`**: a process-shared mutex+condvar channel (`new`/`send`/`receive`/`close`), passed across the boundary **by reference** (the channel pointer travels, not a copy) so producers/consumers on different Ractors rendezvous. Registry-owned for the process lifetime; added to the GC-mark skip list.

A message can now be Integer/Float/bool/nil/Symbol/String/Array/Hash/user-object (recursively), and Ports are shared by reference.

Verification: full suite **933 pass / 0 fail / 0 error** (6 Ractor tests); every demo is GC-verify-, valgrind memcheck-/helgrind-, TSan- and ASan-clean. Earlier review fixes included.

Two documented sharp edges (operator/method overload collisions, both with simple workarounds): `x.send("literal")` parses as a reflective send (use a variable); `port << v` can be mis-inferred as an Array push (use `port.send`).


OriPekelman:
Thanks — good catches on the leak-on-exception paths. Fixed in the latest push:

- **`sp_Ractor_new_args` calloc-fail**: now frees the (ownership-transferred) spawn-arg blob array before raising.
- **`SPR_STR_HASH` deser**: drops the malloc entirely — reads the key from the reader buffer and roots it (boxed) across the value deserialize, so a raise in the value can't leak it; also bounds-checks the key length.
- **`SPR_SYM_HASH` deser**: interns and frees the key buffer *before* the value deserialize (symbols are immediate, no rooting needed), so nothing is live to leak on a longjmp.
- **codegen**: now raises a compile-time error for a Ractor.new block with >16 params (was silently dropping extras), and emits a NULL-check + `sp_oom_die()` after the generated spawn-arg-array malloc.

Suite still 930 pass / 0 fail / 0 error; spawn + hash demos are valgrind memcheck- and GC-verify-clean. (The earlier ubuntu-gcc red was an sccache server-startup timeout, not the build — should clear on this re-run.)

Issue #1450, opened 2026-06-17
String#+ leaks: a hot `s = s + x` loop runs RSS to multi-GB / OOM (in-place `s << x` stays flat; GC doesn't track it)


rubys:
## Summary

`String#+` (binary concatenation) appears to leak its result: a hot `s = s + x`
loop drives RSS to **multiple GB and OOM-kills**, while the byte-identical loop
using in-place `s << x` stays flat at ~4 MB. The leaked memory is **not tracked by
the GC** — `GC.stat["bytes"]` stays tiny and `GC.start` doesn't reclaim it — so it
reads as a native leak of the concatenation results, not a GC-threshold issue.

Reproduced on master `7897906` (fresh `make`), macOS arm64. CRuby runs the same
file in ~50 MB.

## Minimal repro (no fibers / sockets / framework)

```ruby
def build_plus
  s = ""
  k = 0
  while k < 2000
    s = s + "x"          # String#+ : a fresh result string each step
    k += 1
  end
  s.length
end

n = 0
while n < 200000
  build_plus
  n += 1
end
puts "done"
```

```
$ spinel string_plus_leak.rb -o spl && ./spl
# RSS climbs without bound; Killed (OOM) at ~18 GB, never prints "done"
$ echo $?
137

$ ruby string_plus_leak.rb     # CRuby reference
done                            # ~50 MB max RSS
```

## `+` vs `<<` — same output, opposite memory

Swapping only the concatenation operator (build the same string, same iterations):

| builder            | peak RSS | exit         |
|--------------------|---------:|--------------|
| `s = s + chunk`    | ~18 GB   | 137 (OOM)    |
| `s << chunk`       | ~4 MB    | 0 (clean)    |

So the leak is specific to the `String#+` result path; in-place append is fine.

## The GC doesn't see it

Instrumenting a smaller run that survives long enough to print, `GC.stat` shows the
heap essentially flat while RSS explodes:

```
served=64      bytes=217800  old=197536  threshold=790144  cycle=1
served=896064  bytes=218528  old=197536  threshold=790144  cycle=1
...            (bytes ~218 KB, cycle stays 1)        # meanwhile RSS = 3+ GB
```

`bytes` never grows, `cycle` stays at 1 (GC barely runs because the tracked heap
never crosses the threshold), yet RSS is multi-GB. Consistent with the `String#+`
result buffers being allocated outside `sp_gc_bytes` accounting and never freed —
[...]

Issue #1021, opened 2026-05-28
Feature: GC introspection (GC.stat, heap census, alloc-site attribution)


rubys:
## Motivation

Spent today hunting a memory leak in a roundhouse-emitted Tep::Server app where idle RSS grew from 2 MB → 6.6 GB across 5 seconds of `wrk -c16` load while throughput stayed at ~15 K req/sec. Same C build, same shape as the published spinel archive at https://rubys.github.io/roundhouse/browse/spinel.tgz.

The bisection process worked but was slow: short-circuit the dispatcher → still leaks. Short-circuit the controller → leak gone. Re-enable DB query without view → no leak. Re-enable view → leak back. So it's in the view-rendering path *somewhere*, and the fix turned out to be a function-scope refactor in the keep-alive loop ([roundhouse 210a5f6](https://github.com/rubys/roundhouse/commit/210a5f6)) — extracting per-request work into its own method, presumably so each iteration gets its own `SP_GC_SAVE`/`SP_GC_RESTORE` pair rather than sharing the surrounding while-loop's function-level scope.

That fix was empirical. What I actually wanted, but didn't have, was: *"between request N and request N+1000, how many of each class are on the heap?"* If I could have seen "50000 `ActionDispatch::Session` objects on `sp_gc_old_heap`" the diagnosis would have been a one-liner instead of a bisection arc.

## Proposed surface

Three things, in priority order. Each is feasible because spinel's IR already knows the answer at compile time.

### 1. `GC.stat` — Ruby API over the existing globals (cheap)

The runtime already tracks `sp_gc_bytes`, `sp_gc_threshold`, `sp_gc_cycle`, `sp_gc_old_bytes`. Expose them:

```ruby
GC.stat
# {
#   "bytes"        => 47_185_920,      # sp_gc_bytes
#   "old_bytes"    => 33_554_432,      # sp_gc_old_bytes
#   "threshold"    => 67_108_864,      # sp_gc_threshold
#   "cycle"        => 142,             # sp_gc_cycle
#   "full_runs"    => 17,              # sp_gc_cycle / SP_GC_FULL_INTERVAL
# }
```

And `GC.start` to force a full collection — already trivial; `sp_gc_collect()` exists, just isn't reachable from Ruby. Currently the only GC-trigger paths I can see are inside specific codegen sites (bigint writes etc.); a Ruby-callable trigger would have let me distinguish *"objects are pinned"* from *"threshold heuristic hasn't fired yet"* in 30 seconds.

[...]

PR #1460, opened 2026-06-17
GC.stat: expose string-heap footprint (str_bytes / str_count)

rubys:
## What

Add two fields to `GC.stat`:

```ruby
GC.stat["str_bytes"]   # total bytes held in the string heap (sp_str_heap)
GC.stat["str_count"]   # number of live string allocations
```

## Why

Spinel keeps string data on a separate `malloc`'d heap (`sp_str_heap`), and string-heap pressure is **deliberately excluded** from `sp_gc_bytes` so the GC-threshold heuristic isn't skewed by string churn (per the comment in `sp_str_alloc`). That's a reasonable design choice — but a side effect is that `GC.stat`, which reports only the object heap, is **structurally blind to the string heap**. For a string-heavy workload (e.g. an HTTP server assembling responses), that's where almost all the bytes live, so you can sit at multi-GB RSS with `GC.stat["bytes"]` reading a few hundred KB and no introspection to explain it.

This surfaces the string heap in the same hash, turning "RSS huge but `bytes` tiny" into a one-line diagnosis.

I hit this chasing a server memory blow-up (#1450): `bytes` stayed ~218 KB while RSS climbed to ~15 GB. With these fields, `str_bytes` reads the leak directly and `str_count` ("how many strings are retained") points straight at the cause. Quick demo on the leak vs the in-place-append equivalent:

| workload | `str_bytes` | `str_count` | `bytes` (object heap) |
|---|---:|---:|---:|
| `s = s + "x"` ×N | 0 → ~773 MB | 0 → 192,258 | flat ~2.6 KB |
| `s << "x"` ×N | flat ~137 KB | 1 → 39 | ~140 KB |

## Implementation

An O(n) walk of `sp_str_heap` at call time. `GC.stat` is a diagnostic, not a hot path, so this seemed preferable to threading a counter through every alloc/sweep site — but if you'd rather have an O(1) figure I'm glad to maintain a running counter in `sp_str_alloc` / `sp_str_sweep` instead. Test included (`test/gc_stat_string_heap.rb`).

Motivated by the introspection RFC in #1021 — `GC.stat` landed there; this extends it to the string heap.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -1242,7 +1242,14 @@ static mrb_bool sp_StrIntHash_eq(sp_StrIntHash*a,sp_StrIntHash*b){if(!a||!b)retu
 /* Keys are spinel rodata literals (SPL: 0xff marker prefix) so the str-hash
    header cache's s[-1] read is in-bounds -- a bare C literal here would
    overread (and could alias a heap marker on some rodata layouts). */
-static sp_StrIntHash*sp_gc_stat(void){sp_StrIntHash*h=sp_StrIntHash_new();sp_StrIntHash_set(h,SPL("bytes"),(mrb_int)sp_gc_bytes);sp_StrIntHash_set(h,SPL("old_bytes"),(mrb_int)sp_gc_old_bytes);sp_StrIntHash_set(h,SPL("threshold"),(mrb_int)sp_gc_threshold);sp_StrIntHash_set(h,SPL("cycle"),(mrb_int)sp_gc_cycle);sp_StrIntHash_set(h,SPL("full_runs"),(mrb_int)(sp_gc_cycle/SP_GC_FULL_INTERVAL));return h;}
+static sp_StrIntHash*sp_gc_stat(void){
+  /* The string heap (sp_str_heap) is malloc'd separately and deliberately
+     excluded from sp_gc_bytes (see sp_str_alloc). Surface its footprint so
+     GC.stat can explain "RSS huge but bytes tiny" for string-heavy workloads.
+     Prototype: O(n) walk; a production version maintains a running counter. */
+  size_t str_bytes=0; mrb_int str_count=0;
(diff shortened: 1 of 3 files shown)


rubys:
Thanks for the reviews. Taking the three points:

**O(n) walk of `sp_str_heap` (Copilot, Gemini).** I'd like to keep the walk:

- `GC.stat` is an on-demand diagnostic / interval-cadence telemetry call, not a hot path. Metric exporters poll it every 10–60 s; at that cadence an O(live-strings) walk (µs–low-ms) sits far below the noise floor, and nothing samples GC stats per request.
- Heap-census introspection is inherently O(heap). This is the string-heap sibling of `ObjectSpace.count_objects` (the priority-2 item in #1021), which has to walk the heap; CRuby's `count_objects` / `memsize_of` are O(heap) as well. Holding `str_bytes` to an O(1) bar would be inconsistent with the rest of this API family.
- The walk is only ever large when the string heap is pathological — i.e. a leak — which is exactly the moment you've called `GC.stat` to find it. Cost and value are correlated.
- It's also unconditionally correct. A running counter has to be maintained at every alloc / free / grow site (the README's "`<<` promotes to mutable `sp_String` for in-place append" is one such path), and a single missed site makes `str_bytes` silently drift. The walk can't drift.

The PR text already offers the O(1) counter as a follow-up if a real workload warrants it — I'd rather defer that (YAGNI) and keep the correct-by-construction version now.

**`size_t` → `mrb_int` overflow (Copilot).** Not reachable on the shipping config: `mrb_int` is `INT64_MAX` (mruby_shim.h) and each `h->size` is `uint32_t`, so `str_bytes` is bounded by RSS (~10^10 even in the leak case) — many orders below `INT64_MAX`. Happy to add a clamp if a narrow-`mrb_int` build ever becomes a target.

**Single-statement packing (Copilot).** Matches the existing `sp_gc_stat` and the surrounding `sp_runtime.h` house style (dense one-liners); the walk loop is the only part broken out, for clarity.

PR #1470, opened 2026-06-18
Poly dispatch: call method_is_void() in the void-arm guards (follow-up to #1469)

rubys:
Follow-up to #1469, implementing the change @matz suggested on that PR.

#1469 fixed the four poly-dispatch void-arm guards by making each enumerate the no-value set `{VOID, NIL, UNKNOWN}` explicitly. As noted there, the guards can instead call **`method_is_void(&c->scopes[mi])` directly** — it's `!is_scalar_ret(ret)`, which is true for exactly `{VOID, NIL, UNKNOWN}` (plus `initialize`). So the void arm is now taken **iff the override is emitted as a C `void` function** — the *same* predicate `emit_method_signature` uses to decide `void`. The dispatch and the function declaration can no longer disagree, and a future no-value `TyKind` can't reintroduce the mirror-incompleteness #1469 fixed.

**Four sites** (case + default × both dispatchers):
- `codegen_fold.c` instance self-dispatch — `arm_ret`/`def_ret` enum check → `method_is_void(&c->scopes[kmi])` / `method_is_void(m)`
- `codegen_call.c` class-method dispatch — `kr`/`dr` enum check → `method_is_void(&c->scopes[kmi])` / `method_is_void(&c->scopes[defmi])`

(The `arm_ret`/`kr`/`def_ret`/`dr` locals stay — the value arms still use them for box-fn selection.)

No behavior change for the cases #1469 covered; additionally routes a poly-dispatched void `initialize` through the void arm, consistent with its void emission.

`make test`: **950 pass / 0 fail / 0 error**.

### DIFF
--- a/src/codegen_call.c
+++ b/src/codegen_call.c
@@ -1526,7 +1526,7 @@ else { memcpy(dir, sf, n); dir[n] = 0; } }
             if (kmi < 0) continue;
             TyKind kr = (TyKind)c->scopes[kmi].ret;
             buf_printf(b, " case %d: ", k);
-            if (unified == TY_POLY && (kr == TY_UNKNOWN || kr == TY_VOID || kr == TY_NIL)) {
+            if (unified == TY_POLY && method_is_void(&c->scopes[kmi])) {
               /* void-return (raises): call then fall through with nil */
               emit_method_cname(c, &c->scopes[kmi], b);
               buf_puts(b, "(");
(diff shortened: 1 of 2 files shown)


matz:
Thanks @rubys — this is exactly the right follow-up, and it matches matz's suggestion on #1469 precisely. Unfortunately the identical change had already been pushed directly to master in 744da4d5 (`method_is_void` at all four void-arm sites) before this PR was processed, which is why GitHub now shows it as conflicting. That was an unfortunate collision on our end, not a problem with your patch — your version was correct and, in fact, your codegen_fold.c comment was clearer than what landed.

I've adopted your improved comment in 7960f8d9 with a Co-Authored-By credit to you, and referenced #1470. Closing this as already-landed. Apologies for pre-empting the PR, and thank you for the clean fix and the careful write-up on both #1469 and here.

Issue #1456, opened 2026-06-17
Ractor: allow shareable (frozen/immutable) captures instead of always raising IsolationError


OriPekelman:
Follow-up to RFC #1452 (reproduce-CRuby item B4).

**CRuby behavior:** a Ractor block *may* capture outer variables that are shareable
(Integer/Float/Symbol/true/false/nil, frozen String/Array/Hash made shareable, etc.)
without error; capturing an unshareable mutable raises.

**Spinel today:** **any** captured outer local or `self` is a compile-time
`Ractor::IsolationError` (the minimal version's blanket rule). See
`docs/RACTOR-RFC.md` "Isolation rule".

**Proposed change:** refine the capture check to classify each captured var by
shareability (reuse the predicate from B3): shareable-by-value → capture by value into
a fresh per-Ractor cell (the plan's "shareable-by-value capture"); unshareable / `self`
→ keep the `Ractor::IsolationError`.

**Acceptance:** `test/ractor_shareable_capture.rb` — a block capturing an Integer and a
frozen String compiles and runs; capturing a mutable Array still raises
`Ractor::IsolationError`. Depends on B3.

matz:
Closing as out of scope for now: I've decided to revert Ractor from the compiler (master 7d93b12d + d729f09c) and stay on Thread + Fiber for concurrency. Full rationale in #1452 — in short, CRuby-grade Ractor (shared-nothing deep copy + N:M scheduling) is a large subsystem, and a partial Ractor risks giving false isolation guarantees, so Thread + Fiber is the near-term concurrency model. Not a quality judgment; the work is preserved in git history and Ractor would return only as a deliberately designed subsystem. Thanks for the detailed write-up here.

OriPekelman:
**Premise correction (and what shipped instead) — PR #1475.**

This issue assumed CRuby lets a Ractor block capture *shareable* locals (Integer/Symbol/frozen). It does **not**. CRuby 3.4 rejects **all** outer local-variable access in a Ractor block, even an `Integer`:

```ruby
x = 1; Ractor.new { x }   # ArgumentError: can not isolate a Proc because it accesses outer variables (x)
```

Only shareable **constants** are accessible (Spinel already supports those — verified). Copying shareable locals by value would make Spinel a permissive *superset* of CRuby: programs that error in CRuby would silently run here and fail to port. So we kept the faithful rejection and instead improved the diagnostic to name every offending local at once and suggest the fix:

```
[...]

Issue #1455, opened 2026-06-17
Ractor.make_shareable / Ractor.shareable?


OriPekelman:
Follow-up to RFC #1452 (reproduce-CRuby item B3).

**CRuby behavior:** `Ractor.make_shareable(obj)` deep-freezes `obj` and marks it (and
its reachable graph) shareable, so it may be referenced from multiple Ractors without
copying; `Ractor.shareable?(obj)` reports the predicate.

**Spinel today:** `Ractor.shareable?`-style classification exists only implicitly in
the isolation/capture check; there is no `make_shareable`, and nothing is shared by
reference except `Ractor::Port`.

**Proposed change:** implement `Ractor.make_shareable` (deep-freeze + a shareable
bit) and the `Ractor.shareable?` query. This is the prerequisite for B4 (shareable
captures) and for any future move toward CRuby's shareable-global model (RFC #1452 Q4).

**Acceptance:** `test/ractor_make_shareable.rb` — frozen immutable graphs report
`shareable?` true; mutables false; a made-shareable value crosses without a deep copy.
Note: depends on the shareability predicate also used by B4.

matz:
Closing as out of scope for now: I've decided to revert Ractor from the compiler (master 7d93b12d + d729f09c) and stay on Thread + Fiber for concurrency. Full rationale in #1452 — in short, CRuby-grade Ractor (shared-nothing deep copy + N:M scheduling) is a large subsystem, and a partial Ractor risks giving false isolation guarantees, so Thread + Fiber is the near-term concurrency model. Not a quality judgment; the work is preserved in git history and Ractor would return only as a deliberately designed subsystem. Thanks for the detailed write-up here.

OriPekelman:
Implemented in PR #1475. `Ractor.shareable?` (deeply-immutable predicate) and `Ractor.make_shareable` (recursive deep-freeze, pass-through type) over immediates / Class / Ractor::Port / frozen String / frozen Array·Hash·object. One documented caveat: a sent shareable value is still deep-copied at the boundary — Spinel's per-Ractor private heaps make genuine by-reference sharing a separate, larger change; for immutable data a copy is observationally equivalent. Test `test/ractor_make_shareable.rb`.

Issue #1453, opened 2026-06-17
Ractor: deliver the block's return value to the first #take


OriPekelman:
Follow-up to RFC #1452 (reproduce-CRuby item B1).

**CRuby behavior:** a Ractor's block return value is delivered to the first `#take`:
```ruby
Ractor.new { 1 + 41 }.take   # => 42
```

**Spinel today:** only an explicit `Ractor.yield(v)` reaches `take`; the block's
return value is dropped (the trampoline closes the outbox after running the body
without enqueuing the result). See `docs/RACTOR-RFC.md` ("the block's return value
is not delivered to take").

**Proposed change:** in the generated `_ractor_body_N` trampoline, serialize the
body's final value into the outbox before closing it (skip if the body already
yielded? — CRuby still delivers the return value as the *last* take, so just append).

**Acceptance:** `test/ractor_take_return.rb` — `Ractor.new { 21 * 2 }.take` prints `42`
with no explicit yield. Clean under GC-verify / TSan / valgrind.

matz:
Closing as out of scope for now: I've decided to revert Ractor from the compiler (master 7d93b12d + d729f09c) and stay on Thread + Fiber for concurrency. Full rationale in #1452 — in short, CRuby-grade Ractor (shared-nothing deep copy + N:M scheduling) is a large subsystem, and a partial Ractor risks giving false isolation guarantees, so Thread + Fiber is the near-term concurrency model. Not a quality judgment; the work is preserved in git history and Ractor would return only as a deliberately designed subsystem. Thanks for the detailed write-up here.

OriPekelman:
Implemented in PR #1475. A Ractor's block value is delivered to the first `#take` as a final implicit yield (after any explicit `Ractor.yield`, FIFO preserved); void/nil tails deliver nil; an uncaught raise delivers nothing. Test `test/ractor_take_return.rb`; clean under GC-verify, TSan, valgrind.

Issue #1452, opened 2026-06-17
RFC: Ractor API-shape questions needing a decision (whole-program/compile-time vs CRuby runtime)


OriPekelman:
## Context

The minimal Ractor (Milestones 1–6) landed in #1333 + #1448 — pthread-backed real
parallelism, private per-Ractor GC heaps, a deep-copy message codec
(immediates / Symbol / String / Array / Hash / user-objects), spawn args,
compile-time `@@cvar`/`$gvar` isolation, and a shared `Ractor::Port`. Design
record: `docs/RACTOR-RFC.md`.

The remaining work splits cleanly into two kinds:

- **(A) API-shape questions that aren't transparent** — places where Spinel's
  *whole-program, compile-time* resolution collides with Ruby's *runtime* model,
  so the right behavior is a **policy decision**, not just an implementation. Those
  are collected here for a decision.
- **(B) Reproduce-CRuby gaps** — behavior CRuby already defines unambiguously;
  each is filed as its own issue and becomes one focused follow-up PR (checklist
  at the bottom).

> Heads-up: this is an LLM drafting on Ori's behalf, so treat the recommendations
> as proposals to sanity-check, not settled claims. Happy to adjust any of these.

---

### Q1 — `Ractor#send` / `Ractor::Port#send` vs reflective `Object#send`

In CRuby, `Ractor#send` (and `<<`) is the *messaging* method; reflective dispatch
is `__send__`. Spinel resolves `recv.send("m")` reflectively **at compile time**
for *all* receivers (a whole-program parser behavior), so `ractor.send("hello")`
compiles as "call method `hello` on the ractor" rather than "enqueue the string".
Today the workaround is to send via `r << v`, a local (`r.send(v)`), or an Integer
literal.

- **Not transparent because:** disambiguation has to happen in the analyzer's
  reflective-send lowering, keyed on the receiver's static type.
- **Proposal:** when the receiver's static type is `TY_RACTOR` / `TY_RACTOR_PORT`,
  treat `.send`/`.send(arg)` as the message-send (matching CRuby) and route
  reflection through `__send__`. Low blast radius — the receiver type is already known.

### Q2 — `<<` as message-send vs Array push-promotion

`port << v` / `r << v` is currently mis-inferred: the `<<` push-promotion heuristic
types the receiver as an Array. CRuby treats `<<` on a Ractor/Port as `send`.

- **Not transparent because:** it's a precedence rule between two inference passes,
  not a local fix.
[...]

PR #1475, opened 2026-06-18
Ractor follow-ups: block return to #take, make_shareable/shareable?, clearer isolation error

OriPekelman:
Three CRuby-fidelity follow-ups to the merged Ractor work (#1333, #1448), each a self-contained commit. Tracked by the RFC #1452.

## #1453 — block return value delivered to the first `#take`
A Ractor's block value is now pushed to its outbox as a final implicit yield, matching CRuby:
```ruby
Ractor.new { 21 * 2 }.take   # => 42
```
Previously only an explicit `Ractor.yield` reached `take`. Explicit yields are delivered first (FIFO preserved), then the block value; void/nil tails deliver nil; an uncaught raise delivers nothing (unchanged). Tail-value extraction mirrors `emit_fiber_new`. Test: `test/ractor_take_return.rb`.

## #1455 — `Ractor.make_shareable` / `Ractor.shareable?`
```ruby
Ractor.shareable?(obj)      # deeply-immutable?  -> bool
Ractor.make_shareable(obj)  # deep-freeze, returns obj (pass-through type)
```
`shareable?` is true for immediates, `Class`, `Ractor::Port`, a frozen `String`, and a frozen `Array`/`Hash`/object all of whose reachable values are shareable. `make_shareable` recursively freezes and returns the value (immediate unchanged, `String` as a frozen heap copy); an un-freezable value raises `Ractor::Error`. Shape coverage matches the message codec. Test: `test/ractor_make_shareable.rb`.

A documented caveat: a sent shareable value is still **deep-copied** at the boundary — Spinel's per-Ractor private heaps make genuine by-reference sharing a separate, larger change (the RFC's "frozen-capture sharing"). For immutable data a copy is observationally equivalent (identity is already not preserved across `send`).

## #1456 — clearer isolation error (premise corrected)
The original #1456 assumed CRuby lets a Ractor block capture shareable locals. **It does not** — CRuby 3.4 rejects *all* outer local access in a Ractor block (even an `Integer`), raising `ArgumentError`; only shareable **constants** are accessible (Spinel already supports those). So copying shareable locals by value would make Spinel a permissive *superset* that diverges from CRuby. We keep the faithful rejection and instead improve the diagnostic to name every offending local at once and suggest the fix:
```
Ractor::IsolationError: block accesses outer variables (base, factor)
(a Ractor block must be self-contained; pass each value as a
Ractor.new argument -- Ractor.new(base, factor) { |...| } -- or send it instead)
```
Test `test/ractor_const_access.rb` pins the CRuby-faithful positive behavior (constants accessible, spawn args supported).

## Verification
- Full suite: **938 pass / 0 fail / 0 error**.
- New tests clean under `SPINEL_GC_VERIFY=1`, **ThreadSanitizer**, and **valgrind memcheck** (0 errors / 0 definitely-lost).
- Output verified byte-for-byte against reference Ruby 3.4.


OriPekelman:
Closing — Ractor was reverted from master (7d93b12d + d729f09c) as a scope/strategy decision (#1452), so this no longer has a base.

That call makes sense, and the false-isolation-confidence point is the right one to weight: a partial Ractor that "compiles and seems to work" is exactly where non-deterministic data races hide, and Thread + Fiber are honest about what they guarantee. Thanks for the thoughtful writeup and the kind words on the codec.

These three follow-ups (block-return-to-`take`, `make_shareable`/`shareable?`, and the named-variable isolation error) stay on the `ractor-followups` branch and in git history, should Ractor ever return as a deliberately-designed subsystem with a committed isolation + scheduling model. No action needed here.

PR #1476, opened 2026-06-18
spinel-doctor: add an `inference` leg (methods widened to untyped)

OriPekelman:
Adds a sixth leg to `spinel-doctor`: **inference** — methods whose signature spinel widened to `untyped`, i.e. the boxed poly slow path.

### Why
A widened method is correct but runs the slow path, and is often the first visible sign that inference couldn't pin a concrete type (sometimes the leading edge of a latent type gap). `--emit-rbs` already marks each such site with `# spinel: widened to untyped (slow path)`; today you have to run `--emit-rbs` separately and read the RBS to find them. This folds that signal into the one health report.

### Behavior
- **info severity** — does not affect the exit code (a widened program is still correct), matching the `requires` leg. It surfaces a perf/precision signal, not a defect.
- Reuses the existing `--emit-rbs` path — **no new compiler surface**.
- `--only inference` / `--skip inference` work like the other legs; listed in `--help` and the header.

### Example
```
$ spinel-doctor --only inference box.rb
spinel-doctor: box.rb
[info] inference (2)
         def initialize: (untyped) -> untyped # spinel: widened to untyped (slow path)
         def get: () -> untyped # spinel: widened to untyped (slow path)

spinel-doctor: clean
```
A program with no widening reports `[ok] inference`.

Stays within the spinel subset (compiles clean via `spinel tools/doctor.rb`). Verified on master `f3bb9af9`.

### DIFF
--- a/tools/doctor.rb
+++ b/tools/doctor.rb
@@ -7,6 +7,7 @@
 #   build       compile to a binary; report compile / codegen / C-build failure
 #   unsupported codegen gaps that degrade to a stub (stderr "unsupported ...")
 #   unresolved  calls that silently degrade to nil/0 (SPINEL_WARN_UNRESOLVED)
+#   inference   methods spinel widened to untyped (the boxed poly slow path)
 #   requires    non-relative `require`s spinel treats as native / no-op
 #   behavior    (opt) compiled output vs CRuby; needs `ruby` on PATH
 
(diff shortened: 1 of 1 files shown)


matz:
Merged. Built and ran it: `spinel-doctor --only inference` on a `Box.new` widening example lists both widened methods at `[info]` and exits 0; a non-widening program reports `[ok] inference`. The info-severity `report(...)` call matches the existing `requires` leg exactly, so the exit code is unaffected, and it reuses `--emit-rbs` with no new compiler surface. CI green on gcc + clang×2. Thanks @OriPekelman.

PR #1480, opened 2026-06-18
Add Float#clamp

ryanseys:
`Float#clamp` with float bounds previously rejected with `unsupported call (clamp)`, even though `Integer#clamp` and the `Comparable#clamp` string path already compiled. This adds the missing float path.

* `sp_float_clamp` in the runtime, mirroring `sp_int_clamp`.
* a codegen branch in the float-receiver dispatch that emits it.
* float-receiver `clamp` inference returning a float.

All three are gated on both bounds being float-typed. With float bounds the returned value is always one of the bounds (a float), so the inferred type is exact; a mixed or integer bound can return the integer bound, which stays poly and is left to the existing path.

Verified: `make test` (951 pass, 0 fail), `make bootstrap` IR + C fixpoint OK for `analyze.rb` and `codegen.rb`, and the new `test/float_clamp.rb` (expected regenerated from CRuby) passes. Generated C confirms the runtime path (`sp_float_clamp(lv_x, lv_lo, lv_hi)`) is taken for a non-literal receiver, not constant-folded.


matz:
Merged (marked ready first — it was still a draft but complete and verified). `sp_float_clamp` mirrors `sp_int_clamp` exactly, and the inference and codegen gates match (both require both bounds TY_FLOAT), so no mirror-incompleteness. Locally verified: `test/float_clamp.rb` matches both the `.expected` and CRuby, and the generated C takes the runtime path `sp_float_clamp(lv_a, ...)` for the non-folded method-param receiver. CI green on gcc + clang×2. Thanks @ryanseys.

Issue #1449, opened 2026-06-17
Whole-program regression: llama training aborts at step-2 backward — get_rows index OOB; byte-exact on the legacy backend (#1367 class)


OriPekelman:
## Byte-exact → hard abort: a whole-program codegen regression in the training backward path

Building toy's llama trainer at master `5959b64`/`0a35f70` (C backend), `toy train` (FromScratch / WarmStart / LoRA) **aborts at the 2nd step's backward pass**:

```
GGML_ASSERT(i01 >= 0 && i01 < ne01) failed     — ggml get_rows (an out-of-range gather index)
```

The same source compiled on the **legacy/union backend** (`fbb9beb`) completes 5 steps **byte-exact**. So this is a **C-backend codegen regression**, not ggml and not the application. Whole-program only — minimal repros of the shapes involved compile and run fine (the #1367 / whole-program-miscompile class).

## What is and isn't affected (tight scope)
| path | result |
|---|---|
| llama seq-engine train: FromScratch / WarmStart / LoRA | **abort @ get_rows, step-2 backward** (CPU **and** CUDA) |
| GPT-2-engine training, inline trainer, `06_train_from_scratch` | pass |
| ViT-tiny training | pass, byte-exact |
| llama **eval** (forward only) | pass, byte-exact |

So it is specific to the **llama training backward graph**.

## Localization
The step loop reads a fresh `seq_ids` slice per step at an increasing byte offset, and those ids index the embedding `get_rows`:
```ruby
seq_ids = ToyCorpusLoader.read_seq(corpus, byte_offset, CONTEXT)
byte_offset = byte_offset + CONTEXT * 4
loss = recipe.step!(seq_ids, positions, labels, hp, step == 0)   # fwd + bwd
```
- **Step 1** (`byte_offset == 0`) is byte-exact — forward *and* backward fine.
- **Step 2** (first **non-zero** offset): forward `get_rows` is in range (no forward abort), but the **backward** `get_rows` indexes out of range.

Forward-OK / backward-OOB with the **same** token ids means the index buffer the backward graph uses is corrupted — i.e. a **freshly-allocated index buffer saved for the backward pass is not kept alive** across the forward→backward (or step→step) boundary, so step 2 reads stale/garbage indices.

## Why this surfaced now (and the likely family)
[...]

PR #1482, opened 2026-06-18
spinel-reduce: add `--shrink-ints` parameter search

OriPekelman:
Adds a post-reduction pass to `spinel-reduce`: **`--shrink-ints`**.

### Why
ddmin isolates *which code* triggers a failure, but not the *numeric boundary* of a **size-triggered** bug ("miscompiles once the dim crosses ~512") — it removes an integer literal whole or not at all. This pass, after the line reduction, **binary-searches each surviving integer literal** down to the smallest value that still reproduces (assuming the trigger is monotone in the value — what a size threshold is), and reports the boundary. A literal whose value is irrelevant (even `0` still reproduces) is zeroed and flagged `not size-dependent`.

### Style / safety
- **Regex-free**, matching the other tools: a manual scan finds integer-literal spans (a maximal `[0-9_]` run not glued to an identifier/number on its left), edited **right-to-left** so an edit never shifts an unprocessed span's offset.
- Works with any oracle (`--oracle` or `--oracle-cmd`). No new compiler surface; pure post-pass over the reduced lines.
- Stays in the spinel subset (compiles clean via `spinel tools/reduce.rb`).

### Example
```
$ spinel-reduce --oracle-cmd './ge512.sh {}' --shrink-ints -o min.rb sizebug.rb
spinel-reduce: reduced to 1 lines
spinel-reduce: shrinking integer literals (parameter search)
spinel-reduce: size thresholds:
  sizebug.rb:1: 600 -> 512  (threshold; 511 does not trigger)
```
`min.rb` becomes `DIM = 512`. Verified on master `f3bb9af9`.

### DIFF
--- a/tools/reduce.rb
+++ b/tools/reduce.rb
@@ -2,11 +2,15 @@
 # input that still reproduces a chosen failure. Re-runs spinel only.
 #
 # Usage: spinel-reduce [--oracle build|unsupported|unresolved]
-#                      [--oracle-cmd 'CMD {}'] [-o OUT] app.rb
+#                      [--oracle-cmd 'CMD {}'] [--shrink-ints] [-o OUT] app.rb
 #
 #   --oracle <leg>     built-in interestingness (default: build = spinel fails)
 #   --oracle-cmd CMD   custom test: "{}" is replaced by the candidate file;
 #                      the candidate is interesting when CMD exits 0
+#   --shrink-ints      after code reduction, binary-search each surviving integer
+#                      literal down to the smallest value that still reproduces,
+#                      and report the threshold (isolates a size-triggered bug's
+#                      numeric boundary, e.g. "fails once the dim crosses ~512")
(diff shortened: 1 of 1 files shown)


matz:
Merged. Built it and ran a size-threshold oracle (`DIM >= 512`): `--shrink-ints` binary-searched `DIM = 600 -> 512  (threshold; 511 does not trigger)` and wrote `DIM = 512`, exactly as designed. The right-to-left span editing and the monotone binary search are sound; the int-literal scan is heuristic (it can shrink the integer part of a float, e.g. the `1` in `1.5`) but it's oracle-gated so it never emits a non-reproducing result. Regex-free and within the spinel subset, no compiler surface touched. CI green on gcc + clang×2. Nice tool for the whole-program size-boundary bug class (#1449/#1367 family). Thanks @OriPekelman.

PR #1511, opened 2026-06-20
codegen: implement find patterns and case/in in value position

mattn:
Two related pattern-matching codegen gaps.

Find patterns (`case x in [*head, a, b, *tail]`) were parsed into a FindPatternNode but never actually matched: emit_pm_cond had no case for the node, so the arm matched unconditionally, and no bindings were emitted, so the splats stayed nil (e.g. `case [:z] in [*h, :a, :b, *t]` wrongly matched with h/t nil). This scans the scrutinee for the first window of consecutive elements matching the requireds, uses its start position as the arm condition, and on a match binds the leading splat to the slice before the window, each required target to its element, and the trailing splat to the slice after; analyze types the bindings.

Separately, a CaseMatchNode used as a value -- assignment RHS, method receiver, or any expression -- hit "unsupported expression" (or silently produced nil in return position). A value mode added to emit_case_match assigns each arm's body value (boxed to the case's result type) to a result temp, and an emit_expr handler declares the temp, emits the match into the prelude, and yields it.

### DIFF
--- a/src/analyze_pass.c
+++ b/src/analyze_pass.c
@@ -778,6 +778,37 @@ int infer_write_types(Compiler *c) {
         /* in [first, *rest] or in Array(head, *tail) */
         array_pat = pat; array_scrutinee = scrutinee_t;
       }
+      else if (!strcmp(pty, "FindPatternNode")) {
+        /* in [*head, a, b, *tail] -- the two splats bind to arrays of the
+           scrutinee's element type; required LV targets bind to an element. */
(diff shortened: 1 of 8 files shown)


mattn:
Addressed the review: find patterns now fail closed (emit a false condition, never match) when the scrutinee isn't statically an array, instead of falling through to an unconditional match. Added tests covering find-pattern non-match and head/required/tail binding, plus case/in in value position.

One known limitation as follow-up: find patterns only fire when the scrutinee is statically an array type. A boxed (poly) scrutinee currently fails closed rather than unboxing to an array and matching at runtime; supporting that is left for a separate change.

PR #1520, opened 2026-06-22
Type each_cons block params as elements for multi-param blocks

ryanseys:
`arr.each_cons(n).map { |a, b| ... }` typed every block param as the whole n-window array — even with several params — while the codegen bound each param to a window *element* (`sp_IntArray_get(window, i + pj)`). The mismatch made `a + b` compile as array concat and `lv_a = <element>` a pointer-from-int error, so non-trivial bodies were rejected (`unsupported equality`) or miscompiled.

`each_cons` and `each_slice` share the same param semantics: a single param `|w|` binds the whole window/slice (an array); several params `|a, b|` (or a single destructured `|(a, b)|`) split it into element-typed params. This drops the each_cons-only branch that forced the array type, so each_cons follows the same rule — matching its codegen.

Tests: `test/each_cons_map_types.rb` (`|a, b|` sum, comparison, pair-collect; single `|w|`; destructured `|(a, b)|`; 3-window `|a, b, c|`; each_slice unaffected). Full `make test` green (1003 pass), ASAN clean.


matz:
Merged into `master` via merge commit 2e8fa3bc (your commit is in master's history). GitHub flagged it *closed* rather than *merged* due to a local non-fast-forward merge whose message used `Closes` — the change is fully integrated. Combined-state gates: test 1007/0/0, bench 57/57, optcarrot checksum 59662. Thanks!

PR #1551, opened 2026-06-24
Render int-typed nil as nil instead of the SP_INT_NIL sentinel

ryanseys:
An unfilled int-typed block/proc param carries the `SP_INT_NIL` sentinel (`INT64_MIN`). Collected into a typed int array and inspected, it leaked the raw `-9223372036854775808` instead of `nil`:

```ruby
proc { |a, b| [a, b] }.call(1)   # was [1, -9223372036854775808], now [1, nil]
```

This clears the proc nil-fill rendering bug. The fix guards the sentinel in the int-array inspect path (`sp_IntArray_inspect`) and in the sibling poly-render paths (`sp_poly_inspect` / `sp_poly_to_s`), mirroring the nil convention already used across the nullable-int array helpers. A bare unfilled param already rendered `nil` and is unaffected.

Adds `test/block_nil_fill.rb` (regenerated from the ruby oracle). Full `make test` is green (1032 pass, 0 fail, 0 error, -Werror).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


matz:
Landed as 50d85a53 (credited via Co-Authored-By). Reworked onto the post-eviction tree: the manual `sp_IntArray_inspect` hunk is subsumed because IntArray#inspect now routes through the generic `sp_inspect_container` → `sp_poly_inspect` walker, so the `sp_poly_inspect` / `sp_poly_to_s` SP_INT_NIL fix covers the typed-array case too. Thanks!

PR #1640, opened 2026-07-01
codegen: flat_map with a scalar-or-mixed block value

mattn:
flat_map with a block only compiled when the block's value was statically an array. A block returning a bare poly value, or a mix of array and scalar such as `cond ? sub_array : scalar`, fell through to the unsupported-call gate. This routes that case through `sp_PolyArray_flatten_into_n(dst, v, 1)` so an array value is spliced one level and a scalar is appended as-is, matching CRuby's flat_map.


mattn:
Good catch. The poly path now only fires when the inferred flat_map result is `TY_POLY_ARRAY`, so a statically-scalar block return (inferred as a typed array such as `TY_INT_ARRAY`) no longer gets forced into a `sp_PolyArray`; it falls through as before. Fixed in afbf4a0a.

PR #1641, opened 2026-07-01
codegen: find/detect with a block over a poly array

mattn:
find/detect { |x| cond } with a block only compiled for typed arrays. Those forms live inside an array_kind-gated block, and array_kind is NULL for a poly array, so find/detect over one (a mixed-type array, or an array of user objects) fell through to the unsupported-call gate. This adds a poly-array path that walks the elements, tests the block condition through emit_cond, and yields the matching boxed element or nil when none match.


mattn:
Good catch. The result temp is now rooted with SP_GC_ROOT_RBVAL after initialization so a GC in the consuming code can't collect the found element. Fixed in 09e0213f (which also declares the block param in the loop body so the same find works as a parameter default).

PR #1643, opened 2026-07-01
set: a usable Array-backed Set

mattn:
The bundled lib/set.rb only defined initialize, so Set#each, size, include? and the rest fell through to the unsupported-call gate (and a Set#each loop would silently drop). This backs Set with an Array and implements the common surface: construction from an enumerable (de-duplicating via Array#include?), each, include?/member?, add and <<, size/length, and empty?. Uniqueness of compound elements (arrays, structs) is only as precise as Array#include?'s element comparison, which is noted in the file.

### DIFF
--- a/lib/set.rb
+++ b/lib/set.rb
@@ -18,29 +18,23 @@ class Set
   end
 
   def <<(x)
-    @data.push(x) unless @data.include?(x)
-    self
+    add(x)
   end
 
   def include?(x)
     @data.include?(x)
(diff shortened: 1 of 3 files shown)


mattn:
Thanks — addressed all four in 9601b581: Set#each now returns self instead of the backing array (matches CRuby, no representation leak), << delegates to add, and member?/length are aliases of include?/size.

PR #1645, opened 2026-07-01
codegen: runtime-expand a poly splat in a poly array literal

mattn:
Splatting a poly value into an array literal (`[a, *x, b]` where x is a poly value) pushed the boxed value as a single element, so an array held at runtime was nested instead of spliced. Whether a poly holds an array is only known at runtime, so emit a check: splice one level if it is an array, drop nil, otherwise push as-is -- matching CRuby splat semantics.


mattn:
Good catch. sp_PolyArray_flatten_into[_n] didn't handle SP_BUILTIN_SYM_ARRAY, so a poly value carrying a symbol array nested instead of splicing. Added the SymArray case to both flatten helpers and a `g(%i[foo bar])` case to the test. Fixed in 91f6630f.

PR #1647, opened 2026-07-01
codegen: reach array operations through a poly / nullable-array value

mattn:
Values whose array-ness is only known at runtime -- a method returning `array | nil`, a recursive parameter, or a `case ... end` whose arms mix arrays and scalars -- were widened to a bare poly, which stripped every array operation from them. This makes those operations reach the underlying array at runtime: ty_unify keeps `poly_array | nil` a nullable poly array (mirroring the existing object|nil rule, where NULL encodes nil); flat_map on a poly receiver coerces it to a poly array via a new sp_poly_to_poly_array helper and runs the ordinary path; and a find pattern against a poly scrutinee coerces it the same way, so a non-array or nil never matches and the else arm runs.


mattn:
With this PR and the earlier ones merged, [matsuya-generator-ruby](https://github.com/toshia/matsuya-generator-ruby) compiles and runs natively end to end — `Matsuya.order` produces correct, varied menu names matching CRuby's output (including the pattern-rewrite transforms like キムカル丼 and 創業ビーフ).

Getting there took a chain of fixes, each surfacing the next once compilation could proceed further:

- #1640 flat_map with a scalar-or-mixed block value (merged)
- #1641 find/detect with a block over a poly array (merged)
- #1642 find/detect block used as a parameter default (merged)
- #1643 a usable Array-backed Set
- #1644 inline a yield method called on a constant-object receiver
- #1645 runtime-expand a poly splat in a poly array literal
- #1646 Array#sample on a poly receiver
- #1647 (this) reach array operations through a poly / nullable-array value

The last few all stem from one root: an array whose type is only known at runtime (a nilable return, a recursive parameter, a mixed `case` result) was widened to a bare poly, which hid its array operations. One remaining cosmetic difference: Spinel seeds the RNG with `srand(time(NULL))`, so repeated runs within the same second are identical, whereas CRuby varies every run — that is orthogonal to matsuya's logic.

mattn:
Thanks, both are real. Fixed in c7dd9aca:

- The coerced sp_PolyArray from sp_poly_to_poly_array is now SP_GC_ROOT'd in both places (the poly flat_map receiver and the poly find-pattern scrutinee), so a GC during the block/condition evaluation or the arm bindings can't collect it.
- The find-pattern coercion is now restricted to a TY_POLY scrutinee. A statically scalar/object scrutinee is not an sp_RbVal, so coercing it wouldn't compile; that case now fails closed (it can never match an array pattern), with a test covering `case 42 in [*, :a, *]`.

PR #1668, opened 2026-07-03
Add sp_fiber_ctx.h and sp_sched.h to the install lib path

MarioAriasC:
The flies `sp_fiber_ctx.h` and `sp_sched.h` are missing from install lib folder


MarioAriasC:
Thanks

PR #1671, opened 2026-07-03
codegen: index op-assign through a poly receiver with int/dynamic keys

### LINKED ISSUE #1669: Compound assignment fails on nested array when indices are not constants
Compound assignment to a nested (2-D) array element, `m[a][c] += v`, where the receiver `m[a]` is itself an index expression and the key is an int variable fails at compile time with  `unsupported index operator assignment (poly-recv, non-sym/str key): IndexOperatorWriteNode`

The plain nested assignment `m[a][c] = v` compiles fine; only the op-assign forms (`+=`, `-=`, `*=`, ...) fail.
Literal indices also work correctly.

Minimal reproduction path:
```ruby
m = [[0.0, 0.0], [0.0, 0.0]]
a = 0
c = 0
m[a][c] += 1.5
p m
```

bartleusink:
## Summary

- Fixes #1669: `m[a][c] += v` on a nested (2-D) array failed at compile time with `unsupported index operator assignment (poly-recv, non-sym/str key): IndexOperatorWriteNode`, while the plain `m[a][c] = v` form compiled fine.
- `emit_index_op_write`'s poly-receiver arm only handled symbol/string (hash) keys. This adds the missing int/dynamic-key fallback: evaluate receiver and key once into temps, read the slot via `sp_poly_arr_get_hash` (int key) or `sp_poly_index_poly` (boxed key), fold through the tag-dispatching `sp_poly_<op>` helpers (int/float/bigint/str dispatch at runtime), and store back with the matching poly setter — mirroring the structure the IndexOrWrite poly-receiver expression path already uses. Covers `+ - * / % ** << >> & | ^`.
- Adds a regression test, `test/index_op_write_nested.rb`, covering float and int elements, the operator variants (`/= %= <<= |=`), computed receiver indices, a hash-of-arrays receiver, three-level nesting, and repeated op-assigns in loops. Output matches CRuby.

## Testing

- The issue's reproducer compiles and prints `[[1.5, 0.0], [0.0, 0.0]]`, identical to CRuby.
- `make test`: 1207 pass. The 5 failures on my machine (`array_pattern_match_basic`, `bundle_array_f`, `bundle_classd_49`, `dir_methods`, `string_slice_range`) fail identically on unmodified master (macOS/arm64), so they are unrelated to this change.


matz:
Merged, thanks! One follow-up on the note about the 5 tests failing on unmodified master on your macOS/arm64 (array_pattern_match_basic, bundle_array_f, bundle_classd_49, dir_methods, string_slice_range): our macOS CI job is green, so this looks environment-specific and possibly a real portability bug we cannot see from here. If you have a minute, the actual-vs-expected diff of one of them (e.g. `make test 2>&1` around string_slice_range, plus `cc --version` and the locale) would be enough to chase it. No pressure -- filing it as an issue works too.

PR #1676, opened 2026-07-03
parse,codegen: fresh block-locals on every block invocation

khasinski:

In Ruby, a variable first assigned inside a block is block-local and starts nil on every invocation. Spinel emitted it as one function-level C local, so it kept its value across loop iterations. Doom's sprite selection does `sprite ||= get_rotated(...)` per visible object, and the leak made every object on screen reuse the first loaded sprite.

What this does:
- the parser serializes each BlockNode's `locals` list (Prism already tracks it),
- codegen resets the non-parameter locals to their nil representation (NULL string, SP_INT_NIL, NaN float, boxed nil) at the top of every block invocation,
- wired into `emit_stmts` plus the two fold paths that walk block bodies directly,
- parameter names are recognized through shadow renames (`name__bpN`), numbered params (`_1`), and `it`.

Adds a regression test covering `||=` fallbacks, conditional first assignment, and map/each/times shapes. Full suite green (1165 pass).

### DIFF
--- a/src/codegen_fold.c
+++ b/src/codegen_fold.c
@@ -2184,6 +2184,9 @@ int emit_each_with_index_terminal(Compiler *c, int id, Buf *b) {
 
   int body = block >= 0 ? nt_ref(nt, block, "body") : -1;
   int bn = 0; const int *bb = body >= 0 ? nt_arr(nt, body, "body", &bn) : NULL;
+  /* fresh block-locals per iteration (this emitter walks the body itself
+     instead of going through emit_stmts, so reset explicitly) */
+  if (block >= 0) emit_block_locals_reset(c, block, g_pre, din);
   if (is_each) {
     for (int j = 0; j < bn; j++) emit_stmt(c, bb[j], g_pre, din);
   }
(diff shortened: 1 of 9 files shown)


khasinski:
All three addressed:
- the body-to-BlockNode map now lives on the `Compiler` (allocated lazily, freed in `comp_free`), so there is no static state to survive a freed node table and no leak,
- block-local names longer than the stack buffer heap-allocate instead of silently skipping the reset,
- the parser builds the locals list in one pass with a single `cstr` per name and allocation checks.

Suite green.

matz:
Merged via 749b26db (the branch conflicted with #1679's declaration in codegen_internal.h, so I merged locally with both declarations kept -- GitHub should mark this merged now that the head is reachable from master). Verification: suite (1220) + bench green, optcarrot checksum intact and AHEAD in the interleaved A/B (the per-iteration resets are trivial stores). Param recognition through shadow renames/numbered/it is thorough, and symbol locals reset to the (sp_sym)-1 sentinel consistently with #1678. One pre-existing gap intentionally unchanged: is_cell locals (captured by an inner closure) still share one binding across iterations; skipping them was the right conservative call. Thanks!

PR #1677, opened 2026-07-03
codegen: destructure a poly multi-assign RHS at runtime

khasinski:

`lump_name, mirrored = @sprite_index[key]` reads the pair through a poly-typed value. The scalar fast path in multi-assign saw a non-array static type and handed the whole boxed array to the first target, nil-filling the rest, so Doom's sprite index lookups returned garbage.

A TY_POLY value can hold an array at runtime, so it now takes the runtime destructure path (element-by-element with nil-fill for surplus targets). Test covers a poly array pair, a poly hash read, and surplus targets. Full suite green (1165 pass).

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -3048,6 +3048,20 @@ static sp_RbVal sp_poly_each_elem(sp_RbVal a, mrb_int i) {
   }
 }
 /* poly_arr_get/set for PolyPolyHash with integer index key. */
+/* multi-assign element read: `a, b = v` destructures only when the boxed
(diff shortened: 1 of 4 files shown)


khasinski:
Right on both counts, thanks. Added `sp_poly_massign_get`: array builtins destructure element-by-element, any other runtime kind is treated as a scalar (first target takes the whole value, the rest nil-fill), matching Ruby's to_ary semantics. The multi-assign emitter now uses it instead of `sp_poly_arr_get_hash` (which reads Integer bits / String chars / Hash lookups). Test extended with poly-scalar int, string, and hash cases. Suite green.

PR #1678, opened 2026-07-03
codegen: honor the nil sentinel in symbol truthiness

khasinski:

A nilable symbol slot stores `(sp_sym)-1` as its nil sentinel, but truthiness treated every symbol value as always-true. Doom keeps `@exit_triggered = nil` and sets it to a symbol on level exit; the always-true read ended the level on the first tic (instant intermission screen).

Fixes both readers:
- `emit_cond`: symbol condition now compares against the sentinel,
- value-form `&&` / `||`: symbol left arm tests the sentinel, and when the kept arm widens to poly the sentinel boxes as nil instead of a bogus symbol.

Test covers if/&&/|| over a nil-then-set symbol ivar. Full suite green (1165 pass).


khasinski:
Both addressed. Added TY_SYMBOL arms to the ivar and local `||=`/`&&=` emitters (statement and value forms), testing the `(sp_sym)-1` sentinel instead of C truthiness. Also plugged the two escape hatches the sentinel had into user-visible values: `sp_box_sym` now boxes it as nil and `sp_sym_inspect` prints `nil` instead of `:\"\"`. Dropped the dead `Buf _vb` allocation from EMIT_ARM. Test extended with or/and-write coverage on ivars and locals. Suite green.

PR #1679, opened 2026-07-03
codegen: treat value constants in when-clauses as equality tests

khasinski:

Both case emitters resolved any constant in a `when` clause as a class name for `Module#===`. A constant that holds a value (`STATE_TITLE = :title`) is an equality test; the class-test path found no class of that name and folded the arm to `if (0)`. Doom's menu dispatches on `case @state when STATE_TITLE ...`, so the whole menu rendered as a black screen.

Constants registered with a concrete value type now skip the class-test path in both the statement and expression case emitters. Real class names in `when` keep working as class tests, covered side by side in the test. Full suite green (1165 pass).


khasinski:
Good catch, fixed. TY_CLASS constants are excluded from the value-constant check, and `emit_poly_class_when` now resolves class aliases (`Alias = SomeClass`) to the underlying class before the cls_id lookup, so the alias actually matches instances instead of silently failing the class test. Test extended with a class-alias arm. Suite green.

PR #1685, opened 2026-07-03
poly: dispatch numeric/byte query builtins by runtime tag

### LINKED ISSUE #1684: Poly value dispatch is missing common Float/Integer/String builtins (nan?, abs, floor/ceil/round, getbyte, bytesize, ord, bit_length)
A poly value (a boxed element of a heterogeneous container) that at runtime holds a Float/Integer/String can call SOME of that type's methods but not others — the missing ones raise at runtime:

```
unhandled exception: undefined method 'nan?' for poly
```

or fail to compile (`unsupported puts argument: ... CallNode 'floor'`), even though the method exists on the concrete type and works in CRuby.

## Minimal reproduction

```ruby
a = ["ab", 1]          # heterogeneous literal -> poly array; a[0] IS a String
puts a[0].getbyte(0)   # CRuby: 97 ; Spinel: undefined method 'getbyte' for poly

v = [1.5, "x"][0]      # a poly value that IS a Float
puts v.to_i            # works (1)
puts v.nan?            # CRuby: false ; Spinel: undefined method 'nan?' for poly
puts v.floor           # CRuby: 1     ; Spinel: compile error (unsupported puts argument)
```

## Catalogue (CRuby: all work)

| receiver tag | OK today | missing |
|---|---|---|
| String | `length`, `upcase`, `strip`, `empty?`, `[]` | `getbyte`, `bytesize`, `ord` |
| Float | `to_i`, `to_f` | `nan?`, `finite?`, `infinite?`, `abs`, `floor`, `ceil`, `round`, `truncate`, `zero?`, `positive?`, `negative?` |
| Integer | `even?`, `to_f` | `abs`, `bit_length`, rounding family (returns self) |

Poly arises legitimately from ordinary Ruby (heterogeneous containers, mixed-type returns, ivar arrays of objects that the narrowing pass leaves poly), so these are reachable divergences from CRuby, not exotic corner cases.

A non-numeric tag calling e.g. `nan?` should keep raising NoMethodError, as CRuby does (`1.nan?` raises).

bartleusink:
Fixes #1684.

## Summary

A poly value that at runtime holds a Float/Integer/String could call `to_i`/`to_f` and a few string transforms, but common query/rounding builtins were missing from the poly table: `nan?`, `finite?`, `infinite?`, `abs`, `floor`, `ceil`, `round`, `truncate`, `zero?`, `positive?`, `negative?`, `bytesize`, `ord`, `bit_length`, `getbyte`. They raised `undefined method '...' for poly` at runtime — or failed to compile when the untyped result reached `puts` — while CRuby handles all of them.

## Changes

- `lib/sp_runtime.h`: new `sp_poly_*` helpers that dispatch on the runtime tag the way CRuby dispatches on the class. A tag whose class does not define the method raises CRuby's NoMethodError, naming the actual runtime class (`1.nan?` → `undefined method 'nan?' for an instance of Integer`). Semantics match the typed emitters: `round` is half-away-from-zero, `infinite?` returns nil/±1, `getbyte` returns nil out of range. `abs` keeps the receiver's class (Integer stays Integer, Float stays Float, negative bigints negate losslessly) and the no-arg rounding family returns an int/bigint receiver unchanged, so those return boxed values.
- `src/codegen_call_recv.c` (`emit_poly_call`): route the names to the helpers, gated on "no user class defines the name as a method **or attr reader**" — a colliding user definition wins and goes through the normal poly method dispatch (same collision rule as #1683).
- `src/analyze_infer.c`: type the calls **after** the user-class dispatch unification so user definitions keep their inferred return types; predicates are Bool, `bytesize`/`ord`/`bit_length` are Integer, `abs`/rounding/`getbyte`/`infinite?` are poly (class-preserving / nullable results).
- `test/poly_numeric_query_methods.rb`: regression test covering Float (incl. negative-value rounding directions), Integer, and String receivers plus the out-of-range `getbyte` nil; output byte-identical to CRuby.

## Test plan

- [x] Both repros from #1684 compile and match CRuby (`getbyte(0)` → 97, `nan?` → false, `floor` → 1).
- [x] New regression test byte-identical to CRuby.
- [x] NoMethodError path verified: `"x".nan?` through a poly slot raises `undefined method 'nan?' for an instance of String`.
- [x] User-shadowing verified: a class defining `abs` / an `attr_accessor :round` still dispatches to the user definition.
- [x] `make test`: 1209 pass; the 5 failures on my machine are the known pre-existing environment artifacts (no `.rb.expected` snapshot + system Ruby too old as oracle) and fail identically on the unmodified baseline.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -1804,6 +1804,34 @@ static sp_RbVal sp_poly_mul(sp_RbVal a, sp_RbVal b) { if (a.tag == SP_TAG_BIGINT
 static mrb_int sp_poly_to_i(sp_RbVal v) { if (v.tag == SP_TAG_INT || v.tag == SP_TAG_SYM) return v.v.i; if (v.tag == SP_TAG_BIGINT) return (mrb_int)sp_bigint_to_int((sp_Bigint *)v.v.p); if (v.tag == SP_TAG_STR) return (mrb_int)strtoll(v.v.s ? v.v.s : sp_str_empty, NULL, 10); if (v.tag == SP_TAG_FLT) return (mrb_int)v.v.f; if (v.tag == SP_TAG_BOOL) return v.v.b ? 1 : 0; return 0; }
 static mrb_float sp_poly_to_f(sp_RbVal v) { if (v.tag == SP_TAG_FLT) return v.v.f; if (v.tag == SP_TAG_INT || v.tag == SP_TAG_SYM) return (mrb_float)v.v.i; if (v.tag == SP_TAG_BIGINT) return (mrb_float)sp_bigint_to_int((sp_Bigint *)v.v.p); if (v.tag == SP_TAG_BOOL) return v.v.b ? 1.0 : 0.0; return 0.0; }
 static mrb_bool sp_poly_numeric_p(sp_RbVal v) { return v.tag == SP_TAG_INT || v.tag == SP_TAG_FLT || v.tag == SP_TAG_BIGINT; }
+/* Numeric queries / rounding on a poly value: dispatch on the runtime tag the
+   way CRuby dispatches on the class. A tag whose class does not define the
+   method raises CRuby's NoMethodError (e.g. `1.nan?`, `"x".abs`). */
+SP_NORETURN SP_COLD static void sp_raise_poly_nomethod(const char *m, sp_RbVal v) {
+  static char buf[128];
+  snprintf(buf, sizeof buf, "undefined method '%s' for an instance of %s", m, sp_poly_class_name(v));
+  sp_raise_cls("NoMethodError", buf);
+}
+static mrb_bool sp_poly_nan_p(sp_RbVal v) { if (v.tag == SP_TAG_FLT) return isnan(v.v.f) != 0; sp_raise_poly_nomethod("nan?", v); }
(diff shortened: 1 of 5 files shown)


bartleusink:
Addressed all review comments in 818526d:

- **static buffer in `sp_raise_poly_nomethod`** — replaced with `sp_sprintf` (GC-heap result, stack temp), the same mechanism every other raise site in the runtime uses; safe under Spinel's no-GVL threads.
- **NaN/Infinity in `floor`/`ceil`/`round`/`truncate`** — the direct `mrb_int` cast was C UB; now raises CRuby's `FloatDomainError` first, with the exact CRuby message (`NaN` / `Infinity` / `-Infinity`).
- **NULL `char*` in `bytesize`/`ord`** — a NULL under `SP_TAG_STR` is the empty string (as in `sp_poly_to_i`): `bytesize` returns 0, `ord` raises CRuby's `ArgumentError: empty string`.
- **`abs` of `INT_MIN`** — plain `-x` was signed-overflow UB; now goes through `SP_POLY_INT_OP(sub, 0, x)`, so promote mode boxes a bigint (matching CRuby's `9223372036854775808`) and wrap mode keeps the documented wrapping arithmetic. Float `abs` uses `fabs` (covers `-0.0`).

All new paths are covered in `test/poly_numeric_query_methods.rb` (rescue-based checks for the three raise paths); output stays byte-identical to CRuby. `make test`: 1209 pass, same 5 known environment-only failures as the unmodified baseline.

PR #1688, opened 2026-07-04
codegen: typed-array compound index assignment coerces a poly RHS

### LINKED ISSUE #1687: Compound index assignment on a typed array with a poly RHS fails C compilation (float_array[i] += poly; also str_array[i] += and poly_array[i] +=)
## Summary

`float_array[i] += <poly>` — compound assignment into a typed float array where the right-hand side is a poly value (an element of a heterogeneous array) — passes type-checking but generates invalid C: the boxed `sp_RbVal` is added directly to the unboxed `double` slot.

```
sp_FloatArray_set(_t7, _t8, sp_FloatArray_get(_t7, _t8) + (sp_PolyArray_get(lv_b, 0LL)));
                            ^ mrb_float (double)          ^ sp_RbVal (poly)  -> no coercion
error: invalid operands to binary expression ('mrb_float' (aka 'double') and 'sp_RbVal')
spinel: C compilation failed
```

The equivalent plain form `a[0] = a[0] + b[0]` already compiles fine (it folds through the tag-dispatching `sp_poly_add` and unboxes the result), so only the `IndexOperatorWriteNode` emitter is missing the coercion.

## Reproduction

```ruby
a = Array.new(2, 0.0)   # inferred float_array
b = [1.5, "x"]          # heterogeneous literal -> poly_array, so b[0] is poly
a[0] += b[0]            # float_array[i] += poly  =>  emits `double + sp_RbVal`
p a                     # CRuby: [1.5, 0.0]
```

```
$ ruby   repro.rb   # => [1.5, 0.0]
$ spinel repro.rb   # => C compilation failed (invalid operands 'mrb_float' and 'sp_RbVal')
```

Reproduces with a literal index `a[0]`, a variable index `a[r]`, and inside a `while` loop. `int_array[i] += <poly>` fails the same way (`mrb_int + sp_RbVal`).

## Two adjacent gaps in the same emitter

While reducing this I found two more cases in `emit_index_op_write`'s array branch:

- `str_array[i] += x` emits native `char* + char*`, which never compiles — even with a plain String RHS (`s = ["ab", "cd"]; s[0] += "!"`).
- A poly_array receiver (`pa = [1.5, "x"]; pa[0] += 2.0`) falls through to `unsupported index operator assignment (array)` because `array_kind()` has no Poly arm, although the TY_POLY receiver path right below already handles the same fold.

## Workaround

Spell out the plain form: `a[0] = a[0] + b[0]`.

bartleusink:
Fixes #1687.

## Summary

`float_array[i] += b[0]` with `b` a heterogeneous (poly) array passed type-checking but emitted the boxed `sp_RbVal` straight into native arithmetic on the unboxed slot:

```
sp_FloatArray_set(_t7, _t8, sp_FloatArray_get(_t7, _t8) + (sp_PolyArray_get(lv_b, 0LL)));
error: invalid operands to binary expression ('mrb_float' (aka 'double') and 'sp_RbVal')
```

The equivalent plain form `a[i] = a[i] + b[0]` already folded through the tag-dispatching `sp_poly_add` and unboxed the result; only the `IndexOperatorWriteNode` array branch was missing the coercion. `int_array[i] += <poly>` failed the same way.

## Changes

- `src/codegen_stmt.c` (`emit_index_op_write`, array branch): a typed int/float slot with a poly RHS folds through the same box → `sp_poly_<op>` → unbox shape as the plain form (`sp_poly_to_f(sp_poly_add(sp_box_float(get), <boxed rhs>))`); shift/bitwise on an int slot unbox the RHS with `sp_poly_to_i`.
- Two adjacent gaps in the same branch, fixed while here:
  - `str_array[i] += x` emitted `char* + char*` (never compiled, poly RHS or not); now concatenates through `sp_str_plus` with both operands GC-rooted, exactly like the plain form's emission.
  - A poly_array receiver (`pa[i] += v`) fell through to `unsupported index operator assignment (array)` because `array_kind()` has no Poly arm; now folds via `sp_poly_<op>` on the boxed slot like the TY_POLY receiver path right below.
- `test/array_index_op_write_poly_rhs.rb` + committed `.rb.expected` snapshot: float/int slots with poly RHS (literal index, variable index, in a loop, `-=`/`*=`//`=`, `<<=`), str_array `+=` with typed and poly RHS, and `+=` on a poly_array receiver.

## Test plan

- Repro from #1687 compiles and prints `[1.5, 0.0]`, matching CRuby.
- New regression test byte-identical to CRuby; snapshot committed so the harness doesn't fall back to the local Ruby oracle.
- `make test`: 1219 pass; the 5 failures on my machine are the known pre-existing environment artifacts and fail identically on the unmodified baseline.

Made with Cursor


bartleusink:
Both review comments are addressed in 88abbc5:

- **Gemini: str_array with a non-`+` operator falls through to native emission** — confirmed: `s[i] <<= "!"` emitted `char* << char*` (C compile failure), and `s[i] *= 2` demoted the receiver to poly_array where `sp_poly_mul`'s missing String arm silently produced `0` instead of `"abab"` (the plain form `pa[0] = pa[0] * 2` had the same silent miscompile on master). Rather than only rejecting, string slots now take String ops explicitly: `+`/`<<` concatenate through `sp_str_plus`, `*` repeats through `sp_str_repeat`, and anything else is rejected with a clear `unsupported` error as suggested. `sp_poly_mul` gains the STR×INT arm and `sp_poly_shl` the STR×STR arm so the same ops also fold correctly on a poly-array receiver, and the poly-array branch now shares the full operator table with the TY_POLY receiver path.
- **Copilot: receiver temp not GC-rooted across allocating fold helpers** — confirmed: the collector uses explicit root registration (no conservative stack scan), so a receiver that is itself a fresh temporary (`make_array()[i] += v`) could be collected during `sp_poly_<op>` / `sp_str_plus` and freed before the closing `sp_*Array_set`. The receiver temp is now `SP_GC_ROOT`ed immediately after evaluation, matching the string-`+` emitter.

New string-slot cases (`<<=`, `*=`, and the poly_array-demoted variants) are covered in the regression test; snapshot regenerated, output byte-identical to CRuby. `make test`: 1219 pass, same 5 known environment-only failures as the unmodified baseline.

PR #1689, opened 2026-07-04
analyze: unify hash variants across ivars, locals, and empty literals

khasinski:

Spinel's hash variants (sp_StrPolyHash, sp_SymPolyHash, sp_PolyPolyHash, ...) have different struct layouts. Type inference could leave a local and the ivar backing it on different variants, and codegen then casts the boxed pointer across layouts, reading garbage or corrupting the heap. Two shapes from the Doom port:

- `opts = @menu.options` where the ivar is a SymPolyHash but a second assignment widened the local to PolyPolyHash: PolyPolyHash accessors walked SymPolyHash memory and segfaulted.
- `@near_linedefs ||= {}` where the empty literal kept its default variant while other (poly-keyed) writes made the ivar a PolyPolyHash, so the emitted constructor disagreed with the ivar's declared C type.

This adds a three-part pass at the end of analyze_program:

1. Ivar variant back-propagation: a local that unified to PolyPolyHash widens any Str/Sym-keyed source ivar it reads to PolyPolyHash, then re-runs inference to a fixpoint.
2. The reverse direction: a hash-typed local fed by a plain TY_POLY value, or by a reader whose backing ivar finished on a different hash variant, is widened to TY_POLY (with the node-type cache repointed) so subscripts go through the runtime dispatch instead of a layout cast.
3. Empty-literal alignment: an empty `{}` (type unknown) adopts the destination slot's hash variant, including when it sits on the right arm of `||=` / `||`.

Test: test/hash_variant_unify.rb reproduces both shapes; before the patch the compiled binary hangs/corrupts, after it matches ruby. Full suite green.

### DIFF
--- a/src/analyze.c
+++ b/src/analyze.c
@@ -3010,6 +3010,76 @@ void analyze_program(Compiler *c) {
     }
   }
 
+  /* Post-fixpoint: hash-variant back-propagation. A local that unified to
+     TY_POLY_POLY_HASH (e.g. a subscript write with a poly key widened it)
+     while the ivars feeding it stayed Str/Sym-keyed leaves the emitted C
+     with a PolyPolyHash* local pointing at a StrPolyHash object -- every
+     write through it then runs PolyPolyHash_set against the wrong struct
+     layout and corrupts the heap (doom's Animations#update
+     `translation = ... ? @texture_translation : @flat_translation`, which
+     shredded neighboring framebuffer memory into nils). Widen such source
+     ivars to the local's variant and re-run inference. */
+  {
+    int hb_changed = 0;
+    for (int id = 0; id < c->nt->count; id++) {
+      const char *nty = nt_type(c->nt, id);
+      if (!nty || !sp_streq(nty, "LocalVariableWriteNode")) continue;
+      const char *nm = nt_str(c->nt, id, "name");
+      LocalVar *lv = nm ? scope_local(comp_scope_of(c, id), nm) : NULL;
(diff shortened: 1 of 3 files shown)


khasinski:
Addressed the review in 741395c9. All five points were valid and are fixed:

1. NULL-scope guard added before `scope_local` in the back-propagation pass.
2. Implicit-self reader calls (`opts = options`) are now handled in the final safety pass by resolving the receiver class from the enclosing scope's `class_id`. I confirmed the bypass was real: an implicit-reader repro crashed on the previous revision.
3. The ntype-cache patch now also covers `InstanceVariableAndWriteNode`, `InstanceVariableOperatorWriteNode` and `InstanceVariableTargetNode`.
4. The final safety pass now also visits `LocalVariableOrWriteNode`, `AndWriteNode` and `OperatorWriteNode`; `t ||= self.table` reproduced the same hazard (hang on the previous revision) and is covered by the new shape (c) in the test.
5. DFS push guards are now derived from the stack capacity instead of a smaller hardcoded bound.

One note from verifying point 2: after the fix the local correctly falls back to runtime dispatch, but `sp_poly_set_poly` still silently drops a key whose tag does not match the runtime hash variant (e.g. a string key into a SymPolyHash object). That behavior predates this PR (reproducible on master) and is out of scope here; the new test asserts only reads unaffected by it.

Suite: 1224 pass, 1 fail (`system_argument_list`, unrelated known failure).

PR #1690, opened 2026-07-04
runtime: root string parameters of allocating sp_str.c functions

khasinski:

The string helpers in lib/sp_str.c take const char* parameters that are not GC roots. When such a function allocates before the last use of its parameter, a collection triggered by that allocation can free the argument if its only reference is a fresh unrooted temp (e.g. an interpolation or slice result passed inline). The function then reads freed, typically reused memory.

Concrete failure in the Doom port: sp_str_bytes allocates its IntArray first; when the GC ran exactly there, the freshly built argument string was freed and .bytes returned an empty/garbage array. Doom's Colormap silently lost map 9 and the screen went black after a while.

This adds SP_GC_ROOT(param) at entry to the 47 allocating sp_str.c functions that take string parameters, matching the rooting discipline already used by sp_str_chars, sp_str_lines and the split helpers.

The new test (test/gc_str_arg_rooted.rb) calls .bytes and .codepoints on freshly interpolated strings under allocation pressure. It fails deterministically on the unpatched runtime (wrong byte sums/lengths) and passes with the fix, also under SPINEL_GC_STRESS=1. Full suite is green: 1224/1224 PASS.


khasinski:
Thanks for the review. Both points verified against the collector internals; fixed in 863d529e.

**1. bytes/codepoints allocate before the nil-receiver check (gemini)** - Fixed: the `if(!s)sp_nil_recv(...)` now runs before `sp_IntArray_new()` in both. One note: the abandoned array was collectable garbage, not a leak (nothing roots it, the next cycle sweeps it), and the alloc-before-check ordering predates this PR. But raising first is strictly better since it avoids the pointless allocation and any GC it could trigger, and the lines were already touched here.

**2. bare "" NULL fallbacks vs the marker-byte probe (copilot)** - Verified as real, and it is the rooting from this PR that makes it bite: `sp_gc_mark` probes `obj[-1]` for every root slot and rewrites it (0xfe -> 0xfc) when it looks like a young heap string. The `if(!a)a=""` fallbacks in concat3/concat4/crypt are pre-existing upstream code (as is `sp_str_byte_len`'s probe of `""[-1]`), but before this PR those pointers never sat in a root slot, so the collector never dereferenced next to the bare literal. With the params rooted, a collection while a root holds "" reads out of bounds beside the literal and could attempt a rodata write. Fixed with the house idiom: the fallbacks now use `sp_str_empty` (sp_alloc.h, `sp_str_empty_data + 1` with the 0xff literal marker the collector skips), which is what `sp_str_concat` already used. Other rooted params never hold marker-less pointers (checked: the char cache and inline empties all carry 0xff markers, and scrub's fffd fallback is never assigned to the rooted param).

Validation: rebuilt, the regression test passes (also under SPINEL_GC_STRESS=1), nil-receiver raises unchanged, and the full suite is green again: 1224/1224 PASS.

matz:
Merged (b5a16727) with a hardening follow-up (a9ab9d09). The rooting direction is right and your UAF test caught a real class -- thanks. Two soundness issues surfaced under the gates and are fixed on top: (1) SP_GC_ROOT is an OBJECT root, and marking a non-spinel string that way walks a fake header in the collector -- bm_io_wordcount segfaulted through each_line's stack line buffer; string parameters now root through a new SP_GC_ROOT_STR that marks via sp_mark_string (safe on any memory). (2) each_line's line buffer is now one reusable HEAP string per loop, so nothing unmarked crosses the runtime API (still zero per-line allocations). Full battery: suite + bench green, your test passes under SPINEL_GC_STRESS, 40-test clang+ASAN sweep clean, optcarrot byte-identical generated C and fps-even.

matz:
Merged the review-fix commit too (8f691b4f), resolved against master's rooting: the sp_str_empty fallbacks in concat3/4 are adopted -- good catch, a bare "" literal has no marker byte and must not sit in a rooted slot -- while the SP_GC_ROOT_STR string roots stay (a plain object root walks a fake header on a frozen 0xf0/0xf1 string, which is why a9ab9d09 introduced the tagged variant). Suite + bench + ASAN spot checks green.

PR #1691, opened 2026-07-04
analyze,codegen: instance_variable_get on a poly receiver

khasinski:

`receiver.instance_variable_get(:@name)` with a poly-typed receiver had no inference rule and no lowering: inference fell through to an unrelated rule and produced a bogus type, and the whole expression chain was silently dropped (emitted as nil). In Doom, `SpriteManager#mirror_sprite` takes its sprite as a method param that unifies to poly across call sites and reads `sprite.instance_variable_get(:@columns)`, so every mirrored monster sprite rendered blank.

The fix mirrors the existing concrete-receiver lowering:

- inference: for a literal `:@name` symbol/string arg, unify the slot's declared type across every instantiated class that has it (all the same concrete type keeps that type; mixed or none stays poly)
- codegen: evaluate the receiver into a temp and `switch` on its `cls_id`, one case per instantiated class owning the slot, boxing each field read per that slot's declared type (a class without the slot reads as nil, matching CRuby's unset-ivar behavior); when the unified type is concrete, unbox the result

Covered by `test/instance_variable_get_poly_recv.rb` (mixed array/hash slot types plus a concretely-unified case). Full suite stays green.


khasinski:
Addressed the review feedback in ac0f017a:

1. **ty_unify in the inference rule** - confirmed the concern: the exact-match loop widened to poly when a slot was still TY_UNKNOWN mid-fixpoint, while ty_unify treats TY_UNKNOWN as identity. Switched to `uni = ty_unify(uni, t)` with an early poly break. One subtlety this pulls in: ty_unify folds int into bigint (INT+BIGINT => BIGINT), but `emit_unbox_text` for bigint is a raw `v.p` cast, so a boxed-int case would have been misread. The codegen now boxes an int slot under a bigint-unified result via `sp_box_bigint(sp_bigint_new_int(...))`, matching ty_unify's documented contract that emitters convert at that boundary. Covered by a new BigCounter/SmallCounter case in the test.

2. **Empty switch + fld buffer** - both fixed: has_cases is precomputed and when no instantiated class owns the slot the switch is skipped entirely (receiver still evaluated for effects, read yields nil, covered by a new `:@missing` case in the test); fld bumped to 512 bytes.

Suite state unchanged: 1223 pass, with only the two pre-existing failures (system_argument_list, and bundle_hash_b2 whose .expected is not in this repo).

PR #1693, opened 2026-07-04
runtime: honor pack/unpack endianness and size modifiers

khasinski:

Array#pack / String#unpack ignored the `<` `>` `!` `_` directive modifiers. Worse, each modifier char was consumed as if it were the next directive, so it ate an element slot: `pack('l<l<l<l<')` packed only every other value. Doom packs its SDL_Rect as `'l<l<l<l<'`, so half the rect fields were dropped and the framebuffer blit came out black.

This adds a `pk_parse_count_mods` helper that consumes the modifiers after a directive (before any count or `*`), and byte-swaps the `s/S/l/L/q/Q` pack and unpack paths for `>`. `<`, `!` and `_` are native order/size on this little-endian target and only need consuming.

Includes `test/pack_endian_modifiers.rb` with a round-trip check, byte-level pack/unpack checks, and cross-checks against `n/N/v/V`; expected output generated by Ruby. The test fails on master and passes with this change. Full suite green.


khasinski:
Addressed the review in 12c3d725:

1. **Conflicting modifiers (`l><`)**: applied, with one correction to the premise. "Last one wins" is not CRuby semantics: CRuby 3.4.9 raises `RangeError: Can't use both '<' and '>'` for *any* repeated endian modifier (`l><`, `l<>`, even `l<<` and `l>>`). Spinel's pack has no exception path here, so I adopted last-wins as the lenient fallback (TruffleRuby does the same: `l><` is LE, `l<>` is BE); `<` now resets `big` to 0. Because CRuby raises, the suggested test line can't go into the spec file (expected output is ruby-generated, and it would also break `make regen-expected`). I verified it manually against a compiled snippet: `[1].pack('l><').bytes == [1,0,0,0]`, `[1].pack('l<>').bytes == [0,0,0,1]`, `unpack('l><')` little-endian; a comment in the test file documents this.

2./3. **Shift-based serialization**: adopted. The `s/S/l/L/q/Q` pack and unpack paths now use `pk_put_int` / `pk_get_int` helpers that build the bytes with shifts per requested endianness, matching the style of the existing `n/N/v/V` code and independent of host byte order. `pk_bswap` is deleted.

Suite re-run: 1228 pass, 0 errors (only the known `system_argument_list` flake and `bundle_hash_b2`).

Issue #1576, opened 2026-06-26
codegen: poly to_s dispatch passes the user-class receiver by pointer to its by-value method → C compile error (cf #1437/#1438)


rubys:
Defining an instance `to_s` makes a polymorphic `.to_s` send emit a class-id `switch` whose user-class arm calls `sp_T_to_s((sp_T *)v.p)` — passing the receiver **by pointer** — while `sp_T_to_s` is generated to take it **by value** (`sp_T self`). The C doesn't compile. Unlike #1437/#1438 the generic `default:` case *is* present, so the dispatch is otherwise correct; the defect is only the user-class arm's calling convention. The user class is pulled into the switch even when it never reaches the call site (because `to_s` is universal).

### Minimal repro

```ruby
class Env
  def initialize(name)
    @name = name
  end

  def to_s
    @name
  end
end

# `value` is untyped here; `.to_s` has no statically-known receiver type.
def render(value)
  value.to_s
end

puts render(42)
puts render("hello")
```

```
$ spinel -E repro.rb
error: passing 'sp_Env *' to parameter of incompatible type 'sp_Env'; dereference with *
   case 0: _t2 = sp_Env_to_s((sp_Env *)_t1.v.p); break;
note: passing argument to parameter 'self' here
   static const char * sp_Env_to_s(sp_Env self) { ... }
spinel: C compilation failed
```

Note `Env` is never passed to `render` — it's pulled into the switch purely because it defines `to_s`.

### Root cause

`spinel -S` for `value.to_s`:

```c
// WITHOUT a user-defined to_s — generic poly lowering, compiles:
return sp_poly_to_s(lv_value);

// WITH Env#to_s — the class-id switch gains a user-class arm:
return ({ sp_RbVal _t1 = lv_value; const char * _t2 = ...;
   switch (_t1.cls_id) {
     case 0:  _t2 = sp_Env_to_s((sp_Env *)_t1.v.p); break;   // <-- passes sp_Env *
     default: _t2 = sp_poly_to_s(_t1); break;
   } _t2; });
```
with
```c
static const char * sp_Env_to_s(sp_Env self) { ... }          // <-- takes sp_Env (by value)
```

The `default: sp_poly_to_s` arm is correct (and would handle the actual `Integer`/`String` receivers). The defect is the `case 0:` arm passing `(sp_Env *)_t1.v.p` (a pointer) to a by-value method — the dispatch site and the method definition disagree on the receiver's pointer-vs-value representation.

[...]

PR #1695, opened 2026-07-04
codegen: attribute operator-write through a poly receiver

khasinski:

`recv.attr op= value` (CallOperatorWriteNode) was rejected with "unsupported expression". Doom hits this 8 times, e.g. `sector.ceiling_height -= speed` where the sector arrives through a poly value.

Lowering: evaluate the receiver into a temp once, read the attribute, apply the operator, write it back. A concretely typed object receiver accesses the ivar directly; a poly receiver dispatches on cls_id over every instantiated class exposing a reader named `attr`. Poly-typed slots fold through sp_poly_add/sub/mul/div/mod on boxed operands, string slots use sp_str_concat, typed slots use plain C arithmetic with poly RHS coercion.

Adds test/call_op_write_poly_recv.rb covering float, int and string attributes plus a widened (poly) slot. Suite green.


khasinski:
Thanks for the reviews, all ten findings were valid. Addressed in 1e8d9a93:

**Type-safety of the generated C**
- String slot `+=` with a poly rhs now coerces through `sp_poly_to_s` before `sp_str_concat`, in both the concrete-receiver arm and the poly dispatch arms (gemini, lines 3928/3983).
- Bitwise op-assign on a boxed (TY_POLY) slot is now lowered as unbox-op-rebox: `slot = sp_box_int((sp_poly_to_i(slot) op rhs))`, with a poly rhs coerced via `sp_poly_to_i` — the same shape the ivar op-assign handler uses for its poly-bitwise arm (gemini, lines 3934/3992). There are no `sp_poly_and/or/xor` runtime helpers, so unbox-op-rebox keeps parity with the existing ivar path instead of adding new ones.
- Operators a string or boxed slot cannot take no longer fall through to raw C (pointer arithmetic on an `sp_Str *` or a type error on an `sp_RbVal`). The concrete arm rejects them with a compile-time `unsupported` diagnostic; the poly dispatch skips that class arm so the object lands in the default raise, matching the skip-on-static-mismatch shape of the existing poly attr-write dispatch (copilot, lines 3936/3994). Note this also makes Ruby-legal `str_attr *= n` / `str_attr %= args` a loud rejection rather than silent bad codegen; full support can come separately.

**Ruby semantics**
- Both arms now require the writer as well as the reader, since op-write desugars into a reader call plus a writer call; an attr_reader-only attribute no longer silently stores (copilot, line 3913). For a concretely-typed receiver this is a compile-time `unsupported` rejection rather than a runtime NoMethodError: the harness treats a compile failure as ERR with no way to expect one, so this shape is verified manually — both the concrete and poly reader-only cases now fail compilation with "unsupported call operator write", where Ruby raises NoMethodError for `attr=` at runtime.
- An unknown backing slot no longer defaults to TY_INT: the concrete arm fails the lowering, the poly dispatch skips the candidate (copilot, lines 3913/3970).
- The cls_id switch is now keyed with `_t.tag == SP_TAG_OBJ ? _t.cls_id : 0x7fffffff` so a boxed scalar (cls_id 0) can't alias the user class at index 0 — same guard rationale as `emit_poly_dispatch_key` / issue #1576 (copilot, line 3961).
- The switch has a `default:` arm raising through `sp_raise_poly_nomethod("attr=", recv)` instead of silently doing nothing (copilot, line 4008). Verified at runtime: an integer flowing into `o.height -= 1` now aborts with "undefined method 'height=' for an instance of Integer".

The test now also covers a bitwise op on a widened (boxed) slot and a string `+=` with a poly rhs. Suite is green: 1229 pass, the only failure is a pre-existing local expected-file artifact unrelated to this change.

PR #1696, opened 2026-07-04
io: File#seek, #tell, #pos, and #rewind

khasinski:

File/IO handles rejected `seek`/`tell`/`pos`/`rewind` as unsupported calls (4 occurrences block the Doom gem: its WAD reader seeks to lump offsets before reading each lump). This adds the runtime wrappers over `fseek`/`ftell`/`rewind`, the TY_IO codegen dispatch arms, and models `IO::SEEK_SET/CUR/END` (and the inherited `File::SEEK_*`) as int constants so the optional whence argument works; `seek` defaults to absolute. Return-type inference already mapped these methods to int.

Test covers read/tell/pos, absolute seek, SEEK_CUR, SEEK_END with a negative offset, and rewind. Suite green: 1228 pass, only the known system_argument_list flake and the bundle_hash_b2 artifact fail.

### DIFF
--- a/lib/sp_io.c
+++ b/lib/sp_io.c
@@ -124,6 +124,25 @@ mrb_bool sp_File_eof_p(sp_File *f) {
   return FALSE;
 }
 
+mrb_int sp_File_seek(sp_File *f, mrb_int off, mrb_int whence) {
+  if (!f || !f->fp) return -1;
+  /* whence uses the Ruby IO::SEEK_* values (0/1/2), mapped explicitly so we
+     never depend on the platform's SEEK_SET/CUR/END numbering. */
+  int w = (whence == 1) ? SEEK_CUR : (whence == 2) ? SEEK_END : SEEK_SET;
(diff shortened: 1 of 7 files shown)


khasinski:
Addressed the review in 6896856d:

1. **tell on a closed handle**: now returns -1 instead of 0, consistent with seek/rewind. Fixed.
2. **32-bit fseek offsets**: switched to fseeko/ftello (off_t), so >2GB files work where long is 32-bit. The codebase has no Windows ifdefs anywhere in lib/ (unix-only for now), so I did not add an _fseeki64 branch; if Windows support lands, that is a one-line ifdef here.
3. **hardcoded /tmp path in the test**: now writes a pid-suffixed file in the cwd ("spinel_file_seek_test_#{Process.pid}.txt") and deletes it afterwards, matching the style of the other io tests.

Suite rerun after the fixes: 1229 pass, 1 fail (the known bundle_hash_b2 artifact).

PR #1697, opened 2026-07-04
analyze,set: fold false defined?-guarded branches; Set#delete

khasinski:

Doom's gosu_window wraps an MRI-only YJIT toggle in `if defined?(RubyVM::YJIT) && RubyVM::YJIT.enabled?`. RubyVM resolves to nothing in spinel, so the guard is statically false, yet the compiler still walked the branch: call collection marked the unemittable toggle helper reachable and codegen rejected the branch body (4 errors).

This adds a shared `comp_defined_guard_false` check and applies it in two places: `cr_collect_calls` skips the dead branch during reachability, and statement-form `emit_if` drops it via the existing static-guard fold chain. The scope is deliberately narrow: only `defined?` over constants/constant paths (optionally as the left arm of an `&&` chain) that resolve to no class, module, value constant, or known builtin. Dynamic `defined?` forms are untouched.

Also adds `Set#delete` (remove and return self) to the bundled set package, which doom's sector_actions.rb needs.

Tests: `test/defined_guard_dead_branch.rb` (dead if/elsif branches plus a truthy `defined?(String)` control) and `test/set_delete.rb`; both fail before this change. Full suite green (1229 pass; the flaky system_argument_list passes when rerun individually).

### DIFF
--- a/packages/set/set.rb
+++ b/packages/set/set.rb
@@ -26,6 +26,11 @@ class Set
   end
   alias member? include?
 
+  def delete(x)
+    @data.delete(x)
+    self
+  end
+
   def each
     @data.each { |x| yield x }
     self
--- a/src/analyze.c
+++ b/src/analyze.c
@@ -28,13 +28,13 @@ void compute_reachable(Compiler *c) {
   int   *sc_cap      = calloc((size_t)c->nscopes, sizeof(int));
   for (int s = 0; s < c->nscopes; s++) {
     if (c->scopes[s].body >= 0)
-      cr_collect_calls(c->nt, c->scopes[s].body, &scope_calls[s], &sc_n[s], &sc_cap[s]);
+      cr_collect_calls(c, c->nt, c->scopes[s].body, &scope_calls[s], &sc_n[s], &sc_cap[s]);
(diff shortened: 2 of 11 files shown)


khasinski:
Good catch, the fold judged a constant path by its tail name only, so a builtin tail like `defined?(MissingParent::String)` masked the unresolved parent and the dead branch came back for qualified constants (CRuby also returns nil there).

Fixed in ee862ff3: `comp_defined_guard_false` now walks the path tail-to-root and treats any unresolved segment as statically false, declining to fold when a parent is not a plain constant (dynamic paths). Extended the test with a builtin-tail path (`MissingParent::String`) and a nested missing-root path (`MissingRoot::Sub::Thing`), both with uncompilable calls in the dead branch; both fail before the fix.

No truthy constant-path control was added on purpose: spinel's DefinedNode emit currently answers nil for every ConstantPathNode (even `Object::String`, where CRuby says "constant"), so a resolvable-path branch is already runtime-dead upstream and its expected output would not match CRuby. The fold never drops a branch the emit would keep live. Suite rerun shows no regressions beyond the two known flaky/artifact cases.

PR #1699, opened 2026-07-04
runtime,codegen: delete on poly arrays and poly-valued hashes

khasinski:

`delete` was unsupported on two container shapes Doom hits: a poly (mixed-element) array receiver, and a poly-keyed poly-valued hash. sector_actions.rb does `@active_specials.delete(key)`, and the array-backed `set` package needs `Set#delete` (which delegates to `Array#delete` on its poly `@data`).

Changes:
- `sp_PolyArray_delete`: element-wise `sp_poly_eq` removal, returns the deleted value or nil (Ruby semantics).
- `sp_PolyPolyHash_delete` plus `TY_POLY_POLY_HASH` in the hash delete codegen arm (Sym/StrPolyHash were already covered).
- `TY_POLY_ARRAY` delete dispatch next to the existing `delete_at` arm.
- `Set#delete` in `packages/set/set.rb`.

New test `test/poly_container_delete.rb` covers hits and misses for all shapes; suite green.


khasinski:
Both points were valid, fixed in 48182b31. The GC here is precise (explicit root stack, collections triggered from allocation), so unrooted locals are indeed invisible to the marker.

1. `sp_PolyPolyHash_delete`: confirmed. `sp_rbval_hash_key`/`sp_rbval_eql_key` can reach the user `hash`/`==` hooks (`sp_obj_eql_hook`), whose generated code may allocate and trigger a collection mid-rebuild, sweeping and finalizing the unrooted `tmp`. Now roots both `h` and `tmp` (`SP_GC_ROOT`), matching the existing pattern in `sp_StrIntHash_keys` et al and the codegen compact arm, which also does `sp_PolyPolyHash_new(); SP_GC_ROOT(...)` before a set loop. I kept rebuild-and-swap rather than switching to the Str/SymPolyHash backshift: `order[]` in this variant stores slot indices (not keys), so an in-place backshift would additionally have to renumber every `order[]` entry whose slot moved. Rooted, the rebuild has no remaining hazard and stays O(n) like the backshift.

2. `sp_PolyArray_delete`: confirmed. `sp_poly_eq` allocates on the bigint-promotion path, so a collection can run mid-loop while `a`/`v` are possibly reachable only through the call expression. Now `SP_GC_ROOT(a); SP_GC_ROOT_RBVAL(v)` at entry.

`SPINEL_GC_STRESS=1` is honored by the runtime (shrinks both heap thresholds to 2 KB); `test/poly_container_delete.rb` compiled and run under it produces output identical to the expected file. Full suite green (sole failure: `bundle_hash_b2`, a pre-existing copied-expected artifact unrelated to this PR).

PR #1700, opened 2026-07-04
codegen: receiverless respond_to? and related dispatch gaps

khasinski:

Four small gaps, each with a test:

- `respond_to?(:m)` with no receiver (implicit self) inside a method was rejected (doom's `gosu_window.rb`: `self.fullscreen = value if respond_to?(:fullscreen=)`). Resolve it against the enclosing scope's class: instance methods consult the method chain and attr readers/writers (a `:m=` query matches the writer table under its base name), `def self.x` methods consult the class-method chain and singleton attrs. Toplevel keeps the existing fall-through.
- A `case` in value position with an arm ending in `puts` routed the tail through `emit_expr`, which has no expression form for it (doom's `GosuWindow#button_down` debug-toggle arms). A nil/void/unknown-typed tail now runs as a statement and the arm value stays at the slot default, mirroring the if-as-value emitter.
- `String#delete` on a receiver widened to poly had no inference or codegen rule, so `data[offset, 8].delete("\x00").upcase` in doom's WAD texture parser collapsed into an unresolved-call raise. Added the poly rule (concrete `TY_STRING`) plus the matching `sp_str_delete(sp_poly_to_s(..))` emission, both skipped when a user class defines `delete`.
- Bare `new(...)` inside a class method (implicit `self.new`) never marked the enclosing class instantiated, so poly-element member reads dispatched with a missing case arm and read back nil for every field of a valid object (doom's `Texture.parse_texture`).

Suite: 1230 pass; the only failures are the known-flaky `system_argument_list` (passes standalone) and the `bundle_hash_b2` artifact.

### DIFF
--- a/src/analyze.c
+++ b/src/analyze.c
@@ -210,6 +210,18 @@ void compute_instantiated(Compiler *c) {
         disable = 1; break;
       }
     }
+    /* bare `new(...)` (no receiver) inside a class/singleton method (`def
+       self.foo; ...; new(...); end`) is implicit `self.new(...)`, i.e. the
+       enclosing class -- e.g. doom's `Texture.parse_texture` builds each
+       Texture via a bare `new(name, ...)`. Missing this meant such a
+       class's `instantiated` flag never got set unless some other call
+       site also did `Texture.new` explicitly -- so a poly-array element
+       member-read dispatch silently dropped its case arm, reading back
+       nil/0 for every field of an otherwise fully-constructed object. */
+    if (recv < 0 && sp_streq(name, "new")) {
+      Scope *encl = comp_scope_of(c, id);
+      if (encl && encl->class_id >= 0) c->classes[encl->class_id].instantiated = 1;
+    }
     /* raise Cls / raise Cls, msg : constructs an instance of Cls */
     if (recv < 0 && sp_streq(name, "raise")) {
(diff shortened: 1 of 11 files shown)


khasinski:
Addressed both review findings in 0f528999:

1. **Implicit-self respond_to? in class methods missed builtin Class capabilities** - confirmed: the is_cmethod branch consulted only user singleton methods/attrs, so `respond_to?(:new)` inside `def self.x` folded to false. Factored the explicit `Const.respond_to?` logic into a shared `class_responds_to()` helper and used it for the implicit-self class-method form as well, so `:new`, `:name`, `:allocate` etc. now answer correctly (and a module still answers false to `:new`). Added test lines for the class capabilities plus a module singleton case.

2. **compute_instantiated fired on bare `new(...)` in any method** - confirmed: verified against Ruby that bare `new` inside an *instance* method is a NameError unless some method `new` is in scope, so it is never Class#new there. Gated the marking on the enclosing scope being a class method (`encl->is_cmethod`) and adjusted the comment; the doom shape (`Texture.parse_texture`) is a class method, so it is unaffected.

Suite after the changes: 1230 pass, with only the known-flaky `system_argument_list` (passes standalone) and the `bundle_hash_b2` artifact failing.

PR #1701, opened 2026-07-04
analyze,codegen: void-typed values widen to poly instead of C void

khasinski:

Compiling the gem produced C that does not compile: `error: variable/field has incomplete type 'void'`. TY_VOID/TY_NIL have no C storage type, but value-position constructs whose inferred type lands there were still emitted as C `void` declarations. The Doom shapes are `self.fullscreen = value if respond_to?(...)` used as an expression (a writer call as an if arm) and `return @cache[k] = nil` (a nil rhs of `[]=` in value position), with the void-typed values also flowing into ivar slots.

The rule applied here: any slot that would be declared `void` (ivar field, result temp of a value-position if/case/begin, the `[]=` store temp) widens to poly (`sp_RbVal`) at declaration, and a value-less arm tail runs as a statement with the slot left nil.

- analyze: backstop pass widens void/nil ivar slot types to poly (plus a defensive widen at the struct-field declaration)
- codegen_expr: if/unless-as-value (ternary and block forms), begin-as-value, case/in-as-value result temps
- codegen_stmt: case-as-value result temp; value-less arm tails
- codegen_call: `[]=` in expression position holds a nil/void rhs boxed

Each added test fails to compile before this change and matches ruby output after it. Test suite is green.


khasinski:
Addressed review feedback in 29c494a7.

**Double semicolon in emit_case_branch_value (gemini):** not a bug. The callers append `break; } ` (int fast path) or `} ` (if-chain path) after the branch emitter, never a semicolon, and the early-return path uses the exact trailing `"; "` convention of the pre-existing fall-through at the end of the same function. Grepping the generated C of the case test for `;;` finds zero occurrences. Left as is.

**TY_UNKNOWN in emit_case_expr (copilot):** correct in principle, widened in 29c494a7. `emit_ctype`'s fallback prints `void` for TY_UNKNOWN, so an unknown-typed value-position case would hit the same incomplete-type error. I could not construct a Ruby shape that reaches this on the current inference passes (recursive no-base-type arms and similar all resolve to poly or a concrete type before codegen), so there is no accompanying test; the widen is defensive and matches the case/in-as-value path in emit_expr, which already handles TY_UNKNOWN.

Note for the maintainer: the value-less-tail hunk in `emit_case_branch_value` also appears in PR #1700 - both PRs needed the same fix. Whichever merges second will conflict there, but it resolves trivially (the hunks are identical in effect; keep either).

Suite rerun after the change: 1232 pass, 1 fail (`bundle_hash_b2`, pre-existing stale expected file - the binary's output matches `ruby` exactly).

matz:
Merged into master via rebase as 3114d67b (resolved a conflict in `src/codegen_stmt.c` where #1700 had already added the value-less case-arm tail; the two edits were equivalent). Full gate green: 1244 tests pass, 58 benches pass, optcarrot checksum 59662 with fps ~843 (no regression vs baseline). Thanks!

PR #1710, opened 2026-07-05
analyze,codegen: blocks passed to methods keeping a real &blk work end to end

bartleusink:
## Summary

A method's `&block` parameter referenced from an independent closure in the method body failed to compile, and a block passed to a class/module method silently dropped writes to captured locals. CRuby runs all of these fine:

```ruby
module M
  def self.f(&blk)
    t = Thread.new { blk.call(0) }
    t.join
  end
end
M.f { |i| puts "got #{i}" }   # CRuby: "got 0"  |  Spinel: compile error
```

```ruby
module Pool
  def self.work(&blk)
    Fiber.new { blk.call }.resume
  end
end

sum = 0
Pool.work { sum += 1 }
puts sum        # CRuby: 1  |  Spinel (before): 0 -- the write was dropped
```

## Root causes

Four holes, all in the same feature area (blocks passed to methods that keep a real `&blk` param instead of being yield-inlined):

1. **Escape analysis missed independent-closure bodies.** The `&blk` escape scan only treated proc literals and lambdas as capturing bodies, so a `blk` read inside a `Thread.new { }` / `Fiber.new { }` / `Enumerator.new { }` block — or inside a block lifted to a standalone proc — went undetected. The method was then yield-inlined and the closure referenced an `lv_blk` that inlining never materialized.
2. **Class/module-method call sites dropped the block.** The instance-method and bare-call paths lower the literal block to an `sp_Proc *` and pass it, but the `Class.cmethod` emitter did not — including the Stage-1 (`comp_sg_reader_const` fold) and Stage-2 (`comp_sg_reader_candidates` cascade) singleton-accessor dispatch paths. The block was silently dropped and the callee's `lv_blk` dangled.
3. **`a_block_is_lifted` never recognized constant receivers.** It only resolved the callee for instance-typed receivers, so `M.f { }` was not detected as a lifted block and its captured enclosing locals were never celled — writes to them from inside the block were silently dropped.
4. **Celled int compound assignment didn't coerce a poly RHS.** `sum += arr[i]` with `sum` captured and `arr` a poly array generated `mrb_int + sp_RbVal`, which fails to compile.

## Changes

- `src/analyze.c`: the `&blk` escape scan marks the bodies of `Fiber`/`Enumerator`/`Thread` `.new { }` blocks and lifted blocks (via `a_proc_create_or_lifted`) as capture sites, so a `blk` read inside them keeps a heap-materialized `&blk` instead of being yield-inlined. `a_block_is_lifted` resolves `Klass.cmeth { }` / `Mod::Sub.cmeth { }` receivers via `comp_cmethod_in_chain`, so blocks passed to class/module methods cell their captures like instance-method blocks.
- `src/codegen_call.c`: a shared `emit_cmethod_block_arg` helper passes `&block` as an `sp_Proc *` (or `NULL`) on the direct `Class.cmethod` path and both singleton-accessor dispatch paths; the Stage-2 cascade lowers the literal block to a single proc temp shared by all candidate branches.
- `src/codegen_stmt.c`: the celled int `+=`/`-=`/`*=` emitter coerces a `TY_POLY` right-hand side through `sp_poly_to_i`, as the non-celled path already did.
- `test/block_param_into_thread.rb` (+ committed `.expected`): module- and instance-method `&blk` into `Thread.new`, a positional param ahead of `&blk`, `Fiber.new` capture, `&blk`-as-value, a value-returning `blk.call`, `&blk` read from a lifted block, and both singleton-accessor dispatch stages.
- `test/lifted_block_nested_capture.rb` (+ committed `.expected`): module-, class-, and instance-method lifted blocks; direct and nested-block captures; a capture mutated across a `Fiber` boundary; poly-element coercion into a captured accumulator; and a small worker pool.

## Test plan

- All repros and covered variants compile and match CRuby byte-for-byte.
- Both regression tests verified through the harness against committed CRuby snapshots.
- `make test`: 1255 pass, 0 fail, 0 error.
- `make optcarrot`: OK (343 fps, checksum 59662).


bartleusink:
Both review comments addressed in 8e6c2f70:

- **Lifted blocks as capture sites**: the `&blk` escape scan now goes through `a_proc_create_or_lifted`, so a `blk` read inside a block lifted to a standalone proc is treated as a capture-escape too (same failure mode as the Thread case, now fixed).
- **sg-reader dispatch paths**: the Stage-1 `comp_sg_reader_const` fold and the Stage-2 `comp_sg_reader_candidates` cascade now pass the block proc via a shared `emit_cmethod_block_arg` helper (also used by the direct `Class.cmethod` path). The Stage-2 cascade lowers the literal block to a single `sp_Proc *` temp shared by all candidate branches so the proc function is emitted once.

Regression test extended with the lifted-block case and both sg-reader stages (block called from a `Thread.new` inside the callee). `make test`: 1254 pass, 0 fail. `make optcarrot`: OK (357 fps, checksum 59662).

bartleusink:
Folded #1711 into this PR (commit 4c2c741d) — its review fixes made the &blk escape scan depend on a_block_is_lifted, which that commit extends, so the two were no longer independently reviewable. Description updated to cover the full scope. Combined stack: make test 1255 pass / 0 fail, optcarrot OK (343 fps, checksum 59662).

PR #1711, opened 2026-07-05
analyze,codegen: blocks passed to class methods capture correctly

bartleusink:
## Summary

A `{ }` block passed to a class/module method that keeps a real `&block` param (not yield-inlined) did not capture enclosing locals correctly — writes to them from inside the block were silently dropped, and folding a poly value into a captured int accumulator failed to compile:

```ruby
module Pool
  def self.work(&blk)
    Fiber.new { blk.call }.resume
  end
end

sum = 0
Pool.work { sum += 1 }
puts sum        # CRuby: 1  |  Spinel (before): 0 -- the write was dropped
```

Two bugs combined:

- **`a_block_is_lifted` never recognized constant receivers.** It only resolved the callee for instance-typed receivers (`ty_is_object`), so `M.f { }` was not detected as a lifted block and its captured locals were never celled. Depending on the shape, the write was silently dropped or the proc emitter rejected the read as an uncaptured outer variable. Constant receivers now resolve through `comp_cmethod_in_chain`, matching the codegen call path.
- **Celled `TY_INT` compound assignment didn't coerce a poly RHS.** `sum += arr[i]` with `sum` captured and `arr` a poly array generated `mrb_int + sp_RbVal`, which fails to compile. The celled path now coerces through `sp_poly_to_i` exactly as the non-celled int op-write does.

## Changes

- `src/analyze.c`: `a_block_is_lifted` resolves `Klass.cmeth { }` / `Mod::Sub.cmeth { }` receivers via `comp_cmethod_in_chain`, so blocks passed to class/module methods cell their captures like instance-method blocks.
- `src/codegen_stmt.c`: the celled int `+=`/`-=`/`*=` emitter coerces a `TY_POLY` right-hand side through `sp_poly_to_i`.
- `test/lifted_block_nested_capture.rb` (+ committed `.expected`): module-, class-, and instance-method lifted blocks; direct and nested-block captures; a capture mutated across a `Fiber` boundary; poly-element coercion into a captured accumulator; and a small worker pool combining them.

## Notes

Stacked on #1710 (includes its two commits): the call-site block passing added there is what makes these lifted-block scenarios reachable. Reviewing only the top commit shows this fix in isolation.

## Test plan

- The repro and all covered variants compile and match CRuby byte-for-byte.
- New regression test verified through the harness against a committed CRuby snapshot.
- `make test`: 1255 pass, 0 fail, 0 error.
- `make optcarrot`: OK (343 fps, checksum 59662).


bartleusink:
Folded into #1710: its review fixes routed the &blk escape scan through a_block_is_lifted — the function this PR extends — so the two changes were no longer independently reviewable. #1710 now carries this commit (4c2c741d) and both regression tests.

PR #1714, opened 2026-07-05
codegen: multi-assign local targets write through emit_local_ref

khasinski:

A multiple-assignment target that is captured by a later block/lambda lives in a heap cell (`*_cell_<name>`), not a plain `lv_` slot, but the literal-elements multi-assign path wrote the bare `lv_<name>`, which is never declared for celled locals. Doom hits this in `GosuWindow#draw_automap`:

```ruby
min_x, max_x, min_y, max_y = bounds[:min_x], bounds[:max_x], bounds[:min_y], bounds[:max_y]
...
to_sx = ->(wx) { offset_x + (wx - min_x) * scale }
to_sy = ->(wy) { offset_y + (max_y - wy) * scale }
```

`min_x` and `max_y` are closed over by the lambdas, so they are celled, and the generated C referenced undeclared identifiers:

```
lib/doom/platform/gosu_window.rb:1128:3: error: use of undeclared identifier 'lv_min_x'
```

Fix: the multi-assign local-target writes (both the value assignment and the under-filled default fill) now go through `emit_local_ref`, the same cell-aware lvalue helper single assignment already uses.

Regression test: `test/massign_cell_captured_target.rb` (multi-assign targets captured by lambdas, reassigned through the same path).


khasinski:
Copilot's review flagged a real codegen bug here: a multi-assign target that is a captured/celled `TY_PROC` local was written through `emit_local_ref`, which renders the non-assignable READ cast `(sp_Proc *)(uintptr_t)(*_cell_x)`, so `a, b = ...` into such a proc local failed to compile. Fixed in follow-up #1730 (raw int-laundered cell lvalue, mirroring `emit_assign`).

PR #1715, opened 2026-07-05
codegen: inline yielding methods that contain return

khasinski:

A yielding method containing `return` bailed out of inlining (`emit_inline_call_x` checked `scope_has_return`). But a bailed block call falls back to a plain function call against a symbol that is never emitted (yielding methods have no standalone function), producing an undefined-symbol link error. Doom hits this in `PlayerPhysics#each_nearby_linedef`, which has an early `@map.linedefs.each { |ld| yield ld }; return` branch:

```
Undefined symbols for architecture arm64:
  "_sp_PlayerPhysics_each_nearby_linedef", referenced from: ...
```

Changes:

- Inline yielding methods with `return` anyway, funneling the method's own returns to a per-inline exit label (`_yret<tag>`), reusing the existing `g_method_pr_label` funnel machinery. The body is wrapped in its own scope so the funnel goto exits (rather than jumps over) cleanup-attributed GC roots.
- A block spliced at a `yield` site is caller code: its `return` exits the real enclosing function, so `emit_block_invoke` suspends the inline funnel in favor of the real function's funnel, mirrored in the new `g_fn_pr_label` / `g_fn_pr_var` / `g_fn_ret_type` globals (maintained at every fresh function context: method, fiber body, proc body).
- Likewise the spliced block body's `self` is the caller's: the new `g_yield_self_fallback` / `g_yield_self_deref_fallback` record the caller context in the inliners, and `emit_block_invoke` restores it around the block body (doom's `compute_slide` reads `@map` inside a block passed through `each_nearby_linedef` into `MapData#each_linedef_near`, which must not resolve against the inner receiver).
- Nested-inline arg binding: binding an inner inline's params previously zeroed the whole rename table, but the call-site expressions may be an outer inlined body whose locals are renamed. Hide only this inline's renames (`g_nren = saved_nren`) in `emit_inline_call_x` and `emit_ctor_yield_inline`.

Regression tests: `test/inline_yield_method_with_return.rb` (return funnel plus caller-block return) and `test/inline_nested_yield_ivar_self.rb` (nested inline, renamed args, caller-self ivar reads).


khasinski:
Copilot's review flagged two real codegen bugs on the inline-yield path: (1) a `return <expr>` inside a statement-position-inlined yielding method dropped the expression's side effects (the funnel var is NULL there, and `emit_return` only evaluated the argument when it was non-NULL), and (2) `g_yield_self_fallback` aliased the shared static `selfbuf` by pointer, so nested inlining clobbered the caller's self used by a spliced block body. Both fixed in follow-up #1731.

PR #1727, opened 2026-07-05
codegen: box a NULL reference-type object as nil, not a truthy wrapper

khasinski:

A method or hash lookup that can `return nil` yields a genuinely NULL object pointer (a nilable C reference). Boxing it into a poly value with `sp_box_obj` produced an `SP_TAG_OBJ` wrapping a NULL `v.p`, which `sp_poly_truthy` reports as **truthy** -- so an `unless x` / `if x` guard passes and the first field/method read off it dereferences NULL and segfaults.

Fix: box reference-type objects via `sp_box_nullable_obj` so a NULL becomes a proper `SP_TAG_NIL`. A value-type object (a small Struct passed by value) is never NULL and `expr` is not a pointer, so it keeps the plain unconditional box.

In doom this is `TextureManager#[]` boxing `build_composite`'s nullable result: the renderer then segfaulted on `texture.width` at renderer.rb:1228 after `return unless texture` wrongly passed a truthy NULL.

Regression test: test/box_nullable_obj.rb (segfaults without the fix, passes with it).

### DIFF
--- a/src/codegen.c
+++ b/src/codegen.c
@@ -1,5 +1,29 @@
 #include "codegen_internal.h"
 
+/* A reference-backed builtin (IO/Fiber/Thread/Queue/Mutex/ConditionVariable/
+   Enumerator/Exception/Proc/Method) is a genuinely nilable C pointer: an unset
+   ivar, a `return nil` method, or a cache miss yields NULL. It must box via
+   sp_box_nullable_obj so a NULL becomes SP_TAG_NIL rather than a "truthy"
+   SP_TAG_OBJ wrapping NULL v.p -- which passes `unless x`/`if x`, misreads as
+   non-nil (a wrong `nil?`), then segfaults on the first field/method read.
+   Value-type builtins (Range/Time/Complex/Rational, boxed as a by-value copy)
+   and non-pointer scalars are deliberately excluded -- they are never NULL. */
+const char *ty_nullable_builtin_id(TyKind t) {
+  switch (t) {
+    case TY_IO:         return "SP_BUILTIN_IO";
+    case TY_FIBER:      return "SP_BUILTIN_FIBER";
+    case TY_THREAD:     return "SP_BUILTIN_THREAD";
+    case TY_QUEUE:      return "SP_BUILTIN_QUEUE";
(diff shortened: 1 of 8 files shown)


khasinski:
Thanks — both points were valid. Fixed in the "review fixes" commit.

**(a) Built-in reference types.** The original fix only covered user objects (`ty_is_object`). The same NULL-as-truthy bug existed for every reference-backed builtin that `emit_boxed_text` / `emit_boxed` / `emit_box_open`/`close` boxed with `sp_box_obj`: `IO`, `Fiber`, `Thread`, `Queue`, `Mutex`, `ConditionVariable`, `Enumerator`, `Exception`, plus `Proc`/`Method` (which went through the `sp_box_proc`/`sp_box_method` switch cases). These are now routed through `sp_box_nullable_obj` via a single `ty_nullable_builtin_id` helper. Value-type builtins (`Range`/`Time`/`Complex`/`Rational`, boxed as a by-value copy) are deliberately excluded — they are never NULL.

Repro that mis-behaves without the fix (a nilable `IO` boxed into a poly array, guarded):
```ruby
class Sink; def open(p); @io = File.open(p,"w"); end; def io; @io; end; end
s = Sink.new; box = []; box << s.io   # NULL IO boxed via sp_box_obj(..., SP_BUILTIN_IO)
a = box[0]
puts a.nil?          # was `false`, now `true`
puts "boom" if a     # wrongly ran before the fix
```
Added as `test/box_nullable_builtin.rb`.

**(b) Direct `sp_box_obj` bypass paths.** I audited the direct `sp_box_obj(...)` sites that skip the box helpers. Only one boxes a genuinely nilable reference: the poly-slot operator-write dispatch in `codegen_stmt.c` (it boxes a user method result, which can `return nil`) — now nullable-boxed when the class is a reference type. The rest are out of scope because they never produce NULL: `sp_%s_new()` / `sp_Object_new()` (fresh allocations), `sp_Thread_list_at` (a live thread), the `initialize_copy`/`dup` source (the receiver being duped), and container-element boxes (`sp_PolyArray_splice`, nested `IntArray` elements).

PR #1730, opened 2026-07-05
codegen: multi-assign into a captured proc cell needs a raw lvalue

khasinski:
Follow-up to #1714 (Copilot review): the multi-assign write path routed a captured/celled `TY_PROC` local target through `emit_local_ref`, which renders the READ form `(sp_Proc *)(uintptr_t)(*_cell_x)` -- a cast expression, not an assignable lvalue. So `a, b = ...` where one target is a proc local captured by an escaping lambda failed to compile with "expression is not assignable".

`emit_assign` already handles the plain `x = proc` case by writing through the raw int-laundered cell deref (`*_cell_x = (mrb_int)(uintptr_t)(...)` / `*((cap*)_cap)->x = ...`). This ports that shape to both multi-assign write sites (filled value and under-filled nil default) via a small shared helper.

Repro (fails to compile on master, correct after):
```ruby
def make
  p = proc { "orig" }
  esc = lambda { p.call }   # captures p by reference -> p lives in a heap cell
  q = proc { "new" }
  p, n = q, 5               # multi-assign into the captured proc local
  puts n
  esc
end
puts make.call              # => new
```

Regression test: `test/massign_captured_proc_cell.rb` (+ ruby-generated `.expected`).


khasinski:
Fixed in 8a3eec16.

Confirmed the bug: a multi-assign into a captured TY_PROC cell pre-evaluates each RHS element into a temp and then wrote `*_cell = (mrb_int)(uintptr_t)(_tN)`. For a nil/void element the temp is a boxed `sp_RbVal` (`sp_box_nil()`), so casting that struct to `(mrb_int)` is invalid C -- clang rejects it with "operand of type 'sp_RbVal' where arithmetic or pointer type is required".

The write site now extracts the proc pointer before laundering, mirroring how `emit_assign` writes a proc cell:
- nil/void element -> `NULL` (the side effect of a void element already ran when the temp was assigned)
- poly element -> the `.v.p` slot
- otherwise (a bare `sp_Proc *` temp) -> the temp directly

Added `test/massign_proc_cell_nil.rb` covering nil-into-proc-cell multi-assign. Suite: 1260 pass, 2 fail (both the known flakes: system_argument_list, bundle_hash_b2).

PR #1731, opened 2026-07-05
codegen: fix inlined-yield return side effects and nested-inline self

khasinski:
Follow-up to #1715 (Copilot review): two correctness fixes on the inline-yield path.

1. **Dropped return-expression side effects.** A yielding method inlined in statement position sets the per-inline return funnel var to `NULL` (the value is discarded). But `emit_return` only evaluated the return argument when that var was non-NULL, so `return <expr>` inside such an inline skipped evaluating `<expr>` entirely -- dropping its side effects -- and just jumped to the exit label. Now the argument is evaluated for effect (matching the void non-inline return path) before the goto.

   Repro (`side effect happened` is missing on master):
   ```ruby
   def side; puts "side effect happened"; 42; end
   def each_thing
     puts "before"; yield 1; return side()
   end
   each_thing { |x| puts "block #{x}" }
   puts "after"
   ```

2. **Nested-inline self aliasing.** `g_yield_self_fallback` recorded the caller's self BY POINTER, but `g_self` may point at the shared static `selfbuf` that a deeper nested inline overwrites -- so a spliced block body reading an ivar through the caller's self later resolved against the wrong (inner) receiver, e.g. `_t9->iv_tag` where `_t9` is the inner receiver struct (a compile error). Copy the self string into a per-inline stack buffer before assigning the fallback.

Regression tests: `test/inline_yield_return_side_effect.rb`, `test/inline_yield_nested_self_ivar.rb` (+ ruby-generated `.expected`).


khasinski:
Addressed in 6f9fda11.

**1. codegen_iter self still clobbered on a read path? (gemini)** -- Rebutting: the copy is sufficient. The only place the spliced block body's self is established is `emit_block_invoke`, which does `if (g_yield_self_fallback) { g_self = g_yield_self_fallback; ... }` *before* `emit_stmts(bbody)`. So the block body reads `g_self` only after it has been reassigned from the (now string-copied) fallback -- it never reads the shared static `selfbuf` while it is holding an inner receiver.

I verified this with a 3-level repro (`process` inlined, then `@a.run` and `@b.run2` nested inside its spliced block; the deepest block reads `@tag` through the outermost self). With the fix it emits `_t7->iv_tag` (the correct Top receiver). Reverting to the by-pointer alias makes it emit `_t13->iv_tag` (B's receiver) and the C fails to compile (`no member named 'iv_tag' in struct sp_B_s`). Added it as `test/inline_yield_deep_nested_self.rb`.

**2. Same alias in emit_ctor_yield_inline (Copilot)** -- Confirmed and fixed with the same copy-into-buffer approach. No standalone regression test though: `emit_ctor_yield_inline` independently miscompiles any heap-object `initialize` body -- it emits the receiver as a pointer (`sp_X *_t = sp_X_new(...)`) but never sets `g_self_deref`, so the body derefs it with `.` (and value-type receivers hit a value/pointer mismatch). Those are separate pre-existing bugs, out of scope here, and they block a compiling end-to-end test. The alias fix itself is verified by inspection/parity with the codegen_iter path and is strictly safer (a value copy vs a shared-buffer pointer).

**3. Magic 64 vs sizeof selfbuf (Copilot)** -- Done; the codegen_iter fallback buffer is now `char self_fb_buf[sizeof selfbuf]` (and the new ctor fix uses the same form).

Suite: 1261 pass, 2 fail (the known flakes: system_argument_list, bundle_hash_b2).

PR #1734, opened 2026-07-05
poly dispatch: index a Hash reached through a poly value

khasinski:

When a user class defines an instance `[]`, a poly `[]` call is lowered to a per-class dispatch switch on the receiver's runtime class. That switch had array arms and per-class object arms but no Hash arms, so a Hash arriving as a poly value (doom's `@textures = result[:textures]` and `@flats = flats.to_h { ... }`, both widened to poly) fell through the switch and returned nil. Every texture and flat lookup then resolved to nil, so the renderer drew nothing textured.

Adds Hash arms to the per-class `[]` dispatch:
- a poly-key arm routes any hash-kind receiver through `sp_poly_index_poly` (dispatching on the key tag), covering a StrPolyHash indexed by a poly key;
- a string-key arm handles a PolyPolyHash keyed by boxed strings (a `to_h` result), boxing the key.

Both apply only to `[]` (nil on miss); `fetch` keeps its existing default-seed fallthrough so a supplied default is not dropped.

With this plus the NULL-object boxing fix, doom's first frame goes from all-black to about 83 percent drawn, byte-identical to CRuby where it draws.

Regression test: test/poly_hash_index_dispatch.rb (returns nil without the fix, correct values with it).


khasinski:
Thanks — both concerns were valid. Fixed in the "review fixes" commit.

**(b) `fetch` had no arm (Copilot).** Confirmed: the new arms were `[]`-only, so `fetch(existing_key)` on a poly-valued hash fell through to the seeded default and dropped the present value as nil. Repro:
```ruby
h = { "a" => 1, 2 => "b" }        # heterogeneous keys -> PolyPolyHash
ph = [h][0]                        # poly receiver
ph["a"]         # => 1   (worked)
ph.fetch("a")   # => nil (bug; wrong) ; now => 1
```
Fixes:
- Extended the `StrPolyPolyHash` arm and the poly-key arm to handle `fetch`, gating the lookup on a `has_key` check so a present key returns its value and an absent key falls back to the supplied default or raises `KeyError`.
- Added `sp_poly_has_key`, a presence check mirroring `sp_poly_index_poly`'s storage kinds (`sp_poly_index_poly` alone returns nil on a miss, which can't be told apart from a key legitimately mapped to nil).
- The analyzer now types `fetch` on a poly receiver as poly (like `[]`); without that the boxed dispatch result was discarded as nil in value position (`(dispatch, sp_box_nil())`).

**(a) TY_UNKNOWN key consistency (gemini).** Verified the boxing is consistent: a poly key routed through `sp_poly_index_poly` / `sp_poly_has_key` hashes and compares by value — string keys via `sp_str_hash` / `strcmp` in `sp_rbval_hash_key` / `sp_rbval_eql_key` — matching how the `PolyPolyHash` keys were stored (`sp_box_str`). A truly `TY_UNKNOWN` key isn't readily constructible (the analyzer infers concrete or poly types), but since such a key is already held boxed-as-poly (`atmp_ty == TY_POLY`), I widened the poly-key arm guard to `TY_POLY || TY_UNKNOWN` so a Hash reached by one isn't dropped to nil.

New coverage: `test/poly_hash_fetch_dispatch.rb` (fetch-present, fetch-default, `KeyError`-on-absent, a poly key, and a user-class `fetch` arm still winning). Full suite green (the two failures are the pre-existing `system_argument_list` flake and the `bundle_hash_b2` artifact).

PR #2044, opened 2026-07-11
numerics: Rational division, Complex classes, negative exponents, Math parity

ryanseys:
This closes the Rational/Complex/Integer/Math conformance gaps that are representable in the current value-type model:

- **Rational division family** — `quo` (aliases `/`, Float operands divide as floats), `fdiv` (always Float), `div` (floor division to Integer via `sp_rational_idiv`), and `round`/`truncate` with a precision argument (`sp_rational_*_prec`: scale by 10^n, round or chop, scale back). A literal precision picks the result class statically; a non-literal precision boxes to poly and picks the class from the runtime value.
- **Kernel#Rational exactness** — `Rational(2.5)` is the exact value of the double, `(5/2)`, routed through the frexp conversion behind `Float#to_r`; the two-arg form with Float/Rational operands divides the two exact values (so `Rational(1.0, 0.0)` raises `ZeroDivisionError`).
- **Complex.rect / #arg / #polar / #rect** — the component-pair constructor, `atan2`-based `arg`/`angle`/`phase`, and the `[abs, arg]` / `[re, im]` pairs.
- **Complex component classes** — `sp_Complex` gains a two-bit class field so each component keeps its CRuby class: `Complex(1, 0.0).to_s == "1+0.0i"`, `Complex(0, 2).abs == 2` (Integer), `Complex(3, 4).abs2 == 25`, `real`/`imaginary` return the component class, arithmetic propagates per component (add/sub) or floats both (mul/div), and `Complex.polar` resolves exact right angles without cos/sin, preserving the magnitude class. Zero-init keeps every untouched construction site on its current rendering.
- **Negative integer exponents** — `2 ** -2 == (1/4)` typed statically for literal exponents; the poly-dispatched path resolves Integer-or-Rational from the sign at runtime (`sp_poly_pow`, which also now squares-and-multiplies the int/int case exactly instead of round-tripping through `pow(double)`). A statically int-typed runtime exponent keeps the loud `RangeError` -- typing it a sometimes-Rational would cascade poly through every int-arithmetic consumer; `docs/limitations.md` documents the residual. `Integer#pow(negative, mod)` raises `RangeError` with CRuby's message instead of silently returning 1.
- **Float/Math tail** — `Float#numerator`/`#denominator` (exact rational of the double), `Math.frexp` returns its `[fraction, exponent]` pair (was mistyped Float), and Rational arguments to `Math.*` coerce through `sp_rational_to_f` instead of raising `TypeError`.

Deliberately left out (documented, still loud rejects): `Rational#to_c` and Rational-with-Complex arithmetic (need rational complex components; `sp_Complex` is a float pair), `Rational#to_f` past int64 (Rational is int64/int64 by design), and `Float#round/truncate` with a non-literal ndigits (documented divergence in `docs/float-rounding.md`).

Each commit carries an oracle-generated test matrix (happy, edge, and exception paths with exact CRuby classes and messages). Verified with per-test oracle diffs, an ASAN + `SPINEL_GC_STRESS=1` pass over the new tests, and a full gaps-repro sweep showing no classification regressions.


ryanseys:
@matz I send this as a single PR with multiple commits to save some of the inbox noise / overhead of managing many PRs. Let me know if you prefer smaller PRs. I tried to find the right balance here and kept this focused on a particular subset of gaps. Let me know, happy to adjust the granularity as needed.

Also less PRs means hopefully less "You have reached your daily quota limit. Please wait up to 24 hours and I will start processing your requests again!" from gemini in the future 🙈

matz:
Reviewed — the Rational/Complex split with per-component classes and the exact Kernel#Rational conversion all look right, and the deliberate-residuals section matches our limitations policy. CI is green on all three platforms. This is still marked draft, so I'll hold off; mark it ready whenever you are and it will be merged as-is.

PR #2360, opened 2026-07-13
codegen: route Time values in string interpolation through Time#to_s

### LINKED ISSUE #2358: codegen: string interpolation of a Time value is unsupported (explicit .to_s works)
## Gap

Interpolating a `Time` value refuses at the front end, even though `Time#to_s` exists and works:

```ruby
t = Time.at(0).utc
puts "at '#{t}'"   # CRuby: at '1970-01-01 00:00:00 UTC'
```

```
t.rb:2: unsupported interpolation value: node 12 (EmbeddedStatementsNode)
```

## Delineation

- **Explicit call works**: `puts "at '#{t.to_s}'"` compiles and prints byte-identically to CRuby — `Time#to_s` itself is fine; only the interpolation path doesn't route through it.
- **Shape-independent**: local, ivar, and reader-method receivers; single- and multi-line strings; `Time.now`, `Time.now.utc`, and `Time.at(0).utc` — all refuse identically.

## Tests

The repro is deterministic and suite-shaped: `test/time_interp.rb` + `.expected` = `at '1970-01-01 00:00:00 UTC'` (CRuby-generated).

Surfaced AOT-compiling a Rails app (lobsters), same investigation as #2357 (which is fixed at f53e02d4 — thanks!): SQL and cache-key strings embed times, e.g. `(comments.created_at >= '#{period}')`.

Verified at master dfdcd2d5 (clean build, macOS arm64 clang).

rubys:
## Gap

`"#{t}"` on a `Time` refused with `unsupported interpolation value: (EmbeddedStatementsNode)` even though `Time#to_s` exists and the explicit `"#{t.to_s}"` form compiles (#2358):

```ruby
t = Time.at(0).utc
puts "at '#{t}'"   # CRuby: at '1970-01-01 00:00:00 UTC'
```

## Fix

The interpolation dispatch in `emit_interp` (codegen_expr.c) has arms for every scalar kind — Int, Float, Bool, Symbol, Rational, Complex, Regexp, arrays, hashes, objects — but not `TY_TIME`. Add the missing arm calling `sp_time_to_s_v`, the same value-taking helper the method-call path (`codegen_call_recv.c`) already uses for `Time#to_s`: whole-seconds semantics, no fractional part, and it returns a GC string via `sp_str_dup_external` so the dynamic-part rooting in the interp machinery is sound.

## Tests

`test/time_interp.rb` (+ `.expected` from CRuby): local, reader-method, and multiline-string shapes, explicit-`to_s` equivalence, and the fractional-seconds-drop semantics (`Time.at(1234567890.5).utc` interpolates without the `.5`). `make test`: 1754 pass, 1 fail — the failure is `test/array_conformance_batch.rb`, which fails identically on clean master (pre-existing, unrelated).

Fixes #2358

Surfaced AOT-compiling a Rails app (lobsters): SQL and cache-key strings embed times (`(comments.created_at >= '#{period}')`); verified the full app compile clears that site with this fix.


matz:
Thanks! This exact fix already landed on master as commit 73a48642 (issue #2358) — the same sp_time_to_s_v arm in emit_interp's interpolation dispatch. The CI failures here are because the change now conflicts with the version already on master. Closing as already-resolved; appreciate the report.

PR #3123, opened 2026-07-20
analyze: two O(N^2) passes on large trees (empty-hash / Hash.new default)

rubys:
Follow-on to fd13553e (`analyze: index writes in desugar_builtin_class_var_recv`). Profiling the same lobsters ~370-file tree afterwards surfaced two more O(N^2) analyze passes stacked behind it; together they were ~65% of compile time.

**Baseline:** ~82s to analyze the tree, single core pinned, 1.25T instructions. A `sample` showed ~72% of stacks in a single `sm_lookup` loop.

### 1. `mark_empty_literal_args` — freeze the scope index
For every `CallNode` it does `for k in nclasses { comp_method_in_class; comp_cmethod_in_class }`, and it runs in `analyze_program`'s *unfrozen* region, so each lookup falls back to `sm_lookup`'s linear reverse scan over all scopes → O(nodes × classes × scopes). The pass never creates or renames scopes, so freezing the index for its duration lets the lookups hit the existing O(1) hash index. Behaviour-identical (the frozen chain resolves in the same order as the reverse scan). **~82s → ~67s.**

### 2. `hash_new_default_arg` — memoize while frozen
Its `LocalVariableReadNode` branch scans every write node in the program, and `infer_call` calls it for every `x[k]` / `x.default` / `x.values_at` on an untyped receiver during the inference fixpoint → O(N) per call, O(N^2) overall. The result is a pure function of the static AST + (frozen) scope assignments, so it's memoized per receiver node while frozen, stamped with a scope-index generation counter (bumped on every freeze/unfreeze) so the cache is discarded whenever scope shape could have changed. **~67s → ~45s.**

**Net: ~82s → ~45s (~45% faster), 1.25T → 0.67T instructions.**

Both are the same bug class as fd13553e (a per-node helper rescanning the whole node/scope table). Verification: full suite **2219 pass / 0 fail / 0 error**; RBS extractor + seed tests pass; `Hash.new` default folding verified behaviorally.

### DIFF
--- a/src/analyze.c
+++ b/src/analyze.c
@@ -5369,6 +5369,12 @@ static int param_nonstr_index_written(Compiler *c, int mi, int p) {
 static void mark_empty_literal_args(Compiler *c) {
   if (!c->empty_hash_arg) return;
   const NodeTable *nt = c->nt;
+  /* This pass does not create or rename scopes, so the scope-shape index is
+     stable across it: freeze so the per-class method lookups below (called for
+     every CallNode * every class) hit the O(1) hash index instead of the
+     unfrozen O(nscopes) linear reverse scan -- the dominant analyze cost on
+     large apps (lobsters: ~72% of compile time was this triple loop). */
+  comp_scope_index_set_frozen(1);
   for (int id = 0; id < nt->count; id++) {
     if (nt_kind(nt, id) != NK_CallNode) continue;
     const char *nm = nt_str(nt, id, "name");
@@ -5419,6 +5425,7 @@ static void mark_empty_literal_args(Compiler *c) {
       }
     }
   }
+  comp_scope_index_set_frozen(0);
 }
 /* An empty `{}` compared against a hash-typed peer takes that peer's variant,
    so the two sides share a representation and `==` can compare contents
--- a/src/analyze_util.c
+++ b/src/analyze_util.c
@@ -301,7 +301,44 @@ int anon_struct_ci_for_value(Compiler *c, int val) {
    return the default-value argument node, or -1. Lets a literal receiver's
    .default / [] / fetch fold to the default when the hash type never
    narrows (no writes ever reach it). */
+static int hash_new_default_arg_compute(Compiler *c, int recv);
+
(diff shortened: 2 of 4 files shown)


rubys:
Thanks — both addressed in the amended commit (49f1e115):

**1. `sm_gen` bumping on every call (Gemini + Copilot).** Now `comp_scope_index_set_frozen` bumps the generation only on an actual `sm_frozen != f` transition, matching the comment and avoiding needless memo re-inits. Safe because scope shape only ever changes while unfrozen, so every scope change is still followed by a `0→1` transition that bumps the generation before the memo is next read.

**2. Infinite recursion via a local-variable write cycle (Copilot).** Confirmed real: `hash_new_default_arg_compute` recurses through a local's write values, so `a = b; b = a` (both untyped, same scope) re-enters `hash_new_default_arg` for a node whose slot is still `INT_MIN`. The memo now marks each slot resolved-but-dynamic (`-1`) *before* recomputing; `-1` is the correct conservative result for a cyclic (non-single-`Hash.new`) definition and breaks the cycle. Verified with an untyped `a = b; b = a` followed by `a[0]` / `b.default` — compiles and runs instead of hanging.

Full suite still 2219 pass / 0 fail / 0 error; lobsters analyze time unchanged (~45s).

matz:
Merged, thanks — this is a clean continuation of fd13553e and the wins reproduce here.

Verified before merging: no conflicts against current master (30 commits ahead of your base), full suite green (2248 pass / 0 fail / 0 error, gcc and clang), optcarrot checksum 59662, bench 58 pass. On a 320-class synthetic tree heavy on `Hash.new(default)` and empty-hash arguments, analyze time drops from ~5.0s to ~1.7s with **byte-identical generated C**, which is the strongest form of the behaviour-identical claim.

Also confirmed the two safety preconditions directly: `mark_empty_literal_args` performs no node/scope creation or renames (pure marking), and `Compiler` is calloc-zeroed so the new memo fields initialize clean; the freeze-epoch stamp correctly discards the memo across every freeze/unfreeze transition.

Issue #3234, opened 2026-07-22
[Runtime] A value read back out of a container loses part of its method table


KieranP:
> (Disclaimer: I asked Claude Opus 4.8 to find Spinel compile/runtime issues with documented Ruby classes and methods. It found the following issue, which I manually verified for accuracy before publishing this report)

_Reproduced with `spinel d298e1e06ec9` vs `ruby 4.0.3`._

When an Array/Set/Range/Time value is read out of a container (an array element, a hash value, a `.each` block parameter, a destructured pair) and a method is called on it, Spinel drops a subset of that type's methods: most raise `NoMethodError`, and `Array#map!` silently no-ops. A direct local of the same value works, and other methods on the very same container-read receiver still work, so the loss is per-method dispatch against a type-erased receiver. Shares the mechanism of `20260722184742254` (a String read from a nested array loses `ljust`/`rjust`/`center`) and `20260722194429454` (a Method object loses its type in a container).

## Repro

`Array#map!` is a silent no-op on a container-read array (a direct local mutates fine):

```ruby
m = [[1,2],[3,4]]; m[1].map! { |x| x*10 }; p m
```

Ruby: `[[1, 2], [30, 40]]`
Spinel: `[[1, 2], [3, 4]]`

`Set#map` returns `[]` and `select`/`inject` raise on a container-read Set (`size` on the same receiver works):

```ruby
require 'set'
h = { a: Set.new([1,2,3]) }
p h[:a].map { |x| x }
p (h[:a].select { |x| x > 1 } rescue $!.class)
p (h[:a].inject(:+) rescue $!.class)
```

Ruby: `[1, 2, 3]` / `[2, 3]` / `6`
Spinel: `[]` / `NoMethodError` / `NoMethodError`

`Range#sum` returns `0` and `Range#cover?` raises on a container-read Range:

```ruby
r = [(1..5)][0]; p r.sum
c = [(10..20)][0]; p (c.cover?(15) rescue $!.class)
```

Ruby: `15` / `true`
Spinel: `0` / `NoMethodError`

`Time#-` raises `NoMethodError` on a container-read Time:

```ruby
arr = [Time.at(0), Time.at(3600)]; a = arr[0]; b = arr[1]
p ((b - a) rescue $!.class)
```

Ruby: `3600.0`
Spinel: `NoMethodError`

`Array#sum` with a Rational seed raises on a container-read row (no-arg `row.sum` works):

```ruby
[...]

PR #3286, opened 2026-07-23
Gate the container-read pre-arms by name before probing the receiver (perf: Rails-scale compile 23s -> >10min regression)

rubys:
The #3234 container-read pre-arms in `infer_call` evaluate `infer_type(c, recv)` before any name/argc gate, so **every** one-arg call with a receiver pays a recursive receiver-inference probe on top of the normal receiver typing — each level of a call chain re-infers its whole receiver suffix once more, which goes exponential on deep chains. Small trees barely notice; a Rails-scale tree (roundhouse's lobsters emit, ~735 files) went from **23s to >10 minutes** (killed; 100% CPU in `infer_call` self-recursion per `sample`).

This reorders the condition so the four rare arm names and `argc == 1` gate first, and only matching sites pay the `infer_type` probe. Behavior is identical — same arms, same results — so `test/container_read_builtin_methods.rb` continues to cover the semantics.

Measured on the same lobsters tree, same machine:

| compiler | time |
|---|---|
| `caa2426` (pre-#3234-series anchor) | 22.9s |
| `69cf91b^` | 20s |
| `69cf91b` .. current master | >600s (killed; clean-build verified) |
| master + this change | **21.6s** |

Bisected to `69cf91b` (the commit introducing the pre-arms); verified by clean-rebuild A/B against its parent, and by this one-block change restoring master to 21.6s. `make test` green.

### DIFF
--- a/src/analyze_infer.c
+++ b/src/analyze_infer.c
@@ -1588,8 +1588,17 @@ TyKind infer_call(Compiler *c, int id) {
       nt_str(nt, recv, "name") && sp_streq(nt_str(nt, recv, "name"), "Hash")) {
     if (sp_streq(name, "try_convert")) return TY_POLY;
   }
-  /* container-read builtin pre-arms (#3234) */
-  if (recv >= 0 && infer_type(c, recv) == TY_POLY && argc == 1) {
+  /* container-read builtin pre-arms (#3234). The name/argc gates run
+     FIRST: the infer_type(recv) probe recurses through the receiver's
+     own call chain, and paying it for EVERY one-arg call re-infers each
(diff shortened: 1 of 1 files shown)


matz:
Verified and merged. Reproduced the blowup locally with a synthetic depth-22 chain (60 statements): pre-merge master exceeds a 10-minute timeout on it, with this change the same input compiles in 0.08s. The gate list matches the arm body's name set exactly (cover?/gcdlcm/sum/inject/reduce), so behavior is unchanged; full test (2399), bench (58) and optcarrot (checksum 59662, fps in band) are green on the merged master. Thanks for the precise bisect and the A/B table.

Issue #3115, opened 2026-07-20
analyze: analyze_program scales O(N²) on a mid-size tree (desugar_builtin_class_var_recv rescans the node table per receiver)


rubys:
## Symptom

`analyze_program` scales quadratically in program size, so a mid-size tree takes minutes.
On the lobsters emitted tree (~370 files) `spinel main.rb --rbs . -c` does not finish within
200 s (single core pinned at 100%). This is a *separate* hot path from the recursion in #2707
(which is fixed — that tree completes now); the dominant cost has moved.

## Where the time goes

`sample` on master (6b52ab16, freshly built), representative of the whole run — ~85% of stacks:

```
codegen_program
  analyze_program                       (+25116)
    desugar_builtin_class_var_recv
      builtin_class_var_static_name
        nt_kind                         <- innermost, called per node, per receiver
```

Reading the two functions:

- **`desugar_builtin_class_var_recv`** (`src/analyze_pass.c:3651`) loops over the whole node
  table looking for call receivers:

  ```c
  int n0 = nt->count;
  for (int id = 0; id < n0; id++) {
    if (nt_kind(nt, id) != NK_CallNode) continue;
    int recv = nt_ref(nt, id, "receiver");
    if (recv < 0 || nt_kind(nt, recv) != NK_LocalVariableReadNode) continue;
    const char *cn = builtin_class_var_static_name(c, recv);   // <-- per receiver
    ...
  }
  ```

- **`builtin_class_var_static_name`** (`src/analyze_util.c:154`) then does *another* full scan
  of the whole node table for each receiver, hunting the matching `LocalVariableWriteNode`s:

  ```c
  for (int w = 0; w < nt->count; w++) {                        // src/analyze_util.c:161
    if (nt_kind(nt, w) != NK_LocalVariableWriteNode) continue;
    const char *wn = nt_str(nt, w, "name");
    if (!wn || !sp_streq(wn, vn) || comp_scope_of(c, w) != sc) continue;
    ...
  }
  ```

So the pass is **O(receivers × total_nodes) ≈ O(N²)** in node count. On a small tree it's
invisible; on the lobsters tree the two nested full-table walks dominate. The recent extension
of the qualifying clause ("a user class constant qualifies too", refs #2717/#2721) increases how
many receivers reach the inner scan, which likely widened the effect.

## Numbers (master 6b52ab16)

- Trimmed ~60-file models-only tree (the `#2707` repro kit): **~48 s** now (vs the README's
[...]

PR #3292, opened 2026-07-23
Memoize an_user_defines_method: the every-class probe ran per builtin-fallback call site (perf: lobsters 20.3s -> 17.6s)

rubys:
Profiling the post-#3286 compile of roundhouse's lobsters emit (42k lines of plain Ruby, the [#3115](https://github.com/matz/spinel/issues/3115) tree one scale up) puts ~22% of self-time under one helper: `an_user_defines_method`. The ~20 builtin-fallback arms in `infer_call` ask "does the program define this method name itself?" per unresolved call site, per fixpoint iteration — and each ask probes **every class's whole ancestor chain**, O(nclasses × chain) `sm_lookup`s per query. It's the #3149 mole one loop further out: the per-item helper that rescans the whole table.

The answer is a pure function of scope/class shape, so this memoizes it per name while the scope index is frozen, stamped with `comp_scope_index_gen()` — the same invalidation `hash_new_default_arg`'s memo already rides. The fill is O(1) for the common case via two per-epoch sets (every instance-method name any class defines; every alias source name): a name no alias redirects resolves identically from every chain start, so membership in the defined-set answers exactly. Alias-source names — where per-class resolution can genuinely redirect — keep the exact every-class scan, and are memoized like everything else. Unfrozen phases bypass the memo entirely (scope shape can move without the counts changing) and pay what they always paid.

| compiler | time |
|---|---|
| master + this change | **17.6s** |
| master + this + #3293 + #3294 | **9.0s** |

Sibling PRs from the same profiling session, independent and individually mergeable: #3293 (local-write queries), #3294 (proc-literal escape marking). One observation for later, not this PR: passes like `mark_empty_literal_args` freeze/unfreeze around themselves, ticking the epoch without changing scope shape — every epoch-stamped memo (this one and `hash_new_default_arg`'s) resets each time. A shape-generation counter decoupled from the freeze flag would let them survive; that's the next ~8% on this tree.

### DIFF
--- a/src/analyze_util.c
+++ b/src/analyze_util.c
@@ -635,10 +635,131 @@ int class_is_blank_slate(Compiler *c, int ci) {
 
 /* Does the program define a method of this name itself? A builtin fallback for
    a poly receiver must decline then: the user method is the likelier target. */
+/* The scan below is O(nclasses * chain) per query and the fallback arms ask it
+   per call site per fixpoint iteration; memoize per name. The answer depends
+   only on scope/class shape, so stamp entries with the scope-index epoch (the
+   same invalidation hash_new_default_arg's memo rides) plus the scope/class
+   counts for the unfrozen phases where the epoch doesn't tick. */
+static unsigned udm_gen = 0;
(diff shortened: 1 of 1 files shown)


matz:

PR #3293, opened 2026-07-23
Route (scope, name) local-write queries through a shared LWIndex instead of whole-table rescans (perf: lobsters 20.3s -> 16.2s)

rubys:
Three helpers still answer "every write of local X in scope S …" by rescanning the **entire node table** per query: `local_all_writes_empty_hash`, `local_all_writes_empty_array`, and the local-variable branch of `hash_new_default_arg_compute`. The first two sit forty lines above the `LWIndex` that #3123 introduced to fix exactly this shape — they just don't use it. All three are asked from `bind_call_params` / receiver typing for call arguments during the inference fixpoint; on roundhouse's lobsters emit (42k lines) they account for ~11% of compile time (`local_all_writes_empty_hash` alone was 7% of self-time in a `sample` profile).

This routes all three through one long-lived `LWIndex` keyed by (scope, name):

- cached while the scope index is frozen, stamped with `comp_scope_index_gen()`; unfrozen phases rebuild per query, which costs the single table walk they already paid (scope shape can move without the epoch ticking there, so the cache is never trusted across unfrozen mutations);
- `lw_index_build` itself now iterates only the four local-write node kinds via the kind-grouped id lists (`nt_nodes_of_kind`) instead of scanning the whole table — this also cheapens the existing per-pass builds. Chain order becomes kind-grouped rather than globally ascending; every consumer is an order-independent existence/uniqueness walk (`has_write` early-exits, all-writes-match folds, single-write checks);
- the accessors are exported through `analyze_internal.h` so `hash_new_default_arg_compute` (analyze_util.c) can walk the same index.

| compiler | time |
|---|---|
| master + this change | **16.2s** |
| master + this + #3292 + #3294 | **9.0s** |

Sibling PRs from the same profiling session, independent and individually mergeable: #3292 (`an_user_defines_method` memo), #3294 (proc-literal escape marking). This is the "sibling helpers" list #3115 predicted, still paying out.

### DIFF
--- a/src/analyze_internal.h
+++ b/src/analyze_internal.h
@@ -139,6 +139,14 @@ extern int g_yvt_unify_all;
 TyKind method_call_ret(Compiler *c, int mi, int call_id);
 /* is_proc_constant / is_proc_literal are declared in analyze.h (codegen needs them). */
 int is_proc_create(Compiler *c, int id);
+
+/* Shared cached local-write index (analyze_pass.c): bucket walk over
+   (scope, name) for "every write of local X in scope S" queries, instead of
+   a whole-table rescan per query. Callers must still confirm scope/name
+   (hash collisions) and node kind on each record. */
+int lw_shared_first(Compiler *c, const char *name, int scope);
+int lw_shared_node(int rec);
+int lw_shared_next(int rec);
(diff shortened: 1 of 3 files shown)


matz:

Issue #3175, opened 2026-07-21
[Runtime] A lambda that returns a local reassigned from an array index inside a loop is mistyped, returning a garbage value


KieranP:
> (Disclaimer: I asked Claude Opus 4.8 to find Spinel compile/runtime issues with documented Ruby classes and methods. It found the following issue, which I manually verified for accuracy before publishing this report)

_Reproduced with `spinel 14ead3d20a9b` vs `ruby 4.0.3`._

A lambda whose return value is a local that gets reassigned inside a loop from the index of another
call's array result (`out = pp.call(rest); rest = out[1]`) has its return type mis-inferred. Instead of
the actual `String`, Spinel returns a garbage `Integer` (a raw pointer value printed as a number). The
value is a real `String` at runtime — `String#[1..]` produced it — but the lambda hands it back as the
wrong type. This surfaces in parser-combinator style code where a `[value, rest]` pair threads through
a fold.

## Repro

```ruby
def seq(parsers)
  ->(s) {
    rest = s
    parsers.each { |pp| out = pp.call(rest); rest = out[1] }
    rest
  }
end

lit = ->(str) { [str[0], str[1..]] }
r = seq([lit]).call("abc")
p r          # Ruby: "bc"       Spinel: 2151525692 (garbage Integer)
p r.class    # Ruby: String     Spinel: Integer
```

Ruby: `"bc"` / `String`
Spinel: a large Integer / `Integer`

The garbage number varies per run (it is a reinterpreted pointer), so the stable assertion is on
`.class`. Returning `rest` directly without the loop, or without sourcing it from `out[1]`, returns the
correct String.

matz:
Fixed in cfa31459. Proc literals that escape via a method return or via a local stored in a container now get poly params (so a String arg keeps its type), and a Range index on a poly string now slices the substring instead of returning char 0.

PR #3294, opened 2026-07-23
ple_mark_escaped: precollect proc-literal writes instead of rescanning the table per argument (perf: lobsters 20.3s -> 18.2s)

rubys:
`ple_mark_escaped` (#3175) handles the `f = ->(a){...}; g([f])` case by finding the proc literal assigned to an escaping bare local read — with a scan over **every node in the table** per marked argument. `ple_build` invokes the marker for every call argument, every hash value, every array element, and every return value in the program, making the build O(marked-args × nodes). On roundhouse's lobsters emit (42k lines) this was ~18% of compile time — the largest single branch in a `sample` profile, sitting under `infer_block_params` → `proc_literal_escapes_as_arg`.

The fix collects the writes whose value is a proc/lambda literal once at the top of `ple_build`, via the kind-grouped `NT_FOREACH_KIND` walk. Such writes are rare, so a flat (scope, name, literal) triple list walked per mark replaces the whole-table scan while computing the same set: the marker still matches on same-name, same-scope plain local writes only. Invalidation is unchanged — the list lives and dies with the escape bitmap it feeds (rebuilt when the node table grows).

| compiler | time |
|---|---|
| master + this change | **18.2s** |
| master + this + #3292 + #3293 | **9.0s** |

Sibling PRs from the same profiling session, independent and individually mergeable: #3292 (`an_user_defines_method` memo), #3293 (local-write queries).

### DIFF
--- a/src/analyze_pass.c
+++ b/src/analyze_pass.c
@@ -6072,6 +6072,12 @@ static int register_proc_numbered(Compiler *c, int create) {
 static const NodeTable *ple_nt = NULL;
 static int ple_ntc = -1;
 static char *ple_escaped = NULL;
+/* Writes whose value is a proc/lambda literal, collected once per build so
+   marking a bare local read doesn't rescan the whole node table per argument
+   (each such write is (scope, name) -> literal id). Proc-literal writes are
+   rare, so a flat list walked per mark is plenty. */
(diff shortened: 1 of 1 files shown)


matz:

PR #3372, opened 2026-07-25
spin: add a --debug / -g flag for a debuggable build

rubys:
## What

Adds `spin build --debug` / `-g` (and a `SPIN_DEBUG=1` env var) to produce a debuggable `-O0` build, forwarding the compiler's existing `--debug` (`#line` + `-g -O0`).

## Why

`spin build` always compiles a release binary, and there was no way to get a debuggable one through `spin`. Stepping an emitted program in lldb/gdb — for example, to find where an AOT-compiled application raises at runtime — meant reconstructing the underlying `spinel … --debug` command line by hand (dependency `-I` paths, `--link` objects, `--rbs`, …). The flag makes that a one-liner.

## How

Mirrors the existing `--allow-native-build` / `SPIN_ALLOW_NATIVE_BUILD` convention: the flag sets `SPIN_DEBUG=1`, and `compile_cmd` forwards `--debug` to the compiler when it's set. Applies to `build`, `run`, and `test`.

## Verified

On a real project:
- `spin build` → 4.48 MB release binary
- `spin build --debug` (and the `-g` alias) → 6.67 MB binary with `#line`/`-g`, steppable in lldb
- `spin --help` documents the flag

### DIFF
--- a/tools/spin.rb
+++ b/tools/spin.rb
@@ -16,6 +16,8 @@ usage: spin <command> [args]
   remove <name>        drop a dependency + relock
   lock | fetch | vendor  resolve deps / warm the cache / copy into vendor/
   build [target..]     build bin/ executables into build/bin/
+                       (--debug / -g: debuggable -O0 build for lldb/gdb;
+                        also SPIN_DEBUG=1 — applies to build/run/test)
   run [target] [-- a]  build, then run one executable
   test [file..]        build and run test/*.rb against expectations
   trust <name>         always allow <name>'s declared native build steps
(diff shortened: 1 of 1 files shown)


matz:
Merged as 51f8ff34. Thank you.

I fixed one thing in 8c6cd519 before pushing. `--debug` rides an env var into `compile_cmd` rather than through `extra`, so `cmd_build`'s mtime short-circuit -- which skips a rebuild when `extra == ""` -- did not see it. Build once, then ask for a debug build, and spin answered `build app (up to date)` and handed back the release binary. Silently, with a success message, which is the one outcome a debug flag must never produce; the reverse held too. Your verified 4.48 MB / 6.67 MB numbers are consistent with a clean build dir, so this would not have shown up in that check. The mode is now recorded in a stamp beside the binary and a mismatch forces a rebuild either way.

Two smaller ones in the same commit. The `--debug` / `-g` strip ran over the whole argument array, before the `--` split, so `spin run app -- -g` turned the program's own flag into a debug build instead of passing it through; only args before `--` are spin's. And the usage text and new comment used em-dashes, which this project avoids.

`spin test --debug` does work, incidentally: `--debug` is appended after `extra` and the compiler parses left to right, so it overrides the hardcoded `-O 1`.

PR #3376, opened 2026-07-25
IO#write: write the bytes a String says it has

rubys:
`IO#write` and `#write_nonblock` size their operand with `strlen`, so a String holding an embedded NUL is truncated at the first one. It fails quietly — the short count comes back as the write's return value, and no exception is raised.

```ruby
require "socket"
payload = "AB\x00CD"
server = TCPServer.new("127.0.0.1", 0)
t = Thread.new { c = server.accept; got = c.read(5); c.close; got }
sock = TCPSocket.new("127.0.0.1", server.addr[1])
wrote = sock.write(payload)
sock.close
p payload.bytesize, wrote, t.value.bytesize
```

| | CRuby | spinel (before) | spinel (after) |
|---|---|---|---|
| `payload.bytesize` | 5 | 5 | 5 |
| `sock.write` returned | 5 | **2** | 5 |
| peer received | 5 | **2** | 5 |

## Why a socket needs to write a NUL

A socket carries protocol bytes, not text, and NUL shows up in ordinary framing rather than in exotic payloads.

**WebSocket frames (RFC 6455) — and this one needs no binary payload at all.** A frame's length is encoded in binary in its header, whatever the payload is. For a payload of 126..65535 bytes the length goes out as a 2-byte big-endian field, so any length that is a multiple of 256 puts a NUL *in the header*.

A plain Action Cable broadcast is JSON in a text frame (opcode 1). At 256 bytes of JSON its header is `81 7E 01 00` — the write stops after 3 bytes, and the peer is left parsing a header that never finishes. Payloads in the 126..65535 range are entirely ordinary for Action Cable, so roughly 1 broadcast in 256 truncates: often enough to break in production, rare enough to be miserable to track down. Nothing about the application is binary; the framing is.

The header's first byte can be NUL too: it is `(fin ? 0x80 : 0x00) | opcode`, so a continuation frame that isn't final is `0x00` and the write sends nothing whatsoever.

**Static files.** A PNG carries the IHDR chunk length `00 00 00 0D` at offset 8, immediately after the 8-byte signature; `.ico` and `.woff` are the same story. Serving one by reading it and writing it back out delivers 8 bytes of a 207,346-byte file (measured, not hypothetical), and the browser then hangs waiting out the `Content-Length` that was advertised honestly.

The read side already handles all of this: `sp_sock_read_nb` builds its result from `(buf, n)`, and `File.read` and `IO#read` return NUL-bearing strings whole (verified — a 5-byte file with an embedded NUL reads back as 5). So a program can receive a byte string it has no way to echo back, and `#bytesize` will confirm the bytes are there while the write drops them.

The practical effect is that the socket classes can't be used to write a binary protocol from Ruby at all — the only way through is FFI to a length-taking C function, which is what we ended up doing. This change is what would let that code be ordinary Ruby.

## The change

Size the operand with `sp_str_byte_len`, which reads the length the string header already records. It is also O(1) where `strlen` walks the buffer, so a large body is no longer measured twice on its way out.

Paired exactly like `sp_String_append` / `sp_String_append_bin` in `lib/sp_string.h`, which exist for this same reason on the String side. The `_bin` entry is emitted **only** where the operand is a String value — the same value class `String#<<` and `#replace` already size this way. Everything else keeps the plain `strlen` entry, because those really are bare C literals with no marker byte at `s[-1]`:

- runtime-internal callers (`sp_runtime.h`'s poly `<<` arm),
- codegen's own synthesized `""` and `"\n"`,
- the `sp_poly_to_s` arm, which answers `sp_class_val_name` / `sp_sym_to_s` — static names.

**This routing is the part worth reviewing hardest.** `sp_str_byte_len` reads `s[-1]`, so sending a bare literal through the `_bin` entry would be an out-of-bounds read at a section edge — the hazard `codegen_expr.c` calls out where it deliberately keeps `strlen` for interpolation parts. I believe the `TY_STRING` gate is the right line because it is the line `String#<<` already draws, but a second opinion on that predicate is the thing I'd most want.

`IO#puts` and `#print` are left on `strlen`. Their non-socket arm writes through `fputs`, which would have to become `fwrite` first — a wider blast radius, and a separate change.

## Verification

`make check` (i.e. `OPT=-O1`): **2485 pass** of 2486, plus the new `test/socket_binary_write.rb` covering both `write` and `write_nonblock`, with `.expected` generated from CRuby.

The one failure is `socket_addrinfo_and_option`, which is **pre-existing and unrelated** — it reproduces identically on a clean master:

```
make build/test-results/socket_addrinfo_and_option.ok OPT=-O1
```

fails with or without this commit, and passes at `-O0` either way. `Addrinfo#ipv4?` / `#ipv6?` answer `false` at `-O1` and `true` at `-O0`. Happy to open that separately — it looked worth knowing about on its own.


matz:
Merged as 0ff37b7b. Thank you; the motivation section is the clearest bug report I have read in a while, and the WebSocket length-field analysis is what makes the case.

I extended it in 06e02538 before pushing. Four more sites reach a write with an operand whose length is already known and were still truncating, and a couple of the resulting asymmetries were worse than the original bug because the same call answered differently depending on how it was spelled:

- `IO.copy_stream` -- its operand is the result of `sp_File_read`, so it is always a marked heap string that may carry NULs. This is the "serve a PNG" case your description uses as motivation, and it was the one path where the operand's provenance is certain.
- `write` on a poly-carried handle (an `IO.pipe` element), `codegen_call.c`. Textually the same idiom you converted one line below.
- `$stdout.write` / `$stderr.write`, which bypass `sp_File_write` for a direct `fwrite`. After your change, `io = $stdout; io.write(bin)` wrote the bytes and `$stdout.write(bin)` did not.
- `sp_poly_shl`'s IO arm, so `io << binary` no longer depends on whether inference typed the operand. That one is defensive and untested: a poly-typed IO handle is hard to construct now that nil unifies with TY_IO.

I also added test coverage for the forms your test did not reach -- multi-argument `write`, `IO#<<`, `IO.copy_stream`, a non-String argument, and a payload whose FIRST byte is NUL, which wrote nothing at all rather than merely truncating.

Your marker analysis holds: I traced every way a TY_STRING value is emitted and a NUL-containing literal never takes the rodata path, so `sp_str_byte_len`'s `s[-1]` read is always in bounds. `IO#puts` / `#print` stay deferred as you scoped them.

Please do file the `Addrinfo#ipv4?` / `#ipv6?` `-O1` vs `-O0` divergence separately. A predicate that flips with optimization level is a real latent bug, not a test artifact, and I would rather chase it on its own.

matz:
Follow-up: I found the mechanism behind the `Addrinfo#ipv4?` / `#ipv6?` divergence you mentioned, and it is a real latent bug rather than a test artifact. Fixed in 7f844a12.

`sp_str_eq` confirms a `strcmp` hit by comparing byte lengths, because strcmp equality is only prefix equality once a length header records an embedded NUL. Taking a bare C literal's length means reading its `s[-1]` marker, which is out of bounds -- whatever byte happens to precede the literal in rodata. Land on 0xfe / 0xfc / 0xfd / 0xf1 there and `sp_str_byte_len` reads the bytes BEFORE the literal as an `sp_str_hdr`, takes a garbage length, and the comparison answers false for two equal strings. The rodata layout decides, so the answer changes with the compiler and the optimization level.

The predicates were emitted as `sp_str_eq((x)->afname, "AF_INET")` -- heap string on the left, bare literal on the right. It does not reproduce on my box (the preceding byte is 0x00 at every optimization level, so it falls through to strlen and works), which is precisely what made it latent, and why your report was worth chasing rather than filing.

I swept for the rest of the family by scanning the generated C of all 2469 tests for a bare literal reaching any of the fourteen runtime functions that read `s[-1]`. Six sites: the four Addrinfo predicates, two FiberKillSignal checks, and the rescue-arm name compares. All now plain `strcmp` -- every one compares a class, module, or address-family name, none of which carries a NUL, so the length re-check was never doing anything but harm. None remain.

So you can skip filing that one; thank you for noticing it in passing.

Issue #3227, opened 2026-07-22
[Runtime] In-place mutation of a `String` is not observed through an alias


KieranP:
> (Disclaimer: I asked Claude Opus 4.8 to find Spinel compile/runtime issues with documented Ruby classes and methods. It found the following issue, which I manually verified for accuracy before publishing this report)

_Reproduced with `spinel d298e1e06ec9` vs `ruby 4.0.3`._

When a String is bound to a second reference (`s2 = s1`, or stored into a container) and then mutated in place through the first reference (`s1 << …`, `upcase!`, `replace`, …), the second reference does not see the change and `s1.equal?(s2)` becomes `false`. The mutating reference updates correctly, but the alias keeps the old value — as if the in-place mutator produced a fresh string and rebound only the receiver variable. `docs/limitations.md` documents String literals as *mutable* ("Mutable string literals (`s = "x"; s << "y"`) … works"; "Spinel's literals stay mutable") and that aliasing (`b = a`) answers `true`, so aliased in-place mutation should be visible. Array and Hash aliasing are unaffected — the divergence is String-specific.

## Repro

```ruby
s1 = "hello"
s2 = s1
p s1.equal?(s2)   # ruby: true   spinel: false
s1 << " world"
p s1              # ruby: "hello world"   spinel: "hello world"
p s2              # ruby: "hello world"   spinel: "hello"
p s1.equal?(s2)   # ruby: true   spinel: false
```

Ruby: `true` / `"hello world"` / `"hello world"` / `true`
Spinel: `false` / `"hello world"` / `"hello"` / `false`

Storing into a container shows the same drop: `arr = [s1]; s1 << "!"; p arr` → Ruby `["hello!"]`, Spinel `["hello"]`. Passing the string as a method argument and mutating it there IS observed by the caller under both — only local-to-local aliasing and container-store diverge.

matz:
Phase 1 (local aliases) is in on master (12b757f0): `s2 = s1` now shares one mutable String object, so an in-place `s1 << ...` shows through `s2` and `s1.equal?(s2)` stays true. Scoped so only strings that are actually aliased-and-mutated become handles -- everything else keeps the value-typed fast path (optcarrot checksum unchanged). The container-store half (`arr = [s1]; s1 << ...`) is the next phase; leaving this open until it lands.

matz:
Phase 3 (container-held aliases) is now on master (f353883d).

[...]

PR #3393, opened 2026-07-27
Hand back a marked empty string, not a bare literal

rubys:
Every string spinel produces carries a marker byte at `[-1]`. `sp_sym_names[]` entries are emitted as `(&("\xff" "name")[1])`, and `sp_gc_mark` reads that byte to decide whether a pointer is a rodata literal to skip or a heap object whose mark word it should write.

Three functions returned a bare `""` instead, which has no such byte — and callers root the result.

## What goes wrong

A **nil Symbol compiles to id `-1`**, so it takes `sp_sym_to_s`'s out-of-range arm and gets the unmarked literal. The generated code then roots it. From a Rails layout where a nil CSS class reaches `html_escape(title_class.to_s)`:

```c
const char * _t5294 = sp_sym_to_s(lv_title_class); SP_GC_ROOT(_t5294);

const char * sp_ViewHelpers_s_html_escape(const char * lv_s) {
    SP_GC_SAVE();
    SP_GC_ROOT(lv_s);
    return sp_re_gsub_str_str_hash(sp_re_pat_3, lv_s, cst_HTML_ESCAPES);
```

Any collection while that root is live has `sp_gc_mark` read whatever rodata precedes the literal, fail to recognise a marker, conclude it is a heap object, and write its mark word — a protection fault on read-only memory:

```
EXC_BAD_ACCESS (code=2, address=0x1030846d7)
  sp_gc_mark + 176            ->  stur w8, [x0, #-0x10]
  sp_gc_mark_all + 408
  sp_gc_collect + 160
  sp_re_gsub_str_str_hash + 684
  sp_Layouts_s_application + 38684
```

`SPINEL_GC_VERIFY=1` names it before the fault, which is how it was found:

```
[phase=root ctx=0x16d0ed9f0]
  obj = 0x1030846d7
```

and `frame variable -L` in a `--debug` build shows the rooted slot holding a rodata pointer, with `lv_title_class = -1` one frame up:

```
0x16fdf9118: (const char *) lv_s = 0x1004e397d ""
0x16fdf9114: (int) _gc_saved    = 120
0x16fdf9110: (int) _sp_gcr_1161 = 1
```

The rooting is **not** the bug — spinel strings are `const char *` carrying a marker, so `SP_GC_ROOT` on one is by design. Only these returns are.

## The change

`sp_str_empty` (`&sp_str_empty_data[1]`) is the marked empty string.

- `src/codegen.c` — the emitted `sp_sym_to_s`: the out-of-range arm, and the arm used when a program has no static symbols.
- `lib/sp_runtime.h` — the no-symbol-table default, and the `pack` dispatcher's fallback (reachable and equally rootable). The bare `""` at the top of the format-string helper is after `sp_raise_cls` and unreachable, so it is left alone.

## Verification

The application that hit this died after ~20 requests by default, and on the very first under `SPINEL_GC_STRESS=1` — in a different place each run depending on where the collection landed, which is what made it look flaky.

| | before | after |
|---|---|---|
| render loop, plain | crash at iteration 59 | 200/200 |
| render loop, `SPINEL_GC_STRESS=1` | crash at iteration 0 | 200/200 |
| render loop, stress + `GC_VERIFY` | abort at iteration 0 | 200/200, verifier silent |
| 49-route HTTP sweep | dead at route 23 | all 49 complete |
| 200 HTTP requests | died on #21 | survived |

`make test`: **2512 pass, 0 fail, 0 error**.

No standalone reproducer, and the reason is worth stating: a small program's symbols are all in range, so they get marked table entries. Reaching the fallback needs an out-of-range symbol whose `to_s` result is rooted across a collection, which depends on whole-program inference typing the slot `sp_sym`. Several attempts at a synthetic case all compiled to something that never calls the fallback.

### DIFF
--- a/lib/sp_runtime.h
+++ b/lib/sp_runtime.h
@@ -7389,7 +7389,10 @@ static inline const char *sp_poly_pack(sp_RbVal recv, const char *fmt) {
     return sp_PolyArray_pack((sp_PolyArray *)recv.v.p, fmt);
   if (recv.tag == SP_TAG_OBJ && recv.cls_id == SP_BUILTIN_STR_ARRAY)
     return sp_StrArray_pack((sp_StrArray *)recv.v.p, fmt);
-  return "";
+  /* Marked, like every other string this can return: a bare "" literal
+     has no 0xff marker byte, and a caller that roots the result would
+     have the collector mark a rodata pointer. */
+  return sp_str_empty;
 }
 
 /* StringScanner is a native-bound spin package (packages/strscan). */
@@ -7405,7 +7408,7 @@ static inline const char *sp_poly_pack(sp_RbVal recv, const char *fmt) {
    referenced-but-undefined static even when the referring function is dead --
    so supply inert fallbacks that the optimizer prunes with everything else. */
 #ifdef SP_TU_NO_POLY_RENDER
-static const char *sp_sym_to_s(sp_sym id) { (void)id; return ""; }
+static const char *sp_sym_to_s(sp_sym id) { (void)id; return sp_str_empty; }
 static const char *sp_class_to_s(sp_Class c) { return c.name ? c.name : ""; }
 static sp_sym sp_sym_intern(const char *s) { (void)s; return (sp_sym)0; }
 #endif
--- a/src/codegen.c
+++ b/src/codegen.c
@@ -5430,10 +5430,19 @@ char *codegen_program(const NodeTable *nt) {
     /* dynamic intern pool: symbols minted at runtime (Symbol#upcase,
        :"interp", String#to_sym) get ids >= the static count. */
     buf_puts(&b, "static const char *sp_dyn_syms[SP_DYN_SYMS_MAX]; static int sp_ndyn = 0;\n");
+    /* Every arm must hand back a MARKED string. sp_sym_names[] entries
+       carry the 0xff rodata marker, but a bare "" literal does not, and
+       callers root the result (`const char *t = sp_sym_to_s(x);
+       SP_GC_ROOT(t);`). sp_gc_mark then reads the arbitrary rodata byte
+       before the literal, fails to recognise a marker, treats it as a
+       heap object and writes its mark word. A nil Symbol lands on the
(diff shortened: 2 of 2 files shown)


matz:
Merged as 29db74a3. The diagnosis and the fix are both right, and the note about why there is no standalone reproducer is the useful part -- it is exactly why the sweep below was worth doing from the invariant rather than from repros.

Following it, two more commits.

**971efd30** -- the same shape in about thirty other places, one of which is reachable from `p obj.class`. A class VALUE carries its name as a plain `const char *` and `sp_class_to_s` hands that pointer straight back, so every `(sp_Class){id, "Name"}` the generator emitted was building one out of a bare literal:

```ruby
require "stringio"
c = StringIO.new("x").class.to_s
GC.start
p c                                  # SIGSEGV
```

Two of those sites already used `SPL`, which is what the other twenty-five should have been. The rest are the runtime renderers' fallback arms, where a missing or NULL object renders a fixed word: `sp_sym_inspect`, `sp_poly_class_name`'s STRBUF case, `sp_io_kind_name`'s whole table, `sp_exc_class_name`, and the `"nil"` of Proc / Enumerator / MatchData / PtrArray `#inspect`.

Reading the runtime alone did not find the class-name family. What did was a test that holds each value in a live local across an explicit `GC.start`, which is the window that faults -- the deterministic form of your `SPINEL_GC_STRESS` observation.

**388269b1** -- that test then found something older and unrelated to markers. Eleven `#inspect` helpers build their text in a temporary `sp_String` and return that object's buffer. The 0xfd marker means "the owner holds this, skip it", so after the helper returns nothing keeps the owner alive:

```ruby
s = "abc".match(/b/).inspect
GC.start
p s                                  # SIGSEGV
```

They now return a copy on the string heap. The #3227 paths that deliberately hand out a live buffer are untouched -- there the owner is the Ruby object and the aliasing is the point.

One thing worth flagging separately: `lib/strscan.c` is dead (the live one is `packages/strscan/sp_strscan.c`) and its string returns are raw `malloc` buffers. It is not linked, so it is not a bug today, only a trap for the next person running this sweep.

Issue #3397, opened 2026-07-27
A method that uses its block's result as a Hash key is miscompiled unless the receiver is an inline `.new`


rubys:
The same method body compiles three different ways depending on how the receiver
reaches the call. Only the inline-receiver form is correct; the other two are a
segfault and a `NoMethodError` for a method that is plainly defined.

Found while compiling a Rails app to spinel: `ActiveRecord::Relation#group_by`
(`k = yield rec` used as a Hash key) produced
`undefined method 'group_by' for an instance of ActiveRecord::Relation` at
runtime, and `sp_Relation_group_by` was absent from the generated C entirely —
along with `each`, `map` and `inject`, the other block-taking methods on that
class. Its non-block methods (`group`, `group_count`, `to_a`, …) were all
emitted.

spinel rev: `77f5094fc424`
Platform: macOS arm64

### A — correct (inline receiver)

```ruby
class Rel
  def go
    out = {}
    k = yield 5
    out[k] = 5
    out
  end
end
p Rel.new.go { |v| v % 2 }
```

`{1 => 5}` — matches CRuby.

### B — segfault (receiver bound to a local)

Identical to A except the receiver goes through a local:

```ruby
class Rel
  def go
    out = {}
    k = yield 5
    out[k] = 5
    out
  end
end
x = Rel.new
p x.go { |v| v % 2 }
```

Exit 139, no output. CRuby prints `{1 => 5}`.

### C — NoMethodError (genuinely poly receiver)

```ruby
class A
  def go
    out = {}
    k = yield 5
    out[k] = 5
    out
  end
end

class B
  def go
    { 0 => 0 }
  end
end

def pick(n)
  n == 1 ? A.new : B.new
end

x = pick(1)
p x.go { |v| v % 2 }
```

`undefined method 'go' for an instance of A (NoMethodError)`. CRuby prints
`{1 => 5}`. This is the shape the Rails app hit.

### What the generated C shows (A vs B)

`spinel -c` on the two, diffed. A infers the hash precisely:

```c
sp_IntIntHash * lv__y1_out = NULL;
mrb_int lv__y1_k = 0;
lv__y1_out = sp_IntIntHash_new();
lv__y1_k = ({ lv_v = 5LL; ... });
sp_IntIntHash_set(lv__y1_out, lv__y1_k, 5LL);
```

B infers the same literal `{}` as a **String-keyed** hash, then writes a boxed
integer into that String key slot:

```c
[...]

PR #3398, opened 2026-07-27
An empty-hash local keyed by a block result is not String-keyed

rubys:
Fixes the segfault half of #3397.

`x = Rel.new; x.go { ... }` segfaults with no output where `Rel.new.go { ... }` — the same method, same body — is correct:

```ruby
class Rel
  def go
    out = {}
    k = yield 5
    out[k] = 5
    out
  end
end
x = Rel.new
p x.go { |v| v % 2 }      # exit 139; CRuby prints {1 => 5}
```

### Why

The inline receiver infers `out` as `Hash[Integer, Integer]` and `k` as an `mrb_int`. Through a local it stays open through the fixpoint and lands on the post-fixpoint backstop in `analyze_program`, which defaulted it to StrPolyHash. Codegen then reads the key out of the boxed value as `.v.s`:

```c
lv__y2_k = sp_box_int(({ lv_v = 5LL; ... }));
sp_StrPolyHash_set(lv__y2_out, (lv__y2_k).v.s, sp_box_int(5LL));
```

`sp_box_int` writes the union's integer member and `.v.s` reads the `const char *` member back out, so the key `1` is dereferenced as a pointer.

`mark_empty_hash_key_ctx` already refuses exactly this — *"a boxed key ... is only representable by the poly-keyed variant; the StrPolyHash default would hand the boxed value to a const char \* slot and segfault"* — but it runs **during** the fixpoint, and a key that is a block's return value has no type yet when it looks. I confirmed that with a probe: at that point the key node reads `TY_UNKNOWN`, and it only widens to poly later.

The backstop runs after, where the key type IS settled, so it can check its default rather than assume it.

### What changed

- `local_aset_key_type` — the key-side twin of `aset_value_type_ex`, keyed by `(scope, name)` rather than a receiver node, because the backstop holds only the local's WRITE node.
- The backstop consults it and keeps StrPolyHash only for a String key, or when there are no `[]=` writes to say otherwise.

Deliberately narrow: it moves only locals the backstop was already defaulting. A key-write that narrowed the type during the fixpoint still wins, and a String key still gets the String-keyed variant, so no hash that was already precise changes representation.

### Not covered

`out[k] ||= 5` still crashes. The aset write index is built from plain `[]=` CallNodes, so the IndexOrWrite forms are invisible to this. I left it rather than widen the index on a guess about its other consumers — happy to follow up if you'd like it in scope.

### Testing

`make test`: **2517 pass, 0 fail, 0 error**. The new test (`test/empty_hash_block_result_key.rb`) segfaults on master, and covers the crash, the `group_by`-shaped fetch/insert body, an explicit `&blk` receiver, a String key that must keep the String-keyed variant, and the inline-receiver form that must keep its precise `Hash[Integer, Integer]`.

Expectations were generated from CRuby.

### Note on #3397

That issue carries a **second, unrelated defect** which this does not touch: a block-taking method is unreachable through a poly receiver, with no hash involved —

```ruby
class A; def go; yield 5; end; end
class B; def go; 0; end; end
def pick(n) = n == 1 ? A.new : B.new
p pick(1).go { |v| v % 2 }   # undefined method 'go' for an instance of A
```

I am filing that separately so #3397's framing doesn't mislead.


matz:
Merged as fd5048c9, rebased onto tip.

The diagnosis is the useful part: the key-context pass and the backstop ask the same question at two different times, and the fixpoint is exactly when a block-result key has no answer yet. Checking the default against a key type that HAS settled, rather than assuming it, is the right shape.

I took the `||=` gap you flagged rather than leaving it — 8b78c630. `h[k] ||= v` keys the slot exactly as `h[k] = v` does, so a variant chosen without looking at it is chosen on incomplete evidence, and the write index now carries IndexOrWrite / IndexAndWrite / IndexOperatorWrite.

Your instinct to leave the shared index alone was the right caution, and it turns out to be safe here for a specific reason worth recording: those nodes carry the key in `arguments[0]` and the value in a `value` ref rather than `arguments[1]`, so `aset_value_type_ex`'s existing `an < 2` guard already skips them and its answer does not move. Only the key side sees them.

Your test grew three rows for it.

Thanks also for splitting the poly-receiver half out into #3399 instead of leaving #3397 to carry two unrelated defects.

matz:
Closing: merged as fd5048c9 (a merge commit on master, so GitHub did not pick it up automatically). The rebase onto tip is what is on master; the follow-up for the `||=` gap is 8b78c630.

PR #3413, opened 2026-07-28
Give an --rbs nilable scalar somewhere to put nil

### LINKED ISSUE #3412: An --rbs-pinned Integer?/Float?/bool? attribute stores nil as 0/0.0/false (String? is correct)
An attribute declared `Integer?` in a `--rbs` seed stores `nil` as **0**. `Float?` stores `0.0`, `bool?` stores `false`. `String?` is correct. Without `--rbs` all four are correct, so the seed is what changes the behaviour.

It is silent — no warning, no error, just a value that is no longer nil.

## Repro

`m.rb`:

```ruby
class Row
  def initialize
    @i = nil
    @f = nil
    @s = nil
    @b = nil
  end
  def i; @i; end
  def i=(v); @i = v; end
  def f; @f; end
  def f=(v); @f = v; end
  def s; @s; end
  def s=(v); @s = v; end
  def b; @b; end
  def b=(v); @b = v; end
end
r = Row.new
r.i = nil; r.f = nil; r.s = nil; r.b = nil
puts "Integer? nil -> " + r.i.to_s + " nil?=" + r.i.nil?.to_s
puts "Float?   nil -> " + r.f.to_s + " nil?=" + r.f.nil?.to_s
puts "String?  nil -> " + r.s.to_s + " nil?=" + r.s.nil?.to_s
puts "bool?    nil -> " + r.b.to_s + " nil?=" + r.b.nil?.to_s
```

`m.rbs`:

```rbs
class Row
  attr_reader i: Integer?
  attr_writer i: Integer?
  attr_reader f: Float?
  attr_writer f: Float?
  attr_reader s: String?
  attr_writer s: String?
  attr_reader b: bool?
  attr_writer b: bool?
end
```

```
$ spinel m.rb -o m_rbs --rbs .   &&  ./m_rbs
Integer? nil -> 0 nil?=false
Float?   nil -> 0.0 nil?=false
String?  nil ->  nil?=true
bool?    nil -> false nil?=false

$ ruby m.rb
Integer? nil ->  nil?=true
Float?   nil ->  nil?=true
String?  nil ->  nil?=true
bool?    nil ->  nil?=true
```

Drop the `--rbs .` and spinel matches CRuby on all four.

The split falls exactly on unboxed vs reference: the three that lose nil are the ones with a machine zero to collapse into, and each lands on precisely that zero. `String?` — the one with a real null pointer available — is fine. That reads like the seed pinning the ivar to an unboxed slot with no nil representation, rather than to a nilable one.

## Where it turned up

Roundhouse emits a `Row` class per model for hydration, with exactly this shape (ivar seeded nil, assigned from a query row), and RBS declaring the nullable columns `Integer?` — the transpiled lobsters app is built with `--rbs .`.

The result is that every SQL NULL in a nullable integer column arrives as 0. In lobsters' user tree the two root users have `invited_by_user_id IS NULL`; both hydrate as 0, so `users.group_by { |u| u.invited_by_user_id }[nil]` finds nothing, the tree walk never starts, and `/u` renders an empty `<ul>` — **3,017 bytes against CRuby's 292,449**, with a 200 status and no error anywhere.

I chased this through the whole hydration path before suspecting the seed, since every layer looked right in isolation: the SQLite read returns nil correctly, and the nil survives into the row object's setter. It is only reading the attribute back that produces 0.

## Environment

- spinel `0ccc3ceb`
- macOS 15 / Darwin 25.5.0 arm64, Apple clang 21.0.0

rubys:
Fixes #3412.

## What the seed actually changed

The pin is what narrows the slot. Without `--rbs`, `@i` sees a nil write and widens to poly, which is why dropping the seed matched CRuby on all four types. An `Integer?` seed pins it to `mrb_int` instead — and a narrow slot needs a coercion the wide one never did.

A poly RHS assigned into a typed ivar unboxed through `emit_unbox_text`, which reads the union member for the target kind — `.v.i` — with no regard for the tag above it. A boxed nil carries zero in that member. So `@i = v` with `v` nil stored 0, and from that point on it *was* the Integer 0: well-formed C, nothing downstream able to tell. `String?` escaped only because the zero it collapses onto **is** a null pointer, which the string paths already read as nil.

## The split, and why it isn't uniform

`int` and `float` each have a reserved sentinel — `SP_INT_NIL`, a NaN payload — that every `nil?` / `to_s` / boxing site on a nullable slot already tests for. The nil-*literal* arm three branches up in the same function writes exactly those, which is why `@i = nil` in `initialize` was already correct and only the assignment through the setter was not. The store was the wrong half, not the pin: routing the coercion through new `sp_poly_as_int_or_nil` / `sp_poly_as_float_or_nil` makes the sentinel survive the round trip.

`bool?` and `Symbol?` have no such sentinel to reach for. `mrb_bool` is a C `bool` — 0 is `false`, and there is no third value — and symbol 0 is an ordinary symbol, so a pinned slot has nowhere to put nil at all. Both now pin to the tagged union, which is what `bool | nil` means anyway.

`Symbol?` turned out worse than the three in the report. Its nil-init *does* write the `(sp_sym)-1` sentinel, but `nil?` on a symbol receiver is in codegen's "value-typed, never nil" fold list, so it constant-folds to false and the read never consults the sentinel — a nil symbol came back as whichever symbol id happened to sit there (`:initialize`, in the fixture below).

| Token | Pinned to | nil is |
|---|---|---|
| `int?` | `mrb_int` | `SP_INT_NIL` |
| `float?` | `mrb_float` | a NaN payload |
| `string?`, `obj_X?`, `*_array?`, `*_hash?` | the pointer | `NULL` |
| `bool?`, `symbol?` | `sp_RbVal` | `SP_TAG_NIL` |

## Verification

The reported repro, after:

```
$ spinel m.rb -o m_rbs --rbs . && ./m_rbs
Integer? nil ->  nil?=true
Float?   nil ->  nil?=true
String?  nil ->  nil?=true
bool?    nil ->  nil?=true
```

Byte-identical to `ruby m.rb`. A wider fixture covering `Symbol?`, `Box?`, `Array[Integer]?` and `Hash[String, Integer]?` also matches CRuby on all eight.

New regression test `test/rbs-seed/nilable_scalar_ivar.rb`, wired into `make rbs-seed-test`. Against the pre-fix compiler it prints

```
-nil nil nil nil nil          +nil nil nil false nil
-nil nil nil nil nil          +0 0.0 nil false :initialize
-2                            +0        # group_by[nil].length
-1                            +3        # group_by[0].length
```

— the last two lines being the `/u` symptom in miniature. It also greps the emitted C for `sp_RbVal iv_f` / `sp_RbVal iv_y`, so a future change that re-pins `bool?` or `Symbol?` to a flat slot fails loudly rather than silently.

`make test` 2528 pass / 0 fail / 0 error, `rbs-test` 13 pass, `rbs-seed-test` pass, `make optcarrot` OK (341 fps, checksum 59662), `make bench` 59 pass / 0 fail.

## Known adjacent gap, not fixed here

A nilable **parameter** still loses nil, through a different mechanism:

```rbs
def take: (Integer? n) -> Integer?
```
```ruby
def take(n); n; end
[1, nil, 3].each { |v| puts take(v).inspect }   # spinel: 1, 0, 3
```

The call site coerces a poly argument with `sp_poly_to_i`, which answers 0 on nil. This PR doesn't touch it: `parse_seed_type` discards the `?` before anything records it, so codegen can't distinguish `Integer` from `Integer?` at the binding, and the plain sentinel unbox can't just be substituted — that path also does the Float/String→int conversions `sp_poly_to_i` exists for. It needs nilability tracked on the seeded param plus a nil-preserving conversion helper.

The Roundhouse shape doesn't reach it (`attr_accessor` emits only `ivar` seed lines, no `meth` lines), and the existing `nilable_arg_group_by` fixture passes because its argument is a literal `nil`, which has its own arm. Worth its own issue if you'd like one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -690,6 +690,14 @@ rbs-seed-test: $(SPINEL) $(RBS_EXTRACT_BIN) $(SP_RT_LIB)
 	  "$$tmp/yu" > "$$tmp/yu.out" 2>/dev/null; \
 	  cmp -s "$$tmp/yu.out" test/rbs-seed/yield_union_hash_obj.expected || { echo "rbs-seed-test: FAIL (#3278 yield-union output mismatch)"; diff -u test/rbs-seed/yield_union_hash_obj.expected "$$tmp/yu.out" || true; ok=0; }; \
 	else echo "rbs-seed-test: FAIL (#3278 yield-union: C did not compile)"; ok=0; fi; \
+	$(SPINEL) test/rbs-seed/nilable_scalar_ivar.rb --rbs test/rbs-seed/sig \
+	  -c --no-line-map -o "$$tmp/ns.c" 2>/dev/null; \
+	grep -Eq 'sp_RbVal[[:space:]]+iv_f;' "$$tmp/ns.c" || { echo "rbs-seed-test: FAIL (#3412: bool? pinned a slot with no nil)"; ok=0; }; \
+	grep -Eq 'sp_RbVal[[:space:]]+iv_y;' "$$tmp/ns.c" || { echo "rbs-seed-test: FAIL (#3412: Symbol? pinned a slot with no nil)"; ok=0; }; \
+	if $(CC) -O0 -Ilib "$$tmp/ns.c" $(SP_RT_LIB) $(LDFLAGS) -lm -o "$$tmp/ns" 2>"$$tmp/ns.err"; then \
+	  "$$tmp/ns" > "$$tmp/ns.out" 2>/dev/null; \
+	  cmp -s "$$tmp/ns.out" test/rbs-seed/nilable_scalar_ivar.expected || { echo "rbs-seed-test: FAIL (#3412 nilable-scalar ivar output mismatch)"; diff -u test/rbs-seed/nilable_scalar_ivar.expected "$$tmp/ns.out" || true; ok=0; }; \
+	else echo "rbs-seed-test: FAIL (#3412 nilable-scalar ivar: C did not compile)"; ok=0; fi; \
 	rm -rf "$$tmp"; \
 	if [ $$ok -eq 1 ]; then echo "rbs-seed-test: pass"; else exit 1; fi
 endif
--- a/docs/rbs-extract.md
+++ b/docs/rbs-extract.md
@@ -120,6 +120,26 @@ ivar <name> <type>
 
 Lines whose first token isn't a keyword are treated as comments.
 
+### How the analyzer reads a `<T>?` token
+
+`parse_seed_type` pins a nilable token to the base type only where that
+type's C slot still has an inhabitant left to spell nil with:
+
+| Token                         | Pinned to        | nil is        |
+|-------------------------------|------------------|---------------|
+| `int?`                        | `mrb_int`        | `SP_INT_NIL`  |
+| `float?`                      | `mrb_float`      | a NaN payload |
+| `string?`, `obj_X?`, `*_array?`, `*_hash?` | the pointer | `NULL`   |
+| `bool?`, `symbol?`            | `sp_RbVal`       | `SP_TAG_NIL`  |
+
+`mrb_bool` and `sp_sym` have no spare inhabitant — `0` is `false`, and
+symbol `0` is a real symbol — so `bool?` and `Symbol?` pin to the boxed
+tagged union rather than collapsing nil onto `false` / `:""` (#3412).
+A poly value narrowed into one of the sentinel-carrying slots goes
+through `sp_poly_as_int_or_nil` / `sp_poly_as_float_or_nil`, which map
+the nil tag to the sentinel instead of reading the zero payload beneath
+it.
+
 ## Follow-up
 
 Tracked as out-of-scope for the initial spike (see matz/spinel#7 +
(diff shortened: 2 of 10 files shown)


matz:
Merged. Gates re-run here on the rebase: test 2530, bench 59, optcarrot checksum 59662 at 1328 fps, clang, spin-check.

The diagnosis is exactly right, and the part worth saying out loud is the second half of it: the pin was not the bug. Dropping the seed matching CRuby on all four types reads like the seed causing the problem, and the tempting fix is to stop narrowing. But the narrow slot is the whole point of a seed, and the nil-*literal* arm three branches up in the same function was already writing the sentinels correctly. Only the store through the setter was not. Fixing the store keeps the slot.

The `bool?` / `Symbol?` split is the right call rather than a concession. `mrb_bool` genuinely has two inhabitants and `bool | nil` needs three; there is nothing to reserve. Pinning those to the tagged union is not giving up on them, it is what the type says.

`Symbol?` being worse than reported is the useful find. A sentinel that exists but is never consulted is a harder failure than no sentinel: `nil?` folding to false on a value-typed receiver is correct reasoning *given* that symbols are never nil, so the fold and the sentinel each looked right on their own. Removing the nilable-symbol slot restores that premise instead of special-casing the fold, which is the end I would have picked.

The two grep assertions in `rbs-seed-test` are the part I would have left out and been sorry about later. A future change that re-pins those to a flat slot would otherwise pass the output comparison right up until something wrote a nil.

## The parameter gap

Yes, worth its own issue, and I would rather have it than not. `parse_seed_type` discarding the `?` before anything records it is the real obstacle, so a nilability bit on the seeded binding has to come first, and `sp_poly_to_i` cannot simply be swapped for the sentinel unbox given the conversions it also performs. That is a design step rather than a patch, and it should be tracked on its own rather than ride along here.

File it and I will take it.

PR #3456, opened 2026-07-30
Fold the two strbuf mutation probes into one CallNode sweep instead of a whole-table rescan per candidate (perf: lobsters 226s -> 152s)

rubys:
`strbuf_mut_kind` and `strbuf_ivar_mut_kind` both answer "over EVERY call whose receiver names this local (or this ivar), was it mutated in place?" — each by scanning **the entire node table**. `promote_shared_stored_strings` (#3227) asks them per candidate from inside its own whole-table walks, so the pass costs O(candidates × nodes).

On roundhouse's lobsters emit (41k lines of generated Ruby + 5k of RBS — the [#3115](https://github.com/matz/spinel/issues/3115) tree one scale up) a `sample` profile of the ~204s front end reads:

| frame | share of front end |
|---|---|
| `analyze_program` | 99.9% |
| ` ├─ promote_shared_stored_strings` | 38.4% |
| ` │   ├─ strbuf_mut_kind` | 21.6% |
| ` │   └─ strbuf_ivar_mut_kind` | 13.5% |

Codegen proper is ~0.1% — it is all analysis.

Both answers are pure functions of call-site syntax and scope homing, so this computes them for **every key at once** in a single sweep over the `NK_CallNode` kind list, into a `(name, key)` table the probes then read in O(1). Same move as #3293 / #3294: build the grouping once, delete the per-item rescan.

The caching follows `lw_shared_first`'s contract — live while the scope index is frozen, stamped with `comp_scope_index_gen()`, rebuilt when the node table grows. While unfrozen it rebuilds per query, which costs one kind-list walk: strictly less than the whole-table scan it replaces, so no phase regresses.

**Two extra stamp inputs, both load-bearing:**

- `comp_class_index(c, "Toplevel")` — a top-level ivar keys on it, and it becomes answerable *independently* of the scope-index epoch. Resolve it once at build time and the record lands under the wrong class id, silently un-promoting the slot. `test/issue_3227_ivar_shared.rb` (the P4 top-level `@log` block) is exactly that shape and is what caught it.
- `nt->version` — a rename pass rewrites a call's `name` in place via `nt_set_str` without growing the table, and this index is keyed by name. (`lw_shared`'s count-only stamp has that exposure; this one doesn't. It costs nothing measurable: the version does not churn during the frozen fixpoint.)

## Measurements

`spin build` of the same tree, `spin clean` before each run, lanes alternated:

| lane | run 1 | run 2 |
|---|---|---|
| master `c728bfad` | 223.07s | 228.54s |
| this branch | **152.11s** | **151.15s** |

The `cc` stage is 21.3s of each total (identical 13.2 MB translation unit in both lanes), so the front end itself goes **204.5s → 130.3s, −36%**.

## Verification

- `make test`: **2564 pass, 0 fail, 0 error** — same count as master.
- **The generated C for the whole lobsters app is byte-identical** to master's, modulo the compiler's own install path appearing in one `#line` directive.
- **Differential sweep.** With the original scans kept alongside the new tables and *both* consulted on every query, compiling all 2536 `test/*.rb` gives **zero divergence**. That harness is how the top-level ivar case above was found; it is not part of this diff.

## Not in this PR

Same profile, same shape, if you want them as follow-ups:

- `strbuf_slot_eligible_shape` — 16.1%, one whole-table CallNode scan per `(vn, vs)` query. Needs an index rather than a memo: its per-call predicates (`an_unique_scope_by_name`, `comp_byref_param`) read analysis state, so the answer can move within an epoch.
- `promote_append_accumulators` (`bdf60528`) — 8.1%, same per-candidate rescan.
- Three nested whole-table walks inside `promote_shared_stored_strings` itself (`analyze.c:7231`, `:7277`, `:7295`) and one in `convert_byref_handle_params`. The `:7231` one is a `(scope, name)` local-write query — what `lw_shared_first` already answers.

### DIFF
--- a/src/analyze.c
+++ b/src/analyze.c
@@ -6514,56 +6514,153 @@ static int ctor_writes_ivar(Compiler *c, int ci, const char *ivname) {
    it TY_STRBUF (the container then unifies to a poly variant and emit_boxed
    wraps the sp_String* as SP_BUILTIN_STRBUF). Unmarked reads of the same
    local keep demoting to TY_STRING, so ordinary consumers are unchanged. */
+/* Both mutation probes below ask "over EVERY call whose receiver names this
+   local (or this ivar), was it mutated in place?" -- and the promotion asks
+   them per candidate, from inside whole-table walks, so a probe that rescans
+   the node table makes the pass O(candidates x nodes). On roundhouse's
+   lobsters emit (41k lines of generated Ruby, the #3115 tree one scale up)
+   the two together are ~35% of compile time in a `sample` profile.
+
+   The answer is a pure function of call-site syntax and scope homing, so one
+   sweep over the CallNode kind list computes it for EVERY key at once into a
+   (name, key) table the probes then read in O(1). Cached while the scope index
+   is frozen and stamped with comp_scope_index_gen() -- the same invalidation
+   the LWIndex behind local_all_writes_empty_hash rides; while unfrozen it
+   rebuilds per query, which costs one kind-list walk, strictly less than the
+   whole-table scan it replaces. */
+typedef struct {
+  const char **name;  /* receiver name, borrowed from the node table */
+  int *key;           /* scope index (locals) / class id (ivars) */
+  signed char *val;   /* 1 = mutated in place, -1 = disqualified */
+  int *next;          /* next record in the same bucket, or -1 */
+  int *head;          /* hash buckets: head record index, or -1 */
+  int cap;            /* bucket count (power of two) */
+  int n, ncap;        /* records used / allocated */
+} SbMutTab;
+
+static unsigned sb_mut_hash(const char *name, int key) {
+  unsigned h = 2166136261u ^ (unsigned)key;
+  for (const char *p = name; p && *p; p++) { h ^= (unsigned char)*p; h *= 16777619u; }
+  return h;
+}
+
+/* `n` bounds the record count: the sweep adds at most one record per call
+   node, so a table sized to the CallNode population never fills. */
+static void sb_mut_tab_init(SbMutTab *t, int n) {
+  int cap = 16;
+  while (cap < n * 2) cap <<= 1;
+  t->cap = cap; t->n = 0; t->ncap = n > 0 ? n : 1;
+  t->name = (const char **)malloc(sizeof(const char *) * t->ncap);
+  t->key = (int *)malloc(sizeof(int) * t->ncap);
+  t->val = (signed char *)malloc(sizeof(signed char) * t->ncap);
+  t->next = (int *)malloc(sizeof(int) * t->ncap);
+  t->head = (int *)malloc(sizeof(int) * cap);
+  for (int i = 0; i < cap; i++) t->head[i] = -1;
+}
+
+static void sb_mut_tab_free(SbMutTab *t) {
+  free((void *)t->name); free(t->key); free(t->val); free(t->next); free(t->head);
+}
+
+/* Status cell for (name, key); appends a fresh 0 cell when `add` and absent. */
+static signed char *sb_mut_tab_slot(SbMutTab *t, const char *name, int key, int add) {
+  unsigned h = sb_mut_hash(name, key) & (unsigned)(t->cap - 1);
+  for (int r = t->head[h]; r >= 0; r = t->next[r])
+    if (t->key[r] == key && sp_streq(t->name[r], name)) return &t->val[r];
+  if (!add || t->n >= t->ncap) return NULL;
+  int r = t->n++;
+  t->name[r] = name; t->key[r] = key; t->val[r] = 0;
+  t->next[r] = t->head[h]; t->head[h] = r;
+  return &t->val[r];
+}
+
+/* -1 wins over 1 whichever order the calls appear, matching the scans' early
+   `return -1`; 1 only lifts a cell still at 0. */
+static void sb_mut_tab_note(SbMutTab *t, const char *name, int key, signed char v) {
+  signed char *slot = sb_mut_tab_slot(t, name, key, 1);
+  if (slot && (v < 0 || *slot == 0)) *slot = v;
+}
+
+static void sb_mut_tabs_build(Compiler *c, SbMutTab *lt, SbMutTab *it, int toplevel) {
+  const NodeTable *nt = c->nt;
+  int ncalls = 0;
+  nt_nodes_of_kind(nt, NK_CallNode, &ncalls);
+  sb_mut_tab_init(lt, ncalls);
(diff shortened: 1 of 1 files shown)


rubys:
Correcting my own baseline before anyone acts on the headline number.

While looking for the next site I logged the fixpoint iteration count, and on this tree **the inference loop no longer converges** — it runs to the 128 cap on both `pf_round`s. That is the #3134 failure mode again. Bisected on a fixed input, front end only (`spinel --rbs . main.rb -c`):

| commit | landed | front end | fixpoint |
|---|---|---|---|
| `8dd45c27` | Jul 25 | 8.4s | converges at iteration 6 |
| `baaeba58` | Jul 30 14:36 | 13.1s | — |
| `bdf60528` | Jul 30 16:45 | **223.9s** | hits the 128 cap |

`bdf60528` ("Build a string in a loop without copying it every time") introduces `promote_append_accumulators`, and instrumenting which passes still report change at iterations 124–127 names exactly two: **`promote_append_accumulators` and `infer_write_types`**, on every late iteration. That looks like the ping-pong shape #3134 fixed for `void`/override-unification — one pass promoting a local to the append handle, the other widening it back, neither able to win.

So: the 226s → 152s in this PR is measured against a **regressed** baseline. With convergence restored the whole compile should return to ~13s, and this PR's saving applies to a much smaller number. I'd rather say that plainly than have the 226s read as normal.

The PR still stands on its own terms — the two probes really are O(candidates × nodes), that is real at Mastodon scale whatever the iteration count, and the change is behaviour-preserving (byte-identical C, 2564/0/0). But **fix `bdf60528` first**; it is worth ~15× and this is worth 1.35×.

Happy to file the convergence regression separately if you'd prefer it tracked apart from this PR.

rubys:
Closing the loop on the baseline caveat above, now that #3457 exists.

With #3457 applied (the fixpoint converges again), I stacked this PR on top and re-measured the same tree. Full `spin build`, clean each run:

| | run 1 | run 2 |
|---|---|---|
| master | 223.07s | 228.54s |
| #3457 alone | 34.74s | 35.34s |
| #3457 + this PR | **30.68s** | **31.26s** |

`cc` is ~21.3s of each total, so the front end goes 13.7s → 9.7s: **this PR is still worth ~29% of analysis time** once the iteration count is sane. Smaller than the 226→152 headline, exactly as flagged — but it is real, and it is the kind of term that grows with tree size, so it should matter more on Mastodon than it does here.

No rebase needed; the two touch different files (`analyze.c` vs `analyze_pass.c`) and cherry-pick cleanly in either order.

matz:
Merged, after #3457 as you suggested.

Verified locally:

- **Generated C is byte-identical across all 2538 `test/*.rb`** — compiled every one on master and on this branch and diffed the output trees: zero differing files. Also byte-identical for the optcarrot single-file input.
- Front end on that input: 0.23s -> 0.20s. Smaller than your 36% because the tree is 4.3k lines rather than 41k, and the pass is superlinear in exactly the way you describe — which is itself consistent with the diagnosis.
- `make test` 2566, bench 59, optcarrot checksum 59662 at 1522-1532 fps, clang build + test green.

The two stamp inputs you called out are the load-bearing part; the Toplevel-class-id one in particular is not obvious from the code, and the differential harness finding it is the reason I trusted the rest. Thank you for spelling both out in the description.

The follow-ups you listed (`strbuf_slot_eligible_shape` at 16.1%, `promote_append_accumulators` at 8.1%, and the nested walks inside `promote_shared_stored_strings`) look worth having — the last one especially, since `lw_shared_first` already answers that query shape.

rubys:
I went to do the `promote_shared_stored_strings` follow-up you singled out and it does not pay. Correcting my own "Not in this PR" list, since you endorsed it on the strength of that list.

**The walk I flagged is dead code — and it is yours.** The one I described as "a `(scope, name)` local-write query — what `lw_shared_first` already answers" (`analyze.c:7231` at `c728bfad`, now `:7328`) sits inside `if (0) { ... }`. You disabled it in 7ac08b29 (P6, Jul 24) when you replaced the literal-store witness with the settled element-type gate:

```c
    if (contv4->type != TY_STR_ARRAY) continue;
    if (0) {
      /* only proceed when every literal store into the container is a
         string (a conservative element-type witness) */
```

It never executes, so there is nothing to index.

**The two live nested walks are not hot.** Both are the `NK_ReturnNode`-by-scope scan inside the deep-return-alias loop (`:7374` and `:7392`). Counters on the lobsters emit — 142,687 nodes, full 15-pass analysis:

| counter | value |
|---|---|
| candidates reaching the scan | 765 |
| `:7374` inner iterations | 4,133,857 |
| `:7392` inner iterations | 142,415 |

4.28M total. For scale, the `mark_empty_hash_key_ctx` rescan in #3460 is **1.19 billion** on the same tree — this is **~0.36%** of that, single-digit milliseconds.

The guards are why. `:7374`'s `shared_ok &&` condition early-exits after ~5,400 iterations on average rather than running the full 142k, and `an_unique_scope_by_name` gates the outer loop down to 765 candidates in the first place. The `:7392` branch (caller-mutates-the-result) is entered about once per compile. On the optcarrot single-file input both counters are **0** — never reached.

The eight *outer* whole-table walks in the function come to ~17M iterations across the run, so the function as a whole is not worth restructuring either.

So the three walks were on my list by **shape** — they look quadratic — and I gave them no measured share, unlike the 16.1% and 8.1% items. That was the wrong basis for putting them in front of you.

**One question:** do you want the dead `if (0)` block deleted, or is that gate meant to come back? Happy to send the deletion either way; I did not want to remove a superseded gate you might still be holding onto.

**Still worth having, with a caveat I should have flagged earlier:** `strbuf_slot_eligible_shape` (16.1%) and `promote_append_accumulators` (8.1%) were both profiled on the pre-#3457 baseline — the run where the fixpoint hit the 128-iteration cap. `promote_append_accumulators` is itself half of that ping-pong, so its share is the most likely to have shrunk once convergence was restored. Those two want re-profiling against a converging tree before either is picked up.

The real successor to this PR is #3460, which does exactly the substitution you had in mind (`lw_shared_first` for the local-write query, plus an `ivw_shared_first` for the ivar analogue) at the site where it is actually 16.7% of the front end.

<sub>Method: temporary counters in the two loops, `__attribute__((destructor))` dump. The `roundhouse/scratchpad` snapshots no longer compile against master (an internal `bool` vs `String` conflict on `email_replies=`, then an unsupported `size` call), so this was a scratch copy with those 10 setter seeds relaxed to `untyped`; its front end runs 8.3s rather than the 43.3s in #3460. Even allowing 5x more work on the unpatched tree, `:7374` lands near 20M — still under 1%.</sub>

Issue #1302, opened 2026-06-03
spinel_analyze consumes 100+ GB RSS (OOMs the host) on large auto-generated API-SDK gems


OriPekelman:
## Summary

The analysis pass (`spinel_analyze`) can consume **100+ GB of resident memory** compiling certain large, auto-generated gems, running the host out of memory. Instead of failing the compile with a diagnostic, it grows until the kernel OOM-killer fires — which on a shared/dev box can take down unrelated processes.

This was found running the [spinelgems](https://spinelgems.org) compatibility harness (a full-corpus re-probe of ~190k gems against Spinel `master`); the harness's OOMs are how it surfaced.

## Evidence

Five separate OOM-kills of `spinel_analyze` during one corpus sweep, from the kernel log (`journalctl -k`), anonymous RSS at kill time:

```
Out of memory: Killed process … (spinel_analyze) anon-rss: 107 GB
                                                  anon-rss: 115 GB
                                                  anon-rss: 114 GB
                                                  anon-rss: 115 GB
                                                  anon-rss: 112 GB
```

(box has 121 GB RAM — i.e. a single `spinel_analyze` very nearly exhausts the machine before the kernel intervenes.)

## Reproducers

All three are **APIMATIC-auto-generated API-client SDKs** on RubyGems — hundreds of `class Model < BaseModel` definitions with `attr_accessor` fields, `from_hash`/`to_hash`, and **model fields whose types are other models** (a large, mutually-referential type graph):

| gem | files | LOC | entrypoint |
|---|---|---|---|
| `plaid-apimatic-sdk` 1.0.0 | 2059 | 247k | `lib/the_plaid_api.rb` |
| `payrix-api-sdk` 1.0.0 | 1136 | 192k | `lib/payrix.rb` |
| `words_ruby_sdk` 18.6 | 299 | 68k | `lib/words_ruby_sdk.rb` |

```sh
gem fetch plaid-apimatic-sdk -v 1.0.0
gem unpack plaid-apimatic-sdk-1.0.0.gem
spinel plaid-apimatic-sdk-1.0.0/lib/the_plaid_api.rb     # watch RSS of spinel_analyze
```

[...]

PR #3460, opened 2026-07-30
Answer mark_empty_hash_key_ctx's write-site lookups from an index, not a rescan (lobsters front end 43.3s -> 35.5s)

rubys:
`mark_empty_hash_key_ctx` carries the "empty `{}` indexed with a typed key" context back to the slot's write literal, so the hash variant is chosen off the key rather than defaulted. For an ivar (`@h[sym]`) or local (`h[sym]`) receiver it found those writes by **scanning the entire node table per read site** — twice, one loop for ivar writes and one for locals — and it runs on every fixpoint iteration.

This is the same rescan shape as #3456 / #3457, one loop deeper: the [#3115](https://github.com/matz/spinel/issues/3115) villain again, now the single largest self-time frame left in the front end.

On roundhouse's lobsters emit (142k nodes) that is ~730 read sites × 142k nodes × 13 passes ≈ **1.19 billion inner iterations, 16.7% of the front end**.

## The fix

Both answers are already indexed:

- **locals** — `lw_shared_first(c, ln, si)` is the existing `(scope, name)` local-write index. Drop-in.
- **ivars** — the index existed (`ivw_index_build`, added for the #1302 `from_hash`/`to_hash` shape) but only as a pass-local instance. This exposes it as `ivw_shared_first` / `ivw_shared_node` / `ivw_shared_next`, under the same frozen-cache contract as `lw_shared`: live while the scope index is frozen, stamped with `comp_scope_index_gen()`, rebuilt when the node table grows.

The bucket key is the ivar name only — a syntactic property of the write node — so scope re-homing needs no re-key; the caller already confirms node kind, name (hash collisions), and defining class on every hit, and those checks are unchanged. Each inner scan goes from 142k nodes to O(matches).

**Why the result is identical:** both loops are existence checks over write sites — order-independent and idempotent per write. The bucket contains every node the table scan would have matched and nothing else that survives the caller's filters.

## `ivw_index_build` walks the kind lists now

Secondary, but it matters because the cache rebuilds whenever the growing table invalidates it — once per pass in practice. It now iterates the kind-grouped id lists (`nt_nodes_of_kind` over `NK_InstanceVariableWriteNode` / `NK_InstanceVariableOrWriteNode`) the way `lw_index_build` does, instead of filtering the whole table. Rebuild cost drops from O(nodes) to O(ivar writes), and the pre-existing #1302 consumer at `analyze_pass.c:1026` gets the same speedup for free.

## Measurements

`spin build` of roundhouse's lobsters emit, `spin clean` before each run:

| lane | front end |
|---|---|
| master `7b6fa010^` | 43.3s |
| this branch | **35.5s** |

**−18%**, on top of the #3456 + #3457 wins already landed.

## Verification

- **Generated C for the whole lobsters app is byte-identical** to master's — same program, faster in.
- `make test`: no new failures. The lone `system_argument_list` failure in my sandbox needs `ruby` on `PATH` and fails identically on the unmodified baseline.

### DIFF
--- a/src/analyze.c
+++ b/src/analyze.c
@@ -6174,7 +6174,11 @@ static int mark_empty_hash_key_ctx(Compiler *c) {
       Scope *rs = comp_scope_of(c, recv);
       int rcid = rs ? rs->class_id : -1;
       if (!ivn || rcid < 0) continue;
-      for (int w = 0; w < nt->count; w++) {
+      /* Walk only the ivar-write nodes named `ivn` (index bucket) instead of
+         rescanning the whole table per read site -- an O(sites * nodes)
+         quadratic on ivar-heavy model graphs otherwise. */
+      for (int r = ivw_shared_first(c, ivn); r >= 0; r = ivw_shared_next(r)) {
+        int w = ivw_shared_node(r);
         if (nt_kind(nt, w) != NK_InstanceVariableWriteNode) continue;
         const char *wn = nt_str(nt, w, "name");
         Scope *ws = comp_scope_of(c, w);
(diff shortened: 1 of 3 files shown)


matz:
Merged.

Verified locally before merging:

- **Generated C byte-identical across all 2538 `test/*.rb`** (compiled every one on both lanes and diffed the trees: zero differing files), and for the optcarrot single-file input.
- `make test` 2566 pass, bench 59, optcarrot checksum 59662 at 1519-1520 fps, clang build + test green. The `system_argument_list` failure you saw does not appear here — `ruby` is on `PATH` in this environment — and it is unrelated either way.
- Front end on the optcarrot input: 0.20s -> 0.19s. Small because that tree is 4.3k lines; the shape is superlinear in read-sites × nodes, which is exactly why it shows as 18% on the 142k-node tree and barely here.

The argument for why the result is identical — both loops being order-independent existence checks, and the bucket being a superset filtered by the caller's unchanged checks — is the part that made this easy to accept. Same for calling out that the bucket key is syntactic so re-homing needs no re-key.

That is three in this series now (#3456, #3457, #3460). If `strbuf_slot_eligible_shape` is still the next largest, it sounds like the one worth doing even though it needs a real index rather than a memo.

PR #3477, opened 2026-07-31
Root the receiver temporary of a polymorphic send

### LINKED ISSUE #3476: [Compiler] The receiver temporary of a polymorphic send is not GC-rooted, so a collection inside the callee frees the receiver
`spinel 61bf38bd`, Apple clang 21.0.0, Darwin 25.5.0 arm64.

A polymorphic send compiles to a `switch (cls_id)` over a temporary holding the receiver. That temporary is the only live reference to the receiver, and it is not registered as a GC root — so a collection triggered by allocation *inside the callee* frees the receiver while its own method is still running. The next ivar read dereferences freed memory.

## Repro

```ruby
class Holder
  def initialize(params = {})
    @params = params
  end

  def go
    n = 0
    i = 0
    while i < 500            # allocate inside the callee, to trigger a GC
      s = "allocate #{i}"
      n += s.length
      i += 1
    end
    n - n + (@params[:k] || []).length
  end
end

class Other
  def go
    0
  end
end

def pick
  return Other.new if false  # gives the call site a union receiver type
  Holder.new(k: [1, 2, 3])
end

r = 0
i = 0
while i < 100
  r += pick.go               # receiver lives only as the dispatch temporary
  i += 1
end
puts "ok r=#{r}"
```

```
$ ruby repro.rb
ok r=300

$ spinel repro.rb && ./repro
[1]    SIGSEGV (exit 139)
```

```
EXC_BAD_ACCESS (code=1, address=0x0)
frame #0: sp_Holder_go
->  ldr x11, [x10, x9, lsl #3]      ; indexed load, base register = NULL
```

## Control

The same program with the receiver bound to a local instead of left as a dispatch temporary — `r += pick.go` becomes `h = pick` / `r += h.go`. It passes on both runtimes:

```
$ ruby control.rb   ->  ok r=300
$ ./control         ->  ok r=300
```

## Generated C

From `spinel repro.rb -c`:

```c
sp_RbVal _t10 = sp_pick();              /* <- no SP_GC_ROOT_RBVAL */
switch (_t10.tag == SP_TAG_OBJ ? _t10.cls_id : 0x7fffffff) {
  case 0: _t11 = sp_Holder_go((sp_Holder *)_t10.v.p); break;
  case 1: _t11 = sp_Other_go((sp_Other *)_t10.v.p); break;
  default: sp_raise_nomethod(sp_nomethod_msg("go", _t10)); break;
}
```

Every other temporary in the same function gets an `SP_GC_ROOT` / `SP_GC_ROOT_RBVAL` on the following line; this one does not. The control's local emits `SP_GC_ROOT_RBVAL(lv_h)`, which is why it survives.

Instance methods do not root their own `self` — `SP_GC_ROOT(self)` appears only in the `_new` constructors — so the calling frame is the only place the receiver can be rooted.

## Why it can look intermittent

Crash timing tracks GC frequency rather than work done. In a larger application (a Rails-shaped benchmark, ~114 requests per iteration) the same binary survived ~10 iterations before dying; under `SPINEL_GC_STRESS=1` it died before the first request completed. The reduced case above crashes either way.

That application's trigger was the same shape: a method returning a freshly built object, called polymorphically (`Relation | StoryRepository`), whose callee allocates before reading an ivar.

rubys:
Fixes #3476.

A polymorphic send compiles to a `switch (cls_id)` over a temporary holding the receiver. That temporary was not registered as a GC root, so when it is the only live reference (`r += pick.go`) and the selected method allocates enough to trigger a collection, the receiver is freed while its own method is still running — the next ivar read dereferences freed memory. Instance methods do not root their own `self` (only the `_new` constructors do), so the calling frame is the only place the receiver can be kept alive.

## Fix

Emit `SP_GC_ROOT_RBVAL` on the dispatch temporary, the same rooting a bound local (`h = pick` / `h.go`) or the monomorphic call temp already receives, at the three dispatch sites that were missing it:

- `codegen_call.c` — the zero-arg poly method dispatch (`emit_poly_method_dispatch`, the shape in the issue)
- `codegen_call.c` — the with-args branch of the same dispatch, which had the identical gap
- `codegen_stmt.c` — the poly attr accessor dispatch, where a write arm's value expression can allocate with the receiver equally unrooted

The root's `__attribute__((cleanup))` pops at the end of the enclosing statement expression, exactly the span the switch arms run in. (`codegen_iter.c`'s inline-block dispatch already rooted its receiver.)

## Before / after

Generated C for the issue's repro, at the call site `r += pick.go`:

```c
/* before */
({ sp_RbVal _t10 = sp_pick(); mrb_int _t11 = 0; switch (...) { ... } _t11; })

/* after */
({ sp_RbVal _t10 = sp_pick(); SP_GC_ROOT_RBVAL(_t10); mrb_int _t11 = 0; switch (...) { ... } _t11; })
```

The issue's repro segfaulted before (exit 139, also under `SPINEL_GC_STRESS=1`) and now prints `ok r=300` both ways.

## Test

`test/poly_dispatch_recv_gc_root.rb` distills the issue repro and makes it deterministic: each candidate method calls `GC.start` before reading an ivar, through both the zero-arg and with-args dispatches. It segfaults on the pre-fix compiler (verified) and passes with the fix, including under `SPINEL_GC_STRESS=1`.

Full suite: 2579 pass, 0 fail, 0 error.


matz:
Merged as 17293e2e (rebased onto master, so GitHub does not link it automatically).

Thanks -- the diagnosis is exactly right. I confirmed the pre-fix shape as a heap-use-after-free under ASAN (`sp_SymPolyHash_get` reading `@params` from freed storage) and clean after, and checked the two dispatch sites you added the root to are the same ones optcarrot exercises, on its reset path only: fps is unchanged (max of 3, 1509 before vs 1534 after, checksum 59662), and the full gate is green including the ruby/spec legs.

One follow-up went in with it. The test read the collected ivar straight away, and freed storage still holds the old bytes, so it printed the right answer on Linux with and without the fix -- only ASAN or macOS could tell the two apart. It now allocates between the `GC.start` and the ivar read, so the receiver's slot is handed out again and the stale read is a crash; it segfaults on the pre-fix compiler and passes with it, including under `SPINEL_GC_STRESS=1`.

PR #3478, opened 2026-08-01
Keep every build target when `--` comes first

rubys:
`spin build -- --profile` — pass a compiler flag, name no target — dies with:

```
spin: no such executable: bin/--.rb
```

The `--` index is 0 in that case, and `rest[0..(di - 1)]` is `rest[0..-1]`: the whole array, `--` included, which then becomes the first build target. `rest[0, di]` is empty at `di == 0`, which is what "no target named" should mean — build them all.

Naming a target (`spin build blog -- --profile`) works today and still does.

### DIFF
--- a/tools/spin.rb
+++ b/tools/spin.rb
@@ -1602,7 +1602,12 @@ when "build", "run", "test", "clean"
     if rest.include?("--")
       di = rest.index("--").to_i
       extra = rest[(di + 1)..-1].join(" ")
-      rest = rest[0..(di - 1)]
+      # `rest[0, di]`, not `rest[0..(di - 1)]`: with `--` first (`spin build --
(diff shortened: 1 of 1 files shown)


matz:
Merged as 2b02e1c2 (rebased onto master, so GitHub does not link it automatically).

Confirmed the repro on a scaffolded project and all three forms after: bare `--` now builds every bin with the flag applied (the `.symbols.json` that `--profile` writes is there), `spin build app -- --profile` still works, and a plain `spin build` is unchanged.

Added a test with it — the end-to-end run now covers both `--` forms. It asserts on the artifacts rather than the exit status, since the bug printed a message and built nothing; written as an explicit `|| fail` so `set -e` does not abort the script silently. It fails on the unfixed spin with `build -- <flag>: exited non-zero`.

PR #3479, opened 2026-08-01
Report string allocation sites, not just a string total

rubys:
`SPINEL_ALLOC_SITES=1` attributes every type to a calling frame except the one that dominates the bytes. Strings rode a pair of standalone counters (`sp_alloc_str_count` / `sp_alloc_str_bytes`) and came back with no site at all:

```
alloc;String 42355                      # 97% of the bytes, one anonymous line
alloc;...;Hash(String,String) 874       # everything else, attributed
```

So the report can say "this request allocates 1.6 MB of String" and never say by whom — which is the next question every time.

Strings now take a reserved key in the same stats table as everything else, so they follow the same per-site path. With `SPINEL_ALLOC_SITES` off they all land in one slot (site `NULL`) and the dump is byte-identical to the old aggregate lines.

The table grows 2048 → 8192 entries to suit what now fills it: one entry per *(type, site)* pair rather than one per type. A full table merges into the home slot, and silently misattributing the numbers is the one failure this feature must not have. It is BSS, so the untouched tail costs nothing when the report is off.

### What it buys

On a Rails-scale app (lobsters, AOT) the string lines now name the caller directly:

```
626944  sp_Layouts_s_application + 37952;String
510081  sp_Settings_s_index + 57804;String
509384  sp_Settings_s_index + 57772;String
509097  sp_Settings_s_index + 57740;String
508892  sp_Settings_s_index + 57700;String
```

A dozen adjacent offsets in one view, each copying ~24 KB per visit — an O(N²) template append that the per-type report could only show as one large total.

### DIFF
--- a/docs/profiling.md
+++ b/docs/profiling.md
@@ -74,7 +74,10 @@ alloc;(no-scan) 100
 buffer — which the collector never has to walk.
 
 Counters key on the object's GC scan callback, which is the de-facto type
-identity, and are bumped inside the allocator itself. Nothing is sampled, so
+identity, and are bumped inside the allocator itself. Strings have no scan
+callback, so they carry a reserved key of their own and otherwise behave like
+any other row — including on the per-site path below, which matters because
+strings are usually the largest share of the bytes. Nothing is sampled, so
 two runs of the same program report the same numbers.
 
 ### Per-site attribution: `SPINEL_ALLOC_SITES`
--- a/lib/sp_alloc.c
+++ b/lib/sp_alloc.c
@@ -508,9 +508,14 @@ const char *sp_float_to_s(mrb_float f) {
 int sp_alloc_report_on = 0;
 static int sp_alloc_sites_on = 0;
 typedef struct { void *key; void *site; unsigned long long count, bytes; } sp_AllocStat;
-#define SP_ALLOC_STATS 2048
+/* Sized for the per-SITE case, which is what fills this table: one entry per
+   (type, site) pair rather than one per type. Strings alone reach into the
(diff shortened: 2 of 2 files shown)


matz:
Merged as d5b9e70a (rebased onto master, so GitHub does not link it automatically).

Checked the three things worth checking.

**Sites off is unchanged in content.** Same lines, same numbers on a string-heavy program. Their *order* moves, though — the String rows now come out of the table in hash order rather than being printed around the loop — so "byte-identical" is not quite it. Folded output is order-independent and every consumer aggregates, so this is fine; noting it in case anyone diffs two reports.

**Sites on does what it says.** 800 strings from two callers split into two lines with the right counts. Worth recording where they land: the string sites resolve to the runtime helper that built the string (`sp_str_repeat`), not the Ruby method above it. That is the same depth the object rows already report (`sp_StrArray_new`, `sp_StrStrHash_new`), so it is the existing "one frame is not always the frame you want" limit rather than anything new — strings are now consistent with everything else, which is the point. Your lobsters example names user code because those strings are built inline in the generated function rather than through a helper.

**The table growth is free when off**, as claimed: BSS +192 KB, file size unchanged, max RSS 1948 KB before vs 1940 KB after. The tail is never touched.

One test went in with it. `test/alloc-report/sites.rb` allocated arrays and hashes and no strings, so the row this changes was unasserted in both directions — the gate would have passed with strings dropped from the report entirely. It allocates strings now, and the gate checks the String line, its bytes line, and, when the run produced sited lines at all, that String is among them. That last assertion fails on the pre-change runtime.

Issue #2635, opened 2026-07-16
[Runtime] `Time#strftime` `%z`/`%s` use the machine-local offset, not the receiver's


KieranP:
> (Disclaimer: I asked Claude Opus 4.8 to find Spinel compile/runtime issues with documented Ruby classes and methods. It found the following issue, which I manually verified for accuracy before publishing this report)

_Reproduced with `spinel 8f9ea9304295` vs `ruby 4.0.3`._

For a non-local receiver (here a `Time.utc` value), `strftime("%z")` emits the
machine's local UTC offset instead of the time's own offset, and `strftime("%s")`
(seconds since the epoch) is shifted by that same offset. `%Z` and `#to_i` are
correct, so the bug is confined to the two offset-derived directives. On a machine
whose local zone is `+1200` (NZST) the epoch drifts by 43200 s.

## Repro

```ruby
t = Time.utc(2026, 7, 16, 13, 45, 30)
p t.strftime("%z")   # zone offset
p t.strftime("%s")   # seconds since epoch
p t.to_i             # correct epoch seconds, for comparison
```

Ruby:

```
"+0000"
"1784209530"
1784209530
```

Spinel:

```
"+1200"
"1784166330"
1784209530
```

KieranP:
@matz Not fixed on latest master:

```
› ruby 20260716182711110_strftime_offset.rb
"+0000"
"1784209530"
1784209530

› ~/spinel/bin/spinel -E 20260716182711110_strftime_offset.rb
"+1200"
"1784209530"
1784209530
```

matz:
Thanks for re-testing -- you caught a real residual, and a platform-shaped one. Fixed in 0038cced.

What happened: the earlier fix routed the **colon forms** (`%:z`, `%::z`, `%:::z`) through the receiver's own offset, but bare `%z` still rode the C `strftime` whitelist. C's `%z` reads `tm_gmtoff`, which is filled by whichever function built the `struct tm` -- and that varies by platform. glibc's `gmtime` zeroes it, which is why the UTC receiver looked fine on Linux; macOS reported the machine-local zone instead, hence your `+1200`. A fixed-offset receiver (`Time.at(0, in: "+05:30")`) was wrong everywhere, since it has no `tm` representation at all.

Bare `%z` now computes from the receiver's offset exactly like the colon forms:

```
t = Time.utc(2026, 7, 16, 13, 45, 30)
t.strftime("%z")                        # => "+0000"   (any machine zone)
[...]

PR #3492, opened 2026-08-01
Render a fixed-offset Time's civil fields in strftime and iso8601

rubys:
`sp_Time.is_utc` is a 3-state zone kind — `0` host-local, `1` UTC, `2` a fixed
`utc_off` — but `sp_time_strftime`, `sp_time_iso8601` and `sp_time_iso8601_frac`
tested it as a boolean. Kind 2 is truthy, so a fixed-offset Time took the
`gmtime` branch and rendered UTC clock fields. Bare `%z` is computed from the
receiver's own offset (#2635) and so stayed correct, which makes the output
disagree with itself:

```ruby
require "time"
t = Time.at(1689676186).getlocal(-5 * 3600)

p t.hour                  # 5                          ✅ (routes through sp_time_vtm)
p t.to_s                  # "2023-07-18 05:29:46 -0500" ✅

p t.strftime("%H:%M %z")  # "10:29 -0500"   ❌ CRuby: "05:29 -0500"
p t.iso8601               # "2023-07-18T10:29:46Z"
                          #                ❌ CRuby: "2023-07-18T05:29:46-05:00"
p t.strftime("%Z")        # "UTC"           ❌ CRuby: ""
```

`sp_time_vtm` has always resolved all three kinds correctly — which is why
`#hour`, `#to_s`, `#inspect` and `#utc_offset` were already right — the three
formatters just weren't calling it. This points them at it.

While there, `iso8601` and `iso8601_frac` had two copies of the same
`mktime(gmtime(s)) - s` offset-suffix block; both now share one helper that
takes the offset `sp_time_vtm` already resolved, so the `Z`-vs-`±HH:MM` choice
is made from the zone kind rather than from a truthiness test. Net effect on
`lib/sp_time.c` is a small reduction in logic.

`%Z` gets the same 3-state correction: kind 2 was answering `"UTC"`, where
CRuby renders a fixed-offset time's zone name as empty.

### Scope

`Time#zone` still returns `""` for kind 2 where CRuby returns `nil`. That's a
return-type change rather than a formatting one, so it is deliberately left
alone here; the new test notes it.

### Test

`test/time_fixed_offset_formatters.rb` — the `.expected` file is CRuby's own
output, and every case uses an explicit offset or `Time.utc`, so it does not
depend on the host zone. It covers the field directives, the RFC 2822 shape,
a date that rolls back over midnight under the offset, offsets both sides of
Greenwich, `Time.new`'s 7th-argument construction, and the UTC/host-local
kinds as regression guards.

Full suite: 2594 pass, 0 fail, 0 error.


matz:
Merged as the merge commit on master (rebased, so GitHub does not link it automatically).

Confirmed the report and the fix, and spent most of the time on the kinds the change did *not* target, since all three formatters moved to a different code path for them too. Host-local and UTC produce identical output before and after under `TZ=UTC`, `Asia/Tokyo`, `America/New_York` and `Europe/Berlin` — including `%Z`, where the zone name still resolves through the `struct tm` sp_time_vtm builds, and a winter timestamp for the DST boundary.

Also checked the test's independence claim rather than taking it: its expected file matches CRuby's output exactly, and the compiled binary produces byte-identical output under all four zones.

Two notes, neither blocking.

**A pre-existing gap the extreme-value probes turned up.** `%Y` does not zero-pad a year below 1000, so `Time.at(-62135596800).utc.iso8601` renders `1-01-01T00:00:00Z` where CRuby renders `0001-01-01T00:00:00Z`. master does the same on the UTC kind, so it is nothing to do with this change — worth its own issue if you want it fixed.

**`sp_time_vtm` zeroes the `struct tm` when gmtime/localtime returns NULL**, where the old formatter code returned an empty string. So an unrepresentable timestamp now renders epoch-zero civil fields instead of `""`. That is sp_time_vtm's existing contract — `#hour` and `#to_s` have always behaved that way — so this makes the formatters consistent with the rest rather than introducing anything new. Mentioning it only because it is a behaviour change on a path nobody is likely to test.

The `iso8601`/`iso8601_frac` deduplication reads well, and keying the `Z`-vs-`±HH:MM` choice off the zone kind rather than a truthiness test is the right shape.

Issue #3505, opened 2026-08-02
A nil returned through a block typed Integer? is a different Hash key than a literal nil (the yield counterpart of #3493)


rubys:
Under `--rbs`, a nil that reaches a Hash key **through a block** is a different key than a literal `nil`. The same value read *directly* is correct, which is what makes this one narrower than #3493.

This is the yield/block counterpart of the nilable-scalar family: #3412 (ivar), #3458 (return), #3493 (Hash key identity). Those are closed and their fixes work — I re-ran #3493's repro on `bfa17733` and it now matches CRuby exactly. This shape survives them.

## Repro

`main.rb`:

```ruby
class R
  def initialize(p)
    @p = p
  end

  def p_
    @p
  end
end

# The shape ActiveRecord's group_by has: fetch-then-insert, keyed by the
# block's return value.
def group_by_p(arr)
  out = {}
  arr.each do |x|
    k = yield x
    a = out.fetch(k, nil)
    if a.nil?
      a = []
      out[k] = a
    end
    a << x
  end
  out
end

rows = [R.new(nil), R.new(7), R.new(nil)]

# DIRECT read of the same Integer? value — correct everywhere.
d = rows[0].p_
puts "direct  cls=#{d.class} val=#{d.inspect} nil?=#{d.nil?}"

parents = group_by_p(rows) { |x| x.p_ }
puts "groups  #{parents.length}"
parents.each do |k, v|
  puts "key     cls=#{k.class} val=#{k.inspect} nil?=#{k.nil?} n=#{v.length}"
end
puts "[nil]   #{parents[nil].inspect}"
puts "[d]     #{parents[d].inspect}"
```

`main.rbs`:

```rbs
class R
  def initialize: (Integer?) -> void
  def p_: () -> Integer?
end
```

Run: `spinel --rbs . main.rb && ./main`

## Observed

```
                      CRuby            spinel --rbs .
direct read .class    NilClass         NilClass          OK
direct read .nil?     true             true              OK
stored key .inspect   nil              nil               OK
stored key .class     NilClass         Integer           <-- differs
stored key .nil?      true             false             <-- differs
parents[nil]          the 2 rows       nil               <-- differs
parents[d]            the 2 rows       the 2 rows        OK
```

Full output under `spinel --rbs .`:

```
direct  cls=NilClass val=nil nil?=true
groups  2
key     cls=Integer val=nil nil?=false n=2
key     cls=Integer val=7 nil?=false n=1
[nil]   nil
[...]

PR #3506, opened 2026-08-02
Follow a nilable scalar through yield, an inferred return, and poly dispatch

rubys:
Refs #3505. **Partial — deliberately does not close it.** Tested against the original codebase there are still shapes in this family that get through; this lands the three the issue's repros isolate and leaves the rest open.

## The defect

The nilable-int slot *representation* is correct. `k` is an `mrb_int` holding `SP_INT_NIL`, which is why `k.class` / `k.nil?` / `k.inspect` all read right off the local. What's wrong is only the **box at the poly boundary**:

```c
// two.c:284
sp_PolyPolyHash_set(lv_h, sp_box_int(lv_k), sp_box_str("from-block"));
//                        ^^^^^^^^^^ SP_INT_NIL goes out as Integer(INTPTR_MIN)
sp_PolyPolyHash_set(lv_h, sp_box_nil(),     sp_box_str("from-literal"));
```

`sp_box_int_or_nil` isn't the default because it costs throughput — forcing it on every int box measures ~12% on optcarrot. So nilability is a static marking pass (`mark_nullable_int_locals` / `nullable_int_value`) that enumerates the syntactic **shapes** that can leave the sentinel. That enumeration is the hole factory.

## Three gaps

Each probe is `h[k] = …; h[nil] = …; puts h.length`; CRuby answers 1.

| shape | before | gap |
|---|---|---|
| `k = r.p_` (typed local receiver) | 1 ✅ | — |
| `k = yield x` | **2** | no `NK_YieldNode` arm |
| `k = pass(r)`, `def pass(x) = x.p_` | **2** | inferred nilable return not propagated |
| `k = rows[0].p_`, `rows` widened to poly | **2** | callee unresolvable through a poly receiver |

1. **`yield`** — no `NK_YieldNode` case, so `k = yield x` never marks. New `yield_block_tails` is `yield_value_type`'s call-site scan answering with **nodes** instead of a unified type: nilability is a property of the producing expression, not of the type it settles on — an `Integer?` and an `Integer` are both `TY_INT`. Forwarded blocks (`m(&b)` / `m(...)`) delegate to the enclosing method. The callee resolution both scans need is extracted into `yvt_callee_index` rather than duplicated.

2. **Inferred nilable return** — the `CallNode` arm ended at `ret_rbs_nilable`, which `seed_method` sets *only* from an RBS signature. New `Scope.ret_nullable_int` is the inferred twin, seeded from it and propagated through the marking fixpoint from each method's tail and its explicit `return`s. This is what `two.rb`'s `key_of` needs — that repro fails through this gap, not through the yield one.

3. **Poly receiver** — resolution only handled `ty_is_object` or a constant receiver. When the receiver stays `TY_POLY`, `poly_dispatch_nullable` asks the runtime dispatch set instead: any same-named method with a nilable scalar return, or any `attr_reader` over a scalar ivar. Over-marking costs a branch; missing one is the silent-wrong key.

All three hit `Float?` identically.

## Both box sites

`h[yield x] = v` with no local was wrong the same way, so codegen's `call_returns_nullable_int` now consults the analyzer's predicate first and keeps its own arms as the backstop for builtins analyze doesn't model.

## One correction to the issue's table

The `parents[d]` → OK row was passing by coincidence. `main.c:341` emitted `sp_box_int(lv_d)` — the direct read was plain-boxed too (gap 3), matching only because the stored key was equally wrong. Patching gap 1 alone gives:

```
key     cls=NilClass val=nil nil?=true n=2     <- fixed
[nil]   [#<R:...>, #<R:...>]                   <- fixed
[d]     nil                                    <- REGRESSED
```

So gaps 1 and 3 had to land together.

## Fixpoint cap

Method returns now chain through the marking fixpoint, so a 5-deep pass-through chain would have under-marked at the old cap of 4. Cap goes 4 → 32, and **failing to converge is now an error rather than a silent stop** — running out of rounds leaves a slot unmarked, which is exactly the silent-wrong-key this pass exists to prevent, so it refuses to emit.

Instrumented over the 2636 test + benchmark programs: 2563 settle in **1** round, 72 in 2, one in 3. The cap is ~10× the observed max. Verified the check is live, not dead code, by starving the cap to 1 — it fires with a clear message, and the build then fails compiling spinel's own `tools/spin.rb`, so it bites on real code.

## Cost

**None on the hot path.** optcarrot's generated C is byte-identical to master (200 `sp_box_int`, 11 `sp_box_int_or_nil` both ways) and compile time is unchanged at 0.16s. fps readings moved around inside run-to-run noise — the same binary measured 1818–1960 across runs — so the byte-identical codegen is the real signal.

## Verification

- Both `#3505` repros match CRuby byte-for-byte, including the `parents[d]` row.
- 8 probes: 3 local-slot shapes, 3 direct-expression shapes, `Float?`, plus the plain-read control.
- `make test` 2603 pass / 0 fail · `make bench` 60 pass · `make rubyspec-gate` 2109 examples across 6 suites · optcarrot checksum `59662`.

New fixture `test/rbs-seed/nilable_scalar_yield_key.rb` covers all three gaps through both a local and a direct box, plus `Float?`, the `group_by`-then-`[nil]` shape from the report, and a non-nil control so a real `7` isn't swept into nil. Confirmed it's a real guard: on master it fails all seven lines and returns `roots []`.

## Still open

Issue #3505 should stay open after this merges. The marking pass is still a shape enumeration, so the family's root cause — nilability tracked per-syntactic-form rather than as a property of the slot — survives, and more shapes in this family are known to get through. Post them on the issue and I'll extend the same three seams.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -751,7 +751,7 @@ rbs-seed-test: $(SPINEL) $(RBS_EXTRACT_BIN) $(SP_RT_LIB)
 	     -c --no-line-map -o "$$tmp/sxa.c" >"$$tmp/sxa.out" 2>&1; then \
 	  echo "rbs-seed-test: FAIL (a contradicted seed on an ARGUMENT compiled)"; ok=0; \
 	else grep -q "seed contradicted" "$$tmp/sxa.out" || { echo "rbs-seed-test: FAIL (contradicted argument rejected without saying why)"; sed -n 1,5p "$$tmp/sxa.out"; ok=0; }; fi; \
-	for t in hash_kind_widened_return poly_dispatch_arm_arg_type; do \
+	for t in hash_kind_widened_return poly_dispatch_arm_arg_type nilable_scalar_yield_key; do \
 	  $(SPINEL) test/rbs-seed/$$t.rb --rbs test/rbs-seed/sig -c --no-line-map -o "$$tmp/$$t.c" 2>/dev/null; \
 	  if $(CC) -O0 -Ilib $(RBS_SEED_STRICT) "$$tmp/$$t.c" $(SP_RT_LIB) $(LDFLAGS) -lm -o "$$tmp/$$t" 2>"$$tmp/$$t.err"; then \
 	    "$$tmp/$$t" > "$$tmp/$$t.out" 2>/dev/null; \
--- a/src/analyze.c
+++ b/src/analyze.c
@@ -8236,20 +8236,68 @@ static int nullable_int_call_name(const char *nm) {
   for (int i = 0; N[i]; i++) if (sp_streq(nm, N[i])) return 1;
   return 0;
 }
-/* Mark the int locals that can hold that sentinel, so codegen boxes them as
-   nil rather than as INTPTR_MIN. Boxing every int through the nil check costs
-   ~8% on optcarrot -- every pixel goes through it -- so the marking is static
-   and the hot path keeps the plain box. */
+/* A scalar slot on a class the fixpoint could not pin to a receiver still
+   dispatches at runtime: codegen emits a cls_id switch over every class that
+   defines the name. Ask whether ANY of those targets can answer the sentinel.
+   Over-marking here only costs the boxing branch; missing one is the
+   silent-wrong hash key of #3505, so the conservative direction is `yes`. */
+static int poly_dispatch_nullable(Compiler *c, const char *cn) {
+  if (!cn) return 0;
+  for (int si = 1; si < c->nscopes; si++)
+    if (c->scopes[si].name && sp_streq(c->scopes[si].name, cn) &&
+        c->scopes[si].ret_nullable_int) return 1;
+  /* an attr_reader over a scalar ivar: those slots are sentinel-defaulted
+     (ivar_scalar_nil_init), so the read carries the sentinel like a `return
+     nil` would -- the resolved-receiver twin of this lives in codegen's
+     call_returns_nullable_int */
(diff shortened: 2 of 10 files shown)


matz:
Merged as 666c996d. Verification I ran here:

- Both repros from #3505 now match CRuby exactly, including the `parents[d]` row your correction called out.
- Generated C is byte-identical to master for optcarrot, `tools/spin.rb`, and 2631 of the 2637 test + benchmark programs. The six that differ each swap one `sp_box_int` for `sp_box_int_or_nil` on a `<=>` result, where `(a > b) - (a < b)` can never be the sentinel, so it is over-marking with no behaviour change. Compile time unchanged (0.21s optcarrot, 0.11s spin.rb, three runs each).
- `make gate` all green.

One thing needed a follow-up (e76c8fdf, in the merge). The marking walks methods in definition order, so a chain of pass-through methods written in the order the propagation walks backwards advances one link per round. A legal 40-deep chain written that way ran out of the fixed 32 rounds and the compiler refused to emit, which turns a program CRuby runs into one that does not build:

```ruby
def dc40(x); dc39(x); end
...
def dc1(x); x.p_; end
```

The bound is now derived from the program (methods + locals). A round that changes anything sets one of the flags the bound counts and never clears one, so the rounds cannot outnumber the flags; your did-not-converge check keeps its meaning as a non-monotone backstop without being reachable by a legal program. A 300-deep chain compiles in 0.77s. Regression test at test/rbs-seed/nilable_scalar_deep_chain.rb.

Leaving #3505 open as you asked.

PR #3519, opened 2026-08-04
Make GC stress pin the threshold it seeds

rubys:
`SPINEL_GC_STRESS` sets the collection threshold to 2048 bytes — but as a *base*, and the post-collection retunes (`sp_gc_retune_object`, `sp_str_retune`) float the threshold to `live*4` with that base only as a floor. So on any program whose live set outgrows ~512 bytes, stress stops stressing after the first collection: the cadence retunes to within a whisker of the default's. This explains a pattern that runs through #3513 — stress reproduced every *boot-time* defect deterministically (the first collection fires at 2 KB, before anything is live) and never once touched a *request-time* one, and a minor-path bug could sit green on two CI legs and red on the third.

This change makes stress pin what it seeds: a `sp_gc_stress_pin` flag set wherever `SPINEL_GC_STRESS` is detected, and an early-out in both retunes that resets the threshold to its base. Nothing changes unless the env var is set.

**Why now: it turns the #3513 heisenbug into a deterministic repro.** On lobsters (the full frozen sequence, `-O2`, spinel HEAD + this change):

| configuration | result |
|---|---|
| `SPINEL_GC_MINOR=1` alone | dies ~1 run in 12 at 576e3f06, ~14 in 20 at 4b099a1a — layout luck, minutes-to-hours to catch |
| minor + stress, **pinned** | **dies at the first collections during boot, 3/3 runs, ~1 s** (`can't modify frozen Array` — a swept header reused) |
| minor + stress + `SPINEL_GC_VERIFY=1 SPINEL_GC_VERIFY_GEN=1`, pinned | deterministic abort, first collection, `phase=globals`, corrupt object `scan=0x0 size=0` |
| default GC + stress, pinned | **full sequence clean**, exit 0 — the corruption is minor-path only |
| minor + stress, *unpinned* (same tree, pre-change binary) | boot, parity, verify, warmup all pass; dies later at the natural rate — the pin is what determinizes |

The boot-time failure the pin surfaces is not necessarily the same member of the lost-object class as the `int_to_s` crash in the issue thread — pinned stress surfaces members in deterministic order starting from boot. That is the point: it converts "rerun the app N times and hope" into a worklist — fix the first member, rerun, the next appears, each step seconds. It should also let the existing suite catch minor-path holes directly: under the old semantics the suite's stress leg was only stressing programs with sub-kilobyte live sets.

**Cost.** Pinned stress is affordable where it matters: the full lobsters sequence completes under default+stress in under two minutes, and under the minor mark each pinned collection marks only the ~2 KB young set. Suite impact under `SPINEL_GC_MINOR=1 SPINEL_GC_STRESS=1 SPINEL_GC_VERIFY_GEN=1`:

**98 failures against 37 baseline** (macOS/arm64). The delta is the finding, not a cost: **all 63 newly failing tests also fail under pinned stress on the *default* collector** — none are minor-mark regressions. They are pre-existing unrooted locals of the `MatchData#named_captures` kind, concentrated in the Enumerator machinery (`enumerator_ops`, `to_enum_enum_for`, `with_index_enumerator`, the Set family; full list in a comment below), and each reproduces in isolation in under a second, e.g.:

    bin/spinel test/enumerator_ops.rb -c --no-line-map -o /tmp/eo.c
    cc -O2 -Ilib /tmp/eo.c lib/libspinel_rt.a -lm -o /tmp/eo
    SPINEL_GC_STRESS=1 /tmp/eo   # SIGSEGV; passes without the env var

Two tests flip the other way (`bignum_downto_upto_to_a`, `hash_each_value_key_enumerator` fail baseline-only) — flaky under the gate env, not pin-related as far as I can tell.

If the new reds are unwelcome on the stress CI leg until they're fixed, the leg can pin the old behavior by also setting `SPINEL_GC_THRESHOLD_KB` — but red is arguably what that leg is for.

Refs #3513.

### DIFF
--- a/lib/sp_alloc.c
+++ b/lib/sp_alloc.c
@@ -63,6 +63,12 @@ int sp_ffi_bin_len = 0;   /* see sp_alloc.h: byte count for :binstr / :cbinstr *
 size_t sp_gc_threshold = 256 * 1024;
 size_t sp_gc_threshold_init = 256 * 1024;
 int sp_gc_stress_checked = 0;
+/* Stress pins the threshold instead of merely seeding it: the retunes float
+   the trigger to live*4 with the base as a FLOOR, so on any program whose
+   live set outgrows the base, stress stopped stressing after the first
+   collection -- request-time bugs sat behind a cadence identical to the
+   default's while boot-time ones reproduced instantly (#3513). */
+int sp_gc_stress_pin = 0;
 
 #ifdef SP_THREADS
 pthread_mutex_t sp_heap_lock = PTHREAD_MUTEX_INITIALIZER;   /* see sp_alloc.h */
@@ -79,11 +85,11 @@ void sp_alloc_stress_init(void) {
   int stress = (e && *e && *e != '0');
   if (!sp_str_stress_checked) {
     sp_str_stress_checked = 1;
-    if (stress) { sp_str_threshold = 2048; sp_str_threshold_init = 2048; }
+    if (stress) { sp_str_threshold = 2048; sp_str_threshold_init = 2048; sp_gc_stress_pin = 1; }
   }
   if (!sp_gc_stress_checked) {
     sp_gc_stress_checked = 1;
-    if (stress) { SP_GC_CTR_SET(sp_gc_threshold, 2048); sp_gc_threshold_init = 2048; }
+    if (stress) { SP_GC_CTR_SET(sp_gc_threshold, 2048); sp_gc_threshold_init = 2048; sp_gc_stress_pin = 1; }
   }
 }
 
@@ -130,6 +136,7 @@ void sp_alloc_worker_tune(int workers) {
 /* Re-tune the object / string GC thresholds from the pre-collect live bytes
    (the heuristic mirrors the original inline code in sp_gc_alloc / sp_str_alloc). */
 void sp_gc_retune_object(size_t before) {
+  if (sp_gc_stress_pin) { sp_gc_threshold = sp_gc_threshold_init; return; }
   size_t freed = before - sp_gc_bytes;
   if (freed < before / 4) { sp_gc_threshold = before * 2; }
   else if (sp_gc_bytes > 0) { sp_gc_threshold = sp_gc_bytes * 4; if (sp_gc_threshold < sp_gc_threshold_init) sp_gc_threshold = sp_gc_threshold_init; }
@@ -142,6 +149,7 @@ void sp_gc_retune_object(size_t before) {
    cycle -- retuning on the aggregate would grow it geometrically for long-lived
    strings. The single-threaded build works in absolute bytes (N == 1). */
 static void sp_str_retune(size_t before, size_t promoted) {
+  if (sp_gc_stress_pin) { sp_str_threshold = sp_str_threshold_init; return; }
 #ifdef SP_THREADS
   int nw = sp_active_workers; if (nw < 1) nw = 1;
   size_t after = (sp_str_bytes_total() + promoted) / (size_t)nw;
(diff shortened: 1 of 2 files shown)


rubys:
The 63 tests that newly fail with the pin, under `SPINEL_GC_MINOR=1 SPINEL_GC_STRESS=1 SPINEL_GC_VERIFY_GEN=1` — every one also fails with the pin on the default collector (no `SPINEL_GC_MINOR`), so these are pre-existing unrooted locals, not minor-mark regressions:

```
array_callforms_strbuf_cycle
array_chain_no_arg
array_clone_poly
array_combination
array_op_assign_concat
bundle_array_b
bundle_array_c
bundle_array_d
bundle_class_06
bundle_classd_11
bundle_classd_17
container_read_blockless_enumerator
container_read_combined_set_map_bang
custom_enumerable_wave10
each_slice_cons_enumerator
each_with_index_enumerator
enum_map_chain
enum_terminal_chunk_zip_lazy
enumerable_chain_enumerator
enumerable_grep_sum_entry
enumerable_via_each
enumerator_generator_each
enumerator_ops
enumerator_size
enumerator_to_a
enumerator_wave10
gc_poly_local_setjmp
hash_array_string_batch8
hash_ctor_merge_clone_transform
hash_default_array_fetch_concat_wave9
hash_enum_with_index_chain
hash_solo_block_pair
inject_symbol_to_proc
inspect_result_survives_gc
int_array_rotate_window
issue_2874
issue_2982
issue_2993
issue_3160
issue_3171
issue_3225
issue_3228
issue_3305_int_enum_with_index
issue_3312_transpose_reduce
issue_3313_lazy_each_cons
lazy_drop_while_each_slice
object_dup_initialize_copy
pattern_methodobj_kwsplat_undersupply
pkg.set.set_value_semantics
poly_dispatch_keyword_args
poly_hash_surface
proc_method_wave10
ptr_array
range_comparable_wave8
set_basic
set_conformance_batch2
set_divide
set_full_api
set_operations
spaceship_nil_guard_poly_compare
to_enum_enum_for
user_enumerable_each_and_for
with_index_enumerator
```

matz:
Merged as f64e7a8d.

This is the instrument the unrooted-local work was missing, and I had just finished sizing that audit when the PR arrived. My count of candidates by static analysis: 34 locals assigned an allocation and never rooted across another one, plus 129 parameters in the same shape. What I could not do was tell which of those are real, because the generational verifier only sees old-to-young barrier misses -- this class is invisible to it, which is exactly why your lobsters crash only showed up as an `-O2` versus frame-pointer difference. 63 deterministic reproductions in one second each is a different situation entirely.

Your reading of why stress only ever caught boot-time defects is right and it is worse than it looks: the retune runs after the FIRST collection, so the window where stress is actually stressing is however long it takes to allocate 2 KB. Every request-time path in every program was behind a cadence within a whisker of the default. That the suite's stress leg was only stressing sub-kilobyte live sets explains a run of things I attributed to layout luck.

The change itself is right where it is. `sp_gc_stress_pin` reads only in the two retunes, both under stop-the-world, and the one-shot writes are idempotent under a race the way the existing `_checked` flags are. `make check` here is green with the env var unset, which is the only path that can change.

I am running the pinned sweep now and will work the list. A note for when you see the results move: two of the 63 are already gone. `sp_md_named_captures` and `sp_md_named_captures_sym` (602aa0b7) built their hash and then called `sp_str_dup` per named group without rooting it, and `sp_poly_enum_proc` (570e32d3) never rooted the block it drives the receiver with -- that one is on the path your backtrace named, so it may be your `-O2` crash. Worth a rerun on your side before we conclude anything about the remainder.

On the CI question: red is what that leg is for. I would rather see the leg fail on 63 real defects than pass because the instrument was blunt.

PR #3770, opened 2026-08-09
sp_crypto: derive PBKDF2 keys longer than one block

### LINKED ISSUE #3769: sp_crypto: PBKDF2 cannot derive more than one block (dkLen = 32)
# sp_crypto: PBKDF2 cannot derive more than one block (dkLen = 32)

`sp_crypto_pbkdf2_sha256_b64url` hardcodes `INT(1)` in the salt suffix and
copies a single 32-byte `T(1)`, so 32 bytes is the most it can produce.
That is the documented contract (`dkLen > 32 not supported` in
`sp_crypto.h`), so this is a gap rather than a bug — filing it because the
first consumer that needs the other block can't work around it in Ruby.

## Why 64 bytes

Rails derives its signed-cookie and signed-id keys through
`ActiveSupport::KeyGenerator#generate_key(salt, key_size = 64)`, i.e.
`OpenSSL::PKCS5.pbkdf2_hmac(secret, salt, iters, 64, SHA256)`. A spinel
program that wants to *read* a cookie a Rails process minted (or mint one
Rails will accept) has to reproduce that derivation exactly.

PBKDF2 blocks are independent — only the `INT(i)` suffix changes — so the
second block is derivable in principle from the outside. It isn't in
practice: `KeyGenerator` defaults to `2**16` iterations, and that inner
loop has to stay in C to be affordable.

## Repro

Build and run against current master:

```ruby
# pbkdf2_dklen.rb
module Crypto
  ffi_func :sp_crypto_pbkdf2_sha256_b64url, [:str, :str, :int], :str
end

got = Crypto.sp_crypto_pbkdf2_sha256_b64url("password", "salt", 1)

# Both values from Ruby's OpenSSL (the oracle Rails itself uses):
#   OpenSSL::PKCS5.pbkdf2_hmac("password", "salt", 1, N, SHA256.new)
want32 = "Eg-2z_z4syxD5yJSVsT4N6hlSMkszDVICAWYfLcL4Xs"
want64 = "Eg-2z_z4syxD5yJSVsT4N6hlSMkszDVICAWYfLcL4XtNvzovPa0zdyZLt7joMw1O_HRRQYYX2r72g3NTYc3BjA"

puts "got     " + got
puts "dkLen32 " + (got == want32 ? "MATCH" : "differs")
puts "dkLen64 " + (got == want64 ? "MATCH" : "differs")
puts "chars   " + got.length.to_s + " (dkLen=64 would be " + want64.length.to_s + ")"
```

```
$ spinel pbkdf2_dklen.rb -o pbkdf2_dklen && ./pbkdf2_dklen
got     Eg-2z_z4syxD5yJSVsT4N6hlSMkszDVICAWYfLcL4Xs
dkLen32 MATCH
dkLen64 differs
chars   43 (dkLen=64 would be 86)
```

The existing implementation is correct for the case it covers — `T(1)`
matches OpenSSL exactly, and the 64-byte derivation starts with those same
32 bytes. The gap is only that there's no way to ask for `T(2)`.

## Suggested shape

`sp_crypto.h` asks callers to file before adding to the surface, hence this
issue rather than a drive-by patch. Two options:

1. **Add a length variant**, keeping the surface additive and every current
   caller untouched:
   ```c
   const char *sp_crypto_pbkdf2_sha256_b64url_len(const char *password,
                                                  const char *salt,
                                                  int iters, int dklen);
   ```
   with the existing three-argument function kept as its `dkLen = 32`
   wrapper. tep's `Tep::Password` and `test/sp_crypto_basic.rb` both call
   the three-argument form and keep working as-is.

2. **Change the existing arity** to take `dklen`, keeping the surface at
   seven functions. Smaller surface, but it breaks tep's password hashing
   at the call site.

A PR implementing option 1 is attached — the block loop is RFC 8018 §5.2 as
written, clamped to `[1, 64]`, +77/-34 including the test. Happy to redo it
as option 2 if the smaller surface is worth the call-site churn.

rubys:
Closes #3769 (option 1 from that issue — happy to redo as option 2).

`sp_crypto_pbkdf2_sha256_b64url` hardcodes `INT(1)` and copies a single
32-byte `T(1)`, so 32 bytes is the most it can produce. Rails derives its
signed-cookie and signed-id keys at dkLen 64
(`ActiveSupport::KeyGenerator#generate_key(salt, key_size = 64)`), and the
second block isn't derivable from the outside in practice: `KeyGenerator`
defaults to `2**16` iterations, so that inner loop has to stay in C.

Adds `sp_crypto_pbkdf2_sha256_b64url_len(password, salt, iters, dklen)`,
clamped to `[1, 64]`. The existing three-argument function stays as its
`dkLen = 32` wrapper, so tep's `Tep::Password` and the current test call
the same shape they always did — verified unchanged, same 43-char digest.

The block loop is RFC 8018 §5.2 as written (blocks differ only in the
`INT(i)` salt suffix). The base64url encode step moves into a small static
helper, since the tail handling now depends on a length.

## Verification

`test/sp_crypto_basic.rb` gains the dkLen=64 vector plus the two clamp
cases (0 → 32 bytes, 999 → 64). The expected values come from Ruby's
OpenSSL, which is the oracle Rails itself uses:

```ruby
OpenSSL::PKCS5.pbkdf2_hmac("password", "salt", 1, 64, OpenSSL::Digest::SHA256.new)
# => Eg-2z_z4syxD5yJSVsT4N6hlSMkszDVICAWYfLcL4XtNvzovPa0zdyZLt7joMw1O_HRRQYYX2r72g3NTYc3BjA
```

`make` is clean and the test's output matches `.expected` byte for byte.
`T(1)` of the 64-byte derivation is the same digest the 32-byte call
already returned, which is the property that says the two agree.

Context, in case it's useful for judging the shape: this comes out of
compiling a Rails app (Basecamp's ONCE Campfire) to a spinel binary, where
`cookies.signed[:session_token]` is the first thing the auth filter chain
reaches. Matching Rails' derivation is what lets one signed cookie work
against both the Rails reference and the compiled binary in a differential
test — and keeps `signed_id`s that already exist in URLs valid.


rubys:
Rebased onto master (was based on a02ffd70, now 1c3d9989).

The conflict was worth flagging rather than just resolving: 5b615738 ("Root a runtime helper's arguments across any call that can collect") had added `SP_GC_ROOT_STR(password)` to the one-block function, and my rewrite dropped it. It is restored on the new `_len` entry point, on the same reading your sweep made — `password` is read inside the derivation loop, after calls that can collect, while `salt` is copied into `salted` before any of them and never touched again, so it needs no root.

Rebuilt and re-verified after the rebase: `test/sp_crypto_basic.rb` matches `.expected` byte for byte, including the dkLen=64 vector.

rubys:
**Second commit: HMAC-SHA1, and a correction to how I framed #3769.**

I wrote that issue as though Rails' signed messages were SHA-256 throughout. That's true of the *key derivation* (`key_generator_hash_digest_class`, SHA-256 since `load_defaults 7.0`) but not of the *signature*. Rails uses two different digests:

| | key derivation | signature |
|---|---|---|
| `cookies.signed[:x]` | PBKDF2-HMAC-SHA256, salt `"signed cookie"`, 2**16 iters, **dkLen 64** | **HMAC-SHA1** |
| `signed_id` / `find_signed!` | same, salt `"active_record/signed_id"` | HMAC-SHA256 |

The cookie digest is `signed_cookie_digest || "SHA1"` (actionpack `cookies.rb:314`), and no `load_defaults` version sets `cookies_digest` — 8.2 included. So a program reading a Rails cookie has no choice of digest.

I know SHA-1 is deliberately not offered as a building block here, and the header's warning is right. I think this is the same case `sp_crypto_websocket_accept` already is: an existing protocol names the hash, and reproducing it isn't a new security design. The header comment keeps the warning, aimed at callers who *do* have a choice.

With both commits, sp_crypto reads a Rails cookie end to end. Against one generated by ActiveSupport 8.2:

```
payload  eyJfcmFpbHMiOnsibWVzc2FnZSI6Iklu...
rails    de112a9fdf80de03481c3b34baa0c0229ba0f9aa
spinel   de112a9fdf80de03481c3b34baa0c0229ba0f9aa
verdict  VALID
```

Happy to split this into its own PR if you'd rather judge the two asks separately — the dkLen fix stands alone and is needed for signed IDs regardless.

PR #3778, opened 2026-08-09
Makefile: Add COPT override + config.mak for fast -O0 builds

rofl0r:
Add COPT variable (default -O2) overridable via config.mak (gitignored). Replaces all 38 hardcoded -O2 in Makefile.
Updates common.mk OPT to respect COPT.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ packages/*/build/
 .claude/
 CLAUDE.local.md
 /local.mk
+/config.mk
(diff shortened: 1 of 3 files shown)


rofl0r:
CI errors inherited from current master.

rofl0r:
rebased on master to get the CI fix. also renamed config.mak to config.mk which adheres to the existing naming convention.

Issue #3518, opened 2026-08-04
--rbs: an Integer-declared ivar makes a loop-grown int local promote to Bigint (returns garbage) where concrete inference keeps it int


OriPekelman:
`--rbs` seeding an ivar as `Integer` makes a **loop-grown int local promote to Bigint**, which then returns through the method's `Integer` slot as a truncated pointer — where concrete inference (no `--rbs`) keeps it a plain int and computes correctly. Adding a *correct* type declaration makes the result wrong.

## Repro (15 lines + a 6-line sig)

`loopint.rb`:
```ruby
class Backoff
  def initialize
    @mult = 2
    @base = 100
  end
  def compute(n)
    d = @base
    i = 0
    while i < n
      d = d * @mult
      i = i + 1
    end
    d
  end
end
b = Backoff.new
puts b.compute(3)
```

`sig/backoff.rbs`:
```rbs
class Backoff
  @mult: Integer
  @base: Integer
  def initialize: () -> void
  def compute: (Integer n) -> Integer
end
```

```
$ spinel loopint.rb -o a && ./a
800                     # correct (== CRuby)

$ spinel loopint.rb --rbs sig -o b && ./b
loopint.rb:16:10: warning: returning 'sp_Bigint *' from a function with return type 'mrb_int' makes integer from pointer without a cast [-Wint-conversion]
189014993188208         # garbage: Bigint pointer truncated to int
```

Verified at master `85abddd2` (also on later masters).

## Why it matters

The concrete literal `@mult = 2` in `initialize` gives the analyzer enough precision to keep `d = d * @mult` an int. The `--rbs` declaration `@mult: Integer` is *less* precise (generic Integer, not the literal), and that widening flips the loop-grown `d` into Bigint promotion — so the RBS-declared build is strictly worse than the inferred one, for a correct declaration.

This blocks tep#199 (retiring our module-load type-seed block in favor of `--rbs sig`): `--rbs` re-pins uncalled methods' types beautifully, but for methods with a loop-grown int accumulator (our `RetryPolicy#backoff_for`, exponential retry backoff) it introduces this miscompile — the retry backoff silently returns garbage → ~0ms sleep. The call-seed didn't have this problem.

Ask: `--rbs`-seeded numeric types should be at least as precise as concrete inference for the int-vs-Bigint promotion decision — an `Integer`-typed local/ivar in a bounded multiply loop shouldn't promote to Bigint when the un-seeded build keeps it int.

matz:
[...]

PR #3780, opened 2026-08-10
Widen a pinned str-hash return whose body stayed poly

### LINKED ISSUE #3779: RBS-pinned Hash return on a class-level memoized ivar: segfault, or a poly value in a const char *
# RBS-pinned `Hash[K, V]` return on a class-level memoized ivar: segfault, or a poly value in a `const char *`

A `def self.x` that memoizes a hash in a class-level ivar compiles and runs
correctly on its own. Give it an RBS signature pinning the return to a
Hash with concrete element types, and the same code either segfaults or
fails to compile, depending on which way the memo is spelled.

Found reproducing Rails' signed cookies on top of #3770 — the derived-key
cache is exactly this shape. Master at `022c82ab`.

## Repro

```ruby
# memo.rb
module M
  def self.table
    @table ||= {}
  end
end

M.table["a"] = "1"
puts "size " + M.table.length.to_s
```

```ruby
# sig/memo.rbs
module M
  def self.table: () -> Hash[String, String]
end
```

```
$ spinel --rbs sig memo.rb -o memo && ./memo
memo.rb -> memo
$ echo $?
139
```

Without the signature it is fine:

```
$ spinel memo.rb -o memo && ./memo
size 1
```

Before the crash the value reads as garbage — a `length` of `4312194560`
in the program this came from, which looks like a pointer arriving where
an Integer was expected.

## The other spelling: a type error instead

Same signature, memo written as an explicit nil-guard:

```ruby
module M
  def self.table
    @table = {} if @table.nil?
    @table
  end
end

M.table["a"] = "1"
puts "read " + M.table["a"]
```

```
$ spinel --rbs sig memo2.rb -o memo2
memo2.rb:9:212: error: initializing 'const char *' with an expression of incompatible type 'sp_RbVal'
1 error generated.
spinel: C compilation failed
```

The read lowers to `sp_poly_get_str(...)`, i.e. the ivar is a poly hash in
the emit while the signature says `Hash[String, String]` — so the pin
isn't reaching the ivar, and the two spellings differ only in whether the
mismatch shows up at compile time or as a bad value at run time.

## What narrows it

| memo | RBS return | result |
|---|---|---|
| `@t \|\|= {}` | `Hash[String, String]` | **segfault (139)** |
| `@t = {} if @t.nil?` | `Hash[String, String]` | **C compile error** (`sp_RbVal` → `const char *`) |
| `@t \|\|= {}` | `Hash[untyped, untyped]` | ok |
| `@t \|\|= "x"` | `String` | ok |
| `@t \|\|= {}` | *(no RBS)* | ok |

So it needs all three: a class-level ivar, an empty-hash literal memo, and
an RBS return pinned to a Hash with concrete element types. A `String`
memo with a `String` pin is fine, and the same hash memo with an
`untyped`-element pin is fine.

## Workaround

Keep the memo in String-typed slots and pin nothing to a Hash, or drop the
signature for that reader and let inference have it. Both cost the typed
surface the signature was there to provide.

rubys:
Closes #3779.

The arm above this one already handles the mismatch from the other side — a
body that built the *narrower* kind returned through a wider declared one —
and the bignum arm below it handles the same shape for `Integer` (#3518,
"declaring the type correctly made the program worse than not declaring it at
all"). This is the third: a body that stayed **poly** under a pinned
`Hash[String, String]`.

A class-level ivar memoizing an empty hash never narrows. Its only assignment
is `{}`, and every write that would give it element types goes through the
value the reader hands back:

```ruby
module Registry
  def self.table
    @table ||= {}
  end
end
Registry.table["a"] = "1"
```

So the body is an `sp_RbVal` while the signature says string-valued, and the
two don't share a layout. Returning it through a `StrStrHash*` reinterprets
the pointer: the `||=` spelling above crashed on the first read, and the
equivalent nil-guard spelling (`@t = {} if @t.nil?`) didn't compile at all —
`sp_poly_get_str` returning `sp_RbVal` into a `const char *`.

## The change

One `else if` in the existing guard, widening the return to what the body
actually is. The caller loses a typed read it never had.

Only `Hash[String, String]` reproduces — `Hash[String, Integer]` and
`Hash[Integer, Integer]` both come out fine — but the arm sits inside the
existing str-keyed condition, so both kinds behave alike rather than diverging
on a distinction the reporter can't see.

## Test

`test/rbs-seed/memoized_ivar_hash_ret.rb` covers both spellings (`||=` and the
nil-guard), reads back through the hash, and iterates it, so a future
regression shows up as wrong output rather than only as a crash. It fails on
master (`C did not compile`) and passes with the fix. Full suite: 2783 pass,
0 fail, 0 error.

## Where it came from

Reproducing Rails' signed cookies on top of #3770 — the derived-key cache is
exactly this shape (an expensive PBKDF2 result memoized behind a typed
reader). Worth noting it is **not** a regression from that PR: the repro fails
identically at `a02ffd70` (the commit #3770 branched from) and at `27a500f2`
(2026-07-15).


rubys:
Superseded by 708ffbe9 — thank you, and yours is the better fix.

I widened the return to poly, which resolves the layout mismatch by giving up the declared type; pinning the ivar the body answers keeps `Hash[String, String]` actually meaning that at the read, and covers arrays as well. The containers-only guard and the `g_seed_nilable` exclusion are both cases I had not thought about — a scalar slot pinned that way losing the nil a boxed reader answers is exactly the kind of thing my version would have papered over.

Confirmed against my repro and the runtime code it came from.

Issue #3388, opened 2026-07-26
method-local `||=` loses its assigned String in Spinel


gurgeous:
```rb
class Crash
  def to_s
    text ||= ["Usage:", "foo"].join(" ")
  end
end
raise "FAIL" if !Crash.new.to_s.include?("Usage:")
```

matz:
Two separate defects, both in your one-line method.

The `||=` never assigned. The emitter treated every local that was not poly, bool or symbol as already truthy, so it dropped the assignment and emitted a bare read of a variable nothing had ever written. That is only sound for a slot with no nil representation; `const char *` is declared NULL exactly so an unassigned one reads as nil. Same for every array, hash and object local, and `mrb_int` / `mrb_float` carry a sentinel for the same purpose. A local with no definite assignment anywhere now starts at that sentinel rather than the zero value, matching what block-locals already did, and the or-write tests it.

The assignment is emitted through the same path a plain `x = v` uses, so it picks up the same coercions, with its statement-shaped setup captured locally. Routed to the enclosing statement prelude it would run outside the test, which would drag the right-hand side's side effects along with it.

The second defect is why fixing the first alone would still have given you `nil`. A tail local write returned the return type's default rather than the value just assigned, so `def m; x = v; end` answered `""` for a String method and `0` for an Integer one, whether or not `||=` was involved. Emitting the write as a statement and then handing back the slot keeps the value without routing the write through the value path, which is what perturbs nullable-scalar inference.

Fixed in 2b6b1778.

Issue #3317, opened 2026-07-24
Array#concat through a reader misinfers a splat-derived Array


gurgeous:
```rb
class Config
  def str(*opts) = add_flag(Flag.new(opts))
  def prepare! = @help_flag = add_flag(Flag.new(["x"]))
  def add_flag(flag) = flag
end

class Flag
  attr_reader :switches
  def initialize(opts)
    @switches = []
    switches.concat(opts)
  end
end

flag = Config.new.str("y")

raise "FAIL" unless flag.switches == ["y"]
```

```
spinel_63.rb: In function ‘sp_Flag_initialize’:
spinel_63.rb:16:65: error: incompatible types when initializing type ‘sp_PolyArray *’ using type ‘sp_RbVal’
   16 |     switches.concat(opts)
      |                                                                 ^
spinel: C compilation failed
```

Issue #3203, opened 2026-07-21
rbs-seeded self-class return on an overridden class method: C decl returns sp_Base * while the poly-dispatch call site unboxes .v.p


rubys:
Gap: with `--rbs` seeds, a class method whose sig returns its own class (`-> Base`) gets a C declaration returning the concrete struct pointer, while a call site reached through poly dispatch still unboxes the result as a boxed value (`.v.p`) — the generated C fails to compile.

## Repro

`repro.rb`:

```ruby
class Base
  def self.instantiate(_row)
  end

  def self.find_by(conditions)
    instantiate(conditions)
  end
end

class Sub < Base
  def self.instantiate(row)
  end
end

Model.find_by(1)
```

`sig/repro.rbs`:

```rbs
class Base
  def self.instantiate: (Hash[String, untyped] _row) -> Base
  def self.find_by: (Hash[Symbol, untyped] conditions) -> Base?
end
```

```
$ spinel repro.rb --rbs sig -o repro.bin
repro.rb:6:59: error: member reference type 'sp_Base *' (aka 'struct sp_Base_s *') is a pointer; did you mean to use '->'?
    6 |   return (sp_Base *)(sp_Base_s_instantiate(lv_conditions)).v.p;
      |                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
      |                                                           ->
repro.rb:6:60: error: no member named 'v' in 'struct sp_Base_s'
spinel: C compilation failed
```

Current master (caa2426); reproduces unchanged at least back to 395dfe2, so it is not a recent regression. Without `--rbs sig` the program compiles.

Three ingredients are load-bearing — removing any one of them compiles cleanly:

1. the `-> Base?` sig on `find_by` (drop that line → clean);
2. the subclass override of `instantiate` (`class Sub < Base; end` → clean);
3. the unresolved-constant receiver at the call site — `Sub.find_by(1)` compiles, `Model.find_by(1)` fails.

The emitted line shows the disagreement directly: `sp_Base_s_instantiate`'s C return type is the seeded concrete `sp_Base *`, while `find_by`'s tail call still wraps it in the boxed-value unwrap `(...).v.p`.

## Why this started biting Rails-shaped trees on 2026-07-21

[...]

Issue #1365, opened 2026-06-07
Codegen: setter-style `@cache = list` returns the assigned pointer from an mrb_int-inferred method (empty-array ivar reassigned to object array)


rubys:
## Summary

A setter-style method whose body is a tail assignment `@ivar = list` returns the assigned **pointer** value from a function whose return type is inferred `mrb_int`. The ivar was seeded by an empty-array literal `@ivar = []` (which types it as an array), then reassigned an array of objects → fails C compile.

```
warning: incompatible pointer types assigning to 'sp_PolyArray *' from 'sp_PtrArray *'
error:   incompatible pointer to integer conversion returning 'sp_PolyArray *' from a function with result type 'mrb_int' (aka 'long')
```

## Minimal repro

```ruby
class Comment; end
class Holder
  def initialize = (@cache = [])               # empty-array literal seeds @cache as an array
  def preload(list) = (@cache = list)          # tail assignment returns the pointer...
  def cache = @cache
end
h = Holder.new
h.preload([Comment.new, Comment.new])
puts h.cache.length
```

- **CRuby:** prints `2` (works).
- **spinel `bab584f`:** `spinel repro.rb -o /tmp/x` → C compilation failed.

## Root cause (from generated C)

```c
static inline mrb_int       sp_Holder_preload(sp_Holder *self, sp_PtrArray * lv_list);
static inline sp_PolyArray *sp_Holder_cache(sp_Holder *self);
...
self->iv_cache = sp_PolyArray_new();                 // seeded as sp_PolyArray*
...
static inline mrb_int sp_Holder_preload(sp_Holder *self, sp_PtrArray * lv_list) {
  ...
  return (self->iv_cache = lv_list);                 // sp_PolyArray* returned from mrb_int fn
}
```

`preload`'s return type is inferred `mrb_int` (it's a setter with no meaningful return), but its body's tail expression is the assignment `@cache = list`, whose value is the array pointer. The return type should follow the tail-expression value (or setter-style tail assignments shouldn't return through the `mrb_int` path). There's also a secondary `sp_PtrArray*` (param) vs `sp_PolyArray*` (slot) mismatch on the same line.

## Notes

- Surfaces in roundhouse's real-blog `Article#_preload_comments` (`@comments_cache = []` in `initialize`, `@comments_cache = list` in the preload), where it presents as `sp_IntArray* <- mrb_int`.
[...]

PR #3791, opened 2026-08-12
Return the assigned value from a plain tail ivar write

tobi:
## What

A method whose last expression is `@x = v` answers `v` in Ruby. Spinel returned the statement default instead, which is the slot type's nil: `NULL` for an Array — so the caller segfaults on the first `.length` — `0` for an Integer, `""` for a String, `{}` for a Hash.

```ruby
class Memo
  def initialize; @cache = nil; end
  def build
    return @cache if @cache
    @cache = [1, 2, 3]     # answered NULL; caller segfaulted
  end
end
puts Memo.new.build.length   # CRuby: 3    Spinel: Segmentation fault
```

This is the ordinary Ruby memoization shape, so it is easy to hit, and it fails as a crash rather than a compile error.

## Why it was still reachable after #3317

#3317 fixed the **object**-valued case by routing it through the value path, guarded on `ty_is_object(...)`. That predicate covers user-class instances only, so Array, Hash, String, Integer and Float tail writes all still fell through to the statement default.

`def m = (@x = v)` parses as a value, so the **endless** form was already correct — which is why #1365's repro passes today. The ordinary statement form `def m; @x = v; end` was not.

## The change

Route it the way the tail **local** write immediately below it is routed (#3388) rather than through the value path: emitting the write itself as a value is what perturbs the nullable-scalar inference the surrounding comment guards. Emit the write as a statement, then hand back the slot — same value, evaluated once, no inference change — and only when the slot's type reaches the return slot without a conversion this text form cannot express.

Guarded to the plain instance-slot rendering (`self->iv_x`). The class-level, `instance_eval` and toplevel-global renderings each have their own slot text, so they stay on the statement path.

## Test

`test/ivar_tail_write_return.rb` covers Integer / String / Array / Hash / Float tail writes, the value being both returned *and* stored, the memoization guard that segfaulted, a value computed into a local, a value arriving as a parameter, the endless form that already worked, and that a side-effecting value expression is still evaluated exactly once.

On unpatched master the compiled binary segfaults at the first assertion.

## Verification

- Full corpus: **2814 pass / 3 fail / 7 err** of 2824.
- The same 10 non-passing tests fail identically on clean master (`promote_*` bigint, thread/fiber/monitor) — zero regressions. Verified by stashing this patch, rebuilding, and re-running them.
- `rbs-seed-test` also fails on clean master (#3203), unrelated to this change. Note it aborts `make test` before the corpus runs, so the corpus was driven via the per-test `build/test-results/*.ok` targets.

## Context

Found while porting [tobi/try](https://github.com/tobi/try) — a ~1300-line Ruby CLI + fuzzy-matching TUI — to Spinel, under a deliberately strict constraint: **one source tree that runs unchanged under both CRuby and Spinel**, no `RUBY_ENGINE` branching, byte-identical output. That makes a good compiler test, because every divergence has to be either a real bug or a documented limitation — there is nowhere to hide an engine-specific patch.

The port works: a single 900KB native binary, byte-identical output including a 42KB ANSI pty transcript, and 62x faster startup (496µs vs 31ms, hyperfine). This bug bit it twice, in the two places that memoize.

Refs #3317, #3388, #1365


matz:
Merged, thank you. The reasoning for routing it like the tail *local* write rather than through the value path is the part I want to highlight: emitting the write as a value is what perturbs the nullable-scalar inference, and this shape does not need to. Statement, then the slot — same value, once.

The guard set is right too: the class-level, instance_eval and toplevel-global renderings each spell their slot differently, and only the plain `self->iv_x` form can be handed back as text like this.

Segfaulting on the ordinary memoization shape was the worst kind of failure we had left in that family.

PR #3792, opened 2026-08-12
Add a failing test for block.call with multiple arguments dropping all but the first

tobi:
## What

A block invoked through `block.call(a, b)` with **multiple arguments** drops all but the first when a collector takes the block's value as one object. An Enumerable whose `#each` forwards the caller's block through a `&block` parameter therefore collects scalars where CRuby collects pairs.

```ruby
class ViaCall
  include Enumerable
  def each(&block)
    block.call(1, 2)
    block.call(3, 4)
  end
end

ViaCall.new.to_a    # CRuby: [[1, 2], [3, 4]]    Spinel: [1, 3]
```

The `yield a, b` spelling of the same method is already correct, and the test asserts both side by side so the divergence is visible in one file. A single-argument `block.call(1)` still passes its value through unwrapped, which the test also pins.

`block_autosplat.rb` covers the receiving side (an array element splatting across `|a, b|`); this is the emitting side.

## Why no fix

I did not want to guess at the block-argument packing path. Happy to take a pointer to the right site and follow up with a patch.

## Related

`first` on the same class does not compile at all (`error: cannot convert to a pointer type`) — filed separately, since the symptom and likely fix site differ.

## Context

try's fuzzy matcher yields `(entry, positions, score)` from an `#each` that forwards the caller's block, so `.to_a` silently dropped the positions and the score — match results came back as bare entries with no scores and no highlights, with no error anywhere.

The workaround, for anyone hitting this: yield an array literal **at the yield site** (`rows.each { |row| yield [a, b, c] }`). Note that switching to `block.call(a, b, c)` fixes `each` but still breaks `to_a`, so the literal-yield form is the only spelling where both match CRuby.


matz:
Merged with the fix (be8a644d), so the test is green on master.

The collector is sized from the widest hand-over `each` makes, and that sizing scanned `yield` sites only — so your spelling, which never yields, got a one-parameter collector and each element arrived as its first value alone. It now counts `block.call(...)` on the method's own block parameter too.

Thank you for asserting the `yield` spelling beside it in the same file: that is what made the shape of the bug obvious rather than the symptom.

PR #3793, opened 2026-08-12
Add a failing test: first on a multi-arg block.call Enumerable does not compile

tobi:
## What

`first` on an Enumerable whose `#each` invokes the block through `block.call(a, b)` with **multiple arguments** fails to build:

```
error: cannot convert to a pointer type
```

```ruby
class ViaCall
  include Enumerable
  def each(&block)
    block.call(1, 2)
    block.call(3, 4)
  end
end

ViaCall.new.first    # CRuby: [1, 2]    Spinel: does not compile
```

`first` takes the block's value as one object, so the multi-argument call has to pack its arguments into an array — the same packing `to_a` needs. In the companion PR (`block_call_multi_arg_pack`) `to_a` compiles but silently answers only the first argument; here the mismatch reaches the C compiler instead, so no binary is produced at all.

The `yield a, b` spelling of the same method compiles and answers `[1, 2]`; the test asserts both side by side.

## Why filed separately

Same root cause as the `to_a` PR, but a different symptom (build failure vs. silent wrong value) and probably a different fix site. Splitting them so neither blocks the other; they can be merged as one change if that turns out to be simpler.

## Context

Found while porting [tobi/try](https://github.com/tobi/try) — a ~1300-line Ruby CLI + fuzzy-matching TUI — to Spinel, under a deliberately strict constraint: **one source tree that runs unchanged under both CRuby and Spinel**, no `RUBY_ENGINE` branching, byte-identical output. try's fuzzy matcher forwards the caller's block from `#each` and yields `(entry, positions, score)`.


matz:
Merged with the fix, so this test is green on master.

Two things were wrong under it, and your two PRs separated them exactly right. The collector was sized from `yield` sites only (be8a644d, your #3792). This one was the typing: `block.call(...)` answers the call-site block's value, and a yield-inlined method takes the *first* concrete one because it is specialized per site — but here two sites drive the same `each` (Enumerable's collector and its generator) and their blocks answer different types, so one site's text was boxed against the other's type and the C compiler was handed an sp_RbVal cast to a pointer. When the sites disagree the value is only known at run time.

That is why one PR silently answered the wrong thing and the other did not build at all: same root, two consumers.

PR #3794, opened 2026-08-12
Add a failing test for [] defined in a Struct/Data block being ignored

tobi:
## What

A method defined in a `Struct.new` / `Data.define` block overrides the built-in member accessor of the same name in CRuby. `[]` written in the block is ignored: the call goes to the generated member lookup, so a key that is not a member raises instead of running the user's body.

```ruby
Item = Data.define(:data, :score) do
  def [](key)
    key == :score ? score : data[key]
  end
end

i = Item.new(data: { text: "hello" }, score: 1.5)
i[:score]   # 1.5 — happens to agree, it IS a member
i[:text]    # CRuby: "hello"    Spinel: NameError: no member 'text' in struct
```

The same applies to any built-in the block redefines (`to_s`, `==`, ...); `[]` is just where it bit us.

## What the test pins

Both directions, so a fix cannot simply drop the generated accessor:

- the override **is** called, for a member key, a non-member key, and via `Struct.new` as well as `Data.define`;
- the generated readers still work alongside it;
- a Struct that does **not** override `[]` keeps the built-in member lookup — including the integer-offset form, the negative-offset form, and the `NameError` for an unknown name.

## Context

try wraps a Hash payload in a Data and defines `[]` to forward unknown keys into that Hash — a pretty common shape for "struct with a bag of extra attributes". The workaround in the port was to rename the method (`fetch`) and dispatch each key to a named reader, which changes every call site.

### DIFF
--- a/test/struct_block_overrides_bracket.rb
+++ b/test/struct_block_overrides_bracket.rb
@@ -1,0 +1,55 @@
+# A method defined in a Struct.new / Data.define block overrides the built-in
+# member accessor of the same name, as it does in CRuby. `[]` written in the
+# block was ignored: the call went to the generated member lookup, so a key
+# that is not a member raised
+#
+#   NameError: no member 'text' in struct
+#
+# instead of running the user's body. The same applies to any built-in the
+# block redefines (`to_s`, `==`, ...).
+#
+# Found porting tobi/try, which wraps a Hash payload in a Data and defines
+# `[]` to forward unknown keys into that Hash.
+Item = Data.define(:data, :score) do
+  def [](key)
+    if key == :score
+      score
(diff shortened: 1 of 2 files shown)


matz:
Merged with the fix, green on master.

Both sides had to give way, which your test caught: declining the generated surface in the emitter alone made the user's `[]` run but kept typing the call from the member, so the value landed in the wrong slot and the C compiler rejected it. The inference declines the same way now.

Pinning both directions was the right instinct — the Struct that does *not* override `[]` keeps the built-in lookup, including the integer and negative offsets and the NameError, so the fix could not be "drop the generated accessor".

PR #3795, opened 2026-08-12
Add a failing test: module included into a class and at toplevel does not compile

tobi:
## What

A module included **both** into a class and at toplevel is compiled twice, and the two instantiations do not agree on the return type of a method that constructs an object. One degrades to `int`, so the emitted C does not build:

```
error: incompatible types when returning type 'int' but 'sp_Box' {aka 'struct sp_Box_s'} was expected
error: request for member 'iv_ch' in something not a structure or union
```

```ruby
class Box
  attr_reader :ch
  def initialize(ch); @ch = ch; end
end

module Helpers
  def fill(ch); Box.new(ch); end
end

class Screen
  include Helpers          # included into a class...
  def go; fill("="); end
end

include Helpers            # ...AND at toplevel

puts Screen.new.go.ch
puts fill("-").ch
```

Either include **alone** is fine — the class include without the toplevel one, or the toplevel one without the class — and both compile and run. It is the combination that breaks. The test covers two separate modules so it is clear the failure is not specific to one method name.

## Context

try's TUI helper module is mixed into the screen classes and also included at toplevel so the script body can call the same helpers — which is exactly the shape above. The workaround in the port was `module_function` plus qualified calls, which changes how every call site reads.

One correction worth recording, since I got it wrong in my own notes first: I initially believed the trigger was a **default argument** on the module method (`def fill(char = " ")`). It is not. The failure reproduces identically with a required parameter, and this test uses the simpler form.


matz:
Merged with the fix, green on master.

Two things were wrong under one symptom, which is why the two instantiations disagreed:

- a receiverless call *inside the class* reached for the top-level copy — the module's own null-receiver function — rather than the class's transplanted one, whose self and inferred types are what that body was typed against;
- and the module's own function was not emitted at all. Including a module into a class marks the source as copied away, which is right until a top-level include needs to call it directly.

Your repro having both an `include` in a class and one at the top level in the same file is what made both visible at once; either alone compiles.

PR #3796, opened 2026-08-12
Sequence only an argument that has a value

rubys:
`f1edaed6` ("Evaluate a method call's arguments left to right") types each sequencing temp from the argument's own type. An argument with **no value** — an unresolved call, which lowers to `sp_raise_nomethod` — types as void, so the hoist emits `void _tN = ...;`.

## Repro

```ruby
def join(a, b)
  a.to_s + b
end

key = "sid"
puts join(nil.missing_method, "cookie." + key)
```

```
$ spinel t.rb -o /tmp/t
t.rb:6:8: error: variable has incomplete type 'void'
    6 |   void _t3 = sp_raise_nomethod(sp_stage_args_msg("undefined method 'missing_method' for nil", 0, (sp_RbVal[]){sp_box_nil()}));
      |        ^
1 error generated.
spinel: C compilation failed
```

Clean on `489cbde7` (= `f1edaed6^`). It takes **two** side-effecting arguments with the valueless one **not last** to reach the hoist, which is why it took a while to show up.

Before, such an argument was never hoisted (neither `TY_POLY` nor `needs_root`), and the argument slot coerced it to the parameter's type:

```c
const char * _t527 = sp_poly_to_s(sp_raise_nomethod(...));   /* 489cbde7 */
void         _t522 =              sp_raise_nomethod(...);    /* f1edaed6 */
```

## Fix

Skip sequencing an argument that has no C storage. There is nothing to sequence: the expression does not return, so no sibling can observe it — leave it to the argument slot, which coerces it as before.

The gate asks about storage rather than testing `TY_VOID`: an unresolved call types `TY_UNKNOWN`, and `emit_ctype` spells both `void`.

## Verification

- `make test` — **2868 pass, 0 fail, 0 error** (2867 before, +1 for the new `test/call_arg_valueless.rb`)
- the repro compiles and raises the right `NoMethodError`
- roundhouse's `toolchain-spinel` job, red on `9a6a9e1f`, is green again

Found from roundhouse's emitted `cookies.rb`:

```ruby
ActionController::MessageVerifier.verified(
  Rails.application.secret_key_base,          # valueless, and first
  ActionController::MessageVerifier::SIGNED_COOKIE_SALT,
  raw, "cookie." + key.to_s, true             # second side-effecting arg
)
```

Filing separately: `c357050d`'s `_sp_cls` is emitted as a bare identifier inside a lifted proc, which has no such parameter — that one breaks our lobsters AOT build and the fix looks like it belongs in the capture machinery rather than at the call site.


matz:
Merged, thank you — and for tracing it to the commit and explaining why it needed two side-effecting arguments with the valueless one not last. That is exactly the shape my own test missed.

Gating on storage rather than on `TY_VOID` is the right call: the unresolved-call token types `TY_UNKNOWN`, and both spell `void`, so a type test would have left the second half of the hole open.

PR #3798, opened 2026-08-12
add cc version to --version command

kostya:


matz:
Merged, thank you. A report that turns out to be a codegen/compiler interaction was always missing half of what it needed, and this is the cheapest place to put the other half.

Two small things on top, in the merge:

- the compiler's own stderr is dropped, so `--cc=` naming something that cannot be run prints the plain revision line instead of a shell error above it;
- `head -n 1` is gone — `fgets` already takes one line, and that keeps the spawn to just the compiler.

`spin` reads the revision as the second field and is unaffected.

PR #3889, opened 2026-08-13
Embed inline C fragments with ffi_source

SamSaffron:
Accept compile-time string fragments from FFI modules and emit them before
generated function bodies. This lets small native adapters travel in a
single Ruby source file while preserving normal FFI declarations and type
checking.

Document the directive, its supported string forms, and its portability
and trust implications.


SamSaffron:
This enables: 

https://gist.github.com/SamSaffron/5499fc91eb3508859ddc0f897b179c42

matz:
Merged. Verified locally: the example builds and answers 42, two fragments in one module compose, a non-literal argument gives the documented compile-time diagnostic, and the generated C is byte-identical to master's across all 2934 existing tests (the feature only adds emission when a program uses it). `make gate` green under gcc and clang.

Issue #3781, opened 2026-08-10
Trying to compile LangArena benchmark


kostya:
in my benchmark: https://github.com/kostya/LangArena

got many errors: 

script is here: https://github.com/kostya/LangArena/blob/master/ruby/main.rb

how to run
```
git clone https://github.com/kostya/LangArena.git
cd LangArena/ruby
spinel main.rb
```

```
spinel: main.rb:360: unsupported call: node 1329 (CallNode `bench_name`) recv=CallNode/ty48 argc=0
```

```ruby
  class Array < Benchmark
    def initialize
      @program_text = Helper.config_s(self.class.bench_name, "program")

```

Current results, It would be very interesting to know the results of Spinel:
| Ruby | Mode | Time,s | Memory, Mb |
|:------------|:-------|---------------:|-------------:|
| Ruby/CRuby2 | Default | 3551 | 157 |
| Ruby/CRuby2 | JIT | 3092 | 168 |
| Ruby/CRuby3 | Default | 3834 | 121 |
| Ruby/CRuby3 | RJIT | 3825 | 121 |
| Ruby/CRuby3 | YJIT | 2269 | 121 |
| Ruby/CRuby4 | Default | 4180 | 123 |
| Ruby/CRuby4 | YJIT | 2480 | 125 |
| Ruby/CRuby4 | ZJIT | 2987 | 122 |
| Ruby/Truffle | Native | 957 | 880 |
| Ruby/Truffle | JVM | 494 | 1632 |
| Ruby/JRuby | Default | 3097 | 1564 |
| Crystal | Release | 75 | 54 |

matz:
Thanks for the report and the benchmark, it is a good stress test.

Three fixes are on master now:

- `module M; class Array < Benchmark` registered under the bare leaf name `Array`, so every builtin-name test took it for the builtin and `M::Array.new` built a PolyArray. Such a class is now qualified the way a leaf name colliding with another user class already was; that clears the `bench_name` reject at main.rb:360.
- `==` on an operand the type fixpoint could not settle was refused at compile time; it now compiles to the runtime comparison (main.rb:779).
- The same for arithmetic on an untyped operand (main.rb:1076).

The next thing it stops on is `synapse.weight += ...` at main.rb:1104 -- an attribute op-assign through an object whose type is not resolved. I am working through the remaining ones; the file is long, so I will keep pushing fixes as each shape is cleared rather than waiting for the whole benchmark to build.

matz:
[...]

Issue #3501, opened 2026-08-02
[Runtime] A user-defined binary operator other than `+`, `-`, `*` is not called when the receiver is a block parameter


KieranP:
> (Disclaimer: I asked Claude Opus 5 (high) to find Spinel compile/runtime issues with documented Ruby classes and methods. It found the following issue, which I manually verified for accuracy before publishing this report)

_Reproduced with `spinel b4c324ba55b6` vs `ruby 4.0.3`._

When a user object reaches an operator through a block parameter, only `+`, `-`
and `*` dispatch to the class's own method. `/`, `%`, `&`, `|`, `^` return a
built-in integer result instead, `**` returns a `Float`, the comparison
operators raise `ArgumentError`, `<=>` prints nothing, and `==` answers
`false`. The same expressions on the same object held in a plain local are all
correct, and a same-shaped named method (`#div`) is correct in either position.
Spinel emits no diagnostic for any of these.

## Repro

```ruby
class Money
  attr_reader :cents
  def initialize(cents); @cents = cents; end
  def /(n); Money.new(@cents / n); end
  def %(n); Money.new(@cents % n); end
  def <(other); @cents < other.cents; end
  def to_s; "$#{@cents}"; end
end
[Money.new(250)].each do |m|
  puts (m / 5).to_s
  puts (m % 7).to_s
  puts (m < Money.new(300))
end
```

Ruby 4:

```
$50
$5
true
```

Spinel — the first two lines are a built-in integer result, and the third
raises:

```
0
0
```

```
comparison of Money with Money failed (ArgumentError)
```

## Every binary operator, boxed vs unboxed receiver

Each row is the same one-operator class, with the operator body
`def OP(n); "Q#{OP}#{n}"; end`, invoked as `puts(e OP 4)` — once with `e` a
block parameter (`[Q.new(3)].each { |e| ... }`) and once with `e` a plain local
(`e = Q.new(3)`). Ruby prints `Q<op>4` in every cell.

| op | receiver is a block parameter | receiver is a plain local |
|---|---|---|
| `+` `-` `*` | correct | correct |
| `/` `%` `&` | `0` | correct |
| `\|` `^` | `4` | correct |
| `**` | `0.0` | correct |
| `<` `>` `<=` `>=` | `ArgumentError: comparison of Q with 4 failed` | correct |
| `<=>` | prints nothing, no error | correct |
| `==` | `false` | `true` |
| `<<` `>>` | C build aborts — see 20260802215238478 | correct |

[...]

PR #3898, opened 2026-08-13
Ask the write index, not the whole table, for a slot's array evidence

rubys:
`recv_has_array_write` (src/analyze_pass.c) answers "does this receiver's slot have a
write that is definitely an array?" by scanning the whole node table — once per `<<`
call site, once per fixpoint iteration. Its reader arm (`obj.list << x`, #3781) crosses
every class with every node, so each such site costs O(classes × nodes) per iteration.

Three functions above it, `recv_has_scalar_numeric_write` asks the same question about
numeric writes and already goes through the name-keyed write index (`wrn_build`). This
PR points `recv_has_array_write` at that same index:

* **ivar/local arm** — walk the `(name, scope)` bucket instead of the table; identical
  query shape to the numeric sibling.
* **reader arm** — drive from the `@<gname>` write bucket (the small side of the join)
  and keep the per-class `comp_reader_in_chain` walk for the "which class defines this
  reader" test, since alias resolution is per-class.

No behavior change intended, and none observed — see below.

### Measurement

On the Roundhouse-emitted lobsters tree from #3115 (297 files, 44k lines of plain Ruby),
macOS/arm64, `spinel main.rb --emit-rbs`:

| | analyze |
|---|---|
| master `79249df8` | 45.5s |
| this PR | **20.3s** |

(A second machine-state pair on the same two commits: 40.1s → 18.3s; the ratio holds
at ~2.2× either way.)

A `sample` profile of master puts 41% of all samples on this helper's call site
(`analyze_pass.c:1902`); after the change it does not appear. Per-pass timing agrees:
`infer_write_types`, which hosts the call, was 15.0s of the 26.1s fixpoint.

The fixpoint converges in the same 17 iterations before and after, so this is pass cost,
not iteration count.

### Equivalence

On that tree, `spinel main.rb -c --rbs .` produces **byte-identical C** before and after
(121,761 lines), and `--emit-rbs` produces byte-identical signatures (6,798 lines).

`make check OPT=-O1`: 2941 pass, 0 fail. `infer-test` reports 7 failures on this branch
and the same 7 on clean `79249df8` — they are grep drift, not inference drift (the greps
want `static mrb_int sp_F_s_add(mrb_int, mrb_int)` and codegen now emits
`static inline __attribute__((always_inline)) mrb_int sp_F_s_add(mrb_int, mrb_int)`; the
inferred types are the ones the test asks for). The two tests that pin this helper's
semantics —
`user_operator_on_block_param.rb` (#3501/#3502) and `narrowed_element_local_pin.rb`
(#3781) — pass.

Refs #3115.

### DIFF
--- a/src/analyze_pass.c
+++ b/src/analyze_pass.c
@@ -107,20 +107,31 @@ static int recv_has_array_write(Compiler *c, int recv) {
     if (rargc != 0 || !gname) return 0;
     char ivn[300];
     snprintf(ivn, sizeof ivn, "@%s", gname);
-    for (int k = 0; k < c->nclasses; k++) {
-      int rdcls = k;
-      if (!comp_reader_in_chain(c, k, gname, &rdcls)) continue;
-      if (rdcls < 0 || rdcls >= c->nclasses) continue;
-      if (comp_ivar_index(&c->classes[rdcls], ivn) < 0) continue;
-      for (int id = 0; id < nt->count; id++) {
-        if (nt_kind(nt, id) != NK_InstanceVariableWriteNode) continue;
-        const char *wnm = nt_str(nt, id, "name");
-        if (!wnm || !sp_streq(wnm, ivn)) continue;
-        Scope *ws = comp_scope_of(c, id);
-        if (!ws || ws->class_id != rdcls) continue;
-        int v = nt_ref(nt, id, "value");
-        if (v < 0) continue;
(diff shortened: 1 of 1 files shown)


matz:
Merged. I verified it as a pure refactor rather than reproducing the timing: the generated C is byte-identical to master's across all 2934 tests and optcarrot-single, and `make gate` is green.

PR #3900, opened 2026-08-13
Memoize the "does any user class own this name?" question

rubys:
`an_user_defines_or_reads` (src/analyze_infer.c) answers "does any user class define
or read this name?" by crossing every class with the ancestor chain. `infer_call` asks
it for every poly-receiver call node, on every fixpoint iteration — on the
Roundhouse-emitted lobsters tree from #3115 that comes to **388.7 million**
`comp_method_in_chain` calls for a few thousand distinct answers.

Its sibling `an_user_defines_method` (src/analyze_util.c) already keeps a name-keyed
memo (`udm_`) stamped with (scope-index generation, `nscopes`, `nclasses`), on the
argument that the answer only moves when scope or class shape does. This PR gives
`an_user_defines_or_reads` the same memo, with the same stamp and the same
"not while unfrozen" guard — a rename leaves the counts unchanged, so outside the
frozen fixpoint the memo is neither consulted nor populated. The computation itself is
unchanged.

### Measurement

`spinel main.rb --emit-rbs` on that tree (297 files, 44k lines), macOS/arm64:

| | analyze |
|---|---|
| master `79249df8` | 45.5s → **39.2s** |
| with #3898 also applied | 18.1s → **14.8s** |

The second pair is the more interesting one: #3898 removes a full-table rescan that was
hiding this, and together they take the tree back under the 14.3s it analyzed in at
`b67a1266` (2026-07-22).

### Equivalence

Dumped signatures are byte-identical before and after in both configurations
(6,798 lines).

`make check OPT=-O1`: 2940 pass, 1 fail — `nilclass_bool_ops_conversions`. That test is
**flaky on clean master**: same commit, same command, four runs gave FAIL/FAIL/PASS/FAIL
(and FAIL/PASS on this branch). The emitted C for it is deterministic — two compiles of
the same source byte-match — so the flip is below codegen. Happy to open a separate issue
with the loop if it is not already known.

### One dead end, for the record

Memoizing `comp_method_in_chain` itself — the leaf that topped the profile — was worth
only 6%: at that call volume, hashing the name for the memo lookup becomes the cost.
The saving is in not asking the question, not in answering it faster.

Refs #3115.

### DIFF
--- a/src/analyze_infer.c
+++ b/src/analyze_infer.c
@@ -50,19 +50,74 @@ static int an_nonblock_no_exception(Compiler *c, int id) {
 static int an_builtin_only = 0;
 int an_builtin_only_p(void) { return an_builtin_only; }
 
+/* Name-keyed answer memo, the same shape (and the same staleness argument) as
+   udm_ in an_user_defines_method: the question below crosses every class with
+   the ancestor chain, and infer_call asks it for every poly-receiver call node
+   on every fixpoint iteration -- on a 40k-line program that is hundreds of
+   millions of chain walks for a few thousand distinct answers. The answer only
+   moves when scope or class shape does, which the (gen, nscopes, nclasses)
+   stamp tracks. */
+static char **udr_names = NULL;
+static signed char *udr_ans = NULL;
+static int *udr_next = NULL, *udr_head = NULL;
+static int udr_n = 0, udr_cap = 0;
+static unsigned udr_gen = (unsigned)-1;
+static int udr_nscopes = -1, udr_nclasses = -1;
+static unsigned udr_hash(const char *s) {
+  unsigned h = 2166136261u;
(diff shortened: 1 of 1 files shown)


rubys:
Follow-up on the `nilclass_bool_ops_conversions` note above: it is not flaky, it is uninitialized memory. Filed as #3901 with a two-line repro — `sp_poly_to_c_m` builds an `sp_Complex` without setting its `fl` int/float flag, so `-O0` and `-O1` disagree on whether `nil.to_c` prints `(0+0i)` or `(0.0+0i)`. Unrelated to this PR.

matz:
Merged. Verified as a pure refactor: the generated C is byte-identical to master's across all 2934 tests and optcarrot-single, and `make gate` is green.

Issue #3901, opened 2026-08-13
nil.to_c on a poly local renders as a float: sp_poly_to_c_m leaves sp_Complex.fl uninitialized


rubys:
`sp_poly_to_c_m` builds an `sp_Complex` without setting its `fl` member, so whether
`nil.to_c` renders as an integer or a float is decided by uninitialized stack memory.

### Repro

```ruby
# toc.rb
x = nil
p(x.to_c)
p(nil.to_c)
```

```
$ spinel -O 0 toc.rb -o toc0 && ./toc0
(0+0i)
(0+0i)
$ spinel -O 1 toc.rb -o toc1 && ./toc1
(0.0+0i)      <-- wrong
(0+0i)
$ ruby toc.rb
(0+0i)
(0+0i)
```

Stable across repeated runs of each binary; `-O2` and `-O3` behave like `-O1` here. The
emitted C is identical at every `-O` (the flag only reaches `cc`), so the difference is
in the C the runtime compiles, not in codegen.

### Cause

`sp_Complex` carries a per-component int/float flag (lib/sp_types.h:189):

```c
typedef struct{mrb_float re;mrb_float im;unsigned char fl;}sp_Complex;
#define SP_CPLX_RE_F 1
#define SP_CPLX_IM_F 2
```

`sp_complex_inspect` renders each component as a float exactly when its flag bit is set
(lib/sp_format.c:35, `c.fl & SP_CPLX_RE_F`).

`sp_poly_to_c_m` (lib/spinel_rt.h:6210) sets `re` and `im` but never `fl`:

```c
if (v.tag == SP_TAG_NIL) { sp_Complex z; z.re = 0; z.im = 0; return sp_box_complex(z); }
if (v.tag == SP_TAG_INT || v.tag == SP_TAG_FLT) {
  sp_Complex z; z.re = sp_poly_to_f(v); z.im = 0; return sp_box_complex(z);
}
```

`z.fl` is whatever the stack slot held. At `-O0` it happens to be zero, at `-O1`+ it is
not. The literal form takes a different path and is unaffected, which is why
`p(nil.to_c)` is right on the same line as `p(x.to_c)` being wrong — codegen folds it to
a compound literal, and that zero-fills `fl`:

```c
fputs(sp_complex_inspect((*(sp_Complex *)sp_poly_to_c_m(lv_x).v.p)), stdout);  /* x = nil */
[...]

PR #3903, opened 2026-08-13
Walk the write kinds through the kind index in lazy_alias_chain

rubys:
`lazy_alias_chain` asks "is this local a single plain write of a lazy chain?", and it
asked by scanning every node in the table and filtering five local-write kinds out of it
— once per lazy-receiver read, on every fixpoint iteration. That is exactly the loop
`nt_nodes_of_kind` documents itself as replacing, so it now walks the five kinds through
`NT_FOREACH_KIND`.

The answer does not depend on the order writes are visited in — it is "exactly one
matching write, of the plain kind, in the same scope" — so taking them kind by kind
gives what the linear scan gave.

### Measurement

On the Roundhouse-emitted lobsters tree from #3115 (297 files, 44k lines),
`spinel main.rb --emit-rbs`, macOS/arm64, alternating the two binaries within each pair:

| | baseline | this PR |
|---|---|---|
| master alone | 40.7 / 41.5s | 41.0 / 41.2s |
| with #3898 and #3900 | 14.9 / 15.1 / 15.5s | **13.4 / 13.4 / 13.6s** |

Worth saying plainly: **on master this is invisible.** The helper is ~1% of analyze until
the full-table rescan in #3898 and the missing memo in #3900 stop hiding it; after those
two it is ~11%. So this is a follow-on to those, not a standalone win.

Dumped signatures are byte-identical in both configurations (6,798 lines).
`make check OPT=-O1`: 2940 pass, 1 fail — `nilclass_bool_ops_conversions`, which is the
uninitialized `sp_Complex.fl` bug of #3901 (this branch is off master, which still has
it; #3902 fixes it). `infer-test`'s 7 failures are the same grep drift as on master,
noted in #3898.

### Two things I tried first

Memoizing the answer per node for the whole compile is worth about twice this, but it is
not sound: `nt_node_set_ref` / `nt_node_set_arr` do not bump `nt->version`, so a pass
that re-points a write's `value` edge would leave the memo stale.

Memoizing per fixpoint iteration instead (the `sp_narrow_memo_bump` generation) is sound
and measured no faster than the unmemoized original — within one iteration each read is
asked about roughly once, so all of the saving lives across iterations, where the cache
cannot safely go. Hence indexing the scan rather than caching its result.

Refs #3115.

### DIFF
--- a/src/analyze_util.c
+++ b/src/analyze_util.c
@@ -377,17 +377,25 @@ int lazy_alias_chain(Compiler *c, int var_read) {
   if (!vn) return -1;
   Scope *sc = comp_scope_of(c, var_read);
   int val = -1;
-  for (int w = 0; w < nt->count; w++) {
-    NodeKind k = nt_kind(nt, w);
-    if (k != NK_LocalVariableWriteNode && k != NK_LocalVariableOrWriteNode &&
-        k != NK_LocalVariableAndWriteNode && k != NK_LocalVariableOperatorWriteNode &&
-        k != NK_LocalVariableTargetNode)
-      continue;
-    const char *wn = nt_str(nt, w, "name");
-    if (!wn || !sp_streq(wn, vn)) continue;
-    if (k != NK_LocalVariableWriteNode || comp_scope_of(c, w) != sc || val >= 0)
-      return -1;   /* not a single plain write */
-    val = nt_ref(nt, w, "value");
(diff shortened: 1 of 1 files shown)


matz:
Merged. Same verification as the other two: generated C byte-identical across all 2934 tests and optcarrot-single, `make gate` green.

PR #3906, opened 2026-08-13
Answer IO#tty?, #isatty and #winsize on a poly-carried handle

tobi:
`tty?`, `isatty` and `winsize` reach their IO implementation only when the receiver is statically typed IO. A handle that arrives as **poly** — read back out of a container, or a parameter whose call sites disagree — misses the poly-dispatch allowlist entirely, so the call answers

```
undefined method 'tty?' for an instance of IO (NoMethodError)
```

for a value that genuinely *is* an IO.

Smallest reproduction, on current master:

```ruby
[STDOUT].each { |s| p s.tty? }
```

```
undefined method 'tty?' for an instance of IO (NoMethodError)
```

CRuby prints `false`.

## The fix

Two places had to agree, matching how the neighbouring IO predicates (`eof?`, `closed?`, `path`, `rewind`, …) are already handled:

- **`src/codegen_call.c`** — names the three methods in the poly-receiver allowlist and emits the existing `sp_File_tty_p` / `sp_File_winsize` runtime helpers. Both were already present in `lib/sp_io.c` and already reachable from the statically-typed arm; only the poly path could not get to them.
- **`src/analyze_infer.c`** — types them, so a following `size[0]` reads an int array rather than an untyped poly value. `tty?`/`isatty` join the `TY_BOOL` predicates; `winsize` answers `TY_INT_ARRAY` under the `io/console` feature guard, the same shape the `TY_IO` arm gives.

`winsize` needed the inference half specifically: without it the call falls through to a plain poly result and the indexing that almost always follows it reads an untyped value.

## Test

`test/io_tty_winsize_poly_handle.rb` covers the shapes that degrade a handle to poly: a literal-array block parameter, a container read (`handles[0]`, `handles.first`), and a parameter with two disagreeing call sites (one `File`, one `IO.console`). `isatty` is exercised as the second name for the same method.

The test pins **reachability, not terminal geometry**. It runs with stdout redirected to a file, so it must not assume a tty exists — `winsize` is asserted only to answer *without* `NoMethodError`. A non-tty answer legitimately differs between the engines (CRuby raises `Errno::ENOTTY`; Spinel reports zeroes from the ioctl), and the helper accepts either:

```ruby
rescue NoMethodError
  false          # the bug: an IO that claims not to know the method
rescue StandardError
  true           # ENOTTY and friends: reached the method, no tty to measure
```

## Verification

Against master at `e05feeb9`:

- Test is **RED before / GREEN after** — output byte-identical to CRuby (`FalseClass false false false true false true true false`).
- Full corpus with the patch: **2958 pass, 0 fail, 7 error**.
- Those same 7 (`fiber_kill`, `thread_gc_start`, `thread_condvar`, `thread_mutex`, `thread_kill_raise`, `monitor_reentrant`, `block_tail_prelude_inside`) error **identically on unpatched master** — pre-existing and unrelated, all thread/fiber tests.

One caveat on CI, unchanged from my earlier PRs: `make test` aborts at `Makefile:552` because `rbs-seed-test` fails on clean master (#3203) before the corpus runs. The corpus numbers above come from building the `TEST_TARGETS` directly to get past that gate. If the check comes back red, that is the reason, not this change.

### DIFF
--- a/src/analyze_infer.c
+++ b/src/analyze_infer.c
@@ -5410,7 +5410,13 @@ else {
       if (sp_streq(name, "resume") || sp_streq(name, "value") || sp_streq(name, "join"))
         return TY_POLY;
       if (sp_streq(name, "alive?") || sp_streq(name, "dead?") || sp_streq(name, "closed?") ||
-          sp_streq(name, "eof?")) return TY_BOOL;
+          sp_streq(name, "eof?") || sp_streq(name, "tty?") || sp_streq(name, "isatty"))
+        return TY_BOOL;
+      /* IO#winsize on a poly-carried handle: [rows, cols], same as the TY_IO
+         arm. Without this the call falls through to a plain poly result and the
+         `size[0]` that follows reads it as an untyped value. */
+      if (sp_streq(name, "winsize") && sp_feature_enabled("io/console"))
+        return TY_INT_ARRAY;
       if (sp_streq(name, "read") || sp_streq(name, "gets") ||
           sp_streq(name, "readline")) return TY_STRING;
       if (sp_streq(name, "write")) return TY_INT;   /* IO#write: the byte count */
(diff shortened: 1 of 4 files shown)


matz:
Merged. Verified: `[STDOUT].each { |s| p s.tty? }` answers false like CRuby, the generated C is byte-identical to master's across the other 2935 tests, and `make gate` is green under gcc and clang.

Issue #3971, opened 2026-08-16
[Compiler] A local bound from an Array[Integer] element is typed String when the expression also indexes a String constant (cead44bb)


rubys:
A local bound from an `Array[Integer]` element is typed `String` when the same
expression also indexes a `String` constant. The integer arithmetic on it then
compiles to `sp_raise_nomethod(... "undefined method '&' for an instance of
String" ...)`, which lands in an `mrb_int` argument slot and fails the C build.

Bisected to cead44bb ("Alias the element a string read out of a container
binds"); its parent 39cc2af2 is clean.

## Reproduce

Self-contained — a published archive, no other checkout needed:

```sh
curl -fsSL https://rubys.github.io/roundhouse/browse/spinel.tgz | tar -xz
cd spinel
make build          # → spinel main.rb --rbs . -o build/blog
```

```
39cc2af2 (parent of cead44bb)   make build → OK
cead44bb                        make build → 13 errors in runtime/base64.rb, exit 2
master (39b2baca)               same 13 errors
```

I could not reduce this below the whole program: `runtime/base64.rb` compiled
on its own is clean at cead44bb, as is a hand-cut version of the method. It
needs the surrounding program, which is why the archive is the repro.

## The source

`runtime/base64.rb`, `Base64.strict_encode64`:

```ruby
ALPHABET = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"

def self.strict_encode64(s)
  bytes = s.bytes        # Array[Integer]
  n = bytes.length
  out = +""
  i = 0
  while i + 3 <= n
    b0 = bytes[i]
    b1 = bytes[i + 1]
    b2 = bytes[i + 2]
    out << ALPHABET[(b0 >> 2) & 0x3F].to_s
    ...
```

Two containers meet in one expression: `bytes` is `Array[Integer]` and
`ALPHABET` is a `String`.

## What is emitted

```
runtime/base64.rb:26:11: error: incompatible integer to pointer conversion
  assigning to 'sp_String *' from 'mrb_int' (aka 'long') [-Wint-conversion]
   26 |       b0 = bytes[i]
```

so `b0` is declared `sp_String *`. The arithmetic on it then becomes a raise:

```c
sp_String_append_bin(lv_out, ({ const char *_t367 =
  sp_str_char_at_or_nil(cst_ALPHABET,
    sp_raise_nomethod(sp_stage_args_msg(
[...]

PR #3981, opened 2026-08-17
Gate the deep-return strbuf demand on the caller's slot

### LINKED ISSUE #3980: [Compiler] An empty array literal returned from a method through a local is typed String and emitted as sp_String_new_shared(sp_IntArray_new())
An empty array literal returned from a method **through a local variable** loses its
container type and settles at `String`. The literal is then emitted as
`sp_String_new_shared(sp_IntArray_new())` — a String constructor fed an `sp_IntArray *` —
and every use of the result in the caller follows the mistype, so the C build fails where a
`double` or an object pointer reaches a `const char *`.

Returning the very same literal directly (`def pick; []; end`) is correct, which is what
kept this hidden.

## Reproduction path

```ruby
def pick
  trimmed = []
  trimmed
end

as = pick
as << 1.5
```

```
$ ruby   t.rb        # no output, exit 0
$ spinel t.rb -o t
t.rb:2:31: error: incompatible pointer types passing 'sp_IntArray *' to parameter
  of type 'const char *' [-Werror,-Wincompatible-pointer-types]
    2 |   lv_trimmed = sp_String_new_shared(sp_IntArray_new());
t.rb:6:31: error: passing 'double' to parameter of incompatible type 'const char *'
    6 |   sp_String_append_bin(lv_as, 1.5);
spinel: C compilation failed
```

## What is emitted

```c
static inline const char * sp_pick(void) {
    sp_String * lv_trimmed = NULL;
  lv_trimmed = sp_String_new_shared(sp_IntArray_new());          /* [] */
  return (_sp_ret_strbuf = (void *)lv_trimmed, sp_str_concat(sp_String_cstr(lv_trimmed), ...));
}
...
    sp_String * lv_as = NULL;
  lv_as = ... sp_pick() ...;
  sp_String_append_bin(lv_as, 1.5);                              /* as << 1.5 */
```

The mistype is *consistent* all the way down — `<<` becomes `sp_String_append_bin`, `[]`
becomes `sp_str_char_at_or_nil`, `.length` becomes `sp_str_length_m` — so the C compile only
fails at the leaves, and the reported line is not where the wrong decision was made.

## Where it goes wrong

`promote_shared_stored_strings` (`src/analyze.c`), the deep-return alias rule (#3227 P6). Its
branch for *the caller mutates the result* tests only
`strbuf_mut_kind(c, lname3, ls3) == 1`. That helper is keyed on **name + scope only** and is
purely name-based, so it reports a string mutation for any `rr << x` — and `<<` is equally
`Array#push` and `Integer`'s shift. Nothing checks that the caller's slot could hold a String,
so the rule demands the callee's tail local into the shared set and types it `TY_STRBUF`.

Per-pass local types, from a temporary dump:

```
infer_write_types              (top) as      : unknown     -> float_array   # correct
promote_shared_stored_strings  pick  trimmed : unknown     -> strbuf        # the bug
infer_write_types              (top) as      : float_array -> string        # poisoned via pick's return
promote_shared_stored_strings  (top) as      : string      -> strbuf
```

Inference gets `as` **right** first; the demand rule then overrides it through the callee.

## What is and is not required

| Ingredient | Required? |
|---|---|
| Method returns the `[]` **via a local** | **yes** — `def pick; []; end` is correct |
| `<<` on the result | **yes** — `.push(1.5)` is correct |
| A method call in the chain | **yes** — `as = []; as << 1.5` alone is correct |
| A parameter on the method | no — `pick(7)` reproduces and never reads the argument |
| A user-defined element class | no — `1.5`, `1` and `"x"` all reproduce |
| Result assigned back over the argument | no |
| Two call sites at different element types | no |

## Notes

Same two ingredients as #3227's `void_ret_infer_shl_ivar_local_tail` family
(return-via-local + `<<`), and the same underlying hazard as #3971, whose fix added
`strbuf_container_stores_string` with the comment that "a name-keyed mutator table cannot
tell `out << "x"` from `b0 << 4`". A name-keyed mutator table needs a type gate at every use;
this branch was the one site in `promote_shared_stored_strings` without one.

bartleusink:
Closes #3980.

An empty array literal returned from a method **through a local** settled at `String`, so it
was emitted as `sp_String_new_shared(sp_IntArray_new())` and the C build failed where a
`double` or object pointer reached a `const char *`:

```ruby
def pick
  trimmed = []
  trimmed
end

as = pick
as << 1.5
```

Returning the same literal directly (`def pick; []; end`) was always correct, which is what
kept this hidden.

## The defect

`promote_shared_stored_strings`' deep-return alias rule (#3227 P6) has a branch for *the
caller mutates the result* (`rr = mk; rr << x`): it demands the callee's tail local into the
shared-handle set so a later iteration can admit the site. Its only test was
`strbuf_mut_kind(c, lname3, ls3) == 1` — a helper keyed on **name + scope only** and purely
name-based, so it reports a string mutation for any `rr << x`, and `<<` is equally
`Array#push` and `Integer`'s shift. Nothing checked that the caller's slot could hold a
String.

Per-pass local types (from a temporary dump) show inference got the caller **right** first,
and the demand rule then overrode it *through the callee*:

```
infer_write_types              (top) as      : unknown     -> float_array   # correct
promote_shared_stored_strings  pick  trimmed : unknown     -> strbuf        # the bug
infer_write_types              (top) as      : float_array -> string        # poisoned via pick's return
promote_shared_stored_strings  (top) as      : string      -> strbuf
```

That is also why the reported error line was inside the callee and not where the wrong
decision was made.

## The fix

Gate the demand branch on the caller's slot being String-capable
(`TY_UNKNOWN | TY_STRING | TY_STRBUF | TY_POLY`) — exactly the gate this function's other
branches already apply. Every other `strbuf_mut_kind` call site on a local already had one
(8915/8989 require `TY_STRING`; 9083 goes through `strbuf_slot_eligible`; the ivar-alias site
checks explicitly before calling it). This branch was the only site without one. The sibling
branch's duplicate lookup and identical gate are folded into the same two locals.

The gate only ever declines slots already typed non-String, which could never have been valid
strbufs, so no legitimate promotion is lost.

Same underlying hazard as #3971, whose fix added `strbuf_container_stores_string` with the
comment that "a name-keyed mutator table cannot tell `out << "x"` from `b0 << 4`". A
name-keyed mutator table needs a type gate at every use.

## Test

`test/empty_array_return_via_local.rb` covers the defect plus the shapes that must keep
working, so a fix cannot pass by blanket-pinning every empty literal to `Array` or by deleting
the branch:

- a literal returned directly (`def make_direct; []; end`)
- `.push` instead of `<<`
- a generic top-K helper at two element types, result assigned back over the argument (the
  shape this was found in)
- the #3227 P6 shape this branch exists for, including alias identity through the handle
  (`q.equal?(q2)` → `true`)

## Verification

- `make test` — **3056 pass, 0 fail, 0 error**
- `make bench` — **61 pass, 0 fail, 0 error, 0 skip**
- `make gate-optcarrot` — OK, checksum 59662
- `make gate-rubyspec` — all expected-PASS examples still pass (core/string 571, core/hash
  192, core/integer 171, core/range 83)
- `make alloc-report-test` — pass
- The five #3227 / strbuf tests re-run individually: all pass

Two `make check` legs fail, both **pre-existing on master** — I confirmed each fails
byte-identically with this commit's `src/analyze.c` stashed:

- `infer-test`: its `grep -Eq 'static mrb_int sp_M_s_mul\(...'` doesn't match the emitted
  `static inline __attribute__((always_inline)) mrb_int ...`, which is the default now that
  small leaf methods are force-inlined.
- `spin-check`: `spin-e2e FAIL: test (sibling spinel dir): want [1/1 passed] got [0/1 passed]`.

I did not find a `gen2.c`/`gen3.c` closure target in this tree, so I ran `make check`'s legs
instead; happy to run something else if there's a target I missed.


matz:
Merged as fe23a320. Thank you -- the diagnosis is exactly right: `strbuf_mut_kind` is name-keyed, so the demand branch was taking `rr << x` as evidence of a String append when it is equally `Array#push`, and this was the one call site on a local with no type gate in front of it.

Verified here on top of current master: the reproduction compiles and prints `[1.5]`, matching CRuby, and `make gate` is green (3058 tests, 61 benchmarks, optcarrot checksum 59662, every rubyspec leg unchanged).

On the two `make check` legs you saw fail: both are known and unrelated to this change, so nothing was expected of you there.

Issue #3417, opened 2026-07-28
[Compiler] `h[k] += 1` on a captured local is refused inside a lifted block over a group_by result


rubys:
Rebuilt at `4b1ed6d0` — #3416 is fixed, the Lobsters tree gets past `relation.rb`, and stops at the next one.

`h[k] += 1` on a captured local is refused inside a lifted block when the receiver is a `group_by { }` result. `h[k] ||= 0` on the same local is fine, and a non-indexed `n += 1` is fine.

## Repro

```ruby
class Rel
  def each
    yield "row"
  end
end

class T
  def initialize(u, t); @u = u; @t = t; end
  def user_id; @u; end
  def tag; @t; end
end

class R
  def sug
    [T.new(1, "a"), T.new(1, "a"), T.new(2, "b")]
  end

  def tally
    v = {}
    self.sug.group_by { |x| x.user_id }.each do |_u, stg|
      stg.each { |s| v[s.tag] ||= 0; v[s.tag] += 1 }
    end
    v
  end
end

p R.new.tally
p Rel.new.each { |r| r }
```

```
$ spinel main.rb -o m
spinel: main.rb:22: unsupported expression: node 83 (IndexOperatorWriteNode)
```

CRuby: `{"a" => 2, "b" => 1}` / `"row"`.

## Direction of travel, so this isn't misread

This shape was not working before and is not a regression from your fix. At `00da0fda` it compiled and answered **wrong** at run time:

```
undefined method 'tag' for an instance of String (NoMethodError)
```

which is the #3409 shape — the block param bound the wrong thing on the other arm. So the trade here is a silent miscompile becoming an honest refusal. That's the better failure; this issue is just the remaining half of it.

## What narrows it

Same HEAD (`4b1ed6d0`) each time; only the named ingredient changes.

| variant | result |
|---|---|
| as above | **refused** |
| drop `class Rel` (nothing else owns `each`) | correct — inner block splices, nothing lifts |
| `v[s.tag] ||= 0` only, no `+=` | correct |
| non-indexed `n += 1` against the same lift | correct |
| receiver is a method returning a Hash (no `group_by`) | correct |
| receiver is a method parameter (Hash) | correct |

So it takes the lift *and* the `group_by` receiver *and* the indexed operator-write together. The last two rows are what surprised me: a plain Hash receiver through the same lifted nest handles `v[s] += 1` fine, so it isn't the indexed write on a captured local by itself.

## Environment

[...]

PR #3997, opened 2026-08-18
Buffer the prelude an index op-assign's value form needs

### LINKED ISSUE #3996: [Compiler] Index op-assign in value position emits invalid C when the RHS needs a prelude
Follow-on to #3417. That fix let `arr[i] += x` compile in **value position** (the tail of a block lifted to a real proc). It works for a simple RHS, but an RHS that needs a prelude of its own emits structurally invalid C.

## Repro

```ruby
class Outcome
  def map_one(f)
    f + 1.0
  end
end

outcome = Outcome.new
a = [1.0]
->() { a[0] += outcome.map_one(2.5) }.call
puts a[0]
```

```
$ spinel main.rb -o m
/tmp/spinel_out_20367_0.c:243:144: error: expected ')'
/tmp/spinel_out_20367_0.c:243:153: error: expected expression
/tmp/spinel_out_20367_0.c:244:69: error: use of undeclared identifier '_t5'
/tmp/spinel_out_20367_0.c:244:73: error: expected ';' after expression
5 errors generated.
spinel: C compilation failed
```

CRuby: `4.5`.

This is a `cc` failure on the generated C, not a type error or a refusal — the emitted translation unit is syntactically invalid. The RHS's prelude landed inside the `sp_FloatArray_set(...)` argument list:

```c
sp_FloatArray_set(_t3, _t4, sp_FloatArray_get(_t3, _t4) + (  sp_float _t5 = 2.5;
sp_Outcome_map_one((sp_Outcome *)(*((_proc_cap_1 *)_cap)->outcome), _t5)));
```

## What narrows it

Same HEAD each time; only the named ingredient changes. All four ingredients are required.

| variant | result |
|---|---|
| as above | **invalid C** |
| `; nil` after the op-assign (no longer the body's tail) | correct |
| array created inside the proc instead of captured | correct |
| body inlined at the call site (`def r(&b); b.call(0); end`) | correct |
| RHS `2.5`, or a bare local, or `a[0] = a[0] + outcome.map_one(2.5)` | correct |
| RHS a zero-arg method, a top-level `def`, a `module_function`, or any builtin (`a.fetch(0)`, `b.sum { }`) | correct |

So it takes the value-position op-write **and** a captured receiver **and** an opaque proc **and** an RHS whose lowering wants a prelude — a user-defined instance method call with at least one argument, since that path hoists its receiver and spills each argument into a temp.

Not required: a float array (Integer and Hash receivers fail the same way); `+=` specifically (`*=` too); the block parameter as the index (`a[0]` fails too). Also fails when the key is itself a method call (`v[s.tag] += outcome.map_one(2.5)`), which additionally goes through `emit_index_opw_hoist`.

## Cause

`src/codegen_expr.c`, the `IndexOperatorWriteNode` value form. It emits the mutation with

```c
emit_index_op_write(c, id, g_pre, g_indent);
```

but `g_pre` is also the sink the RHS drains its own prelude into, so appending to that buffer mid-statement drops the declaration inside the argument list. The `else` branch immediately below (no prelude buffer available) already handles this correctly — it redirects `g_pre` into a local buffer and splices it in front of a `({ ... })`. The `g_pre` branch needs the same treatment.

## Environment

- `matz/spinel` master @ `9132e140`, clean worktree build (`make`)
- macOS 25.6.0 (arm64)

bartleusink:
Fixes #3996.

`arr[i] op= v` in value position — the tail of a block lifted to a real proc — emits the mutation and then reads the slot back, since Ruby yields the stored value. The mutation went straight into `g_pre`:

```c
emit_index_op_write(c, id, g_pre, g_indent);
```

but `g_pre` is also where the RHS drains a prelude of its own, and a user method call on the RHS has one: that path hoists its receiver and spills each argument into a temp. Appending to the buffer while the statement was still being written to it dropped the declaration inside the `sp_..._set(...)` argument list:

```c
sp_FloatArray_set(_t3, _t4, sp_FloatArray_get(_t3, _t4) + (  sp_float _t5 = 2.5;
sp_Outcome_map_one((sp_Outcome *)(*((_proc_cap_1 *)_cap)->outcome), _t5)));
```

So it failed in `cc` — unbalanced paren, `_t5` used before its declaration — rather than as an honest refusal. After:

```c
sp_float _t5 = 2.5;
{ ... sp_FloatArray_set(_t3, _t4, sp_FloatArray_get(_t3, _t4) + (sp_Outcome_map_one(..., _t5))); }
```

The fix collects the RHS's prelude and the statement in separate buffers and splices them back prelude-first — what the no-`g_pre` branch just below already did for the same reason. Emission order is now the hoisted receiver/key lines, the RHS's prelude, the mutation, then the read-back.

This is the follow-on to #3417, which made the value-position op-write compile at all; it works there for a prelude-free RHS.

## Test

One case per distinct emitted path, each verified to fail against a pre-fix build of this tree and to match CRuby after:

| case | why it is not redundant |
|---|---|
| captured float array, argument spilled to a temp | the reduced original shape |
| `Outcome.new` as the RHS receiver | the prelude is a receiver hoist that **allocates**, so it carries a GC root |
| integer array, op-write's value consumed | different set/get arm, plus the read-back value |
| hash receiver | another arm, with its own key handling |
| method-call key (`v[s.tag] += ...`) | non-cheap receiver/key, so `emit_index_opw_hoist` runs first (the #3417 path) — the ordering interaction none of the others reach |

## Checks

- `make test` — 3081 pass, 0 fail, 0 error
- `make rubyspec-gate` — all expected-PASS examples still pass across all six legs


matz:
Merged as 482136f4. Thank you -- the diagnosis is exact, and the fix is the shape the branch below it already had.

Verified here on top of current master: the reproduction from #3996 compiles and prints 4.5, matching CRuby, and `make gate` is green (3083 tests, 61 benchmarks, optcarrot checksum 59662, every rubyspec leg unchanged).

The test's case-by-case note on why each shape is not redundant is appreciated -- the method-call key one in particular reaches an ordering interaction none of the others do.

PR #3999, opened 2026-08-18
Raise CRuby's ArgumentError for builtin calls with an impossible argument count

blacktm:
## Summary

A builtin method called with a positional argument count CRuby rejects now raises
CRuby's own `ArgumentError` — same class, same message — instead of one of three
wrong behaviors:

1. **Compiler crash.** Several emitters index `argv[]` unconditionally, so a
   missing operand was a compile-time SIGSEGV:

   ```ruby
   (1..3).cover?    # spinel: exit 139
   (1..3).include?  # spinel: exit 139
   File.open        # spinel: exit 139
   File.new         # spinel: exit 139
   :a.<=>           # spinel: misdispatch / crash
   ```

2. **`NoMethodError` for a method that exists.** Builtin dispatch is keyed on
   exact `(name, argc)` pairs, so a wrong count fell off the table and read as an
   unknown method (`File.read` → `NoMethodError`, CRuby: `ArgumentError, "wrong
   number of arguments (given 0, expected 1..3)"`).

3. **Silently absorbed arguments.** Excess or missing arguments were dropped and
   the call answered something plausible (`:a.to_s(9)` → `"a"`, `"x".delete` →
   `"x"`, `Hash.new(1, 2)` accepted).

After this series all of the above behave exactly as CRuby 4.0.6, e.g.:

| call | CRuby / spinel (now) |
|---|---|
| `(1..3).cover?` | `ArgumentError: wrong number of arguments (given 0, expected 1)` |
| `"x".insert(1)` | `… (given 1, expected 2)` |
| `:a.to_s(9)` | `… (given 1, expected 0)` |
| `[1].insert` | `… (given 0, expected 1+)` |
| `"x".chomp("a", "b")` | `… (given 2, expected 0..1)` |
| `File.open` | `… (given 0, expected 1..3)` |
| `Time.at` | `… (given 0, expected 1..3)` |
| `(1..2).first` | `1` (still valid — see below) |

## How

One guard at the top of `emit_call` (`emit_builtin_arity_guard`,
`src/codegen_call.c`), run before the argument *type* guards since CRuby checks
arity at dispatch. When it fires, the emitted code evaluates the receiver and
arguments for effect and raises; the program still compiles, exactly as CRuby
defers the error to run time.

It is driven by two spec tables **probed from CRuby itself** by the new
`tools/gen_builtin_arity_spec.rb` (rerun with `--write` to regenerate against a
newer CRuby):

- counts 0–3 are called with `nil` arguments to find each method's true
  minimum — any outcome other than a "wrong number of arguments" `ArgumentError`
  proves the *count* is accepted. This runs on non-empty receivers, because
  several C methods take an empty-receiver fast path that returns before the
  arity check (`"".delete` never raises);
- a 99-argument call makes CRuby name the accepted range in its own message
  ("given 99, expected 0..1"), giving the maximum and the exact wording. The
  low-side message is captured separately, because branch-implemented methods
  word the two sides differently (`Range#first`: 0 args is a bare read, 1+ args
  is "expected 1");
- every count 0–12 is then verified against the derived range. A method whose
  accepted set is not one contiguous range — `Time.utc` takes 1..8 *or exactly
  10* — is left unguarded rather than guessed.

## When the guard stands down

It fires only when the count is provably wrong, and skips:

- splatted, keyword, or forwarded arguments (runtime count unknown);
- calls carrying a block — a block changes several methods' counts
  (`Array#fill`: 1..3 bare, 0..2 with a block);
- safe navigation (`x&.m` short-circuits before any arity check);
- a reopened builtin that defines the name, and a user class shadowing a
  guarded constant (`File`);
- receivers without a settled static type (an empty `[]`/`{}` literal types as
  unknown — a known follow-up, listed below).

## Validation

- `make test`: full suite green, including two new tests
  (`test/builtin_arity_instance.rb`, `test/builtin_arity_class_methods.rb`)
  whose `.expected` files are CRuby-generated.
- `test/str_method_nil_arg_no_segv.rb` previously pinned stand-in behaviors
  (TypeError / silent no-op) for these counts; its pins now assert CRuby's
  ArgumentError.
- Independently validated against a differential corpus of minimized programs
  run under both CRuby 4.0.6 and spinel: **99 previously-diverging programs now
  behave identically, with zero regressions** across the ~2,100-program corpus.

## Known follow-ups (out of scope here)

- An empty `[]` / `{}` literal receiver types as unknown and skips the guard.
- `send(:<<)` reaches the frozen-string check before the arity guard, so a
  frozen receiver reports `FrozenError` where CRuby raises `ArgumentError`.
- `String#bytesplice` (accepts 2, 3, or 5 positional args — a gapped set,
  though CRuby's own message says "2..5") is covered by the verification rule
  and left unguarded; it could get a hand entry if the table format grows a
  second range.

## Summary by CodeRabbit

- **バグ修正**
  - 組み込みメソッドおよびクラスメソッドで、引数の不足・過剰を正しく検出し、CRuby互換の `ArgumentError` を返すよう改善しました。
  - 不正な呼び出しでも、レシーバーや引数の評価順序を維持します。
  - スプラット引数、キーワード引数、転送引数、ブロック付き呼び出しなどには従来どおり対応します。

- **テスト**
  - `File`、`IO`、`Time`、`Hash`、`Math`、`Dir` などの組み込みメソッドに対する引数検証を追加しました。
  - 文字列メソッドの引数エラーに関する期待結果を更新しました。

### DIFF
--- a/src/codegen_call.c
+++ b/src/codegen_call.c
@@ -8749,6 +8749,336 @@ static int emit_range_step_bad_stride(Compiler *c, int id, int recv, int arg,
   buf_printf(b, "%s; })", dv ? dv : "0");
   return 1;
 }
+/* A call whose argument count CRuby's own method cannot take: evaluate the
+   receiver and arguments for effect, then raise CRuby's ArgumentError --
+   instead of falling into an emitter specialized to another count, where a
+   missing operand read argv[] blind (a compile-time SIGSEGV) and an extra one
+   was silently dropped. Conservative by construction: fires only when the
+   dumped CRuby arity table proves the count wrong, never on a call whose
+   runtime count a splat / kwargs / forwarding keeps open, and never when a
+   reopened builtin owns the name. */
+/* Positional-arity spec for the builtin instance surface, probed from
+   ruby 4.0.6 by tools/gen_builtin_arity_spec.rb (see there for the
+   technique; rerun it with --write to regenerate both tables). max -1 =
+   no upper bound; a NULL exp = that side is never violated. Bare calls
+   only -- a block changes several counts, and block-carrying calls skip
+   the guard. */
+static const struct { const char *cls; const char *m; int min; int max;
+                      const char *lo_exp; const char *hi_exp; }
+sp_builtin_arity_spec_tbl[] = {
+  {"String","%",1,1,"1","1"},{"String","*",1,1,"1","1"},
+  {"String","+",1,1,"1","1"},{"String","+@",0,0,NULL,"0"},
+  {"String","-@",0,0,NULL,"0"},{"String","<<",1,1,"1","1"},
+  {"String","<=>",1,1,"1","1"},{"String","==",1,1,"1","1"},
+  {"String","===",1,1,"1","1"},{"String","=~",1,1,"1","1"},
+  {"String","[]",1,2,"1..2","1..2"},{"String","[]=",2,3,"2..3","2..3"},
+  {"String","ascii_only?",0,0,NULL,"0"},{"String","b",0,0,NULL,"0"},
+  {"String","byteindex",1,2,"1..2","1..2"},{"String","byterindex",1,2,"1..2","1..2"},
+  {"String","bytes",0,0,NULL,"0"},{"String","bytesize",0,0,NULL,"0"},
+  {"String","byteslice",1,2,"1..2","1..2"},{"String","casecmp",1,1,"1","1"},
+  {"String","casecmp?",1,1,"1","1"},{"String","center",1,2,"1..2","1..2"},
+  {"String","chars",0,0,NULL,"0"},{"String","chomp",0,1,NULL,"0..1"},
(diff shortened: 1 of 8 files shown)


matz:
Merged as 82faf1ff. Thank you -- probing the tables out of CRuby rather than hand-writing them is what makes this trustworthy, and the guard's exclusions (block, splat, keywords, forwarding, `&.`, a same-named user class, a reopened builtin owning the name) are the right set.

Verified here on top of current master: every row of your table matches CRuby 4.0.6 exactly, message included. I also threw 39 legitimate calls at it looking for over-rejection and found none -- the one difference in that batch (`Hash.new(0)` reached in a particular context) reproduces on master without this change, so it is a separate defect and I am fixing it now.

Compile time is unaffected: optcarrot's 125KB single-file source compiles in 0.309s with the guard against 0.311s without (best of three), so the table scan does not show. `make gate` green (3086 tests, 61 benchmarks, checksum 59662, all rubyspec legs unchanged).

PR #4002, opened 2026-08-18
Ask the owning class whether an ivar slot is numeric

### LINKED ISSUE #4001: [Compiler] An ivar's element type leaks between unrelated classes that share the ivar name
An ivar's element type leaks between two classes that share only the ivar's NAME. A numeric `@threads` in one class stops an unrelated `@threads = []` in another from ever learning its element type, so the slot keeps the empty literal's default int array and pushing a pointer element into it fails at `cc`.

## Repro

```ruby
@threads = 1
class B
  def initialize
    @threads = []
    @threads << "x"
  end
end
```

```
$ spinel main.rb -o m
main.rb:5:38: error: incompatible pointer to integer conversion passing 'char *' to
    parameter of type 'sp_int' (aka 'long') [-Wint-conversion]
    5 |   sp_IntArray_push(self->iv_threads, ({ static struct { ... } _fzl_0 = ...; _fzl_0.d; }));
1 error generated.
spinel: C compilation failed
```

CRuby runs it fine. Rename either `@threads` and nothing else, and it compiles.

## What narrows it

Same HEAD each time; only the named ingredient changes.

| variant | result |
|---|---|
| as above | **invalid C** |
| either ivar renamed | correct |
| the colliding write removed entirely | correct |
| colliding write is `1.5` instead of `1` | **invalid C** (any scalar numeric) |
| colliding write inside a class instead of top level | **invalid C** |
| declaration order swapped | **invalid C** |
| pushed element is a `Thread` instead of a `String` | **invalid C** (any pointer-like element) |
| `@threads = ["x"]` as a literal, no `<<` | correct |
| reverse pairing: `"s"` scalar in one class, `<< 1` in the other | correct |

The last two rows are the interesting ones. A literal array compiles, so the element type has to be arriving from a later push — this is the empty-literal-with-no-element-type family. And the reverse pairing compiles into two independent fields (`const char *` and `sp_IntArray *`), which shows ivar types are **not** globally name-keyed.

## Cause

Ivar types are already per-class: `comp_ivar_index` is a flat per-class table and a subclass interns its own slot. The global part is one query in `src/analyze_pass.c`.

`recv_has_scalar_numeric_write` decides whether `<<` is `Array#push` or `Integer#<<` (a shift), and vetoes the push promotion for a numeric slot. It indexes writes by NAME and then narrows:

```c
if (is_local && comp_scope_of(c, id) != rscope) continue;
```

For a local that filters by scope. For an ivar it filters by nothing — `qscope` is hard-coded `-1` and no owner is compared — so `@threads = 1` anywhere in the program answers a question asked about `@threads` in class `B`. The veto fires, the push supplies no element evidence, and the bare `[]` keeps its default int array.

Same lesson as the `strbuf_mut_kind` gate: a name-keyed table cannot tell `out << "x"` from `b0 << 4`, so the owner has to be part of the key.

## Environment

- `matz/spinel` master @ `9132e140`, clean worktree build (`make`)
- macOS 25.6.0 (arm64)

bartleusink:
Fixes #4001.

`recv_has_scalar_numeric_write` decides whether `@x << v` is `Array#push` or `Integer#<<` (a shift), and vetoes the push promotion for a numeric slot. It indexes writes by name and then narrows:

```c
if (is_local && comp_scope_of(c, id) != rscope) continue;
```

A local filters by scope; the ivar arm filtered by nothing — `qscope` is hard-coded `-1` and no owner was ever compared. So `@threads = 1` in one class answered a question asked about `@threads` in a class sharing nothing with it but the name. The veto fired, the push supplied no element evidence, the bare `[]` kept the empty literal's default int array, and pushing a pointer element into it failed at `cc`.

Ivar types are otherwise already per-class — `comp_ivar_index` is a flat per-class table, a subclass interns its own slot, and the reverse pairing (a String scalar in one class, an int push in the other) really does emit two independent fields. This query was the global part. The fix compares the owning class, treating a top-level ivar's `-1` as the owner it is.

## Considered and deliberately not implemented

The sibling `recv_has_array_write` has the same unfiltered plain-ivar arm, and it is tempting to fix both as one defect. I left it alone, because the two fail in **opposite directions**:

- `recv_has_scalar_numeric_write` **vetoes** the push promotion. Leaking across owners vetoes too much and loses the element type outright — this bug. Narrowing it strictly removes spurious vetoes.
- `recv_has_array_write` supplies **evidence**. Narrowing it could only remove evidence and let a slot fall back to the empty literal's bottom kind — which is exactly the #3781 regression its reader branch was added to fix.

I tried several shapes to make the missing filter there produce wrong code (a user-defined `<<` crossed with a same-named array write in another class, with and without other type evidence for the receiver) and every one compiled correctly and behaved identically before and after. Rather than add code for something I could not show was broken, I left it as-is. Happy to revisit if someone has a reproducer.

## Test

Three cases, no duplicates. The first two fail against a pre-fix build of this tree; all three match CRuby after.

| case | why it is not redundant |
|---|---|
| numeric `@threads` in one class, `[]`-then-`<<` of a String in another | the bug |
| top-level `@threads` vs a class's `@threads` | a distinct owner kind, and it covers the `-1` owner where a class-id collision would slip through |
| same-class `@bits = 1` then `@bits << 3` | the guard being narrowed must still fire within one class; a mistake here is a silent wrong answer (an int array instead of a shift), not a build error |

Checked and excluded because they behave identically before and after, so they exercise nothing: a parent/subclass pair sharing an ivar, and the user-defined-`<<` shapes above.

## Checks

- `make test` — 3081 pass, 0 fail, 0 error
- `make rubyspec-gate` — all expected-PASS examples still pass across all six legs

Found in a worker pool whose `@threads` collided with an unrelated config object's `@threads`.

## Summary by CodeRabbit

* **バグ修正**
  * 異なるクラスで同名のインスタンス変数を使用した場合に、型情報が誤って判定される問題を修正しました。
  * トップレベルおよび各クラスのインスタンス変数を正しく区別できるようになりました。

* **テスト**
  * クラスごとに同名インスタンス変数の型情報が分離されるケースを追加しました。


matz:
Merged as cb92db9a. Thank you -- the diagnosis is exact, and the reasoning for leaving `recv_has_array_write` alone is the right call: the two queries fail in opposite directions, and narrowing the evidence-supplying one could only take evidence away.

Verified here on top of current master: the #4001 reproduction compiles and prints ["x"], `make gate` is green (3084 tests, 61 benchmarks, optcarrot checksum 59662, every rubyspec leg unchanged). I also checked the two shapes the narrowing could plausibly have broken -- a numeric write in a PARENT with the shift in the child, and one supplied by an included module -- and both still compile to a shift and answer as CRuby does, before and after.

PR #4029, opened 2026-08-19
Validate arguments at builtin boundaries the way CRuby does: nil/false TypeErrors and block-call arity

blacktm:
## Why

When a Ruby program passes a bad argument — a nil that should have been a String, an extra argument a method doesn't take — CRuby stops it right there, at the call, with an error that names the problem. Spinel mostly let it through. The nil became `0` or `""`, the extra argument was dropped, and the program kept running on a wrong value: `File.join("a", nil)` answered `"a/"`, `[10, 20, 30].take(nil)` answered `[]`, `3.times(1) { }` just ran the loop. This is the worst kind of incompatibility, because nothing announces it — a program that raises on CRuby *succeeds* on Spinel with a different answer, and the bug surfaces far from its cause, if it surfaces at all.

We run a differential corpus of ~2,100 minimized programs that behave differently under CRuby 4.0.6 and Spinel (the same corpus behind #3999 and #4003). Missing argument validation was the single largest root cause left in it. This series closes that family: **211 of those programs now behave identically to CRuby, with zero regressions** — the biggest single step the corpus has measured so far.

## What was missing

1. **A known-bad value in a typed slot was silently absorbed.** A nil, a boolean, or a wrongly-classed scalar entering a String- or Integer-typed builtin slot became the slot's zero:

   ```ruby
   [10, 20, 30].take(nil)   # spinel: []            CRuby: TypeError
   File.join("a", nil)      # spinel: "a/"          CRuby: TypeError
   "ab".sub(nil, "x")       # spinel: "ab" (no-op)  CRuby: TypeError
   [1].pack(1)              # spinel: C build fails CRuby: TypeError
   ```

2. **Block-carrying calls skipped the arity guard.** #3999's guard stood down whenever a block was present (a block changes several methods' accepted counts), and its probe surface — the Method#arity dump — missed methods and receiver classes outright:

   ```ruby
   3.times(1) { |i| i }        # spinel: ran the loop  CRuby: ArgumentError
   1.5.div                     # spinel: NoMethodError CRuby: ArgumentError
   Struct.new(:a).new(1).a(2)  # spinel: 1             CRuby: ArgumentError
   ```

After this series both behave as CRuby 4.0.6 does — error class, message wording, and argument-evaluation order included.

## How

**The typed-slot emitters carry the validation** (same boundary as #4003's conversion protocol). A compile-time-known nil / true / false raises CRuby's TypeError at run time; a known wrong scalar class raises the class-naming form — and since every such case previously emitted ill-typed C that stopped the build, that extension cannot cost any working program. The wording follows CRuby's own split: rb_num2long slots say "no implicit conversion from nil to integer", rb_to_int slots say "of nil into Integer", the arithmetic family says "can't be coerced" (a Symbol by its inspect form), and the pattern family says "wrong argument type nil (expected Regexp)".

**The nil-accepting slots are probed, not guessed.** A new generator (`tools/gen_nil_arg_probe.rb`) asks CRuby, per method / slot / count, where nil is legal — `"x".split(nil)`, `Array#fill` offsets, `ENV[k] = nil`, a regexp's nil subject — including a with-block re-probe to see through lazily-validating enumerator returns. Those slots keep their looseness through nilable entry points, and every one of them stays bool-strict (`"x".split(false)` still raises, as CRuby).

**The arity spec learned what a block changes.** Every row of the #3999 tables now carries a second quartet probed WITH an inert block, so block-carrying calls are guarded too (an `&expr` block can be nil at run time, where bare-call arity applies — those still skip). The probe surface grew from the arity dump to every public instance method of the mapped classes, plus NilClass / TrueClass / Rational / Complex, the native package classes (StringIO, StringScanner, Pathname, Set, Mutex — whose loose C dispatch dropped excess arguments), and Object's universal rows, which a user object falls back to for methods its own chain does not define. The guard also runs where emit_call never sees the call: iterator-emitted loops, attr/Struct readers, and Struct's [] / dig / deconstruct_keys.

## Validation

- Full suite green (3,116 tests), including two new regression tests whose expected output is CRuby 4.0.6's own (`test/strict_arg_conversion.rb`, `test/block_call_arity.rb`).
- **211 previously-diverging corpus programs now behave identically, zero regressions** across the ~2,100-program corpus.
- The series was independently reviewed, adversarially, before filing — four reviewers over the emitters, the dispatch arms, the guard, and the tools, every claim checked against CRuby. What they caught (a missed nil-subject slot in the named-capture `=~` form, `super(nil)`'s class-name default, `fill`'s nil length, probe-tool sandboxing) is fixed and documented in the last commit.
- The probe generators found two ways to segfault **CRuby itself** (`StringIO#pread` with mistyped arguments; `StringIO#set_encoding_by_bom` with a block) — both excluded from the probe surface here, and worth reporting to ruby-core separately.

## Reviewing commit by commit

1. *Raise CRuby's TypeError for nil and false in typed argument slots* — the mechanism: strict emitters, probed nilable entry points, runtime forms for boxed values.
2. *Sweep the conversion protocol's straggler slots and family arms* — the call sites: pattern family, Array-operand family, MatchData / Range / String#* / Regexp.new / native module functions, per-family wording in the numeric-literal guard.
3. *Probe the with-block arity and widen the guard's reach* — the second quartet, the widened probe universe and receiver map, the iterator-path and Struct/reader hooks.
4. *Pin the argument-validation behavior with CRuby-generated expectations* — the regression tests.
5. *Harden the validation against an adversarial review* — the edges an independent review of the series caught, each verified against CRuby.
6. *Extend the strict contract to run-time-typed values and Data indexing* — the automated-review response: the boxed runtime forms carry the same wrong-class TypeErrors as the static emitters, and Data's [] is NoMethodError.
7. *Reject boxed builtin values at strict slots, minus the real converters* — the contract's last runtime gap, with CRuby's genuine converters (strbuf-as-String, Rational#to_int, Complex#to_int) carved out.

## Not in this series (deliberately)

- Keyword-argument validation (unknown / missing keywords) — a separate mechanism from positional arity.
- `send` / `public_send` dispatch arity — the dynamic-dispatch tables are their own boundary (the same known gap noted in #3999).
- `coerce` (the numeric protocol) and `to_ary` / `to_hash` at splat boundaries — future series, per the #4003 plan.

## Summary by CodeRabbit

* **バグ修正**
  * 組み込みメソッドの引数検証を強化し、不正な`nil`・真偽値・型に対して適切な`TypeError`を返すよう改善しました。
  * メソッド呼び出し時の引数個数やブロック引数を適切に検証するようになりました。
  * 例外メッセージ未設定時に、空文字列ではなく例外クラス名を表示するよう変更しました。
  * `nil`を許容する引数は、引き続き正しく処理されます。

* **テスト**
  * 引数変換、引数個数、各種組み込みメソッドの互換性テストを追加しました。


matz:
Merged as c1d283eb. Thank you — this closes the largest remaining family in the differential corpus, and the wording split you followed (rb_num2long vs rb_to_int vs the coerce family) is what makes the errors usable rather than merely present.

Reviewed against master before merging: the full gate is green (3120 tests, 61 benchmarks, optcarrot checksum 59662), optcarrot costs 0.03% more instructions and compiles in the same time, and probes over the nil-accepting slots, valid block-carrying calls, the native package classes, and user-defined methods sharing builtin names found nothing this series takes away. Every divergence the probes did turn up reproduces on master unchanged. Three of them the series fixes on its way past, which the description does not claim: a nil subject to `=~`, `Exception#message` defaulting to the class name, and the ill-typed C that `Struct.new(:a).new(1).a(2)` used to emit.

Two follow-ups landed on top, neither a fault of the series:

- 2b9ba7fc splits five `} else {` lines (spinel writes the brace and the else on separate lines, in its own C as in emitted C), and rotates `raise_tail_value_c`'s static buffer so two calls in one `buf_printf` cannot render the same text twice. Every current call site uses it once; the rotation is for the one that later does not.
- 9b5cb856 fixes `join`'s separator, which the sweep did not reach: `[1, 2].join(nil)` and `join(false)` both segfaulted, because the separator went into the `const char *` slot from a raw emit_expr and arrived as NULL. It is a nilable String slot — nil means "" — so it is emitted through the nilable path now, and the typed joins take NULL as the empty string.

The two CRuby segfaults your probe generators found are worth reporting to ruby-core; please do file them.

PR #4050, opened 2026-08-20
Answer Object's identity protocol on the native handle kinds instead of refusing to compile

blacktm:
## Why

Every Ruby object answers `===`, `==`, `!=`, `equal?`, `eql?`, `frozen?` and `freeze` — they come from Object, and programs use them without thinking: a `case` over a Fiber, a `uniq` over Mutexes, an `assert_equal` on two Addrinfos, `frozen?` before a write. On Spinel, a surprising number of these calls did not run and did not raise: they stopped the compiler. `f === f` on a Fiber, `cv.frozen?` on a ConditionVariable, `st.equal?(st)` on a File::Stat, `(1..2) == [1, 2]` — each was an `unsupported call` at build time, where the program could not even rescue it. For someone trying Spinel on existing code that is the first wall they hit, before any wrong answer gets a chance to matter.

We run a differential corpus of ~2,100 minimized programs that behave differently under CRuby 4.0.6 and Spinel (the same corpus behind #3999, #4003 and #4029). Of the programs Spinel refuses to compile at all, this was the largest single family, and the cheapest: one missing fallback, not a missing feature. **40 of those programs now behave identically to CRuby, with zero regressions.**

## What was missing

Each native handle kind — Fiber, Thread, Queue, Mutex, ConditionVariable, Dir, Addrinfo, IO/File/File::Stat, Enumerator, Method, Exception, MatchData, a curried Proc, Random, OpenStruct, Time, Process::Tms, a String range — carried its own hand-written arms for *some* of the seven methods. A receiver without the arm for the method it was sent fell through to the front-end reject:

```ruby
f = Fiber.new { }
p(f === f)                      # spinel: unsupported call   CRuby: true
ConditionVariable.new.frozen?   # spinel: unsupported call   CRuby: false
Addrinfo.tcp("::1", 80).eql?(x) # spinel: unsupported call   CRuby: false
p (1..2) == [1, 2]              # spinel: unsupported equality  CRuby: false
```

And where an arm did exist, a few answered wrong: `r.freeze; r.frozen?` was false on a Random (and on Dir, Addrinfo, IO, Method, Enumerator — the generic `freeze` identity shortcut swallowed it), `Random#eql?` compared generator state where Ruby's is identity, `MatchData#frozen?` was a constant false that `freeze` then contradicted, `File::Stat#==` was pointer identity where CRuby compares modification times, OpenStruct's `eql?` was its `==`, a Float or String range's `!=` was typed as nil, and `case e2 when e` over two Times did not compile.

## How

**One arm, at the very end.** `emit_native_object_protocol` runs only from the tails of `emit_call` and `emit_case_eq_call`, after every typed arm and the user-method routing have declined. That position is the safety argument: it can never shadow an existing answer, only replace a rejection. The inference side has the matching rule so the results type as Bool.

**One decision, shared.** Which calls the arm answers is decided in one function, `ty_object_protocol_answers` (types.c), that the type inference consults too — so the two sides cannot drift, and a kind is added in one place.

**The semantics are CRuby's, probed per kind, not assumed.** A heap handle *is* its pointer: the predicates are pointer identity and `frozen?` reads the bit `freeze` sets — the GC header, or for an IO the handle's own flag, because the standard streams are static storage with no header to flip. Random, OpenStruct, Exception, Time and Process::Tms compare structurally under `==` and `===` (state, members, class and message, instant, fields); `File::Stat` compares by modification time, as Comparable gives it; `eql?` and `equal?` stay identity where the value has one, and OpenStruct's `eql?` is its table's (class-strict per member). Time, Tms and a String range are by-value structs with no identity: `equal?` and (bar the Range, which is always frozen) `frozen?` are not answered for them rather than answered wrongly. `Proc#===` — a curried one too — `Method#===` and `Range#===` call, invoke or cover, and are left alone; a user exception subclass that defines one of these methods keeps its definition.

**Evaluation is Ruby's order, once.** The receiver is evaluated first into a rooted temp, then the operand, then the test — an allocating operand cannot collect a receiver held only in the expression (verified under `SPINEL_GC_STRESS`). A poly operand is unwrapped in place — tag, builtin id, then the pointer or structural test — never re-boxed from the receiver; a kind with no builtin id (Random) has no boxed form at all, so a poly slot never holds one and the answer is false.

**`case`/`when` answers the same way.** The case emitter's raw-pointer fallback now asks the arm first, so `case e2 when e` over two exceptions with the same message matches, as `e === e2` does, and a `case` over two Times compiles (it used to emit C that did not).

**A second tier answers the one question the typed equality arms refused rather than answered:** a Range, Array or Bignum against an operand of another concrete family is simply unequal (`(1..2) == [1, 2]`, `[1] == 1`, `(10**20).eql?("x")`), and two arrays of different storage kinds compare by value through the boxed comparator (`[1] == [1.0]`).

## Validation

- Full suite green (3,144 tests), including a new regression test whose expected output is CRuby 4.0.6's own (`test/native_object_protocol.rb`): every kind × every method, same-kind, cross-kind and poly operands, `freeze` then `frozen?`, `case`/`when`, the static streams, a nil handle slot, receiver evaluation order, and the cross-family tier.
- **40 previously-diverging corpus programs now behave identically, zero regressions** across the ~2,100-program corpus.
- Every semantic decision above was probed against CRuby 4.0.6 first, and the branch was then reviewed adversarially by four independent reviewers (semantics, the inference and cross-family tier, the generated C and GC safety, claims and coverage). What they caught is fixed here: an unguarded downcast in the poly unwrap, the standard streams reaching the GC-header freeze, the operand evaluated before the receiver (and the receiver dropped when the tag test failed), `File::Stat` treated as identity, a curried Proc's `===` treated as identity, `case`/`when` disagreeing with `===`, and by-value `equal?` answered where CRuby says the objects differ.

## Known gaps, stated rather than papered over

- A Random cannot be boxed at all today (`[Random.new(1)][0]` is nil on master), so a comparison against a poly slot holding one answers false. That is a boxing gap worth its own fix; this PR does not pretend otherwise.
- `Exception#==` compares class and message; CRuby also compares the backtrace, which Spinel does not populate (docs/limitations.md), so two exceptions raised at different sites with one message compare equal here.
- `$stdout.freeze` sets the flag and `frozen?` reads it back, but writes to a frozen IO do not raise FrozenError as CRuby's do.
- `Method#==` / `#eql?` (pre-existing arm) compare the receiver and the compiled function rather than receiver and method name, so two `1.method(:+)` are not equal here. Not touched by this PR, listed because the arm above sits next to it.
- A user exception subclass that defines `===`, `==`, `eql?` or `equal?` makes every exception-typed receiver's call to that method a compile-time rejection again (the old behaviour), since a rescued value carries no user class for dispatch.

## Not in this PR (deliberately)

- `hash` / `object_id` on these kinds — a different contract (must agree with `eql?`), left for when a program needs it.
- `Socket::Option` — its constructors are not available yet, so the protocol on it is moot.
- User classes that reopen a builtin handle class — the arm sits after user-method routing, so a user-defined `==` on one already wins.

## Summary by CodeRabbit

* **新機能**
  * ネイティブオブジェクトで `==`、`!=`、`===`、`equal?`、`eql?` の比較が一貫して利用できるようになりました。
  * `freeze` と `frozen?` による凍結状態の設定・確認に対応しました。
  * `File::Stat`、`Time`、`OpenStruct`、例外、IO などの比較精度を向上しました。
  * `case` / `when` でネイティブ型の比較が正しく処理されるようになりました。

* **テスト**
  * 各種ネイティブオブジェクトの比較・凍結動作を検証するテストを追加しました。


matz:
Merged as 9da41143. Thank you — a compile refusal is the worst place for a gap to sit, because the program cannot even rescue it, and this was the largest family of them.

The design choice that makes it reviewable is putting the arm at the tails of emit_call and emit_case_eq_call: it can only replace a rejection, never shadow an answer. Deciding what it answers in one function the inference also consults is the other half — those two drifting is exactly how this kind of fallback usually goes wrong.

Reviewed against master before merging: the gate is green (3150 tests, 61 benchmarks, optcarrot checksum 59662), a user class defining any of the seven keeps its own definition, and ordinary values are untouched. Probes turned up two build failures, and both reproduce on master unchanged, so neither is yours: a `Thread::Mutex` bound to a local cannot be typed at all, and a user `frozen?` returning a String collides with the builtin's Bool.

optcarrot costs 0.09% more instructions. I checked rather than assumed: the generated C for it is byte-identical, so it is a layout shift from the one byte added to `sp_File`, not new work.

One gap it opens, which I recorded in the merge rather than holding the PR for: `IO#freeze` now sets a bit that `frozen?` reads, but the write paths do not consult it, so a frozen `$stdout` still accepts output where CRuby raises FrozenError. Answering `frozen?` at all is much the larger win, and the write guard is a natural follow-up if you want it.

Issue #4097, opened 2026-08-24
[Compiler] A remapped diagnostic carries the C column onto the Ruby line, so the caret lands hundreds of columns past the source


rubys:
`#line` is on by default, so cc's errors are reported against the `.rb`. The **line** is
remapped; the **column** is not — the C column is printed against the Ruby line. Since
spinel emits a whole Ruby line as one C line, that column is routinely in the hundreds
or thousands, and it lands far past the end of the source line it is printed under.

Two things follow. The caret points at nothing, and cc pads its caret line out to the
column it was given, so every error costs a line of whitespace as wide as the C
expression was long.

## Reproduction

The `repro.rb` from #4096 — line 12 is `  x.values`, ten characters:

```
$ spinel repro.rb -o repro          # e58a92132690
repro.rb: In function ‘sp_f’:
repro.rb:12:615: error: incompatible types when assigning to type ‘sp_RbVal’ from type ‘sp_PolyArray *’
   12 |   x.values
      |
spinel: C compilation failed
```

The caret line under `x.values` is 636 characters wide. Column 615 is a position in the
generated C, printed against a line that ends at column 10.

## What would help

Drop the column when the location is remapped. `repro.rb:12: error: …` is honest and
readable; `repro.rb:12:615` claims a precision that does not exist and costs the caret
line to print. Keeping the column would only make sense if it were remapped too, which
for one-C-line-per-Ruby-line is not really available.

## Where it bites

Roundhouse's lobsters AOT lane compiles a tree where these columns run to 3154. A build
log of 23 errors is mostly caret whitespace, and a plain `tail` of it shows one error,
because one error is wider than the terminal. `scripts/lobsters-spinel-replay` carries a
workaround — grep the `error:` lines, trim each to 200 characters, and put the count in
the headline — written after a published run reported a single `sp_box_str` error when
the build had eight. The count is now right, but the columns still say nothing, and the
caret line is still discarded rather than read.

PR #4099, opened 2026-08-24
Ask clang for the caret flag it actually has

rubys:
Since 9e818906 (`Stop printing a C column against a Ruby line`), `make` fails on a clang host.

`-fno-diagnostics-show-caret` is gcc's spelling. clang does not ignore it — it refuses the argument and stops:

```
clang: error: unknown argument: '-fno-diagnostics-show-caret'
```

That is every build on such a host, spinel's own tools included: `make` gets as far as `bin/spinel` and then dies compiling `spin`, `spinel-flatten`, `spinel-doctor` and `spinel-reduce`, since those are compiled by the compiler that just grew the flag.

```
$ make -j8
...
clang: error: unknown argument: '-fno-diagnostics-show-caret'
spinel: C compilation failed
make: *** [bin/spin] Error 1
```

Measured on Apple clang 21.0.0 (clang-2100.1.1.101), arm64-apple-darwin25.5.0.

## The flag clang has does the same job

```
$ clang -c e.c
e.c:1:24: error: use of undeclared identifier 'undefined_thing'
    1 | int main(void){ return undefined_thing; }
      |                        ^~~~~~~~~~~~~~~

$ clang -fno-show-column -fno-caret-diagnostics -c e.c
e.c:1: error: use of undeclared identifier 'undefined_thing'
```

No column, no quoted line, no caret — the file and line the message already names, which is what 9e818906 was after. `-fno-show-column` is common to both compilers and stays where it is.

## The choice this makes

The spelling is picked by the compiler that **built** spinel, since that is the same toolchain it then drives as `cc` wherever the spelling differs. The gcc arm is untouched, so this cannot regress it.

A one-time probe of the configured `cc` would also cover a build whose `CC` differs from its host compiler, at the cost of one compile per run. Happy to write that instead if that combination is meant to be supported — I went with the smaller change since it restores the build without touching the gcc path.

## Verification

`make test` on this branch: **3228 pass, 0 fail, 0 error.**

No test in the diff: the harness compiles and runs Ruby, and this is a choice made about the `cc` command line before any of that. The seam that would hold one is a way to assert on the command `compile_cmd` builds.

## Summary by CodeRabbit

- **改善**
  - Apple clang 使用時の診断メッセージ表示を調整しました。
  - コンパイラに応じて適切な診断抑制オプションを選択し、不要な列・ソース引用・キャレット表示を抑制します。
  - デバッグまたは行番号マッピングが有効な場合にのみ適用されます。

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -538,10 +538,28 @@ int main(int argc, char **argv) {
      true (#4097). The caret goes with it -- cc draws it from the same column,
      so it points at nothing and pads to that width whatever the location says;
      the source-quote line is cc's only way to carry it and goes too, leaving
-     the file and line the message already names. Both gcc and clang take the
-     flags. */
-  if (debug || line_map)
-    s_add(&cmd, "-fno-show-column -fno-diagnostics-show-caret ");
+     the file and line the message already names. */
+  if (debug || line_map) {
+    s_add(&cmd, "-fno-show-column ");
+    /* clang spells the caret flag -fno-caret-diagnostics and REFUSES the
+       gcc spelling outright: Apple clang 21 answers "unknown argument:
+       '-fno-diagnostics-show-caret'" and stops, so on macOS every build
+       fails -- including spinel's own tools, which it compiles with itself.
+       Its own spelling gives exactly what is wanted here:
+       `e.c:1: error: ...`, no column, no quoted line, no caret.
+
+       Chosen by the compiler that BUILT spinel, since that is the same
+       toolchain it then drives as `cc` wherever the spelling differs. The
+       gcc arm is left exactly as it was, so this cannot regress it. A
+       one-time probe of the configured cc would also cover a build whose
+       CC differs from its host compiler, at the cost of a compile per
+       run. */
+#if defined(__clang__)
+    s_add(&cmd, "-fno-caret-diagnostics ");
+#else
+    s_add(&cmd, "-fno-diagnostics-show-caret ");
+#endif
+  }
   if (fiber_frame_guard) s_add(&cmd, "-Wframe-larger-than=65536 ");
   snprintf(tmp, sizeof tmp, "-I\"%s\" -I\"%s%cregexp\" ", lib_dir, lib_dir, PATH_SEP); s_add(&cmd, tmp);
   /* Compile the generated TU with the same threading define as the mt runtime


matz:
Thank you for this, and sorry for the way it ended up. You diagnosed it first and in more detail than I did, and the fix landed on master as 2282bf43 before I looked at the open PRs. That was my mistake: I went looking at a red CI run without checking whether someone had already filed the fix, and the commit does not carry a `Refs #4099` trailer it should have.

The question you raised in the description is the one that decided the shape of it:

> A one-time probe of the configured `cc` would also cover a build whose `CC` differs from its host compiler, at the cost of one compile per run. Happy to write that instead if that combination is meant to be supported

It is meant to be supported, so master went with the probe. `--cc=CMD` is a documented option (it is in `--help`), and `CC=` is an ordinary way to build. Measured, the build-time choice has a hole in both directions:

```
# spinel built by gcc, driving clang through the documented option
$ ./bin/spinel --cc=clang -e 'puts "hi"' -E
clang: error: unknown argument: '-fno-diagnostics-show-caret'

# spinel built by clang, on a host whose `cc` is gcc
$ CC=clang make -j8
cc: error: unrecognized command-line option '-fno-caret-diagnostics'
```

The second one needs no flag at all to reach: on a Linux host `cc` is usually gcc, so a `CC=clang` build of spinel dies compiling its own tools, which is the same failure mode you described from the other side.

The probe is one `cc --version` per run, taken only behind the flag that needs the answer and cached for the process, so a build that never emits the caret flag never pays for it. Your `-fno-show-column` reading carried over unchanged, including the point that it is the whole of what #4097 wanted on gcc, since with no column there is nothing to draw a caret from.

Closing as fixed, with the credit for finding it yours. If the probe turns out to be the wrong trade for some setup you have, please say so and I will revisit it.

matz:
Following up, because comparing the two builds side by side turned up something your version had right and mine did not.

I had passed the caret suppression to clang alone, reasoning that `-fno-show-column` was the whole of it on gcc. That reading came from measuring a diagnostic `#line` had remapped, and it does not generalise: gcc drops the quote line with the column only where the location IS remapped, since it then has no file to read the source line out of. Every other diagnostic still draws a caret from the generated file, and a line of generated C is as wide as a whole Ruby statement, which is the case #4097 was about:

    out.c:4: error: conflicting types for 'strlen'; have 'int(void *)'
        4 | extern int strlen(void *);
          |            ^~~~~~

Your patch passes gcc's spelling unconditionally on the gcc arm, so it never had this hole. Fixed on master in 0ccc9e17, which now passes the flag on both in each compiler's own spelling. Built from your branch and compared: byte-identical output on gcc, and no caret line survives on either compiler.

So the shape of the fix is yours in both halves, and the probe is only about which spelling to reach for. Thanks again, and sorry for the detour.

PR #4101, opened 2026-08-25
Fix String.new invoked with a poly-typed value

mauri:
## What

`String.new(x)` fails C compilation when `x` is a poly value — an element
read out of a nested array, rather than a flat array, a local, or a literal.
The diagnostic names generated runtime internals, not the Ruby expression at
fault:

```ruby
nested = [["class", "x"]]
::String.new(nested[0][1])
```

```
error: passing 'sp_RbVal' to parameter of incompatible type 'const char *'
  ... sp_String_new_shared(sp_str_dup_external(sp_poly_arr_get_hash(...)))
note: passing argument to parameter 's' here
  static inline const char *sp_str_dup_external(const char *s)
```

A flat array element, a local, and a literal all compiled fine — only a
value that stayed poly-typed through codegen hit this.

## Why

`String.new`'s codegen emitted its content argument with `emit_expr`, which
passes the argument through however it is statically represented. A flat
value is a plain `const char *`, matching `sp_str_dup_external`'s parameter.
A value that reaches codegen still poly-typed (e.g. an element pulled out of
a nested array, where inference doesn't narrow past the first level) is
represented at runtime as a boxed `sp_RbVal`, which has no implicit
conversion to `const char *` — hence the C type error.

## The fix

Emit the argument with `emit_str_expr` instead of `emit_expr`
(`src/codegen_call.c`). `emit_str_expr` already handles exactly this
argument-boundary case elsewhere in the codebase (`codegen_call_recv.c`'s
other `const char *`-typed builtin arguments): a plain `TY_STRING` value
still emits directly, unchanged from before, while a poly value is unboxed
through `sp_poly_arg_str_chk` — which raises `TypeError` for a non-String,
non-`to_str` argument the same way CRuby does for `String.new(nil)` /
`String.new(5)`.

## What the test pins

`test/string_new_poly_arg.rb`, golden-checked against CRuby's own output:

- `String.new` on an element read from a **nested** array (the bug) —
  `["class", "x"]` two levels down still compiles and returns `"x"`.
- The same for a **flat** array element, so the already-working case stays
  covered and a future regression shows up as a diff, not a silent
  narrowing.
- The result is an independent mutable copy, not aliased to the source.
- `String.new` on a poly-typed `nil` / `Integer` (reached through a
  variable-indexed heterogeneous array, so the element stays poly rather
  than being narrowed) still raises `TypeError`, with CRuby's exact message.

Confirmed the test actually catches the bug: `ERR` (C compilation failure)
with `emit_expr` reverted, `PASS` with the fix.

## Summary by CodeRabbit

* **バグ修正**
  * `String.new` が配列要素などの多様な値を適切に文字列化できるようになりました。
  * 変換できない値を渡した場合、CRuby と同様に `TypeError` が発生するようになりました。

* **テスト**
  * ネストされた配列、文字列の独立性、変換不能な値に関する検証を追加しました。

### DIFF
--- a/src/codegen_call.c
+++ b/src/codegen_call.c
@@ -6893,10 +6893,13 @@ static int emit_class_new_call(Compiler *c, int id, Buf *b) {
         /* String.new(str = "", capacity:, encoding:): always a mutable heap copy.
            The capacity/encoding keyword arguments are hints that do not change
            the value, so the content is the leading positional argument (when it
-           is not itself the keyword hash). */
+           is not itself the keyword hash). emit_str_expr emits it as a plain
+           `const char *` for sp_str_dup_external, unboxing a poly-typed
+           argument through sp_poly_arg_str_chk (which raises TypeError like
+           CRuby for a non-String, non-to_str value). */
         const char *a0ty = argc >= 1 ? nt_type(nt, argv[0]) : NULL;
         int has_content = argc >= 1 && !(a0ty && sp_streq(a0ty, "KeywordHashNode"));
-        if (has_content) { buf_puts(b, "sp_str_dup_external("); emit_expr(c, argv[0], b); buf_puts(b, ")"); }
+        if (has_content) { buf_puts(b, "sp_str_dup_external("); emit_str_expr(c, argv[0], b); buf_puts(b, ")"); }
         else buf_puts(b, "sp_str_dup_external((&(\"\\xff\")[1]))");
         return 1;
       }
--- a/test/string_new_poly_arg.rb
+++ b/test/string_new_poly_arg.rb
@@ -1,0 +1,36 @@
+# String.new(x) where x is a poly value -- an element read out of a nested
+# array, rather than a flat array, a local, or a literal -- previously failed
+# C compilation: the argument reached sp_str_dup_external (which takes a
+# plain `const char *`) still boxed as an sp_RbVal.
+nested = [["class", "x"]]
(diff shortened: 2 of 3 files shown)


matz:
Merging this. The description is accurate on every point I checked, and the reasoning about why only a poly-typed argument hits it is right.

What I verified beyond the PR's own test:

The bug reproduces on master exactly as described, and the fix answers it. The claim that a plain `TY_STRING` argument emits unchanged holds byte for byte: generated C for a corpus of sixteen `String.new` shapes (no argument, literal, local, expression, flat element, `.first`, method return, and the capacity/encoding keyword forms) is identical before and after, as is the generated C of all 40 tests in the tree that mention `String.new`. So this narrows nothing that was working.

The conversion boundary agrees with CRuby 4.0.4 across the whole range, not only the two the test pins: nil, Integer, Symbol, Float, Array, Hash and true all raise with CRuby's message, a String passes through, and an object carrying `#to_str` is accepted and converted. `make gate` is green on the branch.

The one thing worth saying out loud is that this is one instance of a shape that may have siblings: an emitter that hands `emit_expr` to a runtime function typed `const char *` works until an argument arrives poly. `String.new` is the site you hit; I will sweep the others separately rather than widen your PR, since each one needs its own measurement and its own test.

Thanks for the clear write-up, and for pinning the flat-array case alongside the nested one so a future narrowing shows as a diff.

PR #4102, opened 2026-08-25
Fix an empty collection literal miscompiled in a case branch

mauri:
## What

An empty `[]`/`{}` literal written as a `case`/`when` branch value compiles
to a null pointer statically typed as the branch's result type, rather than
an empty container. Since the pointer is null but the static type says
`Array`/`Hash`, a later method call on it skips the nil check and segfaults:

```ruby
r = case "s" when String then [] else ["x"] end
r.each { |e| puts e }
```

```
CRuby:  prints nothing, exits 0
spinel: Segmentation fault: 11
```

A non-empty array literal in the same position (`["a"] vs ["b"]`) compiles
correctly, as does an empty array reaching the branch through a local or a
method call rather than a literal — only the empty literal written directly
in branch position is affected.

## Why

`emit_case_branch_value` (`src/codegen_stmt.c`) decides a branch's tail is
"value-less" — an arm ending in something like a `puts` call or a writer,
which spinel emits as `void` C with no expression form — whenever the tail's
own cached type is `TY_UNKNOWN`, and in that case runs it for effect and
leaves the result slot at its type's default value.

An empty `[]`/`{}` literal also caches `TY_UNKNOWN` for its own type — not
because it has no expression form, but because an empty literal has no
element type until something supplies one. That's a different reason to be
`TY_UNKNOWN`, but the check couldn't tell them apart, so the literal's actual
construction (`sp_StrArray_new()` etc.) got built and then discarded, and the
branch fell through to the value-less path's default — a null pointer,
still statically typed as the container.

## The fix

Before the value-less-tail check, `emit_case_branch_value` now recognizes an
empty `ArrayNode`/`HashNode` tail specifically and builds it directly against
the case's own merged result type (`rt`, already computed for the whole
`case` expression) — the same role `g_ret_type` already plays for an empty
literal in a method's tail position, and the same thing `emit_ternary_arm`
already does for the identical literal in an `if`/`unless` value position.
When `rt` isn't itself a concrete array/hash type the fix doesn't apply and
falls through unchanged.

That's the scope of this fix: a branch's empty-literal tail merged with a
sibling of the **same container kind** (`[] vs ["x"]`, `{} vs {"a"=>1}`,
`[] vs []` via a concrete sibling). It does not touch the separate case
where `case`'s own type inference (`ty_unify`/`infer_type`) merges an empty
literal against an *incompatible* sibling type (a bare `case ... then []
else "str" end`, or no `else` at all) — that merge already produces a
non-container `rt` upstream of this function, in a different subsystem, and
is a distinct problem from the one this PR addresses.

## What the test pins

`test/case_empty_literal_branch.rb`, golden-checked against CRuby's own
output:

- an empty array literal branch merged with a non-empty array sibling
  compiles to a real empty array, not a crash, in both the if-chain `case`
  form and the `case <int>` switch/jump-table fast path (which lowers
  through the same emitter)
- the same for an empty hash literal branch merged with a non-empty hash
  sibling
- the empty literal branch specifically taken as the `else` arm (not just
  `when`)
- calling a method on the resulting value (`.each`) does not crash
- a non-empty literal in the same position is unaffected (regression guard)

## Summary by CodeRabbit

- **バグ修正**
  - `case` 式の分岐で空の配列・ハッシュリテラルを使用した際、null ポインターではなく、正しく空のコンテナが生成されるようになりました。
  - 通常の分岐や高速な評価経路、`else` 分岐でも一貫して正しい結果を返します。

- **テスト**
  - 配列・ハッシュの空リテラル、非空リテラル、数値・文字列の評価結果を検証するテストを追加しました。


matz:
Merging this. The diagnosis is exactly right, including the part that is easy to get wrong: an empty literal and a value-less tail both cache TY_UNKNOWN for different reasons, and the check could not tell them apart. Naming that is what makes the fix obviously correct rather than merely effective.

What I checked beyond the PR's own test:

The segfault reproduces on master (exit 139) and the branch answers it. Nothing else moves: generated C for the 50 tests in the tree that use `case` is identical before and after, so this narrows nothing. `make gate` is green on the branch.

The in-scope shapes hold across the range, not only the two the test pins: an empty array against a poly sibling, an empty hash against an Integer-keyed and a Symbol-keyed one, a nested sibling, and the empty branch reached through a local. All agree with CRuby 4.0.4.

Your scope note is accurate, and I measured it rather than taking it: `case ... then [] else "str"`, `else nil`, `else 1`, no `else` at all, and both branches empty are unchanged by this patch, and none of them crashes on either side -- they answer nil, "" or 0 where CRuby answers []. So this converts none of them from a crash into a wrong answer; it fixes the crash and leaves a separate, already-quiet problem alone. That is the right split.

I also swept the other value positions an empty literal can sit in, to see whether the same hole had siblings: ternary, if, unless, begin, a rescue arm, ensure, `||`, a while result, a method tail, an if and a case inside a method, and a block result. All fifteen answer [] and none crashes. The one thing that came out of that sweep is unrelated to your change and predates it -- `true && []` answers true rather than [], and the generated C carries a "pointer/integer type mismatch in conditional expression" warning. I will take that separately.

Thanks for the write-up, and for pinning the jump-table form alongside the if-chain one -- they lower through the same emitter, which is not obvious from the source.

PR #4117, opened 2026-08-26
Preserve positional and keyword arguments when forwarding `...`

fc-anjos:
## The bug

Ruby keeps positional and keyword arguments separate when `...` forwards them:

```ruby
def h1(*a, **k) = [a, k]
def h10(...) = h1(...)

p h10(1, k: 2)
```

Ruby returns:

```text
[[1], {k: 2}]
```

Spinel returned `[1, 2]`, because its forwarding lowered the call positionally and flattened the arguments before they reached `*a` and `**k`.

## The fix

When the compiler can determine that a forwarding call targets rest and keyword-rest parameters, it lowers the forwarding definition to anonymous `*` and `**` parameters. This preserves the same argument boundaries as Ruby.

Calls that may forward a block, or whose target cannot be resolved safely, keep the existing forwarding behavior. Block forwarding through `...` is unchanged.

## Tests

Added `test/arg_forwarding_to_rest_callee.rb`, covering positional and keyword arguments forwarded to `*rest` and `**kwrest`.

## Verification

The focused regression matches Ruby 4.0.6 for rest-only, keyword-rest-only, mixed, class-receiver, and fixed-arity forwarding.

## Summary by CodeRabbit

- **新機能**
  - 引数転送（`...`）を、可変長位置引数（`*rest`）やキーワード引数（`**kwrest`）を受け取る呼び出し先へ正しく展開できるようになりました。
  - 固定引数との組み合わせや、位置引数とキーワード引数の混在に対応しました。
  - 多数の引数転送・明示引数も制限なく処理できるようになりました。

- **バグ修正**
  - 匿名キーワード引数が位置引数へ重複して渡される問題を修正しました。

- **テスト**
  - クラス内メソッド、ネストした配列・ハッシュを含む引数転送の検証を追加しました。


matz:

PR #4118, opened 2026-08-26
Make bare block `next` return nil

fc-anjos:
## The bug

Ruby treats a bare `next` as the block's result value, so it contributes `nil` when the block is used by an iterator:

```ruby
p [1, 2, 3].map { |v| next if v == 2; v }
```

Ruby returns:

```text
[1, nil, 3]
```

Spinel returned `[1, 0, 3]` for an integer array. It inferred the element type from the block's other results, then exposed the typed array slot's zero value when the bare `next` skipped an element.

## The fix

Treat a bare `next` as producing `nil` during block-result inference and code generation. The result is then stored using the existing generic collection path, which preserves the `nil` value.

A bare `break` remains unchanged because it exits the iterator rather than producing an element.

## Tests

Added `test/block_next_nil_typed_map.rb`, covering array, range, and polymorphic receivers, including string and float element types.

## Verification

The focused regression matches Ruby 4.0.6 for array, range, and polymorphic block results, including string and float elements.

## Summary by CodeRabbit

* **バグ修正**
  * ブロック内の裸の `next` が、正しく `nil` として扱われるようになりました。
  * `break` や `next` を含むブロックで、戻り値の型がより正確に推論されるようになりました。
  * `next` によるスプラット値が、マッピング結果で配列として正しく保持されます。
  * 型付きマッピングや `compact`、`yield`、`proc` 呼び出しなどで、`nil` を含む結果が正しく保持されます。


matz:

PR #4119, opened 2026-08-26
Match Ruby's hash default behavior for derived hashes

fc-anjos:
## The bug

Ruby has specific rules for defaults on hashes derived from another hash:

```ruby
g = Hash.new { |h, k| h[k] = k * 2 }
p g.dup["yy"]
p g.except("a")["zz"]

h = { 1 => "a" }
h.default = "m"
p h.merge({ "x" => 2 })[9]
```

Ruby returns:

```text
"yyyy"
nil
"m"
```

Spinel dropped the default proc during some copies and merges, and incorrectly retained defaults for `except`. Cross-layout merges also lost a plain default.

## The fix

Derived hashes now follow Ruby's behavior:

- `dup`, `clone`, `merge`, and `compact` preserve the receiver's plain default and default proc;
- `except` returns a hash without either default;
- a cross-layout merge preserves a plain default when the receiver is promoted to the generic hash representation.

A layout-specific default proc cannot be carried across a cross-layout merge because the generic hash uses a different callback representation. Plain defaults are preserved in that case.

## Tests

Added regressions covering string-, symbol-, and generic-keyed hashes through `dup`, `clone`, `merge`, `compact`, and `except`, plus cross-layout merges with a plain default.

## Verification

The focused regressions match Ruby 4.0.6 for defaults through `dup`, `clone`, `merge`, `compact`, and `except`, including cross-layout merges.

## Summary by CodeRabbit

- **改善**
  - ハッシュの `dup`、`clone`、`merge`、`compact` で、デフォルト値やデフォルトブロックが適切に引き継がれるようになりました。
  - 異なる形式のハッシュをマージした場合も、レシーバーのデフォルト値やデフォルトブロックが保持されます。
  - `Hash#except` はデフォルト情報を持たないハッシュを返すようになりました。
  - `Hash#compact` はデフォルト情報を保持したまま、`nil` の値を除外します。
  - ガベージコレクション後も、デフォルトブロックが正しく機能します。


matz:

PR #4120, opened 2026-08-26
Preserve `nil` when nullable scalar results cross into generic values

fc-anjos:
## The bug

Ruby returns `nil` when a read or search has no result:

```ruby
def class_name(value) = value.class.name

a = [1, 2]
h = { "a" => 1 }

p class_name(a[5])
p class_name(h["missing"])
p class_name([].max)
```

Ruby returns:

```text
NilClass
NilClass
NilClass
```

Spinel returned `Integer` for these values. Spinel represents `nil` from typed integer and float operations with internal sentinel values. When such a result crossed into a generic Ruby value, ordinary boxing exposed the sentinel as a real number.

## The fix

Use nil-aware boxing whenever a nullable integer or float result crosses into a generic value, including array and hash reads, searches, selectors, `dig`, numeric conversions, locals, parameters, returns, and polymorphic collections.

Typed results continue to use their existing fast paths.

## Tests

Added `test/int_nil_boxing_at_poly_boundary.rb`, covering missed reads, integer and float arrays, searches, selectors, `dig`, numeric conversions, locals, returns, and polymorphic collections.

## Verification

The focused regression matches Ruby 4.0.6 for missed reads, searches, selectors, `dig`, numeric conversions, and generic-value boundaries.

## Summary by CodeRabbit

- **バグ修正**
  - 配列やハッシュの範囲外アクセス、空コレクション操作などの結果が、数値ではなく正しく `nil` として扱われるようになりました。
  - `first`、`last`、`find`、`detect`、`at`、`dig` などの呼び出しで、`nil` 判定や配列への追加が正しく機能します。
  - ブロック付きの `min`／`max` や空の `min`／`max`／`minmax` でも、結果が適切に `nil` として扱われます。
  - `Integer`／`Float` の非例外変換で返される `nil` も適切に処理されます。

- **テスト**
  - 整数・浮動小数点コンテナの `nil` 処理と、`min`／`max` の動作確認を追加しました。

### DIFF
--- a/lib/sp_alloc.h
+++ b/lib/sp_alloc.h
@@ -790,9 +790,11 @@ static inline const char*sp_encoding_inspect(sp_Encoding e){return sp_sprintf("#
 static inline sp_bool sp_encoding_eq(sp_Encoding a,sp_Encoding b){const char*an=sp_encoding_name(a);const char*bn=sp_encoding_name(b);return strcmp(an,bn)==0;}
 
 /* ---- Box helper prototypes (0 optcarrot uses; bodies in lib/sp_cold.c). ---- */
-sp_RbVal sp_box_int_or_nil(sp_int v);
+/* An int? / float? value crossing into a poly slot: the sentinel is nil, never
+   a number. Inline because every boxed element read goes through here. */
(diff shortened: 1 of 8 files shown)


matz:

PR #4121, opened 2026-08-26
Keep polymorphic locals polymorphic after mutator calls

fc-anjos:
## The bug

Ruby locals can hold values of different types over their lifetime:

```ruby
x = [1, 2]
r = x.pop
r = x.concat([3])
```

The first assignment makes `r` an integer. The second assigns an array, so Ruby simply lets the local change value and type.

Spinel failed to compile this program. A late type-capture pass saw that `concat` returns its array receiver and narrowed `r` to an array, even though an earlier write had already made `r` polymorphic.

## The fix

Only an array-only local may follow the receiver type of a self-returning mutator such as `push`, `concat`, or `sort!`. A local that already contains mixed values remains polymorphic, so later assignments use a compatible representation.

Array-only locals still follow the receiver when a mutator returns it; only mixed locals stay polymorphic.

## Tests

Added `test/local_rebind_after_self_returning_mutator.rb`, covering scalar and mutator writes in both orders, integer and string arrays, and array-only locals that should still follow receiver widening.

## Verification

The focused regression matches Ruby 4.0.6 for mixed and array-only locals, both write orders, and integer and string arrays.


matz:

PR #4122, opened 2026-08-26
Evaluate all output arguments before writing

fc-anjos:
## The bug

Ruby evaluates every argument to `puts`, `print`, and `p` before writing any of them:

```ruby
def a = (puts "eval-a"; 1)
def b = (puts "eval-b"; 2)

p a, b
puts 1, (raise "x")
puts [], 2
```

Ruby evaluates `a` and `b` before printing their values. It also prints nothing before the second call raises, and an empty array argument produces no output.

Spinel evaluated and wrote arguments one at a time. As a result, it could print `1` before evaluating `b`, print `1` before the exception, or print an unwanted blank line for `puts []`.

## The fix

When argument evaluation can have later side effects, collect all arguments first and write them only after evaluation finishes. Calls whose arguments are known to be safe continue to use their typed fast paths.

The runtime now also distinguishes an empty array argument from a call to bare `puts`.

## Tests

Added `test/output_args_eval_before_write.rb`, covering `p`, `puts`, `print`, raising and mutating arguments, splats, empty arrays, and user-class receivers.

## Verification

The focused regression matches Ruby 4.0.6 for `p`, `puts`, and `print` with side effects, exceptions, splats, empty arrays, and mutating receivers.


matz:

PR #4123, opened 2026-08-26
Bind pattern rest variables as ordinary Arrays

fc-anjos:
## The bug

Ruby binds a pattern's rest variable to an ordinary `Array`:

```ruby
def match(value)
  case value
  in [first, *rest] then rest.inspect
  in Integer then "int"
  end
end

puts match([1, 2, 3])
```

Ruby returns:

```text
[2, 3]
```

Spinel could return an empty or corrupted value when the matched array used one of its specialized typed-array representations. The pattern slice retained that representation, but the rest binding was then read as a generic Array.

## The fix

Materialize the pattern rest result as an ordinary Array before binding it. This makes the binding safe regardless of the representation of the value being matched.

The same conversion is used for find-pattern `*pre` and `*post` bindings and for the corresponding array-pattern forms.

## Tests

Added `test/pattern_rest_from_typed_array.rb`, covering ordinary and find-pattern rest bindings from typed arrays.

## Verification

The focused regression matches Ruby 4.0.6 for typed arrays, nested and middle rest bindings, and find-pattern `*pre`/`*post` bindings.


matz:

PR #4124, opened 2026-08-26
Preserve outer exception handlers across `ensure` exits

fc-anjos:
## The bug

Ruby runs an `ensure` clause when a `return` or `next` leaves a rescue or else clause, without removing exception handlers belonging to the caller:

```ruby
def evaluate
  raise "x"
rescue
  return "r"
ensure
  puts "always"
end

p evaluate

begin
  raise "boom"
rescue => error
  p error.message
end
```

Ruby prints:

```text
always
"r"
"boom"
```

Spinel printed the first two results, but then treated the final exception as uncaught. While performing the deferred return, it removed the caller's exception handler.

## The fix

Deferred exits now remove only the frames that still belong to the exited `ensure` region. When a rescue or else clause has already removed that region's frame, the exit may remove zero additional frames instead of forcibly popping the caller's handler.

## Tests

Added `test/rescue_return_with_ensure.rb`, covering returns from rescue and else clauses, `next` from rescue inside a loop, nested `ensure` regions, and later rescue clauses.

## Verification

The focused regression matches Ruby 4.0.6 for returns and `next` through rescue, else, and `ensure`, including nested handlers.

### DIFF
--- a/src/codegen_stmt.c
+++ b/src/codegen_stmt.c
@@ -5097,10 +5097,10 @@ void emit_return(Compiler *c, int id, Buf *b, int indent) {
       }
     }
     {
+      /* inside a rescue/else clause the region's frame is already popped, so
+         0 is a valid count; popping one anyway takes a caller's handler */
       int pops = g_exc_frame_depth - ctx->exc_base;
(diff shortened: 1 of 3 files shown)


matz:

PR #4125, opened 2026-08-26
Do not compile impossible `is_a?` branches for known scalars

fc-anjos:
## The bug

Ruby can answer `is_a?` from a value's known type. That means an impossible branch does not need to be evaluated or compiled:

```ruby
value = h[key]   # known to be an Integer

puts(value.is_a?(Array) ? value.map(&:to_s).join(",") : value.to_s)
```

Ruby takes the `else` branch. Spinel still emitted the `Array` branch, and compilation failed because `map` was being generated for an integer value.

## The fix

Use the known scalar type to fold supported type predicates (`is_a?`, `kind_of?`, and `instance_of?`) in expression-form conditionals, so an impossible branch is not emitted. Cases where the answer depends on the runtime value or receiver identity remain dynamic. Existing user-defined-class handling is unchanged.

## Tests

The regression covers the ternary case above, boolean receivers, module targets, and `instance_of?`, with Ruby 4.0.6 as the reference.

## Verification

The regression passes against Ruby 4.0.6 for scalar ternaries, boolean receivers, module targets, and `instance_of?`.

### DIFF
--- a/src/codegen_expr.c
+++ b/src/codegen_expr.c
@@ -2470,16 +2470,15 @@ else {
       else_stmts = nt_ref(nt, sub, "statements");
     int en = 0;
     const int *eb = else_stmts >= 0 ? nt_arr(nt, else_stmts, "body", &en) : NULL;
-    /* A statically-answered defined? predicate folds to its live arm alone:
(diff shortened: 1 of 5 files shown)


matz:

Issue #4115, opened 2026-08-26
spin: useless cache


rofl0r:
spin defaults to cache object files in some hidden XDG_WHATEVER dot-files path and when those cache file exists it behaves differently than when they don't, making it hard to debug. for example me trying to figure out why it doesn't add -I/path/to/spinel/lib https://github.com/matz/spinel/issues/4105#issuecomment-5420613356 when i run it from the project root directory, but adds it just fine when i run it from the test_curses directory. apart from being annoying, it's also pretty much useless. what it caches are object files compiled from handwritten .c files which compile in milliseconds, in comparison to spinel's .rb <-> .c translation which takes *minutes*. totally caching the wrong thing here.

so my honest ask: add an option to disable the object file caching, the refusal to rebuild when the build artifact already exists, or remove the caching completely. at the very least let the cache path be overridable.

matz:
The include-path half is a plain bug and is fixed in `5db19457`. The cache half gets two knobs, and one part of your criticism stands as written.

## The `-I` difference was not the cache

spin adds `-I <spinel>/lib` when it compiles package C, so `"spinel/runtime.h"` resolves -- but it gave up the moment the compiler came from PATH instead of sitting beside spin. `spin install` puts spin in `~/.local/bin` with no spinel next to it, so from an installed spin every package carrying C failed on a missing runtime.h, while the same package built fine from a tree where the two were co-located. That is exactly the shape you described: curses built once under the test app and then would not build from your repo root.

The cache made it *look* like the cache: the earlier object was still there, so the run that would have failed did not have to compile anything. Two different behaviours with the same cause underneath.

It resolves the compiler through PATH now, and `SPINEL_HDR_DIR` overrides the search for a layout it cannot work out.

## The knobs

- `SPIN_NATIVE_CACHE=<dir>` -- put the objects somewhere you can delete
- `SPIN_NO_NATIVE_CACHE=1` -- recompile package C every time

Your reason for wanting them is the one I would give: a run that behaves differently depending on whether an object happens to be there is the worst thing to be holding while you work out why two runs differ. Both are covered in the e2e check so they do not quietly stop working.

## On "caching the wrong thing"

You are right about the proportions, and it is worse than you guessed. Measured on a 21k-line Rails-shaped application: parse 0.08s, **inference 26.4s (87%)**, code generation ~0, C compilation 4.1s. `-O0` moves that 4.1s barely at all. So yes -- the objects are the cheap half by an order of magnitude.

[...]

PR #4126, opened 2026-08-26
spin tool: realpath fix for PATH-resolved compiler

rofl0r:
`tools/spin.rb`: in `spinel_hdr_dir` (line ~255), after the PATH walk via `which("spinel")`, add `File.realpath` on the found path when it's a symlink. This resolves symlinked installations (e.g. `~/bin/spin -> ~/spinel/bin/spin`, `spin install` output) to their real install root so the downstream `lib` probe finds the runtime headers.

`spinel_bin` first tries `$0`-sibling resolution (`$0` = binary path). When that fails (e.g. `spin` invoked from a different cwd), it falls back to the string `"spinel"`. Previously `spinel_hdr_dir` then called `which("spinel")` and used the raw result - which is the symlink itself. The install layout probe (`File.expand_path("..", bin)` + `lib/`) then walked from the symlink target's parent, missing the real `../lib` that sits beside the compiler binary.
`File.realpath` collapses the symlink chain so the walk lands in the correct install tree.

## Summary by CodeRabbit

* **バグ修正**
  * シンボリックリンク経由で起動された場合でも、ランタイムヘッダーを正しいインストール先から探索できるようになりました。

### DIFF
--- a/tools/spin.rb
+++ b/tools/spin.rb
@@ -252,7 +252,13 @@ def spinel_hdr_dir
   env = ENV["SPINEL_HDR_DIR"].to_s
   return env if env != "" && File.exist?(File.join(env, "spinel_rt.h"))
   bin = spinel_bin
-  bin = which("spinel") if bin == "spinel"
+  if bin == "spinel"
+    found = which("spinel")
+    # PATH returns the symlink as-is; the lib probe below then walks
+    # relative to a symlinked path and misses the install tree.
+    found = File.realpath(found) if found != "" && File.symlink?(found)
+    bin = found
+  end
   return "" if bin == ""
   d = File.expand_path("..", bin)
   a = File.join(d, "lib")


matz:
Merged in `2b1e507d`. Verified in worktrees both ways: with a package whose C includes "spinel/runtime.h", symlinked into a bin directory with no lib beside it, master fails with `native compile failed` and this passes.

Your diagnosis is right and mine was incomplete -- I fixed the PATH resolution yesterday without following the link, so an install that symlinks (which `spin install` produces) still walked from the wrong parent.

Added an e2e case in `af27aee3` so it stays fixed. Worth saying that it took three tries, and two of the failures are interesting:

- the first version used the existing test package, whose carried C includes only `<stdint.h>` -- it compiles with or without the runtime include path, so the check passed against the bug
- the second still passed because an object cached by an earlier case was reused and nothing compiled at all. That is exactly the masking you described in #4115, showing up in our own test suite the day you reported it. `SPIN_NO_NATIVE_CACHE=1` is what makes the case run.

Thanks for both this and #4115.

PR #4132, opened 2026-08-26
[BLOCKED on #4134] Take a URI and an initheader, and open a connection #request needs

rubys:
Three places where a request written in the CRuby spelling did not run under
the `net` package. All three sit in the same four lines a caller writes to POST
to a URL it has already parsed — this is [once-campfire's
`Webhook#deliver`](https://github.com/basecamp/once-campfire/blob/main/app/models/webhook.rb),
reduced:

```ruby
require "net/http"
require "uri"

uri = URI("http://127.0.0.1:9/hook")
req = Net::HTTP::Post.new(uri, "Content-Type" => "application/json")
req.body = "{}"
p req.path
p req["content-type"]
http = Net::HTTP.new(uri.host, uri.port)
begin
  http.request(req)
rescue => e
  puts "#{e.class}: #{e.message}"
end
```

| | before | CRuby |
|---|---|---|
| `Post.new(uri, hdrs)` | `ArgumentError: wrong number of arguments (given 2, expected 1)` | `"/hook"` |
| `req["content-type"]` | `nil` | `"application/json"` |
| `#request` unstarted | `Net::HTTPError: not started` | `Errno::ECONNREFUSED` |

### What changed

* **`HTTPRequest#initialize` takes an `initheader`**, and the five verb
  subclasses forward it. A URI argument contributes its `request_uri` — the
  path with the query — not its `to_s`, which is the whole URL and would go
  out as an absolute request line.

* **`#request` opens a connection when there is none.** CRuby's `#request` on
  an unstarted `Net::HTTP` starts it, runs the request and finishes, which is
  what a caller writes when the connection is configured somewhere other than
  where the request is sent. Doing that here rather than raising keeps the two
  spellings interchangeable, as they are there.

* **Request header lookup is case-insensitive.** A header stored as
  `"Content-Type"` read back as `nil` through `req["content-type"]`. Names are
  case-insensitive on the wire, CRuby's `Net::HTTPHeader` is case-insensitive
  on both halves, and the *response* half of this file already was — the
  request half was the odd one out. The caller's spelling is still what is
  stored, because for a request that is what goes on the wire.

### Test

`packages/net/test/net_http_client.rb` grows the four lines above end to end
against the server it already stands up, so all three are gated. `.expected`
is CRuby's own output for the same file, and the pre-existing lines in it are
unchanged. Reverting `net/http.rb` alone takes the test to FAIL (it was worth
checking — `make`'s exit status is 0 either way; the verdict is the contents of
`build/test-results/pkg.net.net_http_client.ok`).

`make test`: **3254 pass, 0 fail, 0 error** on macOS.

### Not in this PR

Two things this same call site wants that are not package-level fixes, noted
here so the gap is visible rather than filed as a surprise later:

* `open_timeout` / `read_timeout` are stored in `initialize` and never read
  again, and `Net::OpenTimeout` / `Net::ReadTimeout` do not exist — so a
  timeout is accepted and silently does nothing, and the `rescue` campfire
  writes around it cannot be written. That needs a socket-level timeout, so it
  wants its own issue.
* The streaming surface (`#request` with a block, `#read_body` with a block,
  `start(ipaddr:)`) — deliberately absent per the package header. It matters
  for fetching an *untrusted* URL, where reading the body whole is unbounded
  and connecting by hostname re-resolves past a DNS check; I'd rather raise
  that as a use case than assume it.

## Summary by CodeRabbit

- **新機能**
  - ヘッダー名を大文字・小文字にかかわらず同一のものとして扱えるようになりました。
  - 複数の値を持つヘッダーを、カンマ区切りで送信・取得できるようになりました。
  - ヘッダーを設定しても、送信時には指定した表記を保持します。

- **バグ修正**
  - 同名ヘッダーの上書きや取得が正しく行われるようになりました。
  - ヘッダーの重複送信を防ぎ、HTTPリクエストの動作を安定させました。

### DIFF
--- a/packages/net/net/http.rb
+++ b/packages/net/net/http.rb
@@ -85,19 +85,46 @@ module Net
     attr_reader :method, :path
     attr_accessor :body
 
-    def initialize(method, path)
+    # `path` is a request path, or a URI -- CRuby takes either, and
+    # `Net::HTTP::Post.new(uri, "Content-Type" => "application/json")` is the
+    # spelling a caller reaches for when it already parsed the URL. A URI
+    # contributes its `request_uri` (the path with the query), not its `to_s`,
+    # which is the whole URL and would go out as an absolute request line.
+    def initialize(method, path, initheader = nil)
       @method = method
-      @path = (path.nil? || path.to_s.empty?) ? "/" : path.to_s
+      @path = if path.is_a?(URI::Generic)
+        path.request_uri
+      else
+        (path.nil? || path.to_s.empty?) ? "/" : path.to_s
+      end
       @headers = {}
       @body = ""
+      initheader.each { |k, v| @headers[k.to_s] = v.to_s } unless initheader.nil?
     end
 
+    # Header names are case-insensitive on the wire, and CRuby's
+    # Net::HTTPHeader is case-insensitive on both halves -- as the response
+    # half here already is. What is kept, rather than downcased the way the
+    # response stores them, is the caller's spelling: for a request that is
+    # what goes out on the wire.
     def []=(name, value)
+      existing = header_name(name)
+      @headers.delete(existing) unless existing.nil?
       @headers[name.to_s] = value.to_s
     end
 
     def [](name)
-      @headers[name.to_s]
+      k = header_name(name)
+      k.nil? ? nil : @headers[k]
+    end
+
+    # The stored spelling of a header, in whatever case the caller asks for it,
+    # or nil when it is not set.
+    def header_name(name)
+      want = name.to_s.downcase
+      found = nil
+      @headers.each_key { |k| found = k if k.to_s.downcase == want }
+      found
     end
 
     def each_header
@@ -120,32 +147,32 @@ module Net
     # CRuby builds requests as Net::HTTP::Get.new(path) and friends. Same
     # spelling here, so the same code compiles.
     class Get < HTTPRequest
-      def initialize(path)
-        super("GET", path)
+      def initialize(path, initheader = nil)
+        super("GET", path, initheader)
       end
     end
 
     class Post < HTTPRequest
-      def initialize(path)
-        super("POST", path)
+      def initialize(path, initheader = nil)
+        super("POST", path, initheader)
       end
     end
 
     class Put < HTTPRequest
-      def initialize(path)
-        super("PUT", path)
+      def initialize(path, initheader = nil)
+        super("PUT", path, initheader)
       end
     end
 
     class Delete < HTTPRequest
-      def initialize(path)
-        super("DELETE", path)
+      def initialize(path, initheader = nil)
+        super("DELETE", path, initheader)
       end
     end
 
     class Head < HTTPRequest
-      def initialize(path)
-        super("HEAD", path)
+      def initialize(path, initheader = nil)
+        super("HEAD", path, initheader)
       end
     end
 
@@ -259,7 +286,20 @@ module Net
     end
 
     def request(req)
-      raise HTTPError, "not started" unless @started
+      # CRuby opens a connection for a #request on an unstarted Net::HTTP, runs
+      # the request over it and closes it again -- so `Net::HTTP.new(host,
+      # port).request(req)` works without a `start`, and is the spelling a
+      # caller writes when the connection is configured somewhere other than
+      # where the request is sent. Doing that here rather than raising keeps
+      # the two spellings interchangeable, as they are there.
+      unless @started
+        start
+        begin
+          return request(req)
+        ensure
+          finish
+        end
+      end
       # Every request goes out with `Connection: close`, so the server hangs
       # up after answering and the socket a second request would use is dead.
       # CRuby reconnects transparently in that situation, and a caller writing
--- a/packages/net/net/http.rb
+++ b/packages/net/net/http.rb
@@ -99,7 +99,7 @@ module Net
       end
       @headers = {}
       @body = ""
-      initheader.each { |k, v| @headers[k.to_s] = v.to_s } unless initheader.nil?
+      initheader.each { |k, v| self[k] = v } unless initheader.nil?
     end
 
     # Header names are case-insensitive on the wire, and CRuby's
@@ -107,15 +107,24 @@ module Net
     # half here already is. What is kept, rather than downcased the way the
     # response stores them, is the caller's spelling: for a request that is
     # what goes out on the wire.
+    # EVERY write goes through here -- `initheader`, `content_type=` and
+    # `set_form_data` included. Writing to `@headers` directly is how the same
+    # header ends up stored twice under two spellings and goes out on the wire
+    # twice: `Post.new(uri, "content-type" => "text/plain")` followed by
+    # `req.content_type = "application/json"` sent BOTH, and a server is free
+    # to believe either one.
     def []=(name, value)
-      existing = header_name(name)
-      @headers.delete(existing) unless existing.nil?
+      delete_header(name)
       @headers[name.to_s] = value.to_s
     end
 
     def [](name)
       k = header_name(name)
       k.nil? ? nil : @headers[k]
+    end
+
+    def key?(name)
+      !header_name(name).nil?
     end
 
     # The stored spelling of a header, in whatever case the caller asks for it,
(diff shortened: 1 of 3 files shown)


rubys:
Both findings were real; reproduced each before fixing, pushed in bcb4d1e0.

**Duplicate header.** Confirmed against CRuby with a server that records
the raw request lines:

```ruby
req = Net::HTTP::Post.new(uri, "content-type" => "text/plain")
req.content_type = "application/json"
```

CRuby sends `Content-Type: application/json`. This package sent **both**
`content-type: text/plain` and `Content-Type: application/json`. Every
write now routes through `#[]=`, and `#[]=` clears every spelling of the
name rather than the single one `header_name` returns — a collision that
predates the write has to be cleared whole, or the duplicate survives the
overwrite meant to replace it. `#key?` moved onto the same rule since it
is the third reader of it.

The regression test records how many `Content-Type` *lines* the server
saw. That detail matters: the header hash the test already built is keyed
downcased, so the two spellings collapsed there and the duplicate would
have gone unseen. Reverting the header fix alone turns that line from
`content-type-lines=1` to `=2`.

**Socket on a failed start.** Also right. `open_connection` assigns
`@socket` and only then completes the handshake, so a TLS failure raised
with a live socket and nothing to close it. `start` is inside the
begin/ensure now; `finish` is a no-op when nothing is open, which covers
the connect-refused case where it raises before `@socket` exists.

I did **not** add a test for that one, and would rather say so than imply
it is gated: reproducing it needs a peer that accepts TCP and then fails
TLS, which means the openssl package — and `PKG_TESTS` filters out
openssl's tests where the system headers are absent, so a net test
depending on TLS would be one that silently does not run on some
machines. The fix is reasoned, not measured.

`make test`: 3254 pass, 0 fail, 0 error.

One note for @matz rather than a change: the same `start`-outside-the-
`begin` shape is in `self.start`, `self.get_response` and
`self.post_form`, and predates this PR. Same leak, same two-line fix. I
left them alone to keep this PR to what it introduced — happy to fold
them in here or send a separate PR, whichever you prefer.

rubys:
**Blocked on #4134 — please don't merge this yet.**

Chasing the array-header finding turned up a SIGSEGV that this PR
introduces the trigger for, and that has been latent since its first
commit:

```ruby
require "net/http"
a = Net::HTTP::Post.new("/f")
b = Net::HTTP::Post.new("/a", { "content-type" => "text/plain" })
puts "ok"
```

CRuby prints `ok`; spinel exits 139. Constructing the same class once
with the new optional `initheader` and once without. Passing the default
explicitly — `Post.new("/f", nil)` — makes it exit 0, so it is the
omitted-argument path, not the Hash. Before the crash there is a silent
stage: the object built by the omitted call gets a `@headers` that
accepts a write and reads back empty. Details and what I could not
reduce are in #4134.

I have to correct something I wrote earlier in this PR. I reported
`make test`: 3254 pass, 0 fail, 0 error, and that was true — but the
suite never constructed a request both ways, so it never reached the
combination. The green suite was not evidence this was safe, and a new
optional parameter with no test exercising both arities is exactly the
coverage question I should have asked.

### On the array-header finding

The `#[]=` half was right and is fixed locally:
`req["X-Multi"] = %w[one two]` gave `"[\"one\", \"two\"]"` where CRuby
gives `"one, two"` — `to_s` on an Array is its inspect form, so that went
on the wire as `X-Multi: ["one", "two"]`.

The `initheader` half of the suggestion is not CRuby's behaviour, so I
did not implement it as described: CRuby's `initialize_http_header` calls
`value.strip` on each value and raises `NoMethodError` for an Array
rather than joining. A test asserting initheader joins to `"a, b"` would
have encoded something upstream does not do. This package routes
initheader through `#[]=` and so accepts an Array — a deliberate
divergence in the permissive direction, noted in the code.

I also replaced the scan-based header storage with downcased keys plus a
parallel map of the caller's spelling. That makes a duplicate impossible
to construct rather than something each write has to hunt for, turns
lookup into a hash hit instead of an O(n) scan, and matches how the
response half of the file has always stored headers. It is held back with
the array fix until #4134 clears, since the package test cannot run
through the segfault.

rubys:
Correcting myself: I said the array fix was "fixed locally" and "held
back", which told you it existed but left you nothing to take. It is
pushed now — 78676324 — along with the header-storage rewrite, and
`make test` is 3254 pass / 0 fail / 0 error on the branch.

I had assumed gating the array fix required constructing a request
without the optional initheader, which is the #4134 arity pair. It does
not: the assertion rides the request `/dup` already builds, so the fix is
covered without going near the crash. Reverting the join alone turns the
recorded wire line into `X-Multi: ["one", "two"]`.

So the branch now carries three separable things, and only the first is
stuck behind #4134:

1. **initheader + auto-start** (8b7d67fb) — the feature. This is what
   introduces the optional parameter, and so the arity pair that
   segfaults. Blocked.
2. **duplicate header + socket-on-failed-start** (bcb4d1e0) — both real,
   both reproduced against CRuby first.
3. **array join + downcased header storage** (78676324) — independent of
   the optional parameter entirely; it is about what `#[]=` does with the
   value it is given.

2 and 3 stand on their own if you would rather take them ahead of
#4134 — they touch the same file but not the same mechanism, and I am
happy to split them into their own PR if that is easier to land.

rubys:
Agreed on the principle — I made the same criticism of this PR myself a
few comments up: a new optional parameter with no test exercising both
arities is exactly the coverage question I failed to ask.

I can't add it yet, because it is the crash. `Post.new(path)` beside
`Post.new(path, headers)` in one program is #4134 — spinel exits 139 on
it. (The web search in the review couldn't find #4134; it was filed
minutes earlier.)

Your comment did sharpen the bug, though, so thank you for it. It made me
ask why the suite stays green while containing an omitted-initheader
construction already, and the answer is that the trigger is **one class
built both ways**, not two call sites:

| | exit |
|---|---|
| `Post.new("/f")` + `Post.new("/a", {…})` | **139** |
| `Get.new("/f")` + `Get.new("/a", {…})` | **139** |
| `Get.new("/f")` + `Post.new("/a", {…})` | 0 |
| `Post.new("/f")` + `Get.new("/a", {…})` | 0 |

The suite has `Get.new("/chunked")` omitted alongside `Post.new(uri,
headers)` supplied — a cross-class pair, which is why it passes. #4134 is
updated with that table and the two-line repro.

So the test you're asking for is the right regression gate for #4134, and
I'd rather it land there, on a compiler that can run it, than be added
here where it would only turn the suite red for a reason that has nothing
to do with the header semantics this PR changes.

On the other finding in this round — the Array-valued `initheader` — your
own research in that thread quotes CRuby's `set_field` correctly, and
that is what 78676324 implements: `#[]=` joins with ", ". Note the
`initheader` half of the suggestion still doesn't apply, for the reason
in my earlier reply: CRuby's `initialize_http_header` calls `value.strip`
per value and raises NoMethodError on an Array rather than joining, so a
test asserting initheader answers `"a, b"` would encode behaviour
upstream does not have. This package accepts one because it routes
initheader through `#[]=`, which is a deliberate divergence and is
commented as such.

matz:
Merged in `ff6229ad`, after fixing #4134 in `096f5030`.

All three of your cases match CRuby on current master:

| | before | now | CRuby |
|---|---|---|---|
| `Post.new(uri, hdrs)` | ArgumentError | `"/hook"` | `"/hook"` |
| `req["content-type"]` | nil | `"application/json"` | `"application/json"` |
| `#request` unstarted | Net::HTTPError | Errno::ECONNREFUSED | Errno::ECONNREFUSED |

Your `packages/net/test/net_http_client.rb` output is byte-identical to CRuby's for the same file.

The blocker was worth the detour: the bug you reduced was not about optional parameters at all, it was hashes missing from the nullable boxing that objects already had. Your two diagnostics -- explicit `nil` behaving differently from omitting, and needing two call sites -- are exactly what a NULL-boxed-as-truthy produces, and they took the search straight to it.

One note on `#request` opening a connection when there is none: that is right and matches CRuby, and it composes with the reconnect I added on the 26th, so a caller can now write `Net::HTTP.new(...).request(req)` or the `start` block form interchangeably, and repeat either.

`make gate` ALL GREEN, optcarrot checksum 59662.

PR #4144, opened 2026-08-27
Convert, refuse or miss at every typed argument slot the way CRuby does

blacktm:
## Why

This PR makes the implicit conversion protocol work at the typed argument slots of the core methods: an object answering `#to_str`, `#to_int`, `#to_ary` or `#to_hash` handed to a builtin now converts the way CRuby converts it, and a value of the wrong class now raises CRuby's rescuable `TypeError` (or misses, where CRuby's answer is a miss) instead of failing the C build.

Today both of those programs are ones Spinel cannot compile at all. The argument slots of methods like `String#chomp`, `#ljust`, `#split`, `Time#strftime`, `Time.at`, `Array#slice!` or `format` are typed `const char *` or `sp_int` in the generated C, and the arms placed the argument's value in the slot raw — so a program that passes anything else dies in the C compiler with an error about a symbol the author never wrote:

```ruby
class Sep
  def to_str = "-"
end
p "a-b-".chomp(Sep.new)
# CRuby:  "a-b"
# Spinel: error: incompatible pointer types passing 'sp_Sep *' to parameter of type 'const char *'
```

That is one root cause with two user-visible faces. A program using the conversion protocol — the way duck typing crosses into the core classes, and what the last two conversion rounds were for — does not build, even though the helpers that implement the protocol have existed since those rounds; these arms simply never routed through them. And a program passing a genuinely wrong class, which under CRuby raises a `TypeError` it could rescue and handle, cannot even start.

Rather than fix the sites a differential corpus happened to hit, this round enumerates the population. `tools/gen_conv_slot_probe.rb` asks CRuby, for the positional slots of every method in the compiler's arity table (one receiver value per class, up to three arguments), what it does with an object answering `#to_str` / `#to_int` / `#to_ary` / `#to_hash` and with a scalar of the wrong class, then compiles the same call with Spinel. On the master this branch starts from it finds 210 failing probes over 108 argument slots in 71 methods whose generated C does not compile. On this branch, 18 remain — all one already-known inference question, detailed below.

## How

The probe's results fall into three slot families, and each now follows its own CRuby rule instead of a per-method fix:

- **Conversion slots** ask `#to_str` or `#to_int` and raise the implicit-conversion `TypeError` for any other class — the existing `emit_str_expr` / `emit_int_expr` helpers, now reached from every arm the probe flagged. The float slots follow `rb_to_float`, which converts only a `Numeric` and says `can't convert X into Float` for the rest (a plain object's `#to_f` is not asked — CRuby does not either). `String#split`'s separator is the pattern family and says `wrong argument type X (expected Regexp)`; `Float#quo` is arithmetic and says `String can't be coerced into Float`; `File.utime`'s times say `can't convert X into time` and `sleep`'s interval says `can't convert TrueClass into time interval`, naming the class the way `rb_time_interval` does. `Regexp.escape` takes a Symbol by its name, which `rb_reg_operand` accepts where `Regexp.union` genuinely refuses it.
- One ordering rule joins them: a conversion belongs below the nil-receiver dispatch guard. The rooted-temp hold that #4057 put in front of a call would hoist a `#to_str` above the guard's `NoMethodError`, running a conversion CRuby never asks for on a call that does not dispatch; the hold now stands down on a guarded arm and the conversion renders inside the guarded body.
- **Lookup slots** convert nothing. A Hash key, an element for `Array#count` / `#delete` / `#index`, a `Range#include?` operand: a value of a class the table cannot hold evaluates for its effects and misses, the way a boxed key of another tag already did (`emit_hash_key` gains the static arm; `hash_key_misses` and `elem_kind_misses` name the rule, the first by `eql?`, the second by `==`, which is why an Integer still finds a Float element). `fetch` and `fetch_values` name the key itself in their `KeyError`, boxed once.
- **Container slots** ask `#to_ary` or `#to_hash`: `Array#zip`, `#product` and `Hash#merge` take an object that answers them, typed as the container it returns, and raise CRuby's `TypeError` for one that does not.

What stays a compile-time refusal is named as such, where the raw pointer used to reach the C compiler: a user object whose class defines `==` (or `<=>`) looked up in a typed container — the slot has no room for it and CRuby would call the method — a `#to_hash` of another layout than the receiver's, a block that receives a missing key of another class than the hash's keys (`fetch { }`, `delete { }`, `fetch_values { }`), a literal that would widen a typed Array or Hash (`[1, 2].unshift("a")`, `{1 => 2}["a"] = 3` — an inference question), `system` with an environment Hash, and the dual-typed slots (`gets(limit)`, `File.open(fd)`, `readlines(limit)`) that belong to a later round.

## Validation

- `test/typed_slot_conversion.rb` (expectations generated by CRuby) covers the three families, the `KeyError` wording, and single evaluation of a key that misses. Under `SPINEL_GC_STRESS=1` it runs clean except for two lines whose `[e.class, e.message]` the pre-existing rescue-local hole below corrupts.
- The full suite: 3,283 pass, 0 fail, including two new compile-time refusal pins under `test/rbs-seed`.
- Against a differential corpus of ~2,100 minimized programs compared with CRuby: 63 programs go from a C build failure or wrong output to matching, 0 regress (37 from the ill-typed-C family the corpus had flagged, the other 26 nil/boolean/wrong-class cases at the same slots that the rules cover for free).
- `tools/gen_conv_slot_probe.rb` is included so the population can be re-measured. On this branch the 210 failing probes become 18, and all 18 are the same shape: a literal widening a typed Array or Hash (`[1, 2].unshift("a")`, `insert`, `prepend`, `{1 => 2}["a"] = 3`, `store`), which is an inference question rather than a slot one.

## What the probe shows that this round leaves alone

- The probe's `differ` column (a slot that neither converts nor refuses, answering something else) is not swept here; it is the next round's list. The ones worth knowing: `Hash#delete` on a typed-value hash answers the value's zero on a miss (`{1 => 2}.delete(3)` is `0`), `fetch { |k| }` hands its block nil rather than the missing key, a Symbol key on a String-keyed literal is treated as its name (the `Hash.new { }` model, which the new `hash_key_misses` inherits and names), and `Time.utc("x")` parses the String. All of these predate this branch.
- A rescued exception's class can be collected under `SPINEL_GC_STRESS=1` while `[e.class, e.message]` is built — a rescue-local rooting hole on master that the new test's stress run trips and that this branch does not touch.
- Under `--int-overflow=promote` a few of these slots (`Integer#<<`/`>>`/`[]`, `Time`'s calendar fields, `Array#slice!`) route through the promote-mode emitters, which this round does not touch — the raw pointer reaches the C compiler there exactly as it does on master. The suite compiles in default mode, so the swept arms are the ones it exercises.
- `sleep nil` stays the no-op that bare `sleep` already is; CRuby sleeps forever on both, which no gate can pin.

## Summary by CodeRabbit

* **新機能**
  * 型付き引数で `to_str`、`to_int`、`to_ary`、`to_hash` などの暗黙変換に対応しました。
  * 数値・文字列・配列・ハッシュ操作で、変換結果に応じた処理とエラー報告を改善しました。
  * 正規表現による文字列分割や、型付きコンテナのキー・要素判定に対応しました。

* **バグ修正**
  * 不正な型の値を誤って処理せず、適切な `TypeError` などを報告するよう改善しました。
  * 型付きハッシュ検索で、互換性のないキーを正しく未検出として扱うよう修正しました。

* **テスト**
  * 型変換、エラー、コンテナ操作に関する包括的なテストを追加しました。

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -846,6 +846,14 @@ rbs-seed-test: $(SPINEL) $(RBS_EXTRACT_BIN) $(SP_RT_LIB)
 	     -c --no-line-map -o "$$tmp/icnm.c" >"$$tmp/icnm.out" 2>&1; then \
 	  echo "rbs-seed-test: FAIL (an object with no #to_str compiled into a String slot)"; ok=0; \
 	else grep -q "no implicit conversion of Inert into String" "$$tmp/icnm.out" || { echo "rbs-seed-test: FAIL (missing #to_str rejected without saying why)"; sed -n 1,5p "$$tmp/icnm.out"; ok=0; }; fi; \
+	if $(SPINEL) test/rbs-seed/typed_slot_block_key.rb \
+	     -c --no-line-map -o "$$tmp/tsbk.c" >"$$tmp/tsbk.out" 2>&1; then \
+	  echo "rbs-seed-test: FAIL (a foreign key reached a typed block parameter)"; ok=0; \
+	else grep -q "a key of another class than the hash's keys" "$$tmp/tsbk.out" || { echo "rbs-seed-test: FAIL (foreign block key rejected without saying why)"; sed -n 1,5p "$$tmp/tsbk.out"; ok=0; }; fi; \
+	if $(SPINEL) test/rbs-seed/typed_slot_compare_obj.rb \
+	     -c --no-line-map -o "$$tmp/tsco.c" >"$$tmp/tsco.out" 2>&1; then \
+	  echo "rbs-seed-test: FAIL (a comparing user object reached a typed Array slot)"; ok=0; \
+	else grep -q "a user object defining == compared against a typed Array" "$$tmp/tsco.out" || { echo "rbs-seed-test: FAIL (comparing object rejected without saying why)"; sed -n 1,5p "$$tmp/tsco.out"; ok=0; }; fi; \
 	for t in hash_kind_widened_return poly_dispatch_arm_arg_type nilable_scalar_yield_key nilable_scalar_deep_chain nilable_scalar_paths poly_index_hash_dispatch yield_site_scalar_tail poly_container_op_result untyped_param_two_shapes untyped_recv_string_surface seeded_hash_boundary_values seed_hash_value_kind seed_ret_replaced_def seed_ret_empty_literal; do \
 	  $(SPINEL) test/rbs-seed/$$t.rb --rbs test/rbs-seed/sig -c --no-line-map -o "$$tmp/$$t.c" 2>/dev/null; \
 	  if $(CC) -O0 -Ilib $(RBS_SEED_STRICT) "$$tmp/$$t.c" $(SP_RT_LIB) $(LDFLAGS) -lm -o "$$tmp/$$t" 2>"$$tmp/$$t.err"; then \
--- a/src/codegen.c
+++ b/src/codegen.c
@@ -298,6 +298,24 @@ int obj_conv_method(Compiler *c, TyKind t, const char *conv, TyKind want, int *d
   return mi;
 }
 
+/* The container half of the protocol: a class whose #to_ary / #to_hash is a
+   no-parameter method returning a static Array (or Hash) kind converts
+   through a direct call, typed as that container (Array#zip, #product,
+   Hash#merge). Answers the container kind, or TY_UNKNOWN when the class has
+   no such method -- the method's own return type is the answer, so there is
+   nothing to search the kind space for. */
+TyKind obj_container_conv(Compiler *c, TyKind t, const char *conv, int *def) {
+  if (!ty_is_object(t)) return TY_UNKNOWN;
+  int cid = ty_object_class(t);
+  if (cid < 0 || cid >= c->nclasses) return TY_UNKNOWN;
+  int d = -1, mi = comp_method_in_chain(c, cid, conv, &d);
+  if (mi < 0 || c->scopes[mi].nparams != 0) return TY_UNKNOWN;
+  TyKind ret = c->scopes[mi].ret;
+  if (sp_streq(conv, "to_hash") ? !ty_is_hash(ret) : !ty_is_array(ret)) return TY_UNKNOWN;
+  if (def) *def = d;
+  return ret;
+}
+
 /* 1 iff any class in the program defines a usable #to_int / #to_str. A
    NARROWING into a typed slot compiles the conversion in only then: unlike an
    argument slot, a narrowing sits wherever the analysis put it -- including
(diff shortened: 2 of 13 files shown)


matz:
Merged in ef8b5ad4. The measurements hold: I reran your probe on the master
this starts from and on the branch, and got 210 cfail over 108 slots in 71
methods, then 18 -- the numbers in the description, to the probe. `make gate`
is green, the optcarrot checksum holds, and `test/typed_slot_conversion.rb`'s
expectations are CRuby's own output rather than a snapshot of ours, which is
the check that matters most for a conformance round. It also runs clean under
`SPINEL_GC_STRESS=1` here, and under ASAN+UBSAN.

Routing the POSIX-shaped slots through the existing `emit_str_expr` /
`emit_int_expr` rather than adding a parallel path is the right call, and the
ordering rule -- a conversion belongs below the nil-receiver guard, because
CRuby never asks `#to_str` for a call that does not dispatch -- is the kind of
thing that is much cheaper to get right now than to find later.

**One bug, fixed on top in 93d911f5 + d45a595f.** `conv_cls_name_of` reaches
`class_ruby_name`, which built the qualified name into a shared static buffer,
so the pointer was only good until the next call. All six new sites hold it
across an `emit_expr`:

```c
const char *cn = conv_cls_name_of(c, t);
buf_puts(b, "({ (void)("); emit_expr(c, node, b);
buf_printf(b, "); sp_raise_cls(\"TypeError\", \"...%s...\")", cn);
```

Reached through your new slots that is observable, because the intervening
emission can itself ask for a class name:

```ruby
p 1.0.quo(cond ? ((1.0.quo(M::Alpha.new) > 0.0) ? x : x) : x)
# CRuby:  "M::Beta can't be coerced into Float"
# branch: "M::Alpha can't be coerced into Float"
```

The inner conversion never runs -- its branch is not taken -- but it is
emitted, and that was enough to rename the outer error. Same three messages in
the Array and Hash protocols.

Nothing to change in your patch: the fix is at the source, `class_ruby_name`
now builds the name once and keeps it on the class. This is the second time
that buffer has bitten this week (it made every `is_a?` on a namespaced
exception compare its receiver against itself, 09bb126c), so the trap is now
disarmed rather than worked around per site.

**On the differ column**, since it is your next round's list: 43 probes moved
cfail -> differ here, and only two of them are yours.

- 24 are `Time.utc/gm/local/mktime` year padding: `Time.utc(1,1,1).to_s` is
  `"1-01-01 ..."` where CRuby pads to `"0001-01-01 ..."`. Pre-existing on
  master, and your slots merely made it reachable. I will fix it.
- 10 are ENV, and they are **probe artefacts rather than divergences**. The
  baseline search calls `ENV.delete` / `ENV.store` in-process via
  `__send__`, so the probe mutates its own environment; the spawned Spinel
  binary then inherits the mutated one and is compared against a CRuby answer
  taken before. Reproduced by hand, CRuby and Spinel agree on every one.
  Snapshotting and restoring `ENV` around `attempt` would clear them.
- 6 are `system` on a missing command answering `false` where CRuby answers
  `nil`. Pre-existing.
- `Hash#delete` on a typed-value hash is the one you already name.
- `Float#rationalize` now raises `TypeError` where CRuby raises
  `NoMethodError` (it calls `eps.abs` without converting). `rb_to_float` is
  right for the Float slots generally; that one slot is the exception.
- `String#split("ab".split(1, obj_to_int))` -- CRuby returns `["ab"]` because
  `limit == 1` short-circuits before it ever validates the separator. Ours
  validates eagerly. Both are defensible; worth a line in the next round's
  notes rather than a fix.

Thank you for the probe tool being in the PR. Being able to re-measure the
population rather than take the count on trust is what made this reviewable at
this size.

PR #4153, opened 2026-08-27
Six language-semantics fixes: setter values, begin/else, visibility, ENV.fetch, keyword flags, comparator blocks

blacktm:
## Why this PR

Spinel's promise is that a compiled program does what it does under CRuby. This PR makes six rules of the language itself hold that today do not — and every one of the six fails *silently*: no build error, no exception, a program that runs and computes something else. For an AOT compiler that is the worst class of defect there is, because nothing tells the user their program changed meaning.

The rules are ones ordinary code leans on every day: the value of an assignment through a setter (`x = obj.attr = v`, a setter as a method's last line); the value of `begin … rescue … else … end`; `private` and `protected` as a design tool, so a class can keep helpers off its public surface and a program can rescue the `NoMethodError`; reading configuration with `ENV.fetch("PORT", 3000)`, which today hands back garbage bytes; a `chomp:` flag that is anything but the literal `true`; and a comparator block that simply vanished from the generated C.

What merging buys: six classes of silent miscompile closed at their root — each fix is one rule in the compiler's own machinery (an inference rule, one gate, one fold), not a per-method patch — at no cost: the generated C for optcarrot, for all 61 benchmarks and for hello-world is byte-identical to master's, compile time is unchanged, and the change arrives as one commit per rule, each buildable and revertable on its own, each with a test whose expectations CRuby generated. They came out of a root-cause pass over a differential corpus of ~2,100 minimized programs, picked because their fix is a language rule rather than a library surface.

## What was wrong

- `obj.x = v` evaluated to whatever the writer *returned*: `def name=(v); @name = v.to_s; end` handed its caller the String, `x = (k.foo = 8)` bound the body's value, and at a method's tail the method answered its type's zero.
- `begin … rescue … else … end` was typed from the body, so an `else` of another type was stored raw: `else false` read back as the program's first Symbol, `else 3` as its third.
- `private` and `protected` were recorded — `respond_to?` and `public_send` honoured them — but an ordinary call never asked, so `account.secret` reached a private method from anywhere.
- `ENV.fetch("PORT", 3000)` placed the Integer in the `const char *` slot as it stood, behind a pointer-mismatch warning.
- `chomp: 1` did not chomp: the flag was true for the literal `true` alone.
- `[3, 1, 2].sort { |x| 0 }` came back sorted: the comparator fold stood down on a one-parameter block and the blockless arm ran.

## How

One commit per rule, each with a test whose expectations CRuby generated:

- **Setter value** — the inference types a plain setter call on an object as its argument (beside the attr_writer rule that already said so), and the object-call emitter evaluates the argument once into a rooted temp that the dispatch reads through `g_argov` and that is the expression's result. A statement-position setter has no reader for the value and emits as before, so `obj.x = v;` is unchanged C. A poly receiver follows the same rule: the dispatch switch yields the argument's temp instead of its accumulator, and a synthesized writer reached in value position — which raised NoMethodError where its statement form stored — takes the statement form's `cls_id` switch with the value yielded.
- **begin/else** — with an `else`, the else's statements are the value source (an empty `else` is nil); a body that raises still counts as diverging.
- **Visibility** — the `public_send` gate becomes `emit_vis_refusal`, asked by the ordinary explicit-receiver call and by the synthesized-writer arms (so `private :x=` holds on an `attr_writer`), with the language's exemptions: literal `self.`, a caller whose class is the declaring class or a descendant for a protected target, and `send`/`__send__` (already stamped visibility-blind). A method copied in from an included module keeps the module's declaration, and the refusal stages `NoMethodError#name`. Poly receivers and class methods are not asked — the dispatch does not carry the caller, and the scope pass records instance declarations only — and stay as they were.
- **ENV.fetch** — typed as String joined with the default's: a String or nil default keeps the nullable string, anything else boxes both arms; the default evaluates before the lookup, as an argument does. The block form already rode the ENV snapshot's `Hash#fetch`, whose splice bound the key raw into a parameter slot the body may have widened — a C build failure for a plain Hash too — and now boxes it.
- **Keyword flags** — `kw_flag_static` answers a flag's compile-time truth (any literal but false/nil is true; a run-time value is tested at run time) and the `chomp:` arms of `lines`, `each_line`, `gets`, `readline`, `readlines` and `File.readlines` choose from it. Other flags read by node kind are left for their own pass.
- **Comparator blocks** — the sort and min/max/minmax folds bind whichever parameters the block declares instead of standing down and letting the blockless arm run.

## Validation

- Six new tests under `test/` (expectations generated by CRuby), plus one existing expectation (`begin_rescue_else_clause`) that had pinned the `0`-for-nil wart and now matches CRuby.
- Full suite: 3,294 pass, 0 fail, run from a cleared results cache. Every commit builds on its own.
- Performance: the generated C for optcarrot, for all 61 programs under `benchmark/`, and for hello-world is byte-identical to master's after every commit — the whole file, so no function gains a rooted temp or a save frame around it. Identical C is identical instructions, so the callgrind count cannot have moved; I have no valgrind on this machine to re-measure it directly. Compile time for optcarrot is unchanged (0.98 s vs 0.97 s, both compilers built with the same flags).
- Against a differential corpus of ~2,100 minimized programs compared with CRuby: 682 → 690 match, 8 gained and 0 lost; the gains are the setter, `begin/else`, `ENV.fetch` and `chomp:` findings the corpus had filed under other names. Visibility and the comparator-block folds have no corpus finding of their own — they came out of the root-cause pass, not the fuzzer.

## Left alone, on purpose

- Builtin setter arms (`Dir#pos=`, `IO#close_on_exec=`) that answer the converted value rather than the argument: the same rule, a different emitter family (typed argument slots).
- `private_class_method`: the scope pass records instance-method visibility only. A receiver whose class is only known at run time is not asked either — the poly dispatch does not carry the caller.
- `minmax` with a comparator block of fewer than two parameters is refused at compile time rather than answered: CRuby's `minmax` yields its elements in pairs, so a block that sees only the first value answers differently than under `min` or `max`. `max(n) { |x| }` and a comparator answering nil (CRuby raises ArgumentError) are pre-existing and untouched.
- `chomp:` on the block and iteration forms (`each_line { }`, `File.foreach`) and a flag passed by double-splat still read the node kind; the other flags read that way are a pass of their own.

## Summary by CodeRabbit

- **新機能**
  - セッター式が代入値を正しく返すようになりました。
  - `ENV.fetch` がデフォルト値の型を考慮して動作するようになりました。
  - `chomp:` に実行時の条件を指定できるようになりました。
  - メソッドの公開・非公開・保護状態が、さまざまな呼び出し形式で適切に検証されます。

- **バグ修正**
  - `begin...rescue...else` 式の評価結果を修正しました。
  - `sort`、`min`、`max` の比較ブロック引数が少ない場合にも対応しました。
  - 非公開メソッド関連の例外メッセージで、メソッド名が正しく復元されます。

- **テスト**
  - 上記の動作を検証するテストを追加・更新しました。


matz:
Merged in 2af11b46. All six hold up, and the review was worth the time it took
because the six are individually verifiable — thank you for one commit per rule
with its own test.

What I checked, beyond `make gate` (green, optcarrot checksum 59662, 3,298
pass):

- Every one of the seven `.expected` files is byte-for-byte what CRuby prints,
  including the amended `begin_rescue_else_clause`. That is the check that
  matters most for a semantics round and it is the one most often skipped.
- Visibility against twelve shapes — private and protected from outside, from
  a sibling instance, from a descendant, through literal `self.`, through
  `send` / `public_send` / `respond_to?(:m, true)` — matches CRuby including
  both `NoMethodError` messages.
- `ENV.fetch` with Integer / String / nil / Symbol defaults, with a block, and
  the no-default `KeyError`; `chomp:` with `true` / `1` / `false` / `nil` and a
  run-time value; comparator blocks of one and two parameters across `sort`,
  `min` and `max`. All match.
- `SPINEL_GC_STRESS=1` clean on the new tests, and ASAN+UBSAN clean on the
  setter and visibility ones — the setter fix adds a rooted temp, so that was
  the thing to check.
- The inference fixpoint still converges on every program in the suite.

**One claim did not hold, and it is worth knowing rather than fixing.** The
description says the generated C for all 61 programs under `benchmark/` is
byte-identical to master's. optcarrot is. Two benchmarks are not:
`bm_ao_render` and `bm_rbtree`, 456 and 138 diff lines. Most is temp
renumbering, but `bm_ao_render` really does gain a rooted temp, at the tail
setter the fix is about:

```c
/* master */  sp_Isect_n_set((sp_Isect *)lv_isect, _t53);
/* this PR */ sp_Vec * _t54 = _t53;
              SP_GC_ROOT(_t54);
              return ({ (void)(sp_Isect_n_set((sp_Isect *)lv_isect, _t54)); _t53; });
```

That is the fix doing exactly what it says — the method's value changes — so
nothing is wrong with it. Two extra roots across the file, outside the hot
loop, and I measured no run-time difference. But "byte-identical" is the
strongest claim in the description and it is what a reader would trust instead
of re-measuring, so it should say "optcarrot and 59 of 61" and name the two.

---

**A request for next time: please send these as separate pull requests.** This
one would have been six.

The reason is not review load, it is what a merge can be undone by. The six
rules are independent — `begin/else` typing has nothing to do with the
comparator fold — but they arrive as one mergeable unit, so if one of them
turns out to regress something the corpus does not cover, reverting it means
reverting the merge and re-landing the other five. Your commits are already
clean enough to split at no cost: each builds and is revertable on its own,
which is most of the work.

It also lets each land at its own speed. The `chomp:` and comparator fixes are
small and would have been in within the hour; visibility changes what an
ordinary call does to every program in the language and deserved its own thread
— it is the one I spent most of the review on, and it is the one most likely to
surface a follow-up.

Same-shaped fixes across several call sites are one PR, as #4144 rightly was.
Six different language rules are six.

PR #4161, opened 2026-08-28
Add Process.spawn and waitpid2

rofl0r:
## Summary by CodeRabbit

* **New Features**
  * Added support for spawning processes with command arguments, working-directory changes, resource limits, process groups, and standard-stream redirection.
  * Added `Process.waitpid2`, returning the process ID and status.
  * Added `Process::Status` with exit, signal, success, core-dump, PID, comparison, and string-formatting helpers.
  * Added clear errors for failed process execution.

* **Bug Fixes**
  * Improved handling of interrupted process waits and invalid process-status values.

* **Tests**
  * Added coverage for successful and failed commands, signals, status checks, and missing executables.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -335,7 +335,7 @@ build/sp_cold.o: lib/sp_cold.c $(RT_HDRS)
 
 SP_RT_LIB = lib/libspinel_rt.a
 
-RT_MEMBERS = sp_bigint sp_crypto sp_pack sp_time sp_core sp_net sp_system sp_gc sp_alloc sp_dtoa sp_marshal sp_format sp_string sp_inspect sp_array sp_str sp_hash sp_proc sp_exc sp_re sp_random sp_fiber sp_sched sp_io sp_cold
+RT_MEMBERS = sp_bigint sp_crypto sp_pack sp_time sp_core sp_net sp_system sp_gc sp_alloc sp_dtoa sp_marshal sp_format sp_string sp_inspect sp_array sp_str sp_hash sp_proc sp_exc sp_re sp_random sp_fiber sp_sched sp_io sp_cold sp_process
 
 $(SP_RT_LIB): $(RE_OBJ) $(addprefix build/,$(addsuffix .o,$(RT_MEMBERS)))
 	ar rcs $@ $^
--- a/Makefile
+++ b/Makefile
@@ -335,7 +335,7 @@ build/sp_cold.o: lib/sp_cold.c $(RT_HDRS)
 
 SP_RT_LIB = lib/libspinel_rt.a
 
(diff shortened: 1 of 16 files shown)


matz:
Merged as 072e7ae7. Thank you -- the design is right, and I appreciate that you rebased onto the cls_id work and added `SP_BUILTIN_PROCESS_STATUS` to the distinctness switch and the class-name table yourself. The note in `sp_process.c` about why it cannot include `spinel_rt.h` saved me working that out.

I added one commit (a610c878) for four things the branch left open, all found by running the feature through shapes the test does not reach.

**The poly receiver had no inference rule.** The accessors are answered on a poly receiver by `emit_poly_builtin_method`, which is how a status normally arrives -- `waitpid2` returns an Array and its second element is read out of a poly container. Only the `TY_PROCESS_STATUS` receiver had a rule in `analyze_infer.c`, so the two sides disagreed:

```ruby
_, st = Process.waitpid2(pid)
return "Error: exit #{st.exitstatus}" unless st.success?
# error: incompatible type for argument 1 of 'sp_poly_to_s'
```

The interpolation asked `sp_poly_to_s` for what the emitter had produced as an `sp_int`. That is your `tools/shell.rb` shape, so I suspect it is the one you were chasing with the `g_ret_type` commit.

**`exitstatus` and `termsig` carried nil as -1**, so `p st.termsig` printed `-1` where CRuby prints `nil`. spinel already spells a nullable Integer `SP_INT_NIL`, which every renderer and `.nil?` understands.

**`success?` is nil, not false**, when the process did not exit normally. It is tri-state in CRuby and now here. Truthiness is unchanged, so `unless st.success?` reads the same.

**The two prototypes were emitted into every generated program**, so a program that never spawns grew two lines about spawning. They belong in `spinel_rt.h`, which `lib/sp_process.c` deliberately does not include but every generated TU does. An unrelated program is byte-identical to master again, which is the property I check on any PR that touches the emitter.

One note on the `g_ret_type` commit. The spawn half was `save; set; restore` with nothing in between, so it was a no-op with a comment explaining a save that guards nothing, and the waitpid2 half had no reader either -- the call's type comes from `infer_call`, like every other arm there. I dropped both writes rather than wrapping them. Your diagnosis of the leak was right; `g_ret_type` is the enclosing *function's* return type, which is exactly why writing it at a call site leaked.

`make gate` green on the merge: 3312 tests, optcarrot checksum 59662, every rubyspec gate unchanged.

rofl0r:
oh, thank you for the fast merge. i was just about to push a fix for the OSX CI build, and then address the review of the rabbit. the OSX fix is at my branch as 47c2c66f in case you havent incorporated it yourself yet.

PR #4162, opened 2026-08-28
Fix build on MacOS

bartleusink:
`make` currently does not build on macOS. Every `W*()` use in `lib/sp_process_status.c` is rejected — 18 errors, and the object never links:

```
lib/sp_process_status.c:53:10: error: cannot take the address of an rvalue of type 'int'
   53 |   return WIFEXITED((int)s) ? 1 : 0;
      |          ^         ~~~~~~
/…/sys/wait.h:131:34: note: expanded from macro '_W_INT'
  131 | #define _W_INT(w)       (*(int *)&(w))  /* convert union wait to int */
```

## Why it only breaks here

macOS still routes the wait macros through the legacy `union wait` bridge, so `_W_INT` takes the **address** of its argument — and the argument at every call site is a cast expression, `WIFEXITED((int)s)`. You cannot take the address of a cast. glibc casts rather than dereferencing, so the same source is fine on Linux.

The fix copies the `sp_int` argument into an `int` lvalue in each accessor and passes that. Values, branches and answers are unchanged; only what the macro is handed differs. The reason is recorded once, next to the include that imposes it.

## A second macOS-only failure the build was hiding

With the build fixed, `process_spawn` then failed — for an unrelated reason. It spawns `/bin/false` and `/bin/true`, which live in `/usr/bin` on macOS. The raise landed *after* the deliberate `Errno::ENOENT` in case 2 had been rescued, so it surfaced as an uncaught exception part way through the file rather than as an obvious missing-file error.

Both spellings exist on Linux, so the test now uses `/usr/bin`. `/bin/echo` and `/bin/sleep` are present on both and are left alone.

These two are separable if you would rather take only the first — the build fix stands alone, and `make` is green after it. They are together here because the second is only reachable once the first lands, and both are the same platform gap in the same feature.

## On a regression test

The Contributing guide asks for one per fix. For the build half the compile *is* the test: it fails at the first `W*()` on any macOS host, so a `.rb` test could not express it. `process_spawn` is the coverage for the second half, and it now passes rather than raising.

## Checks

On macOS 25.6.0 (arm64):

- `make` — builds clean, no warnings
- `make test` — 3304 pass, 0 fail, 0 error

There is no `gen2.c == gen3.c` closure target on master, so nothing to run for that item.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved process-status handling for compatibility with macOS.
  * Preserved existing process status behavior while ensuring status checks compile reliably across supported platforms.

* **Tests**
  * Updated subprocess test paths to use the correct executable locations on macOS.

### DIFF
--- a/lib/sp_process_status.c
+++ b/lib/sp_process_status.c
@@ -19,6 +19,14 @@
 #include "sp_alloc.h"   /* sp_RbVal, sp_gc_alloc, sp_raise_cls */
 #include "sp_process_status.h"
 
+/* The W*() macros must be handed an int LVALUE, not a cast expression. macOS
+   still defines them through the legacy `union wait` bridge --
+   `#define _W_INT(w) (*(int *)&(w))` -- so the macro takes the ADDRESS of its
+   argument, and `WIFEXITED((int)s)` fails to compile there with "cannot take
+   the address of an rvalue". glibc casts instead of dereferencing, which is
+   why this built on Linux. Each accessor below copies its sp_int argument
+   into `int st` first and passes that. */
+
 /* CRuby answers nil for exitstatus on a signaled status and for termsig on
    one that exited, and spinel already has a spelling for a nullable Integer:
    SP_INT_NIL, which every renderer and `.nil?` understands. -1 is a real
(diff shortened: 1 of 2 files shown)


matz:
Merged as f7f43509. Thank you, and sorry you had to find this: master had been red on macOS since I merged #4161, and the diagnosis is entirely right.

`_W_INT(w)` being `(*(int *)&(w))` is the whole story, and it explains why `lib/sp_system.c` was fine all along -- its `WIFEXITED(status)` is handed a variable, so there is an address to take. Every new call site passed a cast. Copying into an `int` lvalue is the minimal fix and leaves the values, branches and answers alone.

I took both halves. The second is not really separable in practice: a test that raises part way through is worse than one that fails to build, precisely because the raise landed after the deliberate `Errno::ENOENT` had been rescued and so read as something else entirely.

I checked the paths hold on Linux too -- `/usr/bin/true` and `/usr/bin/false` are both present here, and `/bin/echo` and `/bin/sleep` are unchanged -- and ran the gate: 3312 tests, benches 61, optcarrot checksum 59662, every rubyspec gate unchanged.

On the regression-test point, I agree: the compile is the test for the build half, and there is no way to write it as a `.rb`. What should have caught it is me looking at the CI on #4161 before merging it -- that branch was already red on macOS and I merged on a green Linux gate alone. That one is mine, not yours.

Issue #3449, opened 2026-07-30
[Runtime] Many Hash and Enumerable methods are missing on a Hash returned by a seedless Array#reduce


KieranP:
> (Disclaimer: I asked Claude Opus 5 (high) to find Spinel compile/runtime issues with documented Ruby classes and methods. It found the following issue, which I manually verified for accuracy before publishing this report)

_Reproduced with `spinel 15288aff1288` vs `ruby 4.0.3`._

A Hash returned by a seedless `Array#reduce` (or `inject`, block or symbol form) raises
`NoMethodError` for `dig`, `value?`, `invert`, `assoc`, `rassoc`, `filter_map`,
`each_with_object`, `group_by`, `partition`, `zip`, `reduce`, `inject`, `find`, `detect`,
`take`, `drop`, `select`, `reject`, `compact`, `slice`, `except`, `values_at`,
`fetch_values`, `entries` and `flatten`. On that identical Hash `[]`, `fetch`, `keys`,
`values`, `size`, `key?`, `to_a`, `merge`, `map`, `sort_by`, `min_by` and `any?` all work.
Passing a seed (`reduce({}) { … }`) avoids it entirely.

## Repro

```ruby
a = { n: 1, s: { x: 2 } }
h = [a].reduce { |acc, l| acc }
p h.keys
p (h.dig(:s, :x) rescue $!.class)
p (h.select { |k, v| k == :n } rescue $!.class)
p (h.find { |k, v| k == :n } rescue $!.class)
```

Ruby 4:

```
[:n, :s]
2
{n: 1}
[:n, 1]
```

Spinel:

```
[:n, :s]
NoMethodError
NoMethodError
NoMethodError
```

Issue #4170, opened 2026-08-28
Poly dispatch of `to_h` has no arm for a built-in Hash — every sibling (`to_a`, `to_s`, `keys`, `length`) does


rubys:
Poly dispatch of `to_h` has no arm for a built-in Hash — every sibling (`to_a`, `to_s`, `keys`, `length`) does

Spinel master `7863de76`, macOS/clang and Ubuntu 24.04/gcc alike.

## Repro — 12 lines

```ruby
class Box
  def to_h
    { z: 9 }
  end
end

def poly(n)
  n > 0 ? { a: 1, b: 2 } : Box.new
end

p poly(1).to_h
p poly(0).to_h
```

```
$ ruby toh2.rb              $ spinel toh2.rb -o /tmp/toh2 && /tmp/toh2
{a: 1, b: 2}                undefined method 'to_h' for an instance of Hash (NoMethodError)
{z: 9}
```

It COMPILES. The failure is at run time, on the first line, and the receiver it names is the plain Hash literal three lines above it.

## The mechanism, from the emitted C

The poly dispatch switch carries an arm for every user class that defines `to_h` and none for the built-in:

```c
switch (_t1425.cls_id) {
  case 51: ... sp_Flash_to_h(...)
  case 53: ... sp_ActionDispatch__Session_to_h(...)
  ...
  default: sp_raise_nomethod(sp_nomethod_msg("to_h", _t1425)); break;
}
```

A `Hash` receiver reaches `default:` and raises. `Hash#to_h` answers self, and `sp_poly_is_hash_kind` is already how the neighbouring code recognises one.

## It is only `to_h`

Same file shape, same poly receiver, one method changed each time:

| call | CRuby | spinel |
|---|---|---|
| `poly({a: 1, b: 2}).to_h` | `{a: 1, b: 2}` | **NoMethodError** |
| `poly({a: 1, b: 2}).to_a` | `[[:a, 1], [:b, 2]]` | `[[:a, 1], [:b, 2]]` |
| `poly([1, 2]).to_a` | `[1, 2]` | `[1, 2]` |
| `poly({a: 1, b: 2}).to_s` | `"{a: 1, b: 2}"` | `"{a: 1, b: 2}"` |
| `poly({a: 1, b: 2}).keys` | `[:a, :b]` | `[:a, :b]` |
| `poly({a: 1, b: 2}).length` | `2` | `2` |

[...]

PR #4173, opened 2026-08-28
Serve the builtin surface on a boxed receiver from one owner table

blacktm:
## Why

When Spinel cannot pin a receiver's class at compile time -- an element read out of a mixed container, a parameter that can also receive nil -- the value travels boxed, and a call on it is served by a pretence: unbox to one concrete kind, retype the receiver, and re-enter the typed emitter, which is then the implementation. That pretence is the compiler's whole answer for the builtin surface on such values, and it lived in six hand-curated name lists across three files, each with its own copy of the trampoline, the inference typing the same names by rules of its own -- and it keeps growing the same way: 2662ea2a added a copy for `fill` alone, 6c682db1 a third inference pin of the same shape, each right on its own and each one more place to keep in step. A name reaches a boxed receiver only if somebody has added it to the right list, so `sort!`, `uniq!`, `rotate!`, `flatten!`, `transpose`, `to_ary`, `select!`, `reject!`, `keep_if`, `delete_if`, `grep`, `minmax_by` and their siblings raise `NoMethodError` naming Array, the very class that defines them, while the typed emitters implement every one.

This PR replaces the lists with one owner table that both halves of the compiler read, so the result slot and the emission agree by construction, and adds the Array rows as the table's first consumer. The refactor is the first commit and changes no generated byte: the C for every program in the tree is identical to master's. The second commit adds fifteen rows -- fourteen names that raised, and `fill` -- a coercion that admits only an Array, and a write-back through the box that is exact where the old one coerced. The cost on the hot path is nothing -- optcarrot and all 61 benchmarks are byte-identical after both commits -- and the mechanism is what the follow-ups need: a name several kinds own (`concat`, `reverse!`, `slice!` are String's and Array's; `compact!`, `assoc` Array's and Hash's) is unrepresentable in a list keyed by name, and becomes a run-time kind switch over rows. Those come as their own PRs on top of this one.

## What

- `a = rows[0]; a.sort!` on a boxed Array raised `undefined method 'sort!' for an instance of Array`; so did `rotate!`, `uniq!`, `flatten!`, `transpose`, `to_ary`, `sort_by!`, `select!`, `filter!`, `reject!`, `keep_if`, `delete_if`, `grep` and `minmax_by`.
- The poly emitter's lists could only hold a name exactly one class answers -- the comment on one of them says so -- and the inference typed those names by separate hand rules, so a row added to one side without the other put a typed pointer in a poly slot or the reverse.
- The array write-back (`sp_poly_arr_writeback`, written for `map!`) copied `min(len)` elements and left the tail in place, so a mutator that shrinks its receiver could not have written back correctly even with a row; and a value the typed original could not hold (a String into an Array of Integers) it coerced to `0`.
- The Enumerable coercion answered an empty list for a receiver that is no collection, so `grep` on a boxed Integer would have answered `[]`; `transpose` answered `[]` for a row that is no Array where CRuby raises `TypeError`.

## How

- **One table** (`ty_poly_face_tbl` in `types.c`): a row per (name, owner kind, argument count, block shape) with the write-back and argument rules the name needs. The inference's last resort pins the receiver to each owner in turn and unifies the answers, the way the Hash face already did (#3449); the poly emitter looks the name up, unboxes to the owner's representation and re-enters -- under the silent emittability probe the dynamic-send dispatch already uses, so a typed emitter that declines drops the arm rather than the build. The four lists in the poly emitter and the Hash face's list become rows; the face's rows are marked last-resort and keep answering only once every poly-receiver emitter has declined the name. The three inference pins (the Hash face's, the poly array's, and the handle face's from 6c682db1, which already carried the type) become one that carries the kind. The first commit is this refactor alone: byte-identical C for every program in the tree.
- **Array rows**: `sp_poly_array_recv` takes an Array of any element kind, raises CRuby's `NoMethodError` for anything else (a Hash included, whose pairs the Enumerable coercion would have offered to `sort!`), and refuses a frozen original before a mutator runs -- a check at the write-back would have run a block over every element first. `sp_poly_arr_writeback` follows the work array's length and refuses an element the original has no representation for rather than coerce it. `grep` and `minmax_by` take the elements coercion, which now raises for a receiver that is no collection; `transpose` raises CRuby's `TypeError` for a row that is no Array. The `fill` arm from 2662ea2a -- the `map!` trampoline copied once more, for one name -- becomes a row, with the inference rule that went with it; its test keeps passing.
- A call with a splat or a block argument is no count a bounded row can match; the open rows -- the old lists' names -- answer it as the lists did.

## Validation

- A new test, `test/boxed_array_methods` (expectations generated by CRuby 4.0): every mutator writing back into typed originals of each layout, the result flowing into a slot the inference widened to poly, the frozen, `TypeError` and `NoMethodError` cases.
- Full suite: 3,309 pass, 0 fail, run from a cleared results cache, on each commit on its own (the first commit is byte-identical, so its suite is master's).
- Generated C against master (both compilers built at the same commit with the same flags): after the first commit, byte-identical for every program in the tree (every test, optcarrot, all 61 benchmarks: 3,335 programs at that commit). After the second, byte-identical for every program except eight: the new test; `poly_face_nilable_param`, whose `fill` now takes the row; and six where the Enumerable rows' coercion changes from `sp_enum_items_from` to `sp_poly_enum_recv`, the same call with a check in front (`boxed_receiver_enumerable_surface`, `combination_chain_in_destructured_block`, `container_read_blockless_enumerator`, `poly_hash_surface`, `safe_nav_block_calls`, `safe_nav_block_family`); all eight pass with their existing expectations. optcarrot and the 61 benchmarks are byte-identical; identical C is identical instructions, so the callgrind count cannot have moved. Compile time for optcarrot is unchanged (4.06–4.08 s vs 4.11–4.21 s on master, both compilers built with the same flags).
- `boxed_array_methods` and `poly_face_nilable_param` run clean under `SPINEL_GC_STRESS=1` and under ASAN+UBSAN (the re-entry roots the box across the typed emitter's call, so that was the thing to check). The inference fixpoint converges on every program in the suite, in the same number of rounds as on master (3,276 programs, at most 11).
- Against a differential corpus of ~2,100 minimized programs compared with CRuby (the same corpus behind #3999/#4003/#4029/#4050/#4057/#4144/#4153): 692 → 697 match, 5 gained and 0 lost -- `select!`/`reject!`/`keep_if` on an element of a nested Array, `sort_by!` and `minmax_by` on a boxed Array, `minmax_by` on a boxed Struct, and `transpose` on a non-Array row.

## Left alone, on purpose

- The names several kinds own -- `concat`, `prepend`, `reverse!`, `slice!` (String and Array), `compact!`, `assoc`, `rassoc`, `fetch_values`, `merge!` (Array and Hash) -- still raise on a boxed receiver. They need the run-time kind switch over rows, and the String write-back and the Hash write-back that go with them; each is a PR of its own on top of this table.
- `Array#shuffle`/`shuffle!` on a poly array: the poly array has no typed emitter for them yet, so a row would only move the `NoMethodError`.
- A boxed value that is a typed Array takes the mutation back by contents, not by identity: `x = a.sort!` answers the work array, and a later write to `x` does not reach `a`. The write-back keeps the two equal at the call; it cannot make them one object.
- A local read out of a container that also holds a String, then mutated by a name String owns, is narrowed to that String by the shared-mutable-string pass (#3227) before this table sees it. That is a decision of the narrowing pass, not of the dispatch.
- The exclusive-name handle face from 6c682db1 keeps its own lookup (`ty_poly_handle_face`); only its pin is shared. Its names become rows when the table grows owner bits for handle kinds.
- The remaining ~180 (class, method) pairs the poly path lacks (File, Float, Integer, MatchData, StringScanner, Thread, ...) are now rows rather than code, and come in their own rounds with their own coercions.

## Summary by CodeRabbit

- **New Features**
  - Improved support for array and enumerable methods when the receiver’s type is determined at runtime.
  - Added safer handling for typed array updates, including validation of element types and correct length updates.
  - Added clear errors for unsupported receivers, frozen arrays, invalid rows, and incompatible values.

- **Bug Fixes**
  - Improved boxed-array operations, mutation behavior, nested-array methods, and method dispatch consistency.
  - Added coverage for sorting, filtering, filling, transposing, and other array operations.

### DIFF
--- a/lib/sp_cold.c
+++ b/lib/sp_cold.c
@@ -1047,6 +1047,26 @@ sp_StrArray *sp_dir_entries_impl(const char *path, int children) {
   return a;
 }
 
+/* The rows of a transpose are Arrays; a row that is not is CRuby's TypeError,
+   named by the class of what was there. This unit sees the tags alone, so the
+   naming is by tag. */
+static int sp_transpose_row_p(sp_RbVal rv) {
+  return rv.tag == SP_TAG_OBJ &&
+         (rv.cls_id == SP_BUILTIN_INT_ARRAY || rv.cls_id == SP_BUILTIN_FLT_ARRAY ||
+          rv.cls_id == SP_BUILTIN_STR_ARRAY || rv.cls_id == SP_BUILTIN_SYM_ARRAY ||
+          rv.cls_id == SP_BUILTIN_POLY_ARRAY);
+}
+static const char *sp_transpose_row_class(sp_RbVal rv) {
+  switch (rv.tag) {
+    case SP_TAG_NIL: return "nil";
+    case SP_TAG_BOOL: return rv.v.i ? "true" : "false";
+    case SP_TAG_INT: return "Integer";
+    case SP_TAG_FLT: return "Float";
+    case SP_TAG_STR: return "String";
+    case SP_TAG_SYM: return "Symbol";
+    default: return "Object";
+  }
+}
 sp_PolyArray *sp_poly_array_transpose(sp_PolyArray *rows) {
   SP_GC_SAVE();
   SP_GC_ROOT(rows);
@@ -1057,7 +1077,9 @@ sp_PolyArray *sp_poly_array_transpose(sp_PolyArray *rows) {
   int16_t kind = 0; /* 0=unknown, SP_BUILTIN_INT_ARRAY, SP_BUILTIN_FLT_ARRAY, SP_BUILTIN_STR_ARRAY */
   for (sp_int r = 0; r < nrows; r++) {
     sp_RbVal rv = rows->data[r];
-    if (rv.tag != SP_TAG_OBJ) continue;
+    /* a row that is no Array is CRuby's TypeError, not a row of nothing */
+    if (!sp_transpose_row_p(rv))
+      sp_raise_cls("TypeError", sp_sprintf("no implicit conversion of %s into Array", sp_transpose_row_class(rv)));
     sp_int rlen = 0;
     if (rv.cls_id == SP_BUILTIN_INT_ARRAY)  { rlen = ((sp_IntArray *)rv.v.p)->len; if(!kind) kind = SP_BUILTIN_INT_ARRAY; }
     else if (rv.cls_id == SP_BUILTIN_FLT_ARRAY) { rlen = ((sp_FloatArray *)rv.v.p)->len; if(!kind) kind = SP_BUILTIN_FLT_ARRAY; }
--- a/lib/spinel_rt.h
+++ b/lib/spinel_rt.h
@@ -3520,31 +3520,59 @@ static sp_int sp_poly_int_recv(sp_RbVal v, const char *m) {
   sp_raise_nomethod(sp_nomethod_msg(m, v));
   return 0;  /* unreachable: sp_raise_nomethod does not return */
 }
-/* map!'s write-back through a boxed receiver: sp_poly_arr_recv normalizes a
-   TYPED array to a poly COPY, so in-place rewrites never reached the
-   original (#3234). Copy the mutated elements back into the original
-   representation; a POLY_ARRAY receiver shares storage and needs nothing. */
+/* A mutated element that the typed original has no representation for: a
+   String written into an Array of Integers through a boxed receiver. CRuby's
+   Array holds anything; the typed original cannot, and it cannot be re-laid
+   from here (the container slot that holds it is not in reach), so the
+   write-back refuses rather than coerce the value into something it was not
+   -- the earlier map! write-back answered 0 for such a String. */
+SP_NORETURN SP_COLD static void sp_raise_writeback_kind(sp_RbVal v, const char *into) {
+  sp_raise_cls("TypeError", sp_sprintf("can't store %s in an Array of %s through a boxed receiver",
+                                       sp_poly_class_name(v), into));
+}
+/* A mutator's write-back through a boxed receiver: sp_poly_arr_recv
+   normalizes a TYPED array to a poly COPY, so in-place rewrites never reached
+   the original (#3234). Replace the original's contents with the mutated
+   elements, in its own representation; a POLY_ARRAY receiver shares storage
+   and needs nothing. The length follows the work array -- reject!, uniq! and
+   slice! shrink it, fill(n) can grow it. Frozenness was refused at the
+   coercion, before the mutator ran. */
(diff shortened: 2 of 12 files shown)


matz:
Merged as 6af76797. This is the right shape for the problem, and the validation section made it reviewable rather than trusted -- I re-ran the load-bearing claims rather than taking them, and every one held.

**The refactor is byte-identical.** 3,273 programs compiled with master's compiler and with 699d3d97's, C compared: 0 differ. After 59d6ebe2, exactly the seven existing programs you name differ, and the diffs are what you said they are -- `sp_enum_items_from` becoming `sp_poly_enum_recv`, and `fill` taking its row. optcarrot's generated C is byte-identical across all three commits, so the hot-path claim needs no benchmark to believe.

**The gap it closes is real.** `sort!`, `uniq!`, `rotate!`, `flatten!`, `transpose`, `to_ary`, `sort_by!`, `select!`, `filter!`, `reject!`, `keep_if`, `delete_if`, `grep`, `minmax_by` on a boxed Array: all raised on master, all match CRuby now. So do the guards -- FrozenError, the NoMethodError for a Hash, the TypeError for a non-Array row.

**The write-back change is an improvement I want to record.** `[[1,2,3]][0].map! { "s" }` answered `[0, 0, 0]` on master, silently. It now raises `can't store String in an Array of Integer through a boxed receiver`. Still not CRuby's `["s","s","s"]`, but a diagnosis instead of a wrong answer.

The thing I most wanted to check was the trap this area has segfaulted on before (#3445): a name several kinds own, unboxed to one of them, compiling the wrong body against the receiver. Your "left alone, on purpose" list shows you know it, and I confirmed the guard holds where the table gets close to it -- `upto` on a boxed String and `select!` on a boxed Hash both raise, exactly as on master, rather than reinterpreting the receiver. `sp_poly_array_recv` admitting only an Array is what makes the Array rows safe to add at all.

I added one commit (26adc3c4), a comment only. The Array rows are headed "the Array-only names", and four of them are not: Hash defines `select!`, `reject!`, `keep_if` and `delete_if` too. Nothing changes behaviourally -- the coercion raises for a Hash, which is the answer those calls already got -- but this table is now the one place a reader looks to decide whether a name is safe to unbox by owner, so it should not read those four as Array's.

`make gate` green on the merge: 3,313 tests, optcarrot checksum 59662, every rubyspec gate unchanged.

I will look at #4170 next; if it is not already answered by the table, it should be a row.

PR #4174, opened 2026-08-29
Add a bundled `securerandom` package

rubys:
`SecureRandom` is a Ruby default gem with no spinel equivalent, so a program that mints a session token, an API key or a uuid has nothing to call. This adds the package and the one primitive it needs.

## The entropy is already yours

`lib/sp_crypto.c` sources from `arc4random_buf` on BSD/macOS and `getrandom(2)` then `/dev/urandom` on Linux, failing closed. Only the base64url rendering was exposed, and nothing in the tree called it. The fill is factored out into `sp_crypto_entropy` — one place decides what counts as secure — and a second entry point `sp_crypto_random_bin` returns the raw bytes with the count in `sp_ffi_bin_len` for the `:cbinstr` return mode. Random bytes carry a NUL about one time in 256, so `strlen` would cut the draw short.

`sp_crypto_random_bin` **raises** where its b64url sibling returns NULL. The sibling keeps its contract; but a caller asking for random bytes is minting a credential, so there is no answer to give, and a nil arriving in a token is the failure nothing notices.

## Not `Random.urandom`, and this is the part I'd most like checked

It is the obvious-looking primitive and I believe it is the wrong one. `sp_Random_urandom` is a PCG stream seeded from `time()`/`clock()`, and its own comment says *"spinel has no real OS entropy; this is a deterministic-per-run stand-in"* — where CRuby's `Random.urandom` **is** the OS entropy source. A `SecureRandom` built on it would hand back guessable session tokens and never fail. That reading is written into the package header so the next person doesn't have to rediscover it.

If you'd rather `Random.urandom` were itself moved onto the CSPRNG, that's a bigger change than this PR and I'm happy to follow up — but I didn't want to alter an existing primitive's behaviour underneath its callers as a side effect of adding a package.

## Surface

Present: `random_bytes`/`bytes`, `hex`, `base64`, `urlsafe_base64`, `uuid` (+ `uuid_v4`), `alphanumeric`.

Absent, and said so in the header: `random_number`, `uuid_v7` (needs a millisecond clock and a monotonic counter), `base36`, `choose`, and `alphanumeric`'s `chars:` keyword.

`alphanumeric` uses **rejection sampling** rather than `byte % 62`. 256 is not a multiple of 62, so the modulo makes the first eight letters about 5% more likely than the rest — invisible to any test that checks the character set, and a real reduction in a token's entropy. CRuby reaches the same distribution by a different route (base-62 digits of one bounded draw, `Random::Formatter#choose`); the outputs are indistinguishable and only the bytes consumed differ.

A single draw is capped at `SPC_RANDOM_MAX` (256); longer requests are assembled from repeated draws, so `hex(1000)` works and every chunk is independently kernel-sourced.

## Tests

A CSPRNG has no fixed expectation, so these are properties: lengths, alphabets, the uuid's version and variant nibbles, padding present in `base64` and absent in `urlsafe_base64`, uniqueness across 50 draws, all 62 characters reachable, NULs surviving a 4 KiB draw, every byte value reachable, and the 255/256/257 boundary of the single-draw cap.

**Both files were run against CRuby's own securerandom first and the `.expected` generated from that**, so they are a differential test rather than a restatement of this implementation. That is what caught the one asymmetry I would otherwise have tidied away: `random_bytes(-1)` raises `ArgumentError` while `alphanumeric(-1)` answers `""`.

`make test`: **3312 pass, 0 fail, 0 error.**

## One line in codegen.c

`SecureRandom` joins the constant→feature table, so using it without the require names the library instead of failing far from the cause. It earns that more than its neighbours do: Rails apps never write `require "securerandom"`, because ActiveSupport loads it as an implementation detail.

## Where it came from

ONCE Campfire's first write. `Account::Joinable` runs `SecureRandom.alphanumeric(12).scan(/.{4}/).join("-")` in a `before_create`, so `POST /first_run` answered `NoMethodError: undefined method 'join' for unknown` — `String#scan` is fine, the unknown receiver was `SecureRandom`. With this package the account and the first user are created and the write path moves on to the next thing.

## Summary by CodeRabbit

* **New Features**
  * Added built-in `SecureRandom` support for random bytes, hexadecimal strings, Base64 and URL-safe Base64 values, UUIDs, and alphanumeric strings.
  * Added validation and support for larger random-data requests.
  * `SecureRandom` is automatically available when referenced.

* **Bug Fixes**
  * Improved handling when a secure randomness source is unavailable.
  * Ensured concurrent random-value generation remains distinct and reliable.

* **Tests**
  * Added coverage for output formats, lengths, uniqueness, boundaries, concurrency, and invalid inputs.


rubys:
Actionable, and more serious than it looked — thank you. Fixed in 02dcbdab.

I ablated it rather than taking it on faith, since a passing concurrency test proves nothing on its own. With the buffer left process-global, 8 threads drawing 200 tokens each:

| draw | distinct / total |
|---|---|
| `SecureRandom.hex(16)` | **1118 / 1600** |
| `SecureRandom.alphanumeric(24)` | **1095 / 1600** |

About a third of the tokens were another thread's. And it is live rather than theoretical: `src/main.c:533` links the MT runtime automatically for any program that uses `Thread`.

The sharper statement of the defect is that **the buffer and its length were in different storage classes** — `sp_ffi_bin_len` is already `SP_TLS`, and the buffer it describes was not. Both cannot be right. The fix is `SP_TLS` on the buffer, matching `sp_random.c`, which already keeps its generator state per-worker for the same reason; the cost is `SPC_RANDOM_MAX` bytes per thread. `test/securerandom_threads.rb` is the regression and it fails on the previous commit, which is the only reason it earns its place.

I did not take the "caller-owned storage" half of the suggestion: that would change the `:cbinstr` ABI for one function, and the per-worker buffer gets the same guarantee without making this the odd one out.

---

**One thing this turned up that is outside the PR, and I would rather flag than quietly fix.** Every other static buffer in `lib/sp_crypto.c` is process-global too — thirteen of them, none `SP_TLS`:

```
sp_crypto_sha256_hex_buf, sp_crypto_sha256_bin_buf, sp_crypto_sha1_hex_buf,
sp_crypto_sha1_bin_buf, sp_crypto_websocket_accept_buf, sp_crypto_hmac_hex_buf,
sp_crypto_hmac_sha1_hex_buf, sp_crypto_hmac_b64url_buf, sp_crypto_b64url_buf,
sp_crypto_b64u_dec_buf, sp_crypto_pbkdf2_b64url_buf, sp_crypto_random_b64url_buf
```

They have the same window, so a threaded server signing cookies through `hmac_sha256_hex` on two workers can read the other's digest. The consequence is milder than a duplicated token — a wrong signature fails a verification rather than colliding two sessions — but it is the same shape and now demonstrably reachable.

I left them alone because widening a package PR into the crypto file's storage class felt like the wrong call to make unilaterally. Happy to send it as a separate change if you want it, either as `SP_TLS` across the board or as a documented single-threaded contract on the file.

matz:
Merged as 902bb90d, and you were right to ask about `Random.urandom` -- I checked it before anything else, since the whole package rests on that judgement.

`sp_Random_urandom` is a PCG seeded once from `time()`, `clock()` and a stack address. Its own comment says "spinel has no real OS entropy; this is a deterministic-per-run stand-in". Building `SecureRandom` on it would have handed out session tokens that two processes started in the same second could collide on, and never failed to do so. Sourcing from `sp_crypto`'s path instead is correct, and factoring the fill into one `sp_crypto_entropy` is what makes "what counts as secure" answerable in one place.

Verified the rest: the package's three tests pass, `alphanumeric(12).scan(/.{4}/).join("-")` gives your campfire shape, and successive processes draw distinct bytes. `make gate` green.

I took you up on the follow-up rather than leave it (3959afe8): `Random.urandom` now goes through `sp_crypto_entropy` too. Your reasoning for keeping it out of this PR was right -- changing a primitive underneath its callers as a side effect of adding a package is not the same act -- but with the source factored out it is a small change, and CRuby's `Random.urandom` *is* the OS entropy source, so leaving spinel's as a per-run stream was a divergence on a supported method rather than a missing feature. Longer draws are assembled from repeated fills so every chunk is independently kernel-sourced, and it fails closed like yours does.

Two things I want to record because they are the sort of care that is easy to skip:

**Rejection sampling in `alphanumeric`.** `byte % 62` skewing the first eight letters by 5% is invisible to every test that checks the character set, which is exactly why it would have survived. Reaching CRuby's distribution by a different route and saying so in the header is the right shape.

**Generating the expectations from CRuby first.** That is what makes these property tests differential rather than a restatement, and it earned its keep immediately by turning up the `random_bytes(-1)` / `alphanumeric(-1)` asymmetry -- which is real CRuby behaviour and would have been "tidied away" by anyone writing the expectations from the implementation.

Issue #3445, opened 2026-07-30
[Runtime] Many String methods are missing on a String returned by Fiber#resume, Queue#pop or Thread#value


KieranP:
> (Disclaimer: I asked Claude Opus 5 (high) to find Spinel compile/runtime issues with documented Ruby classes and methods. It found the following issue, which I manually verified for accuracy before publishing this report)

_Reproduced with `spinel 15288aff1288` vs `ruby 4.0.3`._

Follows on from _sub / gsub / tr / squeeze are missing on a String returned by Fiber#resume_ (`sub`, `gsub` and `tr` now answer normally; `squeeze` still does not, along with a wider set of methods).

A String obtained from `Fiber#resume` raises `NoMethodError` for `squeeze`, `squeeze!`,
`index`, `rindex`, `byteindex`, `partition`, `rpartition`, `slice`, `insert`, `hex`, `oct`,
`tr_s`, `crypt`, `casecmp` and `casecmp?`. On that identical value `[]`, `slice`'s operator
form, `length`, `ord`, `delete`, `upcase`, `split`, `sub`, `gsub`, `tr` and `squeeze("abc")`
(the one-argument form of the same method) all return the expected result. `Queue#pop`,
`Thread#value`, and the elements of a `filter_map` whose block ends in a short-circuiting
`&&` all produce a String with the same missing methods.

## Repro

```ruby
def fv
  f = Fiber.new { Fiber.yield("aabbcc"); nil }
  f.resume
end
p (fv.squeeze rescue $!.class)
p (fv.index("b") rescue $!.class)
p (fv.partition("b") rescue $!.class)
p (fv.casecmp?("AABBCC") rescue $!.class)
p (fv.squeeze("abc") rescue $!.class)
```

Ruby 4:

```
"abc"
2
["aa", "b", "bcc"]
true
"aa"
```

Spinel:

```
NoMethodError
NoMethodError
NoMethodError
NoMethodError
"aa"
```

PR #4175, opened 2026-08-29
Dispatch a name String and Array share on the receiver's run-time kind

blacktm:
## Why

The owner table that serves the builtin surface on a boxed receiver (#4173) can hold a name several kinds own -- the one thing the lists it replaced could not -- but nothing yet reads such a row. So `concat`, `prepend`, `reverse!` and `slice!` on a boxed value, which String and Array both answer, still raise `NoMethodError` for a String and for an Array alike, naming the class that defines them; and `setbyte`, `scrub!`, `bytesplice` and `append_as_bytes` on a boxed String raise because the String arm has no write-back at all.

This PR makes a multi-owner name dispatch on the receiver's run-time kind -- the box bound once, one re-entry per owner inside its own branch, the `NoMethodError` the call raised before as the default -- and gives the String mutators a write-back into the receiver's variable, so a container the value came from observes the change. Eight names that raised now answer as CRuby does on a receiver that is a variable, with one exception named under "left alone" (a boxed Array whose element type cannot hold the argument raises a store `TypeError` rather than widening); a boxed receiver of the wrong kind still raises what it raised before, and an arm ruled out by an argument's static type raises CRuby's `TypeError`. The generated C for optcarrot and every benchmark is byte-identical to master's; the tests whose C changes are the ones that reach these names, and all pass with their existing expectations.

## What

- `concat`, `prepend`, `reverse!` and `slice!` on a boxed value raised for a String and for an Array alike: a list keyed by name cannot hold a name two classes own, since the receiver's class is unknown when the list is read.
- `setbyte`, `scrub!`, `bytesplice` and `append_as_bytes` on a boxed String raised: the String arm had no write-back.
- `Array#concat` with a boxed argument that is no Array appended nothing where CRuby raises `TypeError`.

## How

- **Several owners**: the box is bound once and its kind tested at run time, one re-entry per owner inside its own branch with its own prelude, and the `NoMethodError` the call raised before as the default -- nothing that raised starts silently succeeding. Because the re-entry probe already restores everything it touched (#4173), a declined owner just drops its arm and the others stand. An arm ruled out by an argument's static type (`concat`'s, and `String#prepend`'s, arguments are of the receiver's own kind) raises CRuby's `TypeError`, naming the value as CRuby names it -- `nil`, `true` and `false` spell themselves, which matters because a Boolean is `true` or `false` only at run time -- and so does a boxed argument of `Array#concat` that is no Array (`sp_poly_set_operand`, which now spells those three the same way). Owners that answer different types leave the call poly and each arm boxes its value.
- **String write-back**: the new contents go back into the receiver's variable, a local or an instance variable: a shared handle absorbs them, so a container the value came from observes the change; a plain string box is replaced, the way the typed path replaces its own. A receiver that is no variable -- an element read, a Hash value -- takes them only from a mutator whose value is those contents (`concat`, `prepend`, `reverse!`, `bytesplice`, `append_as_bytes`, `scrub!`; the typed emitter leaves `slice!`'s and `setbyte`'s in the receiver's temp for a variable alone) and only through a shared handle (`sp_poly_str_become_handle`); a plain box there has nowhere to send them, and the call raises the `NoMethodError` it raised before, rather than a mutation that silently goes nowhere. The #3227 tests that reached `prepend`, `concat` and `reverse!` on a shared handle through the builtin-method arms now take this arm, with the same output. `scrub!` answers self in CRuby, so its value is the new contents, and contents that are the receiver's own mean it scrubbed nothing: no write, so no frozen check (`"abc".freeze.scrub!` answers `"abc"` in CRuby too). That is a row flag (`PF_SAME_OK`) for `scrub!` alone, because `sub!` with no match answers its own contents too and CRuby raises for it; the value-form bang rows are untouched, and every bang name on a frozen, unchanged boxed string raises exactly as CRuby does.

## Validation

- A new test, `test/boxed_string_methods` (expectations generated by CRuby 4.0): the shared names on a parameter that is a String at one call site and an Array at another, the mutators writing back through a local, an instance variable and a shared handle, the `TypeError` cases with the argument named at compile time and at run time, and the `NoMethodError` case.
- Full suite: 3,311 test cases pass, 0 fail, from a cleared results cache.
- Generated C against master, over the 3,276 suite programs, optcarrot and the 61 benchmarks: byte-identical for every program except the new test and seven existing ones -- `bang_result_through_shared_handle`, `concat_boxed_source`, `issue_3227_container_read_mutation`, `issue_3227_corners`, `issue_3227_matrix`, `issue_3227_poly_mutators`, `issue_3317_reader_concat_splat` -- which are the programs that reach `concat`, `prepend` or `reverse!` on a shared handle, or `Array#concat` with a boxed argument, and all pass with their existing expectations. optcarrot and the 61 benchmarks are byte-identical; optcarrot's compile time is unchanged (4.0-4.2 s either way on this machine).
- `boxed_string_methods`, `boxed_array_methods` and `issue_3227_container_read_mutation` run clean under `SPINEL_GC_STRESS=1` and ASAN+UBSAN, with the expected output. The inference fixpoint converges on every suite program (at most 11 rounds), in the same number of rounds as under master for every one of them.
- Against the differential corpus (~2,100 minimized programs; the same corpus behind #3999/#4003/#4029/#4050/#4057/#4144/#4153/#4173): 697 → 705 match, 8 gained (the seven shared or String-only names on a boxed receiver, and `scrub!` on a String read out of an Array) and 0 lost.

## Left alone, on purpose

- A boxed Array whose element type cannot hold what `concat` or `prepend` brings: `def c(x, y) = x.concat(y); c("a".dup, "b"); p c([1], ["z"])` raises `can't store String in an Array of Integer through a boxed receiver (TypeError)` where CRuby answers `[1, "z"]` and master raised `NoMethodError`. The typed origin cannot hold the element and the inference cannot see the call widen it -- the same diagnosis #4173 gives `map!`: proof against a wrong answer, not yet CRuby's answer.
- A plain (non-handle) String box mutated through a method parameter does not reach the caller's variable: `def tack(x, y) = x.concat(y); a = "ab".dup; tack(a, "c"); tack([1], [2]); p a` prints `"ab"` where CRuby prints `"abc"` (with the second call site gone, `x` is typed and the typed path's return handle keeps the aliasing). The boxed path has no handle to write through. Pre-existing for `upcase!`, wider now.
- A String mutator on a receiver that is no variable and no shared handle -- `h[:k].concat("x")` on a Hash of plain strings, a global -- raises the `NoMethodError` it raised on master, where CRuby mutates in place: a plain box has nowhere to send the new contents, and a silent no-op would be worse than the raise. `slice!` and `setbyte` on any receiver that is no variable (`arr[0].slice!(1)`) raise the same way, since their value is not the new contents.
- A local read out of a container that also holds a String, then mutated by a name String owns, is narrowed to that String by the shared-mutable-string pass (#3227) before the table sees it; `rows = [[1, 2], "x"]; a = rows[0]; a.reverse!` therefore still compiles the String body and answers `""` where CRuby answers `[2, 1]`, as on master.
- The builtin-method arms for `prepend`/`concat`/`bytesplice` on a shared handle in `codegen_call.c` stay: a call with a splat argument still reaches them, since a row that does not ask for a block declines a non-plain argument list (that shape fails to compile on master too, unchanged here). The `reverse!` row of that dispatch's bang table cannot be reached any more and is dropped.
- `String#slice!(i, 0)` answers `nil` where CRuby answers `""`, through the box as on the typed path: the typed emitter's, not the table's.
- The names Array and Hash share (`compact!`, `assoc`, `rassoc`, `fetch_values`) and the Hash mutators (`merge!`) need the Hash write-back, and come next.

## Summary by CodeRabbit

* **New Features**
  * Improved runtime dispatch for boxed String and Array mutation methods.
  * Added support for shared mutation through variables and instance variables.
  * Improved handling of polymorphic receivers and return values.

* **Bug Fixes**
  * Corrected operand type errors for `nil`, `true`, and `false`.
  * Improved `NoMethodError` and `TypeError` behavior for unsupported operations.
  * Fixed mutation handling for change-only String methods and shared aliases.

* **Tests**
  * Added comprehensive coverage for boxed String mutations, dispatch, return values, and error messages.

### DIFF
--- a/lib/spinel_rt.h
+++ b/lib/spinel_rt.h
@@ -3505,7 +3505,7 @@ static sp_PolyArray *sp_poly_set_operand(sp_RbVal v) {
   if (v.tag == SP_TAG_OBJ && sp_poly_is_array_kind(v.cls_id))
     return sp_poly_to_poly_array(v);
   sp_raise_cls("TypeError", sp_sprintf("no implicit conversion of %s into Array",
-                                       sp_poly_class_name(v)));
+                                       sp_convert_src_name(v)));
   return NULL;  /* unreachable: sp_raise_cls is noreturn */
 }
 /* Coerce a poly value that a container-read Array method (find/reject/sort/
@@ -5691,6 +5691,26 @@ static sp_RbVal sp_poly_str_become(sp_RbVal v, const char *s) {
   if (sp_poly_is_strbuf(v)) { sp_String_set_bin((sp_String *)v.v.p, s); return v; }
   sp_str_check_mutable(v.v.s);
   return sp_box_str(s);
+}
+/* The same through a receiver that is no variable -- an element read, a Hash
+   value: a shared handle absorbs the new contents and every alias observes
+   the change; a plain string box has nowhere to send them, and the call
+   raises the NoMethodError it raised before the row existed, rather than a
+   mutation that silently goes nowhere. */
+static sp_RbVal sp_poly_str_become_handle(sp_RbVal v, const char *s, const char *name) {
+  if (sp_poly_is_strbuf(v)) { sp_String_set_bin((sp_String *)v.v.p, s); return v; }
+  sp_raise_nomethod(sp_nomethod_msg(name, v));
+  return v;
+}
+/* Are the contents a mutator answered the receiver's own? Then it changed
+   nothing, and a mutator that writes only when it changed something (scrub!)
+   writes nothing -- and raises no FrozenError for a write that never happens,
+   as CRuby does not. Only such a row asks: sub! with no match answers its
+   receiver's own contents too, and CRuby raises for it. */
+static int sp_poly_str_is_own(sp_RbVal v, const char *s) {
+  if (v.tag == SP_TAG_STR) return v.v.s == s;
+  if (sp_poly_is_strbuf(v)) return ((sp_String *)v.v.p)->data == s;
+  return 0;
 }
 static sp_RbVal sp_poly_insert(sp_RbVal v, sp_int i, sp_RbVal x) {
   /* String#insert on a boxed receiver: splice at a character index, where a
--- a/src/codegen_call.c
+++ b/src/codegen_call.c
@@ -4083,7 +4083,7 @@ static int emit_poly_builtin_method(Compiler *c, int id, Buf *b) {
       {"capitalize!","capitalize",0,1},{"swapcase!","swapcase",0,1},
       {"strip!","strip",0,1},{"lstrip!","lstrip",0,1},{"rstrip!","rstrip",0,1},
       {"chomp!","chomp",0,1},{"chop!","chop",0,1},{"squeeze!","squeeze",0,1},
-      {"reverse!","reverse",0,0},{"succ!","succ",0,0},{"next!","succ",0,0},
+      {"succ!","succ",0,0},{"next!","succ",0,0},
       {"delete_prefix!","delete_prefix",1,1},{"delete_suffix!","delete_suffix",1,1},
       {"delete!","delete",1,1},
       {"gsub!","gsub",2,1},{"sub!","sub",2,1},{"tr!","tr",2,1},{"tr_s!","tr_s",2,1},
--- a/src/codegen_call_recv.c
+++ b/src/codegen_call_recv.c
@@ -1902,8 +1902,9 @@ int emit_array_call(Compiler *c, int id, Buf *b) {
         if (from) { buf_printf(b, "%s(", from); emit_expr(c, argv[ai], b); buf_puts(b, ")"); }
         else if (at == TY_POLY || at == TY_UNKNOWN) {
           /* a boxed argument (a rest param widened to poly): unbox to the
-             working array through the runtime kind dispatch (#3317) */
-          buf_puts(b, "sp_poly_to_poly_array("); emit_boxed(c, argv[ai], b); buf_puts(b, ")");
+             working array through the runtime kind dispatch (#3317); one
+             that is no Array is CRuby's TypeError, not an empty list */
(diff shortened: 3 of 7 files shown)


matz:
Merged as 99c89835. This is the piece #4173 was built for, and it lands on the one shape this area has segfaulted on before, so I checked that first.

**A wrong-kind receiver still raises, and raises what CRuby raises.** `concat` and `reverse!` on a boxed Hash, Integer, Symbol, Float and nil: output identical to CRuby, line for line. That is the property #3445 was about -- a name two classes own, unboxed to one of them, compiling the wrong body against the receiver -- and the run-time kind switch is the answer to it rather than a way around it.

**The measurements hold.** 3,275 programs compiled with the base compiler and with this one: exactly the seven you name differ, no others. optcarrot's generated C is byte-identical. `boxed_string_methods` matches its expectation under `SPINEL_GC_STRESS=1`, and under ASAN+UBSAN with the stress on -- I rebuilt the runtime with both sanitizers rather than take that one, since the write-back into a shared handle is where a mistake would hide.

**The shared names work on both kinds.** `concat`, `prepend`, `reverse!`, `slice!` on a parameter that is a String at one call site and an Array at another, all raising `NoMethodError` on master, all answering now.

Both divergences I ran into were already in your "left alone" list, and I want to record that they are, because each cost me a test before I recognised it: a plain String box mutated through a method parameter does not reach the caller's variable, and a boxed Array whose element type cannot hold the argument raises the store `TypeError`. Written down as limits, both are honest; the thing that would have worried me is a silent wrong answer, and neither is one.

The `scrub!` reasoning is the detail I would have got wrong. "Contents that are the receiver's own mean it scrubbed nothing, so no write, so no frozen check" is right and non-obvious, and confining it to a row flag rather than to the whole bang family -- because `sub!` with no match answers its own contents and CRuby *does* raise for it -- is exactly the distinction that keeps the rest correct.

PR #4176, opened 2026-08-29
Process.spawn inherits unset fds

rofl0r:
sp_process.c:apply_redirect used to redirect -1 (unset) slots to /dev/null, which silently swallowed child stderr in the AOT runtime. That matched nothing in CRuby: an unset :in/:out/:err slot is inherit. Fix: do nothing when src_fd < 0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Unset standard input, output, and error streams now inherit their parent process settings as expected.
  * Prevented child-process error output from being silently discarded when no redirection is configured.

### DIFF
--- a/lib/sp_process.c
+++ b/lib/sp_process.c
@@ -61,16 +61,18 @@ static const char *sp_errf_errno(const char *prefix, int err) {
 }
 
 /* Apply the redirect in the child: dup2 src_fd onto target_fd, close src.
-   If src_fd is -1 (false), redirect to /dev/null. */
+   src_fd < 0 means the caller did not pass that slot in the opts hash
+   (the codegen initialises every slot to -1 and overwrites only the ones
+   the user set); in that case the fd is left inherited from the parent.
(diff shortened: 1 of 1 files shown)


matz:
Merged. You are right on both counts: an unset slot is inherit in CRuby, and `/dev/null` there is not a conservative default but a silent one.

Measured before and after on `sh -c 'echo out; echo err 1>&2'` with only `out:` set: master swallowed `err` **and** reported `success? == false`; with your change both match CRuby.

I added one commit (f7f454b2) for a second fd bug your change makes reachable. `apply_redirect` closed the source right after its `dup2`, so naming one descriptor twice --

```ruby
Process.spawn("/bin/sh", "-c", "echo o; echo e 1>&2", out: w, err: w)
```

-- left the second `dup2` working on a closed fd, silently, and the child's stderr went nowhere. That is idiomatic enough that I did not want to leave it next to the fix for the same class of problem. The close is now a separate pass after all three redirects, skipping a descriptor already closed under another slot.

The test carries both shapes: an unset `:err` inheriting, and one pipe named by two slots.

PR #4179, opened 2026-08-29
Give every sp_crypto return buffer its own thread, not the process's

rubys:
The `SP_TLS` on `sp_crypto_random_bin_buf` in #4174 fixed one buffer. The other twelve in `lib/sp_crypto.c` were process-global for the same reason and had the same window; I flagged them there rather than widening that PR, and this is the follow-up.

## The window is the copy, and it is on every function in the file

Nothing in these functions races. The race is at the call site: each one returns a pointer into its own file-static buffer and the caller copies out of it **after the call has returned** — a `:cstring` `native_func` dups onto the GC heap, an ffi `:str` is read by whatever reads it next. A second thread entering the same function inside that window overwrites the first one's answer before the first has taken its copy.

`Digest::SHA256.hexdigest`, `Digest::SHA1.hexdigest`, `OpenSSL::Digest::SHA256.hexdigest`, `OpenSSL::HMAC` — every one of them is a plain Ruby method over one of these buffers, and `src/main.c:533` links the MT runtime for any program that uses `Thread`.

## What it costs, measured

`test/sp_crypto_threads.rb` is the regression. It cannot check a shape, because a clobbered digest is still 64 hex characters and a clobbered HMAC is still a signature — the wrongness only shows up as a failed verification somewhere else, later, with nothing pointing back here. So it is an oracle: compute the answers with one thread running, where no clobber is possible, then recompute them under contention and count the disagreements. 8 threads × 25 rounds × 64 inputs × 3 functions = 38,400 comparisons.

| | disagreements / 38,400 |
|---|---|
| before | **11,819 · 11,718 · 12,450 · 12,109** (four runs) |
| after | **0 · 0 · 0** |

About 31%, which is the same fraction the `SecureRandom` ablation in #4174 showed for a duplicated token. The test also prints SHA-256("abc") against the FIPS 180-4 vector, so a run in which every digest came out uniformly wrong could not agree with its own oracle and pass.

## Why not the contract that was already written

`sp_crypto.h` did document this — *"NOT thread-safe … apps that thread need to caller-side serialize"* — so "leave the storage class and keep the contract" was a real option. I don't think it is a contract a caller can keep. There is no call site to serialize: `Digest::SHA256.hexdigest(data)` is an ordinary method call in ordinary Ruby, and nothing about it tells its author they are inside a critical section shared with every other thread in the process. A contract that can only be honoured by knowing which Ruby methods happen to bottom out in this file is one that will be broken by the next person who writes a threaded server, exactly as it was broken here.

Both header comments now say the new rule instead: the clobber rule still holds, per thread.

## Cost

`SP_TLS` is `__thread` only under `-DSP_THREADS`, so a program that never threads pays nothing — I compiled `lib/sp_crypto.c` at both revisions with the single-threaded rule's flags and the objects are byte-identical (`b6f9783dbab262e110d212ecd47b195f96a48adc`).

The threaded variant pays the buffers per worker. ~32 KiB of that is the two 16 KiB base64url buffers; all ten others together are under 600 bytes. The pool is bounded by `min(cores, SPINEL_WORKERS)`, not by the number of Ruby threads, and the MT build already compiles with `-ftls-model=initial-exec`.

I did not shrink the two 16 KiB buffers or move anyone to caller-owned storage. Both are defensible changes and neither is this one: the first changes a documented input limit, the second changes the `:cstring`/`:cbinstr` ABI.

## Outside this PR, same shape

Two more static return buffers with the copy-after-return contract live outside the crypto file: `sp_status_buf` in `lib/sp_process_status.c` (`Process::Status#to_s`/`#inspect`) and `sp_err_buf` in `lib/sp_process.c` (an errno message on its way into an exception). The consequence is much milder — a wrong string in a message, not a wrong signature — and neither is a security surface, so I have left them alone rather than widen this into a sweep of `lib/`. Happy to send them, or file them, if you want them.

`make test`: **3316 pass, 0 fail, 0 error.**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved thread safety for cryptographic operations by isolating returned hash, HMAC, encoding, password derivation, WebSocket, and random-data buffers per thread.
  * Prevented results from being overwritten unexpectedly when multiple threads perform cryptographic operations concurrently.

* **Tests**
  * Added multithreaded coverage verifying consistent SHA-1, SHA-256, and HMAC-SHA-256 results under contention.


matz:
Merged. I reproduced the race and the fix before taking it: on master your test reports 10,705, 15,663 and 16,054 disagreements out of 38,400 across three runs; on this branch, 0, 0, 0. The FIPS vector agrees in every run, so the oracle is not uniformly wrong in the way that would let a broken build pass. The single-threaded object is byte-identical here too.

Your argument for changing the contract rather than keeping it is the part I want to endorse explicitly, because "it was documented" is usually where this stops. A contract that can only be honoured by knowing which Ruby methods happen to bottom out in one C file is not a contract a caller can keep -- `Digest::SHA256.hexdigest(data)` is an ordinary method call, and nothing about it tells its author they are in a critical section shared with the whole process. Documenting a hazard is not the same as giving someone a way to avoid it.

The test earns its keep by being an oracle rather than a shape check. A clobbered digest is still 64 hex characters; that is exactly why this survived until someone went looking, and why a test that checked the shape would have gone on passing.

I took the two you offered, in a follow-up commit (9f36ee57): `sp_status_buf` and `sp_err_buf`. You were right that they are milder and right not to widen the PR, but the reasoning transfers unchanged and the cost is nothing -- I checked that the single-threaded assembly for both files is identical to before. A wrong string in an exception message is a smaller harm than a wrong signature, and still not one anyone can defend against from the call site.

PR #4181, opened 2026-08-29
spin: realpath the spinel binary that <$0-dir>/spinel found

rofl0r:
When spin is invoked through a symlink, $0 is the symlink path, and the "<dir-of-$0>/spinel" probe in spinel_bin also yields a symlink to the real binary. spinel_hdr_dir then walks from that symlink path looking for lib/spinel_rt.h, finds none of its fallbacks, and returns empty. The cc command for a package that includes spinel/runtime.h has no -I for the header and the build dies on a missing-header error.

The same one-line realpath the spinel_hdr_dir PATH branch already does for which("spinel") fixes it for the <$0-dir> branch: when cand is a symlink, return its realpath so the lib probe walks from the real install tree.

### DIFF
--- a/tools/spin.rb
+++ b/tools/spin.rb
@@ -78,7 +78,13 @@ def spinel_bin
   # macOS) instead of falling through to the installed binary (#3407).
   me = File.expand_path($0)
   cand = File.join(File.expand_path("..", me), "spinel")
-  return cand if File.file?(cand) && File.executable?(cand)
+  if File.file?(cand) && File.executable?(cand)
+    # Realpath: a symlinked install (e.g. /home/user/bin/spin ->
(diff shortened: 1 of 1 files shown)


matz:
Merged. Reproduced first: with `spin` reached through a symlink, the `cc` line for a package's carried C has no `-I <lib>` and the build dies on `fatal error: sp_alloc.h: No such file or directory`. With the fix, the `-I` is there and the build goes through.

The reason this is clearly right rather than merely working: the PATH branch of `spinel_hdr_dir` already does the same realpath, with a comment saying why -- "PATH returns the symlink as-is; the lib probe below then walks relative to a symlinked path and misses the install tree". The `<$0-dir>` branch had the identical exposure and not the identical fix. Restoring the symmetry is the whole change.

`make gate` green.

Issue #3782, opened 2026-08-10
[Runtime] `next` inside a `Hash#each` block loops forever


blacktm:
Found while porting [Ruby 2D](https://github.com/ruby2d/ruby2d) to Spinel.

## Description

Taking `next` inside a block passed to `Hash#each` does not advance the iterator, so the program hangs. The block runs, but reaching `next` restarts the same entry rather than moving to the following one.

`Array#each` is unaffected. `Hash#each_pair` fails the same way.

This is a regression. `git bisect` over the 78 commits since `1c3d998` gives a first bad commit of `ffb0587c` ("Refuse a new key, and keep iterating past a deleted one, in `Hash#each`").

## Reproduction

```ruby
{ a: 1 }.each do |k, v|
  next
  puts "unreachable #{k}#{v}"
end
puts 'ok'
```

**Ruby 4.0.6:**
```
ok
```

**Spinel (c70ed332):**
```
(hangs; no output, must be killed)
```

## Additional Findings

**Working — `Array#each` with the same block:**

```ruby
[1, 2].each do |v|
  next
  puts "unreachable #{v}"
end
puts 'ok'                 # prints ok
```

**Working — `Hash#each` whose `next` is never taken:**

```ruby
{ a: 1 }.each do |k, v|
  next unless v.is_a?(Numeric)
  puts "#{k}=#{v}"
end
puts 'ok'                 # prints a=1, then ok
```

The hang depends on `next` actually firing, not on its presence: rewriting the guard to the equivalent `if` form makes the same loop terminate.

**Failing — `each_pair`, same shape:**

```ruby
{ a: 1 }.each_pair do |k, v|
  next
end
```

We hit this in a validation helper that skips non-numeric entries while iterating a keyword-splat hash, which put every shape constructor into an infinite loop.

## Environment

- Spinel commit: `c70ed332` (good at `1c3d998`)
- Ruby version: 4.0.6
- Platform: macOS 26 (arm64), Apple clang 21

Issue #3569, opened 2026-08-08
[Runtime] Mutating a Hash's key set during `#each` neither raises on insert nor completes on delete


KieranP:
> (Disclaimer: I asked Claude Opus 5 (high) to find Spinel compile/runtime issues with documented Ruby classes and methods. It found the following issue, which I manually verified for accuracy before publishing this report)

_Reproduced with `spinel 8b029022e663` vs `ruby 4.0.3`._

Adding a key to a Hash from inside its own `#each` block raises `RuntimeError: can't add a new
key into hash during iteration` in Ruby; Spinel adds it silently. Deleting the current key —
which Ruby supports and which empties the Hash here — visits only the first entry under
Spinel and leaves the rest in place.

## Repro

```ruby
h001 = { a: 1, b: 2 }
r001 = (h001.each { |_k001, _v001| h001[:c] = 3 } rescue $!.class)
p r001
```

Ruby 4:

```
RuntimeError
```

Spinel:

```
{a: 1, b: 2, c: 3}
```

Deleting the current key stops the iteration early:

```ruby
h002 = { a: 1, b: 2 }
h002.each { |k002, _v002| h002.delete(k002) }
p h002
# Ruby:   {}
# Spinel: {b: 2}
```

Mutating only the values during iteration matches:

```ruby
h003 = { a: [1], b: [2] }
h003.each { |_k, v| v << 9 }
p h003                                    # => {a: [1, 9], b: [2, 9]}
h004 = { a: 1, b: 2 }
h004.each_key { |k| h004[k] = 0 }
p h004                                    # => {a: 0, b: 0}
```

PR #4183, opened 2026-08-29
Filter the general hash in place, and a boxed one

blacktm:
## Why

`select!`, `filter!`, `reject!`, `keep_if` and `delete_if` work on every typed hash and raise `NoMethodError` on the general one, the hash the inference could not fit to one of the six typed layouts: keys of more than one kind, a key kind no layout has, such as a Float or nil, or Integer keys whose values it could not narrow to Integer or String. The in-place filter emitters stop at `TY_POLY_POLY_HASH` because they read the key from the iteration order itself, and the general hash keeps a slot index there. The same loop reads its predicate as a C condition, which is right for a boolean and nothing else: a block whose last value is poly does not compile, and a nilable Integer reads its sentinel as true and its 0 as false. And it has the two faults the `each` loop was cured of in #3782 and #3569: a `next` in the block runs the same pair forever, and a pair the block deletes makes it skip the next. These all live in the two filter emitters, so they are one change: the loop becomes one function both forms call, the general hash gets it, and every variant gets the predicate read by Ruby truthiness, `next`, and CRuby's rules for a hash changed under it.

With the general hash filtering, the five names on a boxed Hash, which #4178 left for exactly this, become Array-or-Hash rows in the run-time kind switch, the way `compact!` did there.

The generated C for optcarrot and every benchmark is byte-identical to master's.

## What

- `select!`, `filter!`, `reject!`, `keep_if` and `delete_if` on the general hash raised `NoMethodError`, in expression position and as a statement.
- `{"a" => 1, "b" => nil}.select! { |k, v| v }` on a typed hash failed to compile (`invalid argument type 'sp_RbVal' to unary expression`; `reject!` and `delete_if`, which emit the condition without the `!`, failed as `statement requires expression of scalar type ('sp_RbVal' invalid)`), and a predicate whose value is a nilable Integer read its sentinel as true and its 0 as false, so `pos = "ab"; {a: "a", b: "z"}.select! { |k, v| pos.index(v.to_s) }` answered `{b: "z"}` where CRuby answers `{a: "a"}`.
- `next` in the block ran the same pair forever, on every variant: `{"x" => 1, "y" => 2}.select! { |k, v| next false if k == "y"; true }` never returns.
- A pair the block deleted, its own or an earlier one, made the loop skip the pair that slid into its slot, and a drop then deleted whichever pair had slid in; a key added during the iteration went in silently where CRuby raises.
- A receiver that is a temporary — a method's value, a literal — was not rooted for the loop, and neither were the block's parameters: with a block that allocates, `mk.select! { |k, v| "s#{v}".size > 1 }` on a method's value is a use-after-free under `SPINEL_GC_STRESS=1`, and a block that stores over its own pair and then allocates reads a freed value.
- The filter as a method's last expression answered nil: the tail took the statement form, which has no value.
- The same five names on a boxed Hash, one read from an element of a heterogeneous Array or a parameter that is an Array at one call site and a Hash at another, raised `NoMethodError` naming Hash: their rows were Array's alone.

## How

- **One loop for both forms**: `emit_hash_filter_loop` in `codegen_iter.c`, called by the expression form in `emit_hash_call` and the statement form in `emit_iteration_stmt`, which had each carried a copy. It emits the receiver in a temp, the frozen check, the pair count before and after, and the loop; the expression form answers nil or self from the counts as before.
- **The key and the value at an order position**: `hash_order_key` and `hash_order_val` name them as C text for every variant, the key itself in `order[]` for a typed hash, `keys[order[i]]` and `vals[order[i]]` for the general one, and the `TY_POLY_POLY_HASH` exclusion goes. The block's parameters take the variant's key and value types, so a general hash's block sees `sp_RbVal`s; a one-parameter block takes the key alone, as before.
- **The predicate**: `emit_cond` in place of the bare expression, the way the array filters read theirs: `sp_poly_truthy` for a poly value, the sentinel test for a nilable scalar, and for a boolean the expression as it was. `emit_cond` takes an Enumerator and an OpenStruct as the pointers they are, which the bare condition had accepted and it refused.
- **The loop's state lives outside it**, and the advance or the delete is the `for`'s own third clause, as the `each` loop's is since #3782: a `next` in the block is a C `continue`, which runs the third clause and skips whatever the body ends with. The verdict is the predicate's, or the value a `next` left, delivered through the block's next-value slot, or nil for a bare `next`. The third clause raises CRuby's `can't add a new key into hash during iteration` when the pair count grew, deletes by the key the block was given when the verdict is drop, which a block that deleted it already made a no-op, and otherwise advances only while the slot still holds that key (#3569), so a pair the block deleted, its own or an earlier one, is the only one skipped.
- **The roots**: `SP_GC_ROOT` on the receiver temp through `emit_gc_root_tmp`, and on a collectable key or value the block holds, poly or String, as the `each` loop's parameters are rooted. The collector does not scan the C stack: a value is a root only while it is on the shadow stack, and a temporary has no holder once its own expression is done, while the block body, or the general hash's delete, which rebuilds its storage, collects before the loop is through.
- **The tail**: the five names join `tap`, `then` and `yield_self` among the block-taking calls whose statement form has no value the tail could use, so a filter that is a method's last expression takes the value path.
- **The rows**: a Hash row beside each of the five Array rows, taking the Hash re-entry and write-back of #4178: `sp_poly_hash_recv` refuses a frozen original before the loop runs, the general copy is filtered, and `sp_poly_hash_writeback` puts the survivors back into the typed original in its own layout. A filter only removes, so no entry is ever refused. The value is the receiver itself, as `compact!`'s is, or nil when a `!` form removed nothing, so a write through it reaches the original. The comment 26adc3c4 left on the Array rows, that four of them are not Array's alone, goes with them (`filter!`, `select!`'s alias, is the fifth).

## Validation

- A new test, `test/poly_poly_hash_filter_bang` (expectations generated by CRuby 4.0): the five names on a general hash with Integer, String, Symbol, Float and nil keys, in expression position and as statements, with the nil-or-self answers; the order of what stays; a one-parameter block and numbered parameters; a receiver that is a method's value; a predicate whose last value is nil, false, 0, "" or a String; a multi-statement block; a typed hash with a poly value and a predicate whose value is a nilable Integer, read the same way; the filter as a method's value; `next` with a value, bare, and as the block's last statement; the key added during the iteration that raises; the pair deleted during it, the block's own, an earlier one, a later one; and the frozen receiver whose block never runs. `test/boxed_hash_methods` grows a section: the five names writing back into a typed hash of each layout and into the general hash through a box, the nil-or-self answers, a write through the value reaching the original, `equal?`, a parameter that is an Array at one call site and a Hash at another, a slot the inference widened to poly, the frozen original, and the `NoMethodError` a receiver of neither kind raises.
- Full suite: 3,319 pass, 0 fail, from a cleared results cache.
- Generated C against master: byte-identical for every program except the two tests above and four others: `hash_delete_if_select_reject_bang`, `hash_filter_bang_return` and `hash_int_keyed_delete`, which filter a typed hash in place and take the new loop shape, and `boxed_array_methods`, where the filters on a boxed Array now go through the kind switch (the Array arm the face emitted before, a Hash arm, and the `NoMethodError` default) instead of straight into the Array re-entry. optcarrot and the 61 benchmarks are byte-identical, so the hot paths do not change.
- The two tests, `test/hash_filter_bang_return`, and probes with temporary receivers, allocating predicates, blocks that store over their own pair, and hashes large enough to rehash on each layout run clean under `SPINEL_GC_STRESS=1` and ASAN+UBSAN with CRuby's output; the receiver probe is a use-after-free on master. The inference fixpoint converges in the same number of rounds as master on every program in the suite (one program, `test/enumerator_block_returns_self`, varies between 3 and 4 from run to run under master's own binary as well), and in 5 on the new test.
- Against the differential corpus of ~2,100 minimized programs the campaign has been working from: 707 → 707 match, 0 gained and 0 lost. No program in it filters a general hash in place; the one that filters a boxed Hash, read from an element of a heterogeneous Array, now answers `keep_if` and `delete_if` as CRuby does and still differs on `shift` and `rehash`.
- Front-end compile time of optcarrot (`-c --no-line-map` to `/dev/null`) is unchanged: 1.02–1.03 s on master and 1.02 s on this branch over three runs of each.

## Left alone, on purpose

- The general hash's delete rebuilds and rehashes the whole table into a fresh allocation for each removed pair, where the typed variants shift their order array in place. Both are linear per delete, so a filter that empties a large hash is quadratic either way, but the general one is some 40× the constant (about 1 s against 0.03 s to empty an 8,000-pair hash). A delete that compacts in place is a runtime change of its own.
- The array in-place filters take the same unrooted receiver temp and lower `next` to the same bare `continue`: `mk.select! { |v| "s#{v}".size > 1 }` on a method's value is a use-after-free under `SPINEL_GC_STRESS=1`, and `[1, 2, 3].select! { |x| next true if x == 2; false }` answers `[]` where CRuby answers `[2]`, on master and here alike. They are this loop's fixes in a different emitter and belong with the array filters.
- The blockless forms, an Enumerator in CRuby, raise `NoMethodError` as they did on every variant, boxed or not; a block parameter with a default (`|k, v = nil|`) is not read by the filter emitters on any variant, on master as well.
- The FrozenError message names the class without the receiver's inspect (`can't modify frozen Hash`), as every Hash mutator's does; the tests print the class.
- `Hash#shift` and `rehash` on a boxed Hash, the other two names of that corpus program: `shift` has an Array arm of its own in the builtin-method dispatch that the table would have to subsume, and the typed `rehash` answers a copy of the receiver where CRuby answers the receiver, a difference of its own that comes before any row.
- A block that writes through its own boxed receiver (`x[k] = v` inside `x.select! { }`) does not compile through the kind switch, for a boxed Array on master and here alike, and so for a boxed Hash now; a read through it (`x[k]`, `x.size`) is fine. And a boxed empty literal written to with an Integer key (`h = gh({}); h[1] = "a"`) is stored to as an Array on master and here alike, so a filter after it never sees a Hash. Both are the face and the poly `[]=`, not the rows.
- Two runtime faults the probes met on master and this branch alike, outside the filter: a key that is a temporary in `h[k] = v` has no holder while the value expression allocates, and a Symbol interned from a temporary String can be collected before the copy.

## Summary by CodeRabbit

- **New Features**
  - Added in-place filtering support for all hash types, including dynamically typed hashes.
  - Hash filtering methods now support consistent block behavior, chaining, deletion, and return values.
  - Added runtime dispatch for hash and array filtering methods.

- **Bug Fixes**
  - Improved handling of frozen hashes, iteration changes, truthiness, and deleted entries.
  - Corrected truthiness checks for enumerators and open structures.

- **Tests**
  - Expanded coverage for hash filtering, mixed key/value types, error cases, and receiver identity.


matz:
Merged as b75baf51, rebased on master.

This is the first one in the series that fixes existing behaviour rather than only adding rows, so I checked the three claims that carry consequences, and each reproduced exactly as you describe.

**`next` in the block hangs on master.** `{"x" => 1, "y" => 2}.select! { |k, v| next false if k == "y"; true }` runs forever under the current compiler -- I had to cap it with a timeout -- and answers `{"x" => 1}` here. A hang is the worst failure mode in the set and it was reachable from ordinary Ruby.

**The use-after-free is real.** Your receiver probe on master, under ASAN and `SPINEL_GC_STRESS=1`: `heap-use-after-free ... READ of size 8`. Clean on this branch. Your reasoning for why -- the collector does not scan the C stack, so a temporary has no holder once its own expression is done, while the block body and the general hash's delete both allocate -- is the part that makes it a rooting bug rather than a lifetime accident.

**The predicate.** `{"a" => 1, "b" => nil}.select! { |k, v| v }` did not compile, and the nilable-Integer case answered `{b: "z"}` where CRuby answers `{a: "a"}` -- a wrong answer from reading a sentinel as true. Both match CRuby now.

The measurements hold too: 3,283 programs compiled with both compilers, exactly the four existing ones you name differ, optcarrot byte-identical. Your two tests plus `hash_filter_bang_return` run clean under ASAN+UBSAN with the GC stress on.

Two things about the shape of the change I want to record, because they are why it is one change rather than five.

**One loop for both forms.** The expression form and the statement form had each carried a copy, which is how a fix could land in one and not the other. The `next`, the predicate, the rooting and the deleted-pair rules are all properties of the loop, so they arrive together and cannot drift apart again.

**Making the advance the `for`'s third clause.** That is what turns `next` into a plain `continue` instead of something the loop has to know about, and it is the same shape #3782 gave the `each` loop. Reusing a fix's *form* and not just its conclusion is what stops the second occurrence.

PR #4193, opened 2026-08-29
Keep what the block said in an array filtered in place

blacktm:
## Why

The in-place array filters, `select!`, `filter!`, `reject!`, `keep_if` and `delete_if`, run their block in a C loop that writes the kept element down at the end of the loop body. A `next` in the block is a C `continue`, which skips that write, so `[1, 2, 3].select! { |x| next true if x == 2; false }` answers `[]` where CRuby answers `[2]`, and any `next` drops the element whatever value it carries. Whatever leaves the loop early — a `break`, a `raise`, a `return`, a `throw` — leaves the array as the compaction had it: `[2, 2, 3, 4, 5]` for a break at 3, where CRuby answers `[2, 3, 4, 5]`, and a `delete_if` whose block raises at 3 has lost the deletion of 2 it already made, which CRuby keeps. And the receiver is not rooted for the loop, so a filter on a method's value with a block that allocates is a use-after-free under `SPINEL_GC_STRESS=1`. These are the same loop, emitted twice, for the typed array and the poly array; this makes it one, `emit_array_filter_loop`, and cures the three the way the hash filters were cured in #4183, whose loop this one sits beside.

The generated C for optcarrot and every benchmark is byte-identical to master's.

## What

- `next` in the block of an in-place array filter dropped the element, whatever value the `next` carried, on every array kind: `[1, 2, 3].select! { |x| next true if x == 2; false }` answered `[]`; `%w[a bb ccc].delete_if { |s| next false if s == "bb"; s.size > 2 }` answered `["a"]` where CRuby answers `["a", "bb"]`; an array of Structs is a poly array and takes the same loop, so `[S.new(1), S.new(2), S.new(3)].select! { |s| next true if s.a == 2; false }` answered `[]`.
- Whatever left the block early left the array half-compacted with its old length: `a = [1, 2, 3, 4, 5]; a.select! { |x| break if x == 3; x > 1 }` left `[2, 2, 3, 4, 5]`; `a = [1, 2, 3, 4]; a.delete_if { |n| raise "boom" if n == 3; n == 2 }` rescued left `[1, 2, 3, 4]` where CRuby leaves `[1, 3, 4]`.
- A frozen receiver was refused by the first kept element's write, after the block had run for it, where CRuby refuses it before the block runs — and a filter that kept nothing refused nothing: `[1, 2, 3].freeze.select! { false }` answered `[]` and left the frozen array empty. A block that wrote to its receiver at the element it was given had the write undone by the compaction, where CRuby writes an element down only when it moves.
- A receiver that is a temporary — a method's value — was not rooted for the loop: `mk(30).select! { |s| "x#{s}".size > 2 }` on a method's value is a use-after-free in `sp_StrArray_set` under `SPINEL_GC_STRESS=1`. The element the block holds was not rooted either, and the loop writes it back after the block has run: a block that stores over its own element and then allocates (`a.select! { |x| k += 1; a[k] = "X"; "pad" * 400; true }` over 30 Strings) is a use-after-free in `sp_StrArray_scan` under `SPINEL_GC_STRESS=1`.

## How

- **One loop for both arms**: `emit_array_filter_loop` in `codegen_iter.c`, called from both arms of `emit_array_call` in `codegen_call_recv.c`, the typed array's and the poly array's, which had each carried a copy; each answers nil or self from the loop's before-and-after lengths as before.
- **The frozen check and the store**: a frozen receiver raises before the loop, as CRuby's `rb_ary_modify_check` does, and a kept element is written down only when `i1 != i2`, as CRuby's is. The predicate goes through `emit_cond`, which the poly arm already used and the typed arm had its own truthiness switch for.
- **The advance and the keep in the `for`'s third clause**, as the `each` loop's advance is since #3782: a `continue` runs it. The verdict is the predicate's, read by `emit_cond`, or the value a `next` left, delivered through the block's next-value slot, or nil for a bare `next`.
- **An ensure region around the loop**: the protocol of `emit_begin`'s ensure clause — the frame, the landing, the `_ensure` label, the deferred `return`, the re-raise, the chain to an enclosing region — with an ensure body that is C alone: the elements from the one the block was given on are moved down behind the kept ones and the length is cut, as CRuby's `select_bang_ensure` does. A `break` reaches it either way, as a throw when the call sits in a break frame or by falling out of the `for` into the label; a `raise` lands in it; a `throw` and a proc `return` pass through it by `sp_unwind_resume`; a normal exit runs the same body with nothing to move. At the ensure stack's limit the loop goes without the region, as `emit_begin`'s clause does, rather than decline. It costs the loop one `setjmp` per call, which CRuby's `rb_ensure` costs too.
- **The roots**: `emit_gc_root_tmp` on the receiver temp, and a root on the element the block is given when it is collectable, a String or a poly value. The collector does not scan the C stack: a value is a root only while it is on the shadow stack, and a temporary has no holder once its own expression is done. The two indexes are `volatile`, since an early exit reads them past the `setjmp`.

## Validation

- A new test, `test/array_filter_bang_next_break` (expectations generated by CRuby 4.0): `next` with a value, bare, as the block's last statement, and inside a nested `if`, on Integer, String, Float, Symbol, mixed and Struct arrays, with each of the five names; `break` bare and with a value, on each kind, and a `raise`, a `return` and a `throw`, with the array as each leaves it, and `next` and `break` in one block; a frozen receiver refused before the block runs, a block that writes to its receiver at its own element; the nil-or-self answers as before; a receiver that is a method's value with a block that allocates; a block that reassigns the receiver's variable.
- Full suite: 3,327 pass, 0 fail, from a cleared results cache.
- Generated C against master: byte-identical for every program except the new test, the five that filter an array in place (`array_select_bang`, `array_filter_bang_return`, `poly_filter_bang_stmt_block`, and `boxed_array_methods` and `boxed_hash_methods`, whose Array arms do), and 47 that use Set, whose `delete_if`, `select!` and `keep_if` in the prelude take the new loop; past the loop itself those diffs are renumbering — temps, write-barrier temps and inlined-block locals all draw on one counter — since the loop declares five more. optcarrot and the 61 benchmarks are byte-identical, so the hot paths do not change.
- The new test, `test/array_select_bang`, `test/array_filter_bang_return`, and probes with method-value receivers and allocating predicates, `next` inside a `case` and across an `ensure`, `break` across an `ensure`, a block that appends to the receiver, the filter as a method's value, and nested and nil-holding arrays run clean under `SPINEL_GC_STRESS=1` and ASAN+UBSAN with CRuby's output; the receiver probe is a use-after-free on master. The inference fixpoint converges in the same number of rounds as master on every program in the suite (one program, `test/enumerator_block_returns_self`, varies between 3 and 4 from run to run under master's own binary as well).
- Against the differential corpus of ~2,100 minimized programs the campaign has been working from: 707 → 708 match, 1 gained and 0 lost; the one is the corpus's program for an interrupted `delete_if`, whose block raises after a deletion it had already made.
- Front-end compile time of optcarrot (`-c --no-line-map` to `/dev/null`) is unchanged: 1.00 s on master and 0.99–1.00 s on this branch over three runs of each, both built the same way.

## Left alone, on purpose

- A block that appends to its receiver during the filter is visited to the new end, as CRuby's is. A block that deletes from its receiver, or clears it, answers CRuby's array only when the shift happens to leave the write index where CRuby's ensure would: `b = [1, 2, 3, 4]; b.select! { |x| b.delete_at(0) if x == 1; x > 0 }` leaves `[1, 3, 4]` where CRuby leaves `[2, 3, 4]`, on master and here alike. Matching CRuby there means holding the visited elements aside, which is the loop's shape, not its verdict.
- A filter on a boxed receiver (#4175's Array rows) works on the poly copy a typed original is normalized to and writes it back once the value is taken, so an early exit from the block leaves the original as it was; that is the write-back's shape, shared with every boxed mutator, not this loop's.
- A block that clears or replaces its receiver during the filter: the compaction then writes past the new length and the array grows back with nils (`a.select! { |x| a.clear; true }` leaves `[nil, 2]` where CRuby leaves `[]`), on master and here alike; the write is the loop's shape again.
- A `next` inside a `while` inside the block reaches the block's next-value slot, so `a.select! { |x| i = 0; while i < 2; i += 1; next true if i == 1; end; next }` keeps every element where CRuby keeps none; the `while` emitter leaves the slot armed, and `map` and `select` on master show the same, so it is that emitter's, not this loop's.
- The blockless forms, an Enumerator in CRuby, are as they were.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved array filtering operations including `select!`, `filter!`, `reject!`, `keep_if`, and `delete_if`.
  - Correctly preserves array contents when blocks use `next`, `break`, `raise`, `return`, or `throw`.
  - Ensures frozen arrays raise `FrozenError` before block execution.
  - Preserves expected return values and behavior across typed and general arrays.


matz:
Merged in 65dd8039. Verified: the generated C differs from master's in 58 suite programs, every one a user of the in-place filters, and optcarrot's is byte-identical (md5). Your unrooted-temporary claim reproduced on master with `make.select! { ... }` under `SPINEL_GC_STRESS=1` (corrupted output) and holds on the branch; an adversarial sweep -- `next` with a value on all three array kinds, `break` leaving the array as CRuby's select_bang_ensure leaves it, a `raise` keeping the deletions already made, a `return` through the filter, a frozen receiver refused before the block, the no-change `select!` answering nil -- matches CRuby line for line, under stress included.

Folding the statement and expression forms into one loop is the part that pays rent: #4183 showed the same fix landing in only one of two copies, and now there is one. The ensure-region reuse of emit_begin's protocol, with the keep in the for's third clause so `next` runs it, is the right shape for the family.

Issue #4187, opened 2026-08-29
`Array()` ignores a user object's `#to_ary`/`#to_a` when the argument arrives through an untyped parameter (#3721 covered the static path only)


rubys:
`#3721` taught `Array()` to ask a user object for `#to_ary` / `#to_a`, and that
fix is on the STATIC path only. When the argument reaches `Array()` through a
parameter whose type is not known — an `untyped` param, a poly slot — codegen
falls to `sp_kernel_array`, which wraps the object instead of converting it.

Same object, same method, two answers, decided by whether the call site could
see the class.

## Repro

```ruby
class Box
  def to_ary
    [1, 2, 3]
  end
end

class Bag
  def to_a
    [4, 5, 6]
  end
end

def wrap(x)
  Array(x)
end

p wrap(Box.new)   # through an untyped param
p wrap(Bag.new)
p Array(Box.new)  # directly, class statically known
p Array(Bag.new)
```

```
$ ruby arr.rb
[1, 2, 3]
[4, 5, 6]
[1, 2, 3]
[4, 5, 6]

$ spinel -E arr.rb          # 5a63653cf613
[#<Box:0x0000000103405de0>]
[#<Bag:0x000000010105a3f0>]
[1, 2, 3]
[4, 5, 6]
```

The last two lines are #3721's arm working. The first two are the same
conversion asked for through a parameter.

## Where I think it is

`src/codegen_call.c:17469` — the `Array()` arm dispatches on
`comp_ntype(c, av[0])`, and its `to_ary` / `to_a` case is guarded by
`ty_is_object(at)`. A poly argument is not `ty_is_object`, so it takes the
final `else` at :17480:

```c
else { buf_puts(b, "sp_kernel_array("); emit_boxed(c, av[0], b); buf_puts(b, ")"); }
```

`sp_kernel_array` (`lib/spinel_rt.h:3685`) handles nil, a poly array, and the
three typed arrays. For anything else — including a `SP_TAG_OBJ` user object —
it pushes the value into a one-element array. Its comment says it runs "for
nil/scalars and for a poly value that may hold an array at run time"; a poly
value that holds a CONVERTIBLE OBJECT is the case that isn't in that list.

The runtime has the shape this needs already: `sp_obj_to_str_fn` /
`sp_obj_to_int_fn` / `sp_obj_to_path_fn` are exactly "ask a class id for its
conversion method", installed from the generated TU by `emit_conv_bridge`
[...]

Issue #1261, opened 2026-06-01
AOT Compiler Limits: What Spinel Cannot Do


neidiom:
keeping a track here - https://github.com/neidiom/spinel_docs

matz:
Thanks for keeping this catalogue. I went through all 30 entries against current master, and a big chunk of the doc describes the **old Ruby self-hosted compiler** — many of those limits are gone now. I've added a current, canonical version in-tree: **[`docs/limitations.md`](https://github.com/matz/spinel/blob/master/docs/limitations.md)** (and [`docs/INCOMPATIBILITIES.md`](https://github.com/matz/spinel/blob/master/docs/INCOMPATIBILITIES.md) for the intentional deviations).

**Already works now (the doc lists these as limits — stale):** mutable string literals (`s << x`, `"x".frozen?` → false), hash missing key → `nil` (str/int keyed), `define_method(:literal)`, block-param arity (un-yielded → `nil`), closures through containers (`{op: ->(a,b){a+b}}[:op].call(2,3)`), `String#oct`, `Array#first` on empty → `nil`, and hash-variant widening. There's also no "bootstrap fixpoint" constraint anymore — the C compiler is master.

**Relaxed in this pass:** implicit-self `send(:literal)` / `__send__("literal")` / `public_send(:literal)` now lower to a direct call (a37a7908). Previously only the explicit-receiver form (`"x".send(:upcase)`) worked. Done on the AST, so a `send(:` inside a string or comment is left untouched.

**Genuinely fundamental (won't change without abandoning AOT):** `eval`, `method_missing` on arbitrary receivers, `ObjectSpace`, `TracePoint`/`binding`, refinements, `callcc`, `Class.new(parent)`, general reflection, `require` of stdlib `.rb`, user `#hash`/`#eql?` for hash keys, mixed encodings, arbitrary-object `Marshal`.

**Relaxable next (additive, by effort):** non-empty `backtrace`/`caller` from a compile-time line side-table; `Marshal` of primitives; lifecycle hooks (`included`/`inherited`/`extended`) emitted at startup; external `Enumerator` (`.next`) via Fiber suspension. These are catalogued in `docs/limitations.md`.

Closing this in favour of the in-tree doc, which I'll keep current. If you spot a specific limit that matters for your use, open a focused issue with a repro and I'll look at relaxing it.

PR #4194, opened 2026-08-29
Consult the receiver's class for Time#min through a poly slot

### LINKED ISSUE #4192: `Time#min` through a poly slot answers nil (or raises), where every sibling accessor is correct
`Time#min` through a poly slot answers `nil` — or raises — where every
sibling accessor is correct.

## Repro

No `.rbs` needed; a mixed-element array boxes its members.

```ruby
box = [ Time.now, "x" ]
t = box[0]
p t.min      # CRuby: 56   spinel: nil
p t.hour     # 11 on both  <- control
```

```
$ spinel nosig.rb && ./nosig
nil
11
$ ruby nosig.rb
56
11
```

Verified on `22ee6ca0`.

## It is `min` alone

Every other `Time` accessor through the same poly slot agrees with CRuby.
Sweeping them (`def self.poly: (untyped t, String name) -> untyped`):

| | `year` | `month` | `mon` | `day` | `mday` | `hour` | **`min`** | `sec` | `usec` | `wday` | `yday` | `to_i` |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| spinel | 2026 | 8 | 8 | 29 | 29 | 11 | **(nil)** | 43 | 53700 | 6 | 241 | 1788019003 |
| CRuby | 2026 | 8 | 8 | 29 | 29 | 11 | **56** | 43 | 53700 | 6 | 241 | 1788019003 |

A typed receiver is correct: `Time.now.min` answers the minute. Only the
poly path diverges.

## Two symptoms, one root — and the silent one is the default

Whether it answers `nil` or raises depends on something unrelated to the
call: whether any *user* class in the program happens to define `min`.

```ruby
class Clock
  def min; 42; end
end

module M
  def self.poly(x) = x.min          # (untyped x) -> untyped
end

M.poly(Clock.new)   # 42   — correct
M.poly(Time.now)    # NoMethodError: undefined method 'min' for an instance of Time
```

With `Clock` present the switch has a user arm and its `default:` raises.
With no user class defining `min` (the first repro) the call routes to
`sp_poly_min` and returns `nil` with no error. **A user class defining
`min`, `max`, `sum`, `first`, `last` or `sample` dispatches correctly
through a poly slot in both cases** — measured, all six. The gap is for a
*built-in* that owns one of those names.

## Where it goes

`src/codegen_call.c` — on a `TY_POLY` receiver with `argc == 0`, the name
routes straight to the enumerable helper:

```c
if (recv >= 0 && rt == TY_POLY && argc == 0 && nt_ref(nt, id, "block") < 0) {
    const char *pm = NULL;
    if (sp_streq(name, "sum")) pm = "sp_poly_sum";
    else if (sp_streq(name, "min")) pm = "sp_poly_min";
```

and `sp_poly_min` (`lib/spinel_rt.h:7235`) asks whether the box
enumerates:

```c
default: { sp_PolyArray *ue = sp_poly_user_elems(v);
           return ue ? sp_PolyArray_min(ue) : sp_box_nil(); }
```

A boxed `Time` has no elements, so it takes `sp_box_nil()`. Nothing
between those two points asks whether the receiver's own class defines
`min`.

Worth noting the neighbours in that same list already carry the guard the
report is about: `value` / `resume` / `join` route to the Fiber helpers
only *"when no user class defines the name (#1261)"*. `min` / `max` /
`sum` / `first` / `last` / `sample` have no such condition.

`Time` is the case I hit; the shape is "a built-in whose own method is
named like an Enumerable one", so a fix that consults the receiver class
first would cover it more generally than a `Time` arm — the same move
`to_a` / `to_ary` took with `sp_obj_to_a_fn` / `sp_obj_to_ary_fn`
(#3234, #4187).

## Where it came from

Rails stores a timestamp as TEXT and formats it with `%02d` per field.
The formatter's parameter is `untyped`, so the receiver is poly:

```ruby
def self.format_db_time(value)     # (untyped value) -> String?
  return nil if value.nil?
  if value.is_a?(Time)
    t = value.utc
    ...
    return format("%04d-%02d-%02d %02d:%02d:%02d.%06d",
                  t.year, t.mon, t.mday, t.hour, t.min, t.sec, micros)
```

`t.min` arrives as `nil` and `format` reports it one step later, as
`TypeError: can't convert nil into Integer` — with nothing naming `min`.
In the app this is `Session#before_create` writing `last_active_at`, so
signing in raises a 500 and no session row is ever written.

rubys:
Fixes #4192.

## The bug

`Time#min` read out of a poly slot answers `nil`, where every sibling accessor is correct:

```ruby
box = [ Time.now, "x" ]   # a mixed array boxes its members
t = box[0]
p t.min      # CRuby: 56   spinel: nil
p t.hour     # 11 on both  <- control
```

`year month mon day mday hour sec usec wday yday to_i` all agree with CRuby through the same slot, and a typed `Time.now.min` is right.

## One omission, two symptoms

**The receiver's own class is never asked.** `min`/`max`/`sum`/`first`/`last`/`sample` on a `TY_POLY` receiver take an enumerable fast path (`codegen_call.c`) guarded on whether a *user* class owns the name. That guard is correct as far as it goes — but a **builtin** owning the name is never considered, and `Time` owns `min`.

Which symptom you get depends on something unrelated to the call:

* **No user class defines `min`** — routes to `sp_poly_min`, whose `default:` asks a boxed Time for `sp_poly_user_elems` it does not have, and answers `nil`. Silent.
* **Some user class defines `min`** — takes `emit_poly_method_dispatch` instead, whose container guard is `is_array_kind || is_hash_kind`. A boxed Time matches neither, so it falls past to the `NoMethodError` default.

The emitted C shows it two lines apart inside the same `is_a?(Time)` guard:

```c
hour →  _t6.cls_id == SP_BUILTIN_TIME ? sp_time_hour(*(sp_Time *)_t6.v.p) : raise_nomethod(...)
min  →  sp_poly_min(lv_value)          // no class check at all
```

## The fix

`sp_poly_min` gains a `SP_BUILTIN_TIME` arm, and the container-kind guard gains Time for `min`, so both paths reach the same answer. This is the `sp_obj_to_a_fn` / `sp_obj_to_ary_fn` shape from #3234 and #4187: consult the class, then fall back to the generic helper.

The other five names are left alone — no builtin in the tree owns them.

## A correction to the issue

#4192 proposes that these six names are missing the `#1261` guard that `value`/`resume`/`join` carry. **That is wrong** — they carry it already. What is missing is that the guard counts user classes only. I've corrected the issue.

## Verification

* `test/time_min_poly_slot.rb` covers both symptoms, the four container kinds `sp_poly_min` already served (int / string / float arrays and a hash), and a typed control.
* Ablated: with the fix reverted the test dies on its first line with `undefined method 'min' for an instance of Time`.
* `make test`: **3326 pass, 0 fail, 0 error**.

Found walking campfire's `POST /first_run`, where `Session#before_create` writes `last_active_at` and a nil `min` surfaces two frames later as `TypeError: can't convert nil into Integer` from `format("%02d", t.min)`, with nothing naming `min`.

https://claude.ai/code/session_01RxsSmoZPGHx5F8pDi7Ye65

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed polymorphic `Time#min` calls to return the minute component instead of `nil`, raising an error, or calculating an enumerable minimum.
  * Preserved correct `min` behavior for arrays, hashes, and other supported value types.
  * Confirmed related `Time#hour` behavior remains unchanged.


matz:
Merged in 1b2fe685. Verified before merging: the generated C across the 3,288-program suite differs only at the guard your commit extends (one test program, `poly_dispatch_builtin_container_arm`, still matching its expectation), and your test holds under `SPINEL_GC_STRESS=1` with CRuby's output.

Your diagnosis of the two symptoms and their one root was exactly right, and the correction you posted to the issue -- that the #1261 guard is present but counts user classes only -- saved me from chasing the wrong predicate.

Probing the same switch turned up the identical pair of symptoms for the receiver kind next to Time: a boxed Range answered nil for min/max/first/last (sum alone had an arm from #3234), and with a user class owning any of those names it raised NoMethodError instead, exactly like your second symptom. Follow-up in f5da8cb7: Range arms for the four (first is the begin with no emptiness check, last is the end with exclusivity untouched, both placed before the user_elems read that would materialize the range and answer the enumerated element), SP_BUILTIN_RANGE admitted to the dispatch guard for all five names, and the min/max defaults now raise NoMethodError for a receiver that is nothing the switch knows -- `[nil, 7][0].min` was a silent nil where CRuby raises. The full gate holds, so nothing was leaning on that nil.

Also: campfire's `POST /first_run` as the provenance trail, ending in a `format("%02d", t.min)` TypeError two frames from the cause, is a good argument for the silent-nil default being the worse of the two symptoms. It is gone now.

PR #4209, opened 2026-08-30
io: honor autoclose: false in close and finalizer; auto-detect mode i…

### LINKED ISSUE #4208: `IO.for_fd(fd, autoclose: false); io.close` closes the underlying fd
## Summary

In Spinel AOT, calling `IO.for_fd(fd, autoclose: false)` and then closing
the wrapper (`io.close` or letting the wrapper go out of scope) closes
the underlying fd. CRuby leaves the fd open. The `autoclose: false`
flag has no effect under AOT.

## Where

`spinel/lib/sp_io.c:34` (`sp_File_fin`):

```c
static void sp_File_fin(void *p) {
  sp_File *f = (sp_File *)p;
  if (f->fp) { fclose(f->fp); f->fp = NULL; }
}
```

`sp_File_fin` is the GC finalizer. It always calls `fclose(f->fp)`,
ignoring `f->no_autoclose`.

`spinel/lib/sp_io.c:230` (`sp_File_close`):

```c
sp_int sp_File_close(sp_File *f) {
  if (f && f->fp && f->fp != stdout && f->fp != stderr && f->fp != stdin) {
    fclose(f->fp); f->fp = NULL;
  }
  return 0;
}
```

Same problem: `fclose` is called unconditionally; `no_autoclose` is
not checked.

`spinel/lib/sp_cold.c:2986` (`sp_io_for_fd`) sets the flag correctly
when `autoclose: false` is passed, but nothing reads it:

```c
sp_File *sp_io_for_fd(sp_int fd, const char *mode, sp_bool autoclose) {
  if (fd < 0 || fcntl((int)fd, F_GETFD) < 0)
    sp_raise_cls("Errno::EBADF", "Bad file descriptor");
  sp_File *f = sp_io_fdopen_ex((int)fd, mode && *mode ? mode : "r", 0);
  if (f && !autoclose) f->no_autoclose = 1;
  return f;
}
```

The fix: in `sp_File_close` and `sp_File_fin`, check `!f->no_autoclose`
before `fclose(f->fp)`.

## Reproducer

Single self-contained Ruby file at `/tmp/autoclose_bug.rb`:

```ruby
FIFO = ARGV[0] || "/tmp/autoclose_test_fifo"
fd = IO.sysopen(FIFO, File::WRONLY | File::NONBLOCK)
io = IO.for_fd(fd, autoclose: false)
io.close

begin
  io2 = IO.for_fd(fd, autoclose: false)
  io2.puts "hello"
  io2.flush
  io2.close
  puts "ok: wrote through fd=#{fd}"
rescue => e
  puts "FAIL: #{e.class}: #{e.message}"
end
```

Setup (a reader must be waiting on the FIFO so the open does not ENXIO):
```
mkfifo /tmp/autoclose_test_fifo
( cat /tmp/autoclose_test_fifo & )
```

Build and run:
```
/home/user/spinel/bin/spinel /tmp/autoclose_bug.rb -o /tmp/autoclose_bug
/tmp/autoclose_bug
```

Expected (CRuby):
```
ok: wrote through fd=3
```

Observed (Spinel AOT):
```
FAIL: IOError: fdopen failed
```

Under AOT, the first `io.close` triggers `fclose(fd)`, which closes
fd 3. The second `IO.for_fd(3, ...)` calls `fcntl(3, F_GETFD)`,
which returns 0 (the fd table still has a slot), then `fdopen(3, "r")`
on a closed fd returns NULL, which becomes `IOError("fdopen failed")`.

## Build environment

- Spinel HEAD: 5a63653c

rofl0r:
…n for_fd

Three related fixes for IO.for_fd / IO.new wrap semantics under AOT:

  1. sp_io_for_fd (lib/sp_cold.c): when the user does not pass a mode (or passes ""), pick a mode that matches the fd's actual open flags instead of defaulting to "r". CRuby's IO.for_fd is lenient here: a write-only fd wraps cleanly with no mode arg, but libc fdopen(fd, "r") on a write-only fd fails. Probe F_GETFL and pass "w" / "r+" / "r" to match.

  2. sp_File_close and sp_File_fin (lib/sp_io.c): honour the no_autoclose flag set by sp_io_for_fd. The flag was set but never read, so io.close and the GC finalizer would fclose a fd that the caller opened via IO.sysopen and asked us not to take ownership of. Honour it and the caller's fd survives the wrap.

  3. IO.for_fd codegen (src/codegen_call.c): detect the user-written string literal at the second positional by AST node type (StringNode) rather than the inferred type (TY_STRING). In larger inference contexts the inferred type widens to poly and the old check misfired, silently dropping the mode arg and defaulting to "r". When the user did not write a string literal at the call site, emit NULL so the runtime can auto-detect instead of being lied to with a hard-coded "r".

Together these bring the AOT behaviour in line with CRuby's IO.for_fd / IO.new semantics: a write-only fd wraps cleanly with no mode arg, autoclose: false keeps the underlying fd open across io.close and finalization, and an explicit mode survives the inference pass.

closes #4208

## Summary by CodeRabbit

* **Bug Fixes**
  * File wrappers now automatically select read, write, or read/write access based on the underlying file descriptor when no mode is specified.
  * Explicit file modes continue to work as provided.
  * Files opened with automatic closing disabled are no longer closed unexpectedly.
  * Improved handling of mode detection for file descriptor operations.

### DIFF
--- a/lib/sp_cold.c
+++ b/lib/sp_cold.c
@@ -2981,7 +2981,22 @@ sp_PolyArray *sp_io_pipe(void) {
 sp_File *sp_io_for_fd(sp_int fd, const char *mode, sp_bool autoclose) {SP_GC_ROOT_STR(mode);
   if (fd < 0 || fcntl((int)fd, F_GETFD) < 0)
     sp_raise_cls("Errno::EBADF", "Bad file descriptor");
-  sp_File *f = sp_io_fdopen_ex((int)fd, mode && *mode ? mode : "r", 0);
+  /* When the user did not pass a mode (mode is NULL or ""), CRuby picks a
+     mode that matches the fd's actual open flags via a "first fdopen
+     that does not fail" fallback. Mirror that: probe the fd's flags and
+     pass the closest fopen mode, so wrapping a write-only fd works
+     without forcing the caller to spell the mode. */
+  const char *m;
+  if (mode && *mode) {
+    m = mode;
+  } else {
+    int fl = fcntl((int)fd, F_GETFL);
+    if (fl < 0) m = "r";
+    else if ((fl & O_ACCMODE) == O_WRONLY) m = "w";
+    else if ((fl & O_ACCMODE) == O_RDWR)   m = "r+";
+    else                                   m = "r";
+  }
+  sp_File *f = sp_io_fdopen_ex((int)fd, m, 0);
   if (f && !autoclose) f->no_autoclose = 1;
   return f;
 }
--- a/lib/sp_io.c
+++ b/lib/sp_io.c
@@ -31,7 +31,14 @@ extern void *sp_gc_alloc(size_t sz, void (*fin)(void *), void (*scn)(void *));
 extern SP_NORETURN void sp_raise_cls(const char *cls, const char *msg);
 extern const char *sp_sprintf(const char *fmt, ...);
 
-static void sp_File_fin(void *p) { sp_File *f = (sp_File *)p; if (f->fp) { fclose(f->fp); f->fp = NULL; } }
+static void sp_File_fin(void *p) {
+  sp_File *f = (sp_File *)p;
+  /* f->no_autoclose is set by sp_io_for_fd when the user passed
+     autoclose: false. The runtime owns the underlying fd in that case
+     (closing it would surprise the caller, who is wrapping a fd they
+     opened via IO.sysopen or a similar call). Honour the flag. */
+  if (f->fp && !f->no_autoclose) { fclose(f->fp); f->fp = NULL; }
+}
 static void sp_File_scan(void *p) { sp_File *f = (sp_File *)p; if (f->path) sp_mark_string(f->path); if (f->mode) sp_mark_string(f->mode); }
 
 /* The two-argument form is the permission form with CRuby's default bits:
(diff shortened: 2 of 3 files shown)


rofl0r:
fixed by 0c6684518ea465754b1a61b4d6a2676559883408

matz:
Apologies for the collision -- your PR and my fix for #4208 landed within the same quarter hour, and I only saw this after pushing. Thank you for closing it so gracefully, and more for the report quality that made two independent fixes converge.

Where the two differed, for the record:

- **The close path**: skipping fclose honors the flag but leaks the FILE structure (libc has no fdclose) and, in the finalizer, skips the flush -- buffered writes on an abandoned wrapper would be lost. The merged version wraps a dup(2) of the fd at for_fd time instead: close and finalizer fclose the dup (flushing on the way), your descriptor never closes, and offset/flags stay shared through the one file description. The `io.autoclose = false` runtime spelling, which has no dup to retreat to, takes your shape -- with an fflush before abandoning.
- **The mode auto-detect**: we wrote the same F_GETFL derivation independently (the merged one also maps O_APPEND to "a").
- **Your codegen observation was the piece mine lacked**: the TY_STRING-only guard indeed drops a mode variable that widens to poly in a larger program. Adopted in the follow-up just pushed, one step further than the StringNode test -- a poly mode unboxes through sp_poly_to_s, so a mode VARIABLE (which is not a StringNode) rides through as well, and only a genuinely absent mode falls to the fd-derived default.

Credited in the follow-up's commit message.

Issue #4211, opened 2026-08-30
Queue#pop dispatched as NoMethodError on poly-typed ivar receiver


rofl0r:
## Summary

A `Queue` held in an instance variable (ivar) initialized with `Queue.new`
in `#initialize` reads as `poly` / `untyped` at every read site instead of
the expected `queue` / `Thread::Queue`. The codegen for Queue methods
gates on `comp_ntype(recv) == TY_QUEUE`, so a poly receiver falls
through to the generic call handler and raises

    undefined method 'pop' for an instance of Thread::Queue (NoMethodError)

A local variable with the same `Queue.new` keeps the static `queue` type
and dispatches to the dedicated `sp_Queue_*` runtime functions. The
fall-through is unique to the ivar path.

## Reproducer

`test/ivar_queue_inference_bug.rb`:
```ruby
    def local_queue
      q = Queue.new
      q.push("a")
      q.pop
    end

    print "ok: local\n" if local_queue == "a"

    class Message
      def initialize(content); @content = content; end
      def role; "user"; end
      def content; @content; end
    end

    class IvarQueue
      def initialize
        @q = Queue.new
      end

      def take
        @q.pop
      end

      def submit(msg)
        if msg.is_a?(Message)
          if msg.role == "steer"
            @q.push(["steer", msg.content || ""])
            return
          end
          @q.push(["user", msg.content || ""])
          return
        end
        @q.push(["user", msg])
      end
    end

    q2 = IvarQueue.new
    q2.submit(Message.new("b"))
    print "ok: ivar\n" if q2.take == ["user", "b"]
```

## Observed (AOT)

    ok: local
    test/ivar_queue_inference_bug.rb:52: undefined method 'pop' for an instance of Thread::Queue (NoMethodError)

## Expected (CRuby)

    ok: local
    ok: ivar

## Where the failure lives

The crash is in the codegen at `src/codegen_call.c` (the Queue arm,
gated on `comp_ntype(c, recv) == TY_QUEUE`). With the ivar widening to
poly, the gate fails and the call falls through to the generic
dispatch, which produces the runtime NoMethodError. The body of the
failing method (`sp_IvarQueue_take` in the generated C) reads:

    static sp_RbVal sp_IvarQueue_take(sp_IvarQueue *self) {
      return sp_poly_pop(self->iv_q);   /* poly path, not sp_Queue_pop */
    }

## Why the ivar is typed `poly` (not `queue`)

`--emit-types` on the reproducer:

[...]

PR #4214, opened 2026-08-30
io: codegen Queue#pop(true) via sp_Queue_pop_nb

rofl0r:
Queue#pop with a single argument is no_wait: CRuby raises ThreadError on an empty queue rather than blocking, and the typical pattern is `begin; q.pop(true); rescue ThreadError; break; end` to drain a worker queue. The codegen arm for Queue methods (src/codegen_call.c) handled argc == 0 with sp_Queue_pop and argc == 1 with sp_Queue_push but had no arm for argc == 1 pop, so the call fell through to the generic dispatch and produced

    unsupported call: node N (CallNode `pop`) recv=.../ty35 argc=1

Add a matching arm that emits sp_Queue_pop_nb, and a runtime helper in lib/sp_sched.c that mirrors sp_Queue_pop but raises ThreadError on an empty queue (returning nil for a closed-and-drained queue, the way sp_Queue_pop does).

Tests:
- test/queue_local_basic.rb: Queue#push / #pop / #size / #empty? on a local Queue, dispatched through the existing TY_QUEUE arm.
- test/queue_local_pop_nb.rb: Queue#pop(true) on a local Queue, covering the empty (ThreadError), single-pushed, and drained cases.
- test/queue_ivar_pop_nb.rb: the same drain-loop pattern on a Queue held in an ivar, the case that was previously falling through.

## Summary by CodeRabbit

* **New Features**
  * Added non-blocking support for `Queue#pop(true)`.
  * Empty open queues now raise `ThreadError`; closed empty queues return `nil`.
  * Available items are removed in FIFO order without waiting.
  * Queue operation arguments and receivers are evaluated exactly once, preserving expected blocking behavior.

* **Tests**
  * Added coverage for local and instance-variable-backed queues, including empty queues, populated queues, ordering, size, emptiness, and truthiness scenarios.


matz:
Merged in 41c530b8. Verified: the generated C across the 3,300-program suite is byte-identical to master's (the new arm fires only for the one-argument pop no existing program used), your three tests pass, and the drain-loop pattern on an ivar Queue works through the typed arm (#4211's fix keeps that slot TY_QUEUE, so your ivar test rides the same sp_Queue_* dispatch as the local).

One semantic correction landed as a follow-up (24bf5195), on the arm's own comment: `pop(false)` is CRuby's spelling of "block, explicitly", not a mistake for `pop` -- a consumer written `q.pop(false)` waits for its producer, and routing every one-argument pop to the non-blocking helper turned that wait into a ThreadError. A literal false/nil now picks the blocking pop and a literal true the non-blocking one at compile time; a non-literal argument decides at run time by its own truthiness, and the argument expression is evaluated either way so its side effects stay the caller's. The follow-up's test pairs each spelling with a late-pushing producer thread; your sp_Queue_pop_nb runtime helper is untouched and does exactly what it says.

rofl0r:
thanks for the quick merge @matz . i may be wrong but it appears the follow up commits addressing the AI review's findings didnt make it into master.

matz:
Merged, including your two follow-up commits (truthiness dispatch and the blocking-branch test). Our trees crossed again: I had pushed an equivalent truthiness fix in 24bf5195 before noticing your 130227a3, so the merge keeps that version, which is the same runtime ternary plus compile-time fast paths for literal true/false/nil arguments. Both of your tests are in and pass, as does a producer-paired pop(false) blocking test (test/queue_pop_false_blocks.rb). Thanks for catching the pop(false) semantics gap.

PR #4215, opened 2026-08-30
Hold a multiple assignment's values while the rest are built

blacktm:
## Why

`a, b = "x#{i}", f(i)` builds each value into a temporary before any target takes one — that is what makes `a, b = b, a` a swap — and nothing holds the first temporary while the second is built. The collector does not scan the C stack: a value is a root only while it is on the shadow stack, and a temporary has no holder once its own expression is done. So a collection the second value's build runs sweeps the first when it is a fresh object. Under a value that collects, 200 rounds of `a, b = "x#{i}", gcv(i)` left, in about a third of runs, 188 distinct Strings in `a` where 200 went in and a hash filled from the pairs whose entries read freed memory — the others crashed, or came out whole, by where the sweep fell; `a, b, c = "x#{i}", "y#{i}", gcv(i)` answered wrong on every round; and `h["k#{i}"], h["m#{i}"] = "x#{i}", gcv(i)` stored the freed first value under its key, so `h["k50"]` read back its own key's text, or a later round's value, from the memory. The test that does all of these segfaults on master, in the collector's mark, with wrong answers before it; under ASAN the probe behind it is a use-after-free in the set's key compare.

The index targets add an order of their own. Ruby evaluates each target's receiver and index before any value, left to right, then the values, then assigns: `r(1)[k(1)], r(2)[k(2)] = v(1), v(2)` runs r, k, r, k, v, v. The emitter ran the values first and built each receiver and index inside its store, so the calls ran in the wrong order. Evaluated before the values, as Ruby has it, a key that is a fresh object is held by nothing while the values build, so the new order needs the roots a single store's key has (#4203). And a hash index target never refused a frozen hash: `fz["b"], x = 2, 3` wrote into it, where a single store checks before its set and CRuby raises; an attribute target and an instance variable target wrote into a frozen object the same way, where the single writes raise.

## What

- A value of a multiple assignment that is a fresh object was swept while the values after it were built: 200 rounds of `a, b = "x#{i}", gcv(i)`, with `gcv` collecting the object heap and allocating past the string heap's own trigger, left, in about a third of runs, 188 distinct Strings in `a` and a hash filled from the pairs whose entries read freed memory, and in the others crashed or came out whole in about equal measure; the three-value form answered wrong on every round; `arr[gci(i) % 10], arr[(gci(i) + 1) % 10] = "x#{i}", gcv(i)` over an Array of 10 left 6 distinct where CRuby leaves 10; `h["k#{i}"], h["m#{i}"] = "x#{i}", gcv(i)` stored the freed first value, so `h["k50"]` read back its own key's text or a later round's value. The same across instance variable, global variable, attribute, splat and nested targets, and with the assignment as a value. Each is whole now; the test that exercises them all segfaults on master partway through, with wrong answers before it.
- The targets' receivers and indexes are evaluated before the values, left to right, as Ruby evaluates them: `r(1)[k(1)], r(2)[k(2)] = v(1), v(2)` printed the values first, and `o(1).x, o(2).y = v(5), v(6)` likewise.
- `fz["b"], x = 2, 3` on a frozen hash wrote into it and answered as if nothing were wrong; it raises `FrozenError` after the values are built, with `x` untouched, as CRuby does. `o.x, x = "p", 9` on a frozen object, and `@x, @y = ...` in a method on a frozen `self`, wrote too, where `o.x = "p"` and `@x = "p"` raise; they raise as the single writes do.

## How

- **The value temps** (the tuple path of `MultiWriteNode` in `codegen_stmt.c`): a temp is rooted, through `emit_gc_root_tmp`, while something after it can collect — a later value that allocates, a store that runs user code (the general hash's key hooks, a receiver typed at run time), or a store that allocates (the rest array, a by-value struct such as a Range boxed into a slot that is not a local or an instance variable of its own type). Nothing holds the temp but its slot, and whatever can allocate can run user code that drops its other holder: a call, or the `to_s` an interpolation runs, which no side-effect predicate sees, so a temp that is a plain read of a global is rooted too — `a, b = $g, "wrap#{$kk}"` with a `to_s` that clears `$g`. A literal that is rodata is never rooted; a Bignum literal is built, so it is. `a, b = b, a`, `a, b = 1, 2` and every other assignment whose values are reads, literals or arithmetic, with no such store after them, emit as they did.
- **The targets' parts**: once any target's receiver or index is a call, a literal that allocates, a write, or a read of a variable a target before it assigns (`i, a[i] = 1, 2`), every target's receiver and index — the plain ones too, so the order among them holds — is evaluated into a temp before the values, in Ruby's order, and the stores read the temps. The temps go into the statement's prelude, ahead of what a value hoists there (a literal's build), each part emitted whole before its declaration is written and the index after the receiver's line, so what a part hoists lands before its own declaration and after the receiver — `r(1)[[k(1)]], x = v(1), v(2)` runs r, k, v, v. The receiver's temp and the index's are rooted outright: a part that is a read is held by its variable only until a later part or a value runs user code — `h1[$k], h2["x#{$kk}#{i}"] = 1, 2` with a `to_s` that clears `$k` — and the hoist happens only where some part is a call or a build, so the roots land where a single store's would (#4203). When every receiver and index is a plain read, a literal, or arithmetic over them, they all stay at their stores, so `a[i], a[j] = a[j], a[i]` over Integers and `a[i % n]` are emitted as they were. `subtree_may_allocate` counts a builtin operator over scalars as a call; `masgn_part_allocates` reads one back as arithmetic, since `i % n` builds nothing. And `subtree_may_allocate` gains the interpolated Symbol, `:"k#{i}"`, which builds a String to intern, the interpolated Regexp, the lambda literal, `-> { }`, which allocates its Proc (`lambda { }` is a call and counted already), and the last-match reads, of which `$~`, `` $` `` and `$'` build a MatchData or a String. A `[]` read is a call to it, so the same swap over Strings gains one root, on the first value while the second is read.
- **The frozen checks**: the hash arm emits the check a single store makes, `sp_gc_is_frozen` before the set, at the store; the attribute and instance variable arms emit `emit_frozen_obj_guard`, the single write's guard, which is nothing on a class whose instances are never frozen.
- The other paths of the emitter — a call, a splat or a typed array on the right, destructured at run time — are untouched. `subtree_may_allocate` is shared with the single store's and the argument roots' sites; one program, `issue_3146`, changes through them, as above.

The generated C changes for 10 of the 3,369 programs of master's suite, the benchmarks and optcarrot: 7 tests gain a root or the frozen check, one (`poly_multi_write_index_target`) has its attr-read receivers hoisted and rooted, with a root on its first value, one (`issue_3146`) has a single store's interpolated-Symbol key, `big[:"k#{i}"] = i * 3`, take #4203's hoisted form now that the key counts as allocating — the Symbol is interned, so nothing there is rooted — and optcarrot gains one root, in `APU#initialize`, on the first of `@pulse_0, @pulse_1 = Pulse.new(self), Pulse.new(self)` while the second is built — construction time, once. Every benchmark is byte-identical.

## Validation

- A new test, `test/multi_assign_operand_gc_root` (expectations generated by CRuby 4.0), a segfault on master partway through: two and three values into locals, instance and global variables, with a value that collects; index targets with a built key on String-keyed hashes with Integer, String and poly values, Integer- and Symbol-keyed hashes, the general hash, an Array and a container typed at run time, with a call as its index too; attribute targets whose receiver is a call, evaluated once; a splat and a nested target; a value that drops the only other holder of an earlier one; the assignment as a value; the order of the targets' receivers and keys against the values, a value built in the statement's prelude among them, and a target assigned before a later target's plain index, a nested target among them; the assignment's own stores collecting, through a hash hook, a key a later target's index cleared, a receiver an earlier target dropped, and a value held by a global alone; the `to_s` an interpolation runs clearing the holder of a hoisted receiver, a hoisted key and a value; a Bignum literal, which is built, an interpolated Symbol, a lambda and the last match, which allocate, a store that boxes a Range into a poly slot and one that boxes a `Process.times`, and the receiver before an index built in the prelude; a swap through plain indexes; and a frozen hash, and a frozen object under an attribute and an instance variable target, refused after the values are built.
- Full suite: 3,346 pass, 0 fail, from a cleared results cache, at `-O2` and at the gate's `-O1`.
- The new test runs clean under `SPINEL_GC_STRESS=1` and ASAN+UBSAN with CRuby's output; so do the probes behind each number above.
- Generated C against master: 10 of 3,369 programs differ, as listed under How; the benchmarks are byte-identical, optcarrot differs by the one root.
- Against the differential corpus of ~2,100 minimized programs the campaign has been working from: 712 → 712 match, 0 gained and 0 lost.
- The gate's other legs: 61 benchmarks pass, optcarrot's checksum is 59662, and its frame rate is master's within the run-to-run spread — 1,847 fps against 1,845, each the median of five with a spread under 2%.

## Left alone, on purpose

- A receiver or index that is a plain read is still read at its store, after the values, so a value that reassigns it — `x[0], x[1] = (x = other; 1), 2` — stores into the new container where Ruby stores into the old, and `h[$1], x = (s =~ /(y+)/), 2` stores under the value's own capture; the hoist is for what a plain read cannot be, a call or a build, as #4203's is.
- The values are built in order, but a value that hoists a build into the statement's prelude — an Array literal, an interpolated Regexp — runs before an earlier value that does not — `a, b = v(1), [v(2), v(3)]` runs v(2) and v(3) first — on master and here alike; the targets' parts now come before both.
- A Range index into an Array target, `a[1..2], x = 9, 1`, is refused at compile time, as `unsupported`, where the C compiler refused it; a Range that arrives through a poly slot, `a[q[0]], x = 9, 1`, is coerced as an index and raises a `TypeError` at run time, where it did not compile; the single store answers CRuby's slice. A two-argument index, `a[k(1), k(2)], x = 9, 1`, drops its second argument without a word, as before. Both want the slice store, not a root.
- The run-time-destructure paths — `a, b = f()`, `a, *b = arr` — evaluate an attribute target's receiver at its store as before; their values sit in the rooted array, so there is nothing to sweep, and the order is observable only when the receiver is a call.
- A container typed at run time that is a Hash drops a multiple-assignment store — `sp_poly_arr_set` on a Hash box stores nothing, where the single store's path finds the hash — on master and here alike; the test's run-time-typed section uses two Array kinds.
- A Symbol-keyed hash target whose key is a poly value (`h[syms[i]], x = ...`) fails to compile, on master and here alike: the masgn arm passes the key as it is, where a single store coerces it through `emit_hash_key`.
- Index targets after a splat (`a, *r, h[k] = ...`) are skipped without a word, as before; that arm has no index case.
- A namespaced constant assigned before a later target's index, `M::X, a[M::X] = 2, 9`, is not among the targets the plain gate consults — a plain `X, a[X] = 2, 9` is — so the index reads the new value, on master and here alike.
- The general hash's key hooks running over a value that is a temporary is #4201's; here such a store counts as one that can collect, so the value temps before it are rooted.
- A root pushed for a temp stays on the shadow stack to the end of the enclosing C scope, as the rest array's does, so a recursive method with such an assignment holds more roots per frame: one more for `a, b = "s#{n}", "t#{n}"`, where the shadow stack's 65,536 slots hold about 21,800 frames instead of about 32,800; four more for `$h[k(n)], b = "v#{n}", "w#{n}"` on a general hash, whose two parts and two values are all rooted, about 13,100 frames. The stack overflows silently, into a dropped root. Bracing the assignment would give that back, at a brace around every one that roots.
- A value that is a rooted temp of its own prelude — an Array literal — is rooted again by its temp; a bare `_tN` in a prelude is not always rooted, so the second root stays.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multiple-assignment behavior to preserve Ruby’s left-to-right evaluation order.
  * Prevented values, receivers, and indexes from being collected during assignments.
  * Added safeguards for frozen objects and hashes, and rejected invalid range-based array indexes.
  * Improved handling of additional allocating expressions, including interpolations, lambdas, regular expressions, and match references.
* **Tests**
  * Added comprehensive coverage for garbage collection, evaluation order, nested assignments, swaps, indexed targets, and frozen-object errors.

### DIFF
--- a/src/codegen_stmt.c
+++ b/src/codegen_stmt.c
@@ -6291,6 +6291,145 @@ void emit_sg_activate(Compiler *c, int node, int recv, Buf *b, int indent) {
   buf_printf(b, ")->cls_id = %d;\n", ci);
 }
 
+/* The receiver and the index of a multiple-assignment target that has them
+   (`a[i]`, `o.x`); -1 for a target that is a slot. */
+static void masgn_target_parts(const NodeTable *nt, int t, int *recv, int *key) {
+  const char *ty = nt_type(nt, t);
+  *recv = *key = -1;
+  if (!ty) return;
+  if (sp_streq(ty, "IndexTargetNode")) {
+    *recv = nt_ref(nt, t, "receiver");
+    int args = nt_ref(nt, t, "arguments"), n = 0;
+    const int *av = args >= 0 ? nt_arr(nt, args, "arguments", &n) : NULL;
+    if (n >= 1) *key = av[0];
+  }
+  else if (sp_streq(ty, "CallTargetNode")) *recv = nt_ref(nt, t, "receiver");
+}
+/* Whether evaluating a part or a value can allocate, and so collect. A builtin
+   operator over scalars is a CallNode to subtree_may_allocate, but `a[i % n]`
+   builds nothing; what has no effect and answers a scalar is arithmetic. */
+static int masgn_part_allocates(Compiler *c, int id) {
+  if (!subtree_may_allocate(c->nt, id)) return 0;
+  if (subtree_has_side_effect(c, id)) return 1;
+  TyKind t = comp_ntype(c, id);
+  return !(t == TY_INT || t == TY_FLOAT || t == TY_BOOL);
+}
+/* Whether the subtree reads the variable a target of the given kind writes. */
+static int masgn_reads(const NodeTable *nt, int id, const char *rty, const char *name) {
+  if (id < 0) return 0;
+  const char *ty = nt_type(nt, id);
+  if (!ty) return 0;
+  if (sp_streq(ty, rty) && nt_str(nt, id, "name") && sp_streq(nt_str(nt, id, "name"), name)) return 1;
+  int nr = nt_num_refs(nt, id);
+  for (int i = 0; i < nr; i++)
+    if (masgn_reads(nt, nt_ref_at(nt, id, i), rty, name)) return 1;
+  int na = nt_num_arrs(nt, id);
+  for (int i = 0; i < na; i++) {
+    int n = 0;
+    const int *ids = nt_arr_at(nt, id, i, &n);
+    for (int j = 0; j < n; j++)
+      if (masgn_reads(nt, ids[j], rty, name)) return 1;
+  }
+  return 0;
+}
+/* Whether the target, or one nested in it, writes a variable the subtree
+   reads. */
+static int masgn_target_writes(const NodeTable *nt, int t, int id) {
+  static const char *const VARS[][2] = {
+    { "LocalVariableTargetNode", "LocalVariableReadNode" },
+    { "InstanceVariableTargetNode", "InstanceVariableReadNode" },
+    { "GlobalVariableTargetNode", "GlobalVariableReadNode" },
+    { "ClassVariableTargetNode", "ClassVariableReadNode" },
+    { "ConstantTargetNode", "ConstantReadNode" } };
+  const char *ty = nt_type(nt, t), *nm = nt_str(nt, t, "name");
+  if (!ty) return 0;
+  if (sp_streq(ty, "MultiTargetNode")) {
+    static const char *const SIDES[] = { "lefts", "rights" };
+    for (size_t s = 0; s < sizeof SIDES / sizeof SIDES[0]; s++) {
+      int n = 0;
+      const int *ts = nt_arr(nt, t, SIDES[s], &n);
+      for (int i = 0; i < n; i++) if (masgn_target_writes(nt, ts[i], id)) return 1;
+    }
+    return 0;
+  }
+  if (!nm) return 0;
+  for (size_t v = 0; v < sizeof VARS / sizeof VARS[0]; v++)
+    if (sp_streq(ty, VARS[v][0])) return masgn_reads(nt, id, VARS[v][1], nm);
+  return 0;
+}
+/* A part that answers the same whenever it runs: a read, a literal, or
+   arithmetic over them -- unless a target assigned before its store writes
+   what it reads (`i, a[i] = 1, 2`). */
+static int masgn_part_plain(Compiler *c, int id, const int *lefts, int before) {
+  if (masgn_part_allocates(c, id) || subtree_has_side_effect(c, id)) return 0;
+  for (int i = 0; i < before; i++) if (masgn_target_writes(c->nt, lefts[i], id)) return 0;
+  return 1;
+}
+/* A literal that is not built -- rodata, which no sweep touches; a Bignum
+   is built. */
+static int masgn_rodata(Compiler *c, int id) {
+  return node_is_pure_literal(c->nt, id) && !subtree_may_allocate(c->nt, id) && comp_ntype(c, id) != TY_BIGINT;
+}
+/* The root push for a temp, after the `;` of the statement that set it, when
+   `emit_gc_root_tmp` has one for the type. */
+static void masgn_root(Compiler *c, TyKind t, int tmp, Buf *b) {
+  Buf rb; memset(&rb, 0, sizeof rb);
+  emit_gc_root_tmp(c, t, tmp, &rb);
+  if (rb.p && rb.p[0]) buf_printf(b, " %s", rb.p);
+  free(rb.p);
+}
+/* Whether the slot value `vi` lands in boxes a by-value struct of type `vt`
+   on the heap: a local or an instance variable of that type holds it as it
+   is; any other slot may box it. */
+static int masgn_slot_boxes(Compiler *c, int id, const int *lefts, int ln, int vi, TyKind vt) {
+  if (vi >= ln) return 1;
+  const char *ty = nt_type(c->nt, lefts[vi]), *nm = nt_str(c->nt, lefts[vi], "name");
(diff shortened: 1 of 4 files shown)


matz:
Merged in 83d35f81. Verified before merging: 9 suite programs differ from master's generated C, every one a multiple assignment gaining its temp roots or the Ruby evaluation order; optcarrot differs by exactly one line -- the APU's constructor-pair assignment gains one SP_GC_ROOT, at startup, nothing per-frame. The plain-shape byte-identity your part-analysis buys is what makes this cheap to take, and the target-order fix (receivers and indexes left to right before any value, frozen checks where a single store makes them) is the part reviewers of masgn code will be gratefulest for.

Your CI was red on all three platforms, and the cause is worth knowing for future tests: the suite runs each test under a ten-second timeout, and this one -- a GC.start plus a 1.2MB allocation per round, two hundred rounds, fifteen sections -- took 7.6s on a fast machine and past the limit on every runner, so all three legs diffed an empty output. The follow-up (7e5fa8a1) trims it to sixty rounds with 450KB padding: still failing on pre-merge master, still clean under SPINEL_GC_STRESS=1, and one second on the clock. (Two hash-inspect lines in the expectations are restored to the `{1 => 1}` spacing your interpreter and spinel share; my local 3.2 prints the old one, which is how they briefly regressed.)

This was also the gap #4203's "left alone" list named, so the store family is now rooted end to end.

PR #4220, opened 2026-08-30
Give HMAC its raw spelling, the one another primitive reads

rubys:
`Digest::SHA256` has had both `hexdigest` and `digest` since the file was written. `HMAC` had only the hex one, which is the half a human or a protocol header reads.

The raw half is what another primitive reads. HKDF (RFC 5869) is HMAC over raw bytes twice, so a caller with only hex has to decode between the two rounds — a lossless detour with a chance to get the decode wrong. Same for anything comparing a MAC against bytes off a wire: a webhook signature, a JWT's HS256 half, an AWS SigV4 chain.

## What's here

`sp_crypto_hmac_sha256_bin` / `_sha1_bin` are the twins of the `sha256_bin` / `sha1_bin` beside them, and follow the same contract: the result is not NUL-terminated, its length rides `sp_ffi_bin_len`, and each gets its own `SP_TLS` buffer so a caller that hexes one MAC and takes the bytes of another doesn't have the second clobber the first.

`OpenSSL::HMAC.digest` answers exactly what `hexdigest` answers, for the same two algorithms, and refuses the rest the same way (`DigestError`, so `rescue OpenSSL::OpenSSLError` still catches it).

The length mattering rather than a terminator is the whole point of the binary return: a MAC is uniform random bytes, so roughly one in eight contains a NUL that a NUL-terminated return would truncate at.

## Tests

RFC 4231 test case 1 for the vector, plus the property that keeps the two spellings honest — `hex(digest) == hexdigest` for both algorithms — at both the `sp_crypto` level (`test/sp_crypto_basic.rb`) and the Ruby surface (`packages/openssl/test/openssl_digest.rb`).

`make test`: **3357 pass, 0 fail, 0 error.**

## One thing worth knowing about the suite on macOS

The `OPENSSL_AVAILABLE` probe uses a bare `$(CC)`, which doesn't search `/opt/homebrew/include`, so on a Homebrew Mac `packages/openssl/test/*` is silently filtered out of `make test` — a green run there has not exercised that package. I ran it with `CPATH=/opt/homebrew/include LIBRARY_PATH=/opt/homebrew/lib` so the openssl-side expected file is a real run rather than a hand-written guess. Happy to send a probe tweak separately if you'd want one; it's orthogonal to this change.

## Why now

This is the smallest slice of a larger question I've opened as an issue alongside it — the `PKey` / `Cipher` half of the package, where the design call is yours. This piece stands on its own regardless of how that goes.

## Summary by CodeRabbit

- **New Features**
  - Added binary HMAC digest support for SHA-256 and SHA-1.
  - Added `OpenSSL::HMAC.digest` for retrieving raw digest bytes.
  - Unsupported algorithms now report a consistent digest error.

- **Tests**
  - Added coverage confirming digest sizes, byte-for-byte accuracy, embedded NUL handling, and unsupported algorithm errors.

### DIFF
--- a/lib/sp_crypto.c
+++ b/lib/sp_crypto.c
@@ -372,6 +372,19 @@ const char *sp_crypto_hmac_sha256_hex(const char *key, const char *msg) {SP_GC_R
     return sp_crypto_hmac_hex_buf;
 }
 
+/* The same MAC as its own bytes rather than as hex. Its own buffer, per
+ * the file header's rule: a caller that hexes one MAC and takes the raw
+ * bytes of another must not have the second clobber the first. */
+static SP_TLS char sp_crypto_hmac_sha256_bin_buf[32];
+
+const char *sp_crypto_hmac_sha256_bin(const char *key, const char *msg) {SP_GC_ROOT_STR(key);SP_GC_ROOT_STR(msg);
+    sp_crypto_hmac_sha256((const uint8_t *)key, sp_str_byte_len(key),
+                          (const uint8_t *)msg, sp_str_byte_len(msg),
+                          (uint8_t *)sp_crypto_hmac_sha256_bin_buf);
+    sp_ffi_bin_len = 32;
+    return sp_crypto_hmac_sha256_bin_buf;
+}
+
 /* HMAC-SHA1 (RFC 2104). SHA-1 is here for the same reason
  * sp_crypto_websocket_accept is: an existing protocol names it, and
  * reproducing that protocol is not a new security design. Rails signs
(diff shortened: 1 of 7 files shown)


matz:
Merged as-is. Reviewed against the CRuby oracle (RFC 4231 case 1 vector and hex(digest) == hexdigest for both algorithms match CRuby's OpenSSL::HMAC), ran both test files plus the full gate locally, and the new sp_crypto test also passes under SPINEL_GC_STRESS=1. The per-function SP_TLS buffer and the sp_ffi_bin_len contract follow the file's existing rules exactly, and the change is purely additive. Thanks.

On the macOS OPENSSL_AVAILABLE probe not searching /opt/homebrew/include: yes, please send that tweak separately, it is worth having the openssl package tests actually run on a Homebrew Mac.

PR #4223, opened 2026-08-31
String#split(nil, limit): pass NULL, RT does whitespace

rofl0r:
The poly-receiver String#split codegen called emit_str_expr on the separator arg, which inserts a 'raise on nil' guard. CRuby's `s.split(nil, n)` is the documented 'split on whitespace' form; sp_str_split_limit at lib/sp_str.c:823 explicitly handles !sep by dispatching to sp_str_split_ws_limit. The guard was firing before the runtime's branch was reached, so every s.split(nil, n) call crashed with

  no implicit conversion of nil into String (TypeError)

in the AOT binary, even though the runtime would have answered correctly.

Detect a nil-typed separator in the poly-receiver split arm and emit NULL directly, the same way the argc==1 arm already special- cases the 'split(nil)' / 'split(" ")' literal pair as a whitespace mode. The runtime does the rest.

The String-receiver split (TY_STRING dispatch at codegen_call_recv.c:6950) and the multi-arg split at the same place keep their existing paths -- the issue only fires through the poly arm, where emit_str_expr runs unconditionally on the separator.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `String#split` so calls with a nil separator correctly use whitespace-based splitting.
  * Preserved existing behavior when a non-nil separator is provided.
  * Ensured separator expressions are evaluated correctly before splitting.


matz:
Merged (e327cc38), with a follow-up in bf4163d9 that finishes the family your fix opened:

- The one-argument poly arm (`v.split(nil)`) had the same raise; your test's receiver was a String literal, which never takes the poly arm, so it stayed green on master without the fix. The regression test now drives a genuinely poly receiver through argc 0/1/2, a side-effecting separator expression, and counts the evaluation.
- A separator that is nil only at run time (`sep = cond ? "," : nil` in a poly slot) was worse than the raise: the loose string conversion rendered nil as "" and silently turned the call into a character split. Both receiver paths now route the separator through the split-only slot (emit_str_pattern_expr), which passes a runtime nil as NULL via a new sp_poly_arg_str_or_null and keeps the strict TypeError for a wrong class.

Your diagnosis of the guard-before-runtime-branch ordering was exactly right, and the side-effect-preserving shape you chose is the one the consolidated slot kept. Full gate green, CI green. Thanks.

PR #4226, opened 2026-08-31
codegen: route IO#write through sp_File_write on a poly receiver

rofl0r:
A user class that defines #write forces the per-class poly dispatch into a switch on cls_id. The runtime already maps SP_BUILTIN_IO to sp_File_write (which itself dispatches to sp_sock_write for sockets via the is_sock flag), but the codegen had no arm for that tag, so any Socket or File receiver fell through to the raise default.

Add the SP_BUILTIN_IO arm to the switch, mirroring the existing SP_BUILTIN_* arms for length/push/join/include on arrays. The arg is emitted as a String when its analyse-time type is TY_STRING, otherwise unboxed via sp_poly_to_s so a poly arg writes its to_s.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `IO#write` calls through polymorphic receivers so they work correctly with both files and sockets.
  - Ensured side-effecting arguments are evaluated only once.
  - Preserved embedded NUL bytes in strings and correctly handled non-string values.
  - Socket writes now return the correct number of bytes written.

- **Tests**
  - Added coverage confirming successful socket writes and expected output.


matz:
Merged. Verified the test raises NoMethodError on pre-merge master and passes after, including under SPINEL_GC_STRESS=1; reusing the materialised arg temp so side effects fire once matches how the sibling builtin arms handle it, and the byte-length write for a String arg keeps embedded NULs. Full gate green on Linux, and your branch's macOS run was green before I pushed. Thanks.

rofl0r:
the review found some new issues, they should be addressed in 740a6bc9

Issue #4221, opened 2026-08-30
The crypto half of packages/openssl: PKey::EC, ECDH, ECDSA and AES-GCM — which shape?


rubys:
`packages/openssl` says what it is in its own header, and it is honest about it:

> `OpenSSL::Digest`, `Cipher`, `PKey` and most of X509 are not here, and a program that uses them fails at COMPILE time with an unresolved call rather than at run time somewhere deep.

I'd like to help close part of that, and I want to ask about the shape
before writing any of it, because the answer decides most of the work.

## What the gap costs today

Web Push (RFC 8291 payload encryption + RFC 8292 VAPID) is a useful
yardstick, not because it is special but because it is entirely
standard, widely implemented, and needs nothing exotic. Taking it apart
against what spinel has now:

| step | primitive | today |
|---|---|---|
| ephemeral keypair | P-256 keygen | missing |
| shared secret | ECDH | missing |
| derive CEK + nonce | HKDF-SHA256 | HMAC exists; #4220 adds the raw spelling |
| encrypt payload | AES-128-GCM | missing |
| VAPID JWT | ES256 (ECDSA P-256 + SHA-256) | missing |
| base64url everywhere | | `sp_crypto` has it |
| deliver | HTTPS POST | `net/http` + TLS, since 6a7107d6 |

Four of seven steps are already there, and the two ends of the pipeline
— the encoding and the transport — are the parts I would have expected
to be the hard ones. The same four missing primitives also cover ES256
JWTs on their own (extremely common: Apple, Google, and most modern JWT
issuers), ECDH key agreement, and AES-GCM for anything at rest.

## What makes it cheaper than it looks

`libcrypto` is already on the link line — `OpenSSL::Native` declares
`ffi_lib "crypto"` beside `ffi_lib "ssl"`, and `spin.toml` notes that
both reach the link line only through a `require "openssl"`. So this is
Ruby surface plus C glue over a library the package already links, and
a program that does not require openssl still links exactly what it did
before. The `OPENSSL_AVAILABLE` probe already gates the whole object.

## The question: which shape?

The package header says the CRuby spelling is deliberate — "the goal is
that a program written against CRuby compiles here, so inventing a
shorter name would be worth nothing." Taken literally for these
primitives, that is a fair amount of surface:

```ruby
key = OpenSSL::PKey::EC.generate("prime256v1")
key.dh_compute_key(OpenSSL::PKey::EC::Point.new(group, OpenSSL::BN.new(peer, 2)))

[...]

PR #4228, opened 2026-08-31
Give the openssl package EC keys, ECDH and HKDF

rubys:
The shape you settled in #4221: one-shot over raw bytes, CRuby's names where CRuby has a spelling.

```ruby
key  = OpenSSL::PKey::EC.generate("prime256v1")
key.private_key_bytes            # 32 raw
key.public_key_bytes             # 65 raw, X9.62 uncompressed
key.dh_compute_key(peer_bytes)   # CRuby's name, bytes not a Point
OpenSSL::PKey::EC.from_private_bytes("prime256v1", bytes)

OpenSSL::KDF.hkdf(ikm, salt:, info:, length:, hash:)   # CRuby's name and keywords
```

No `BN`, no `EC::Group`, no `EC::Point`; a program that names one still fails at compile time the way the package's other gaps do. `openssl/pkey.rb`'s header says that out loud, and why.

The C glue is in `packages/openssl/sp_openssl.c`, behind `OPENSSL_AVAILABLE` with the rest. Nothing reaches `lib/`, and a program that does not require openssl links exactly what it did before.

## Two decisions worth calling out

**No handle table.** A key is fully described by its bytes, so nothing on the C side outlives a call — no slot to release, no table to exhaust. That is the deliberate contrast with the connection table above it in the same file, and a key abandoned mid-protocol costs nothing.

**`EC_GROUP` / `EC_POINT` / `BIGNUM`, not `EC_KEY` / `ECDH_compute_key`.** Those two are deprecated in OpenSSL 3.0, and their replacements (`EVP_PKEY_get_bn_param` and friends) did not exist at this file's 1.1.0 floor. The low-level three are current in 3.x and present in every version the file compiles against. Points cross the boundary as `0x04 || X || Y` — what every protocol naming raw EC keys means by "the public key" — which also makes ECDH a slice: the shared secret is the X coordinate.

## Three refusals that are load-bearing

- **A peer point off the curve is rejected before any multiply.** An ECDH against an attacker-chosen off-curve point leaks the private scalar a few bits at a time, and that check is the whole defence. Small-subgroup confinement needs no separate check on these prime curves, whose cofactor is 1.
- **A scalar outside [1, n-1] is rejected** rather than multiplied into a point that is not the caller's key: zero gives the point at infinity, and `d >= n` is a silent alias for `d - n`.
- **A scalar of the wrong width** is not a key on this curve.

## HKDF has no C

It is HMAC over raw bytes twice, and #4220 gave HMAC its raw spelling. Being Ruby also keeps it outside the `OPENSSL_AVAILABLE` gate, the same way `OpenSSL::Digest` is.

## Tests

RFC 8291 §5 and Appendix A — a complete Web Push example with fixed keys, so every value is published rather than computed by the code under test. The public halves derive to the RFC's, ECDH agrees in both directions on the RFC's `ecdh_secret`, and the RFC's IKM, CEK and nonce come out of three HKDF calls. RFC 5869 test case 1 pins HKDF by its own vector. A generated pair agrees with itself, and the byte lengths follow the curve rather than a constant — P-384 answers 48 and 97.

`make test`: **3365 pass, 0 fail, 0 error**, with the openssl package tests running. The new test also passes under `SPINEL_GC_STRESS=1`.

The byte comparisons in that test need #4224, now merged — before it they were all false, because a `:cbinstr` return was tagged as text and an EC public key always has a byte >= 0x80. Writing this surface is what turned that up.

## Not here

**ES256.** Same namespace, but it brings the DER to raw `r || s` conversion with it, which you asked to live in the C glue, and it reviews better on its own.

**AES-GCM.** The other half you named. I have one question about `Cipher`'s state across the FFI to put on #4221 before writing it — `SSLSocket` gets away with an int handle into a fixed table because `close` is in its protocol, and `Cipher` has no `close` in CRuby's.

## Summary by CodeRabbit

* **New Features**
  * Added support for elliptic-curve key generation, public-key handling, and Diffie-Hellman shared secrets.
  * Added HKDF key derivation with SHA-1 and SHA-256 support, including validation for invalid parameters.
  * Added access to EC key and curve byte representations.

* **Bug Fixes**
  * Added validation for invalid curves, key sizes, scalar values, and peer keys.

* **Tests**
  * Added coverage for RFC-based EC, ECDH, and HKDF scenarios and error handling.

### DIFF
--- a/packages/openssl/openssl.rb
+++ b/packages/openssl/openssl.rb
@@ -1,14 +1,18 @@
-# Spinel bundled `openssl` -- the SSL half only.
+# Spinel bundled `openssl`.
 #
 # The spelling is CRuby's, deliberately: the goal is that a program written
 # against CRuby compiles here, so inventing a shorter name would be worth
 # nothing. What is missing is missing the way a subset is missing things --
-# OpenSSL::Digest, Cipher, PKey and most of X509 are not here, and a program
-# that uses them fails at COMPILE time with an unresolved call rather than at
-# run time somewhere deep.
+# OpenSSL::Cipher and most of X509 are not here, and a program that uses them
+# fails at COMPILE time with an unresolved call rather than at run time
+# somewhere deep. Digest, HMAC, KDF and PKey::EC are here; EC is a subset of
+# CRuby's in a second sense, spelled out in openssl/pkey.rb -- it keeps
+# CRuby's method names but trades BN, EC::Group and EC::Point for the raw
+# bytes every protocol that names an EC key means by them.
 #
-# Spinel implements no TLS. sp_openssl.c is glue over the system libssl, and
-# the trust anchors are the operating system's; nothing is bundled.
+# Spinel implements neither TLS nor any of the arithmetic below it.
+# sp_openssl.c is glue over the system libssl and libcrypto, and the trust
+# anchors are the operating system's; nothing is bundled.
 #
 # SSLSocket is not an IO, exactly as in CRuby: it is an ordinary object that
 # answers #to_io, which is why IO.select had to learn that protocol. The
@@ -16,6 +20,8 @@
 # is handed to a garbage-collected world.
 require "openssl/buffering"
 require "openssl/digest"
+require "openssl/kdf"
+require "openssl/pkey"
 
 module OpenSSL
   module Native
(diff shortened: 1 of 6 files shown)


matz:
Merged. This is exactly the #4221 shape, and the review found nothing to change: the glue stays package-side behind OPENSSL_AVAILABLE with nothing reaching lib/, the no-handle-table decision is right for a value that is fully its bytes, and the three refusals (off-curve peer before any multiply, scalar range, scalar width) are the load-bearing kind. I cross-checked the RFC 8291 IKM derivation against CRuby's own OpenSSL::KDF.hkdf independently of your vectors, ran the test under SPINEL_GC_STRESS=1, and the full gate is green on Linux with your branch's macOS run green before the push. The HMAC-key observation that makes salt: "" and the RFC's zero-salt default the same PRK is a nice property to have left implicit in code and explicit in the test.

On the Cipher question you are about to put on #4221: worth asking it there as planned, matz decided Cipher stays CRuby-shaped, so the state question is real. My first reaction: a Cipher is small fixed state (key, iv, aad, mode) that only matters between calls in one thread; holding it Ruby-side as plain ivars and making final/auth_tag the one C call that sees everything at once would keep the no-handle property you just used for EC. But put it on the issue so the trade is on the record.

PR #4230, opened 2026-08-31
Reach every typed receiver with the builtin arity guard

blacktm:
## Why

A builtin called with a count CRuby rejects is a rescuable `ArgumentError` there, message and all. The guard that raises it here (33ae1e98's, widened in 505009f1) runs only on the receiver kinds its map names — String, Integer, Float, Symbol, Range, Time, nil, the booleans, Rational, Complex, Mutex, the arrays and hashes — and on a native package object or a user object. Everything else the compiler types goes past it, and lands one of four ways: `/a/.source(1)`, `m.begin(0, 1)`, `q.size(1)`, `pr.arity(1)`, `Foo.superclass(1)` and `GC.start(1)` refuse to compile as unsupported calls; `Thread.list(1)`, `File.file?` and `1.respond_to?` lower to a NoMethodError; `e.message(1)`, `$stdout.fileno(1)`, `f.alive?(nil)` and `throw :t, 1, 2` drop the argument and answer; `Foo.name(1)` and `Mod.name(1)` answer nil, a value the in-range call never gives. All four report a supported method as something it is not, the first stops the build, and the last is quietly wrong.

The rows come from CRuby itself — the generator probes each method's accepted counts and takes CRuby's own wording for the range — so reaching a receiver kind is a matter of probing its class and mapping the kind to the rows. This does that for Regexp, MatchData, Proc, Method, Fiber, Thread, Queue, ConditionVariable, IO, Dir, Exception, Enumerator, Random, Struct and Data instances, for a class constant — a user class's own Class surface, a Struct or Data class's readers, `GC`, `Fiber`, `Thread` — for a bare Kernel function, and for the universal names every builtin inherits from Object, which only a user object consulted before.

## What

- A wrong count on a Regexp, MatchData, Proc, Method, Queue, ConditionVariable, Dir or Enumerator receiver, on a Struct or Data instance, on a class or module constant or on a bare Kernel function refused to compile — `spinel: unsupported call: node 4 (CallNode \`source\`) recv=RegularExpressionNode/ty17 argc=1 arg0ty3` for `/a/.source(1)`, and the like for `/a/.match`, `q.size(1)`, `pr.arity(1)`, `mt.name(1)`, `cv.signal(1)`, `en.next(1)`, `S.members(nil)`, `Foo.superclass(1)`, `GC.start(1)`, `Process.times(nil)`, `Integer()`; `unsupported MatchData method` for `m.begin(0, 1)` and `m.length(1)`. Each is CRuby's `ArgumentError: wrong number of arguments (given 1, expected 0)` now, with the range CRuby names — `(given 0, expected 1..2)` for `Integer()`, `(given 3, expected 1..2)` for `throw :t, 1, 2`.
- A wrong count on a Thread or Fiber, on a class constant's own surface, or on a name every object answers lowered to a NoMethodError at run time: `Thread.list(1)` was `undefined method 'list' for class Thread`, `Fiber.current(nil)` the same for Fiber, `File.file?` with no path the same for File, `s.members(1)` on a Struct instance `undefined method 'members' for an instance of S`, `1.respond_to?` with no name `undefined method 'respond_to?' for an instance of Integer`, and `1.public_send(nm, nil)` with `nm = :itself` — the candidate a dynamic send synthesizes for a name it is handed — `undefined method 'itself'`.
- A wrong count on an Exception, an IO, a Fiber or a Kernel function dropped the argument and answered: `e.message(1)` printed the message, `$stdin.eof?(1)` answered true, `f.alive?(nil)` true for an unstarted fiber, `throw :t, 1, 2` threw 1, `x = rand(1, 2)` assigned 0; `Foo.name(1)` and `Mod.name(1)` answered nil where the in-range call answers the name.
- `1.5.respond_to?(:clamp)` and `:abc.respond_to?(:=~)` answered false where CRuby answers true: the inference probe behind `respond_to?` synthesizes a zero-argument call, and a count the method refuses typed unknown, which read as "no such method". A refused count names a method that exists, so the probe's call types nil and the answer is true.
- The in-range shape beside each still answers; the guard stands down for a def of the name — at top level, in a reopened Object or Kernel, in the enclosing class's chain, on a Struct class in either singleton spelling, on a module whose instance method `extend self` or `module_function` makes its own, and for a class object's own `class << self` accessors — whatever the count; and the receiver and the arguments are evaluated first, in order, as the guard always did.

## How

- **The rows** (`tools/gen_builtin_arity_spec.rb`): the probe receivers grow by the handle kinds the guard maps — a Regexp, a MatchData, a Proc, a Method, an unstarted Fiber, a finished Thread, a Queue, a SizedQueue, a ConditionVariable, an open File and Dir, a RuntimeError, an Array's Enumerator, a Random, a Struct and a Data instance, a Class and a Module — and the class-method targets by GC, Fiber, Thread, the Kernel functions (`throw`, `catch`, `raise`, `rand`, `srand`, `caller`, `lambda`, `proc`, `binding`, `block_given?`), File's predicates and metadata calls, `Process.times`, `spawn` and `waitpid2`, and the three Socket packers; the readers a class `Struct.new` or `Data.define` answers are keyed `StructClass` and `DataClass`. A row must hold for every value of its kind: Queue's rows are the ones Queue and SizedQueue agree on, since one TyKind stands for both; the IO rows are probed on a File, the surface any IO handle can answer. The names whose counts belong to what the receiver forwards to — `Proc#call`, `Method#call`, `Enumerator#each` and its `to_a`, `entries`, `to_h`, `each_entry`, `reverse_each`, `Class#new` — keep a row that never fires, the sentinel quartet the tables already had for a shape the probe cannot characterize, so the guard sees the class owns the name and does not read Object's row for it (`Proc#===` is an Object name too); a probe Struct's own accessors get no row. `freeze` joins the surface, since a fresh receiver per probe makes it harmless. 1,214 instance rows become 1,809 (fourteen of them the sentinels) and 191 class-method rows 240; every row that was there is in the regenerated tables unchanged but `waitpid2`'s, and the generator reproduces today's instance table from this Ruby before the change.
- **The map** (`emit_builtin_arity_guard`, `src/codegen_call.c`): each new kind names its rows; a Struct or Data instance its kind's; a name a builtin inherits from Object unchanged — `respond_to?`, `is_a?`, `freeze`, `instance_of?` — checks Object's row, but only when the class has no row of its own, so `12.to_s(2)` keeps Integer's. A bare call resolves the way the bare-call arm above the guard does — the enclosing class's chain, then a top-level def or a reopened Kernel or Object, then Kernel's rows. A user class constant keeps the class methods it defines and its `class << self` accessors (a singleton def on a class `Struct.new` answered, `def S1.members` or a def or `define_method` under a statement list, a branch, a loop or a block inside `class << S1`, is collected before that class exists and never reaches its chain, and a def under a conditional inside `class << self` reaches none either; the singleton defs written on a constant, and a module body's `extend self`, are read off the tree once and asked by name) and checks Class's or Module's surface (`name`, `superclass`, `ancestors`) with Object's beneath for the rest, or its Struct or Data readers; a native class or a bare constant (`GC`, `Fiber`, `Thread`) checks the class-method rows as before. A row Object supplied is checked against a reopened Object the way a class's own row is checked against a reopened class — `class Object; def itself(a)` keeps `5.itself(1)` — and, in either arm, against a top-level def, which is Object's private method and CRuby's NoMethodError, not a count; a kind that stands for two classes (the booleans, File and IO, the queues) asks the other class too. A module that extends itself owns its instance methods too, so a def of the name on such a module stands the guard down as well; `module_function` registers the name class-side and is found there. The two tables share a row type now, `SpAritySpec`, and one lookup, `arity_spec_row`.
- **The decision and the emission** are separate functions, so the analyzer can ask the first: `infer_call` types a refused call nil when the inference has no type for its shape. A dynamic `public_send` synthesizes a call per candidate name and drops one that types unknown, which is how `1.public_send(nm, nil)` became a NoMethodError; typed nil it is kept, and lowers to the raise. Any type is sound for a call that raises before it answers.

## Validation

- A new test, `test/builtin_arity_reach` (expectations generated by CRuby 4.0.6; 90 rescued ArgumentErrors, 3 ms of run time): a wrong count on each of the thirteen handle kinds, on a Struct and a Data instance and their classes, on a user class and a module constant, on `GC`, `Fiber` and `Thread`, on eight Kernel functions, `catch` bare and with a block, on the universal names over eight receiver kinds, and on a name handed to `public_send` in a variable; the in-range shape beside each, and a class's own row outranking Object's (`12.to_s(2)`); a top-level def shadowing a Kernel function, a reopened Object owning `itself` on an Integer and a String, a class object's `class << self` accessor, a module function, and an instance reader on a class that does not reach the class object; the receiver and the argument evaluated, in order, before the raise. On master it stops at its first assertion as an unsupported call.
- Full suite: 3,368 pass, 0 fail, from a cleared results cache, at `-O2`; the same 3,368 pass at the gate's `-O1`.
- Generated C against master, over the suite, the benchmarks and optcarrot (3,386 programs): 3 differ — `test/block_call_arity` (a guarded call types nil now, so `p` prints its nil path after the raise), `test/format_raise_dollarbang_symproc` and `test/issue_3053` (a symbol-proc dispatch on a Symbol read from a variable gains the arm for `+` with no argument, CRuby's ArgumentError where it said `undefined method '+'`). The 61 benchmarks and optcarrot are byte-identical, so no hot path changes; optcarrot's checksum is 59662 and its binary is master's.
- Against the differential corpus of ~2,100 minimized programs the campaign has been working from: 715 → 757 match, 42 gained and 0 lost. Forty are the wrong-count programs — a Regexp, MatchData, Proc, Method, Fiber, Thread, Queue, ConditionVariable, Exception or Enumerator receiver, a Struct class, a user class or module constant, `GC`, `Process`, `Thread`, `Fiber`, `Integer()`, `throw`, `respond_to?`, `public_send` with a name in a variable — and two are the `respond_to?` answers above.
- The generator reproduces master's instance table byte-for-byte on this Ruby before the change, and every row that was there is in the regenerated tables unchanged but `waitpid2`'s (see below).
- Front-end compile time of optcarrot (`-c --no-line-map` to `/dev/null`, three runs each, both compilers built the same way): the same to within run-to-run noise (1.00–1.01 s on both in one session, 1.04–1.06 s on both in another). The row lookup is a linear scan and the tables are half again as long, with a second scan for Object's row when a class has none of its own; that is a few microseconds per call site, and the singleton-def index is one pass over the tree per table; invisible on a program this size.

## Left alone, on purpose

- An in-range shape no arm matches — `[1].find(1) { }`, `[1].zip { }`, `h.update` with no argument, `File.delete` with no path, `GC.stat(nil)`, a blockless `Mutex#synchronize`, `Fiber.new` without a block — is a different mechanism: a fallback for a supported method's unmatched shape, where this raises only where CRuby raises. Those still refuse to compile or miss as they did.
- `Queue#push`, `<<`, `enq` and `SizedQueue#max`: one kind stands for both queues and their counts differ (`push` takes 1 on a Queue, 1..2 on a SizedQueue), so they have no row; `q.push` with no argument still refuses to compile. `SizedQueue#push(v, true)` and `Thread#join(1)` are in-range shapes without an arm, found while writing the test.
- A method the prelude defines in Ruby owns its name the way any def does: `Set#each(1)` and `Symbol#to_proc(nil)` are answered by those defs today, and a wrong count there is the user-method arity check's, not this table's.
- A bare Kernel call reaches the guard where its value is consumed — `x = rand(1, 2)`, `p rand(1, 2)` — but not in statement position, where the emitter drops a call without effect, nor as a method's last expression, which the return path's own arm answers: `def g; rand(1, 2); end` still answers 0.
- `Method#call`, `Proc#call`, `Enumerator#each` and `Class#new` carry their target's counts, so they never fire; `m.call` with too few arguments either fills the missing parameters in or fails the C build, by where the method is defined, on master and here alike, `pr.===(1, 2)` on a lambda of two is the unsupported call it was (the `===` arm below the guard has the one-argument shape), and `m.===()` on a Method still answers nil.
- A CRuby rejection worded otherwise than "wrong number of arguments" — `format` with no arguments ("too few arguments"), `:a.upcase(1)` ("invalid option") — the probe leaves unguarded, as it always has; a bare `format` is a NameError here.
- `Data#with(1)`: Data's own emitter takes the call, which is keywords-only, before the guard sees it. A singleton def on a class `Struct.new` answered — `def S1.members(x)`, or the same inside `class << S1` — is collected before the struct class exists and nothing dispatches it, on master and here alike; the guard stands down for it, so `S1.members(9)` is the unsupported call it was rather than a raise.
- A rescue modifier whose value is nil-typed fails the C build on master — `x = (nil rescue nil)` declares a `void` temp — and a refused Kernel call under one, `x = binding(1) rescue nil`, now types nil and reaches that failure where master refused the call as unsupported; both stop the build, and the modifier's temp is the emitter's own.
- The guard counts before the method checks its receiver, as the rows are CRuby's counts alone: `$stdout.flock` with a wrong count (the IO rows are a File's), `f.wait_writable(1, 2)` on a file opened for reading and `$stdout.wait_priority(1, 2)`, `5.define_singleton_method(:a, :b, :c)` raise ArgumentError where CRuby's is NoMethodError, IOError, TypeError; master refused the first as unsupported, answered the next two, and refused the last with its own diagnostic. Likewise a reopened *subclass* of a kind — `class KeyError; def itself(a)` — or an `attr_reader` in a reopened builtin — `class Regexp; attr_reader :match` — is not asked, and the row fires where master refused the program; a reopened Exception or a def in a reopened Regexp is.
- A blockless iterator with an argument, `(1..5).each_with_index("x")`, types poly and takes the run-time route ahead of the guard, so it stays a NoMethodError miss; `p Fiber.current(nil)`, `p cv.signal(1)` and `p t.join(1)` are refused because `p` has no emitter for a handle value, whatever the count; `x = cv.signal(1)` raises.
- Five class-method rows that ba4a3415 and cea93a40 wrote into the table by hand — `Process.spawn`, `Process.waitpid2`, `Socket.pack_sockaddr_in`, `Socket.sockaddr_un`, `Socket.pack_sockaddr_un` — come from the probe now, so a regeneration keeps them. Four are the same; `waitpid2`'s said "expected 1" and CRuby accepts 0..2, so `Process.waitpid2` with no argument, which waits for any child in CRuby, is no longer an ArgumentError here but the unsupported call it is (the emitter has the one-argument arm).
- Found on the way, not touched: a class-level `def srand(a, b)` called bare inside its class is answered by the Kernel `srand` arm, arguments dropped, on master and here alike (the bare-call arm asks the enclosing chain only for the names it routes to `emit_method_call`, and the guard stands down for that def as it should); `12.method(:to_s).call(2)` answers `"12"`, the argument dropped; a `class << self; attr_accessor :name` on a class does not dispatch (`Foo.name` answers the class's name on master and here), so the guard's stand-down for it defers to an arm that is not there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved argument-count validation for a broad range of built-in methods and classes.
  * Invalid calls now consistently raise clear `ArgumentError` messages with the provided and expected counts.
  * Expressions used as receivers or arguments are evaluated correctly before an arity error is raised.
  * User-defined methods and overrides continue to dispatch correctly, including top-level and extended methods.
* **Tests**
  * Added comprehensive coverage for valid and invalid built-in method calls across core types and APIs.

### DIFF
--- a/src/analyze_infer.c
+++ b/src/analyze_infer.c
@@ -1011,7 +1011,21 @@ static int an_bare_call_class_owned(Compiler *c, int id) {
          comp_reader_in_chain(c, cid, name, NULL);
 }
 
+static TyKind infer_call_inner(Compiler *c, int id);
+
+/* A builtin call whose count the arity guard refuses raises before it
+   answers, so any type is sound for it; nil lets the shapes that need one
+   -- a send candidate, which is dropped when it types unknown, an argument
+   -- compile down to the raise instead of a miss. The guard runs after a
+   few arms of emit_call, so none of those may claim a call it refuses, or
+   the value typed here is not the one emitted. */
 TyKind infer_call(Compiler *c, int id) {
+  TyKind t = infer_call_inner(c, id);
+  if (t == TY_UNKNOWN && builtin_arity_violation(c, id)) return TY_NIL;
+  return t;
+}
+
+static TyKind infer_call_inner(Compiler *c, int id) {
 
   /* a yielder push (`y << v` inside an Enumerator.new generator) lowers to a
      Fiber.yield, whose value is boxed -- never the array append it looks like */
--- a/src/analyze_internal.h
+++ b/src/analyze_internal.h
@@ -69,6 +69,7 @@ int is_builtin_class_name(const char *n);
 int is_builtin_module_name(const char *n);
 int is_builtin_exception_name(const char *n);
 int builtin_method_known(const char *cls, const char *m);
+int builtin_arity_violation(Compiler *c, int id);
 int is_handler_proc_block(Compiler *c, int id);
 int builtin_class_id(const char *name);
 int builtin_object_method_known(const char *m);
--- a/src/codegen_call.c
+++ b/src/codegen_call.c
@@ -9850,26 +9850,33 @@ static int emit_range_step_bad_stride(Compiler *c, int id, int recv, int arg,
   buf_printf(b, "%s; })", dv ? dv : "0");
   return 1;
 }
+/* A row of the two spec tables below; the instance table's comment says
+   how to read one. */
+typedef struct { const char *cls; const char *m; int min; int max;
+                 const char *lo_exp; const char *hi_exp;
+                 int blk_min; int blk_max;
+                 const char *blk_lo_exp; const char *blk_hi_exp; } SpAritySpec;
+
 /* A call whose argument count CRuby's own method cannot take: evaluate the
    receiver and arguments for effect, then raise CRuby's ArgumentError --
    instead of falling into an emitter specialized to another count, where a
    missing operand read argv[] blind (a compile-time SIGSEGV) and an extra one
    was silently dropped. Conservative by construction: fires only when the
    dumped CRuby arity table proves the count wrong, never on a call whose
-   runtime count a splat / kwargs / forwarding keeps open, and never when a
-   reopened builtin owns the name. */
+   runtime count a splat / kwargs / forwarding keeps open, and never when
+   the program owns the name -- a def in the enclosing chain, at top level
+   or in a reopened builtin, a singleton def or accessor on the constant, a
+   module's own method behind `extend self`. */
 /* Positional-arity spec for the builtin instance surface, probed from
    ruby 4.0.6 by tools/gen_builtin_arity_spec.rb (see there for the
    technique; rerun it with --write to regenerate both tables). max -1 =
    no upper bound; a NULL exp = that side is never violated. The first
    quartet describes the bare call, the blk_ quartet the block-carrying
    call (several counts change under a block: Array#fill 1..3 vs 0..2,
-   Integer#step); a quartet the probe could not prove is -1,-1,NULL,NULL
-   and never fires. */
-static const struct { const char *cls; const char *m; int min; int max;
-                      const char *lo_exp; const char *hi_exp;
-                      int blk_min; int blk_max;
-                      const char *blk_lo_exp; const char *blk_hi_exp; }
+   Integer#step); a quartet the probe could not prove, or a name whose
+   counts are its target's (Proc#call, Enumerator#each), is
+   -1,-1,NULL,NULL and never fires. */
+static const SpAritySpec
 sp_builtin_arity_spec_tbl[] = {
   {"String","%",1,1,"1","1",1,1,"1","1"},
   {"String","*",1,1,"1","1",1,1,"1","1"},
@@ -9930,6 +9937,7 @@ sp_builtin_arity_spec_tbl[] = {
   {"String","encoding",0,0,NULL,"0",0,0,NULL,"0"},
   {"String","eql?",1,1,"1","1",1,1,"1","1"},
   {"String","force_encoding",1,1,"1","1",1,1,"1","1"},
+  {"String","freeze",0,0,NULL,"0",0,0,NULL,"0"},
   {"String","getbyte",1,1,"1","1",1,1,"1","1"},
   {"String","grapheme_clusters",0,0,NULL,"0",0,0,NULL,"0"},
   {"String","gsub",1,2,"1..2","1..2",1,2,"1..2","1..2"},
@@ -10234,6 +10242,7 @@ sp_builtin_arity_spec_tbl[] = {
   {"Array","flat_map",0,0,NULL,"0",0,0,NULL,"0"},
   {"Array","flatten",0,1,NULL,"0..1",0,1,NULL,"0..1"},
   {"Array","flatten!",0,1,NULL,"0..1",0,1,NULL,"0..1"},
+  {"Array","freeze",0,0,NULL,"0",0,0,NULL,"0"},
   {"Array","grep",1,1,"1","1",1,1,"1","1"},
   {"Array","grep_v",1,1,"1","1",1,1,"1","1"},
   {"Array","group_by",0,0,NULL,"0",0,0,NULL,"0"},
@@ -10361,6 +10370,7 @@ sp_builtin_arity_spec_tbl[] = {
   {"Hash","first",0,1,NULL,"0..1",0,1,NULL,"0..1"},
   {"Hash","flat_map",0,0,NULL,"0",0,0,NULL,"0"},
   {"Hash","flatten",0,1,NULL,"0..1",0,1,NULL,"0..1"},
+  {"Hash","freeze",0,0,NULL,"0",0,0,NULL,"0"},
   {"Hash","grep",1,1,"1","1",1,1,"1","1"},
   {"Hash","grep_v",1,1,"1","1",1,1,"1","1"},
   {"Hash","group_by",0,0,NULL,"0",0,0,NULL,"0"},
@@ -10708,6 +10718,7 @@ sp_builtin_arity_spec_tbl[] = {
   {"Object","eql?",1,1,"1","1",1,1,"1","1"},
   {"Object","equal?",1,1,"1","1",1,1,"1","1"},
   {"Object","extend",1,-1,"1+",NULL,1,-1,"1+",NULL},
+  {"Object","freeze",0,0,NULL,"0",0,0,NULL,"0"},
   {"Object","frozen?",0,0,NULL,"0",0,0,NULL,"0"},
   {"Object","hash",0,0,NULL,"0",0,0,NULL,"0"},
   {"Object","inspect",0,0,NULL,"0",0,0,NULL,"0"},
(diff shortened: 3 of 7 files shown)


matz:
Merged. I reproduced your validation independently before pushing: the generated C over all 3,326 test programs differs in exactly the three you named (block_call_arity, format_raise_dollarbang_symproc, issue_3053), all three still answer their expected output, optcarrot's generated C is byte-identical (md5 f05afac9), the new test passes including under SPINEL_GC_STRESS=1, and the full gate is green. A review of this quality, with the blast radius measured and the deliberate non-fixes written down, is a pleasure to merge. The Left-alone list (the queue-kind split, the in-range unmatched shapes, waitpid2's corrected row) reads as the right calls, and the respond_to? consequence falling out of the decision/emission split is the kind of side effect one wants. Welcome, and thanks.

PR #4231, opened 2026-08-31
io: CV#wait(mutex, timeout) -- 2-arg form routes to non-blocking drain

rofl0r:
The 1-arg #wait(mutex) parks on the CV and re-acquires the mutex
on wake. The 2-arg form #wait(mutex, 0) in CRuby returns
immediately without parking, used as a 'drain any stale signal'
check before the next blocking wait. AOT's sp_CondVar_wait
ignored the timeout (the codegen dropped argv[1] with a comment
that said so), so the 2-arg form parked unconditionally and the
worker thread that called it never woke up.

Add sp_CondVar_wait_nb(cv, m): release the mutex, drain a
pending signal from the waiters list, re-acquire the mutex. The
codegen now routes argc==2 to this entry, leaving argc==1 on
sp_CondVar_wait. A real clock-driven timeout (t > 0) is not yet
supported in the scheduler; the 0 case is the only shape user
code uses for the drain check.

## Summary by CodeRabbit

* **New Features**
  * Added support for condition-variable waits with a literal zero timeout.
  * Zero-timeout waits now release and reacquire the associated mutex while checking for pending signals.

* **Bug Fixes**
  * Condition-variable waits now validate argument counts and reject unsupported non-zero or dynamic timeouts instead of silently ignoring them.

* **Tests**
  * Added coverage confirming zero-timeout waits complete successfully and produce the expected output.


matz:
Merged, after verifying your test deadlocks on pre-merge master (ThreadError: no runnable thread) and passes after. The no-park handoff mirrors sp_CondVar_wait's release half, and raising NotImplementedError for a timeout the scheduler cannot honor beats the silent forever-park it replaces. Two nits I left as-is: the sp_sched.h comment says "drain a pending signal" where the implementation deliberately drains nothing (the .c comment explains why correctly), and the codegen comment says rejected "at compile time" where the raise is at run time. Worth folding into the clock-driven-timeout work when the scheduler grows it. Full gate green.

PR #4234, opened 2026-08-31
Give the openssl package ES256, and a key that is only public

rubys:
> **Stacked on #4233, which is stacked on #4232.** This branch carries their commits too; the last one is the reviewable change here. Merge in order and I will rebase.

The last primitive Web Push needed. With this a VAPID (RFC 8292) `Authorization` header is buildable end to end: `k=` is `#public_key_bytes`, the `t=` JWT is base64url over JSON with an ES256 signature, and `net/http` already delivers it over TLS.

```ruby
key = OpenSSL::PKey::EC.from_private_bytes("prime256v1", vapid_private)
sig = key.sign_raw("SHA256", "#{b64u(header)}.#{b64u(claims)}")   # 64 bytes, r || s
```

**Signatures are raw `r || s`, never DER, and the conversion is in the C rather than in each caller** — a DER signature is a well-formed String that every JWT verifier rejects, so getting it wrong fails silently and remotely, in someone else's stack, long after the call.

## Why `#sign_raw` and not `#sign`

This is the one place I did not keep CRuby's name, and I would rather flag it than let it pass.

Everywhere else the package keeps CRuby's name for a different *argument* shape: `dh_compute_key` takes bytes where CRuby takes a `Point`, and answers **the same bytes CRuby answers**. `#sign` is not that — CRuby's answers DER — so keeping the name would compile a CRuby program here and hand it bytes CRuby never produces. Failing at compile time is what this package offers instead, so the name differs precisely because the bytes do.

Say the word and I will rename it to `#sign` if you would rather have the collision. The *argument* shape is CRuby's — a digest name and the data, not a pre-computed hash — so a caller cannot accidentally sign something unhashed.

## `EC.from_public_bytes`

A key that is only public: everything a verifier has, and the side a push service is on. `#verify_raw` works on one; `#sign_raw` and `#dh_compute_key` raise, because there is no scalar to do them with. The point is checked where the key is built rather than at first use, so an ECDSA verify against something not on the curve is not reachable at all.

## The deprecated call, which is a decision and not an oversight

`ECDSA_do_sign` / `ECDSA_do_verify` are `OSSL_DEPRECATEDIN_3_0`, and the 3.0 replacement — `EVP_PKEY_fromdata` into `EVP_DigestSign` — did not exist at this file's 1.1.0 floor. Unlike ECDH, where `EC_POINT_mul` happened to span both, **no single ECDSA API does.**

A version shim would mean two signing paths of which only one is ever compiled on a given host, and a crypto path that never runs where the tests run is a crypto path nobody has checked. So this takes the deprecated call everywhere and silences the warning at the two sites that earn it, with the reason written above them. If the floor moves to 3.0 the EVP form is a small single-path replacement. Everything either side of the call — `ECDSA_SIG_get0`, `ECDSA_SIG_set0`, `BN_bn2binpad` — is current in 3.x and present at 1.1.0.

Happy to take the shim instead if you would rather carry both.

## Testing a randomised signature

ECDSA draws a fresh `k` per signature, so there is no fixed vector and the expected-output harness cannot pin one. **That is why `#verify_raw` ships alongside `#sign_raw`** rather than waiting for a caller to need it — it is what makes the test self-checking.

The test signs and verifies, checks that two signatures over the same input differ and both verify, and checks the six things that must **not** verify: altered data, a mangled `r`, a short signature, an empty one, another key's signature, and this key's signature under another key. `#verify_raw` answers false for a malformed signature as well as a wrong one — a caller asking "is this good" must not be able to read "the input was strange" as yes.

Cross-checked against CRuby by hand in both directions, since the harness cannot:

```
CRuby verifies spinel's ES256 signature: true
and rejects it over altered data:       false
spinel verifies CRuby's ES256 signature: true
and rejects it over altered data:        false
```

P-384 signs and verifies at its own width, 96 bytes, so the halves follow the curve and not a constant.

`make test`: **3372 pass, 0 fail, 0 error**, green under `SPINEL_GC_STRESS=1`.

## Summary by CodeRabbit

- **New Features**
  - Added AES-128-GCM and AES-256-GCM encryption and decryption support through OpenSSL.
  - Added random key and IV generation, authenticated data, authentication tags, and clear validation errors.
  - Added ECDSA signing and verification, including public-key-only EC key support.
- **Bug Fixes**
  - Binary encoding is now preserved across concatenation, joining, repetition, slicing, and empty-string results.
  - Improved concurrent and empty-message authentication handling for AES-GCM.
- **Tests**
  - Added coverage for encryption vectors, authentication failures, EC signatures, public keys, and binary string behavior.


matz:
Merged, and keep sign_raw. Your reasoning is the deciding one: everywhere else the package keeps CRuby's name for a different argument shape while answering the same bytes, and #sign would answer bytes CRuby never produces. A compile-time failure the caller can see beats a DER signature that fails silently in someone else's verifier. Same for the deprecated ECDSA pair: one path everyone runs is worth more than two paths of which each host tests one, and the comment above the call says so for whoever moves the floor to 3.0.

I verified the direction your harness cannot pin: CRuby verifies a signature produced by this code (over the RFC 8291 key) and rejects it over altered data. from_public_bytes checking the point where the key is built, so a verify against an off-curve point is unreachable rather than merely refused, is the right place for it. Full gate green, test green under SPINEL_GC_STRESS=1.

With this the #4221 surface is complete as scoped, so I am closing that issue; anything further (a mode, a curve, the ES256 header helpers) is its own request.

Issue #4071, opened 2026-08-22
A user class's own `join` is not called when the receiver is widened to untyped


elct9620:
## What happens

When a method's receiver is widened to `untyped`, a call to `join` does not enter the
receiver's own `join`. It answers the receiver's string form, and nothing is reported at
compile time. An Array reaching the same call is answered correctly, separator and all.

The emitted C shows the call lowered to a bare `sp_poly_join`, with no `switch` on the
receiver's class — where the same file with the method renamed emits one carrying an arm
for the class.

## Reproduction

`spinel repro.rb -o repro`, no requires. `@ran` records whether the method was entered.

```ruby
class Row
  def initialize = @ran = false
  attr_reader :ran

  def join(sep)
    @ran = true
    "row"
  end
end

def joined(v) = v.join("-")

r = Row.new
p joined(r)
p r.ran
p joined(["x", "y"])
```

```
$ ruby repro.rb
"row"
true
"x-y"
```

```
$ ./repro
"#<Row:0x...>"
false
"x-y"
```

```
def joined: (untyped) -> String # spinel: widened to untyped (slow path)
```

## The same file with the method renamed

`join` written `glue` throughout, which no builtin carries. The last line becomes
`begin; glued(42); rescue NoMethodError; end`, since an Array has no `glue` either and
something has to widen the parameter:

```ruby
class Row
  def initialize = @ran = false
  attr_reader :ran

  def glue(sep)
    @ran = true
    "row"
  end
end

def glued(v) = v.glue("-")

r = Row.new
p glued(r)
p r.ran
begin; glued(42); rescue NoMethodError; end
```

```
$ ruby repro.rb
"row"
true
```

```
$ ./repro
"row"
true
```

```
def glued: (untyped) -> String # spinel: widened to untyped (slow path)
```

Same signature as the reproduction, so the inferred types are not what separates them.
Widening the first file with `42` in place of the Array gives the same answer as the
Array version does.

## Generated C

`spinel repro.rb -c -o repro.c`, the body of the widened helper. Whitespace added, the
string literal's expansion elided.

`join`:

```c
static inline const char *sp_joined(sp_RbVal lv_v) {
  return sp_poly_join(lv_v, /* "-" */);
}
```

`glue`, from the file above:

```c
switch ((_t1.tag == SP_TAG_OBJ ? _t1.cls_id : 0x7fffffff)) {
  case 0: _t2 = sp_Row_glue((sp_Row *)_t1.v.p, _t3); break;
[...]

Issue #3941, opened 2026-08-16
[Runtime] A mutable String read out of an Array or Hash into a local is a copy, so in-place mutation never reaches the container


KieranP:
> (Disclaimer: I asked Claude Opus 5 (high) to find Spinel compile/runtime issues with documented Ruby classes and methods. It found the following issue, which I manually verified for accuracy before publishing this report)

_Reproduced with `spinel 6eadf28a69a4` vs `ruby 4.0.3`._

Reading a mutable String out of a container into a local (`r = rows[0]`) and mutating it
in place updates only the local: the Array or Hash keeps the original value, and
`r.equal?(rows[0])` answers `false` where Ruby answers `true`. It affects `upcase!`,
`sub!`, `<<` and `replace` alike, on both Array elements and Hash values.

`docs/limitations.md` states the opposite as a guarantee — aliased in-place mutation is
observed "across every storage shape: local aliases, array elements and hash values
(stored or read back …)".

## Repro

```ruby
rows = [+"abc"]
r = rows[0]
r.upcase!
p rows
p r
p r.equal?(rows[0])
```

Ruby 4:

```
["ABC"]
"ABC"
true
```

Spinel:

```
["abc"]
"ABC"
false
```

The same loss occurs through a Hash value (`v = h[:a]; v << "Z"`) and with `sub!` and
`replace`. Mutating *through* the container (`rows[0] << "Z"`), mutating a block
parameter (`rows.each { |r| r << "Z" }`), and aliasing the string before storing it
(`s = +"abc"; rows = [s]; s[1] = "*"`) all match Ruby.

Issue #3459, opened 2026-07-30
Poly dispatch over (builtin | user class) loses the builtin arm: Array#first/#last and Hash#keys/#values raise NoMethodError, Array#map/#select silently run the user arm


rubys:
When a call site is polymorphic over **a builtin container and a user-defined class that defines the same method name**, the builtin arm is lost. Two symptoms depending on whether the method takes a block:

- **non-yielding** — `Array#first`, `Array#last`, `Hash#keys`, `Hash#values` raise `NoMethodError` on a genuine Array/Hash receiver
- **yielding** — `Array#map`, `Array#select` **silently dispatch to the user arm**: the block never runs, the result is wrong, no error is raised

This looks like the residue of #3409, which fixed the same shape for a yielding method. `each` is correct now; `map` and `select` are not.

A union of two *builtins* (`Array | Hash`) dispatches correctly — a user-defined class has to be in the union to trigger it. No `--rbs` seed is needed; plain `spinel main.rb` reproduces.

## Repro — non-yielding (NoMethodError)

```ruby
class Shadow
  def keys; nil; end          # a user class that shadows a builtin name
end

class Src
  def self.pick(w)
    if w
      h = {}
      h["a"] = 1
      h
    else
      Shadow.new
    end
  end
end

h = Src.pick(true)            # a Hash at runtime
puts "ok " + h.keys.to_s
```

```
$ ruby main.rb
ok ["a"]
$ spinel main.rb -o m && ./m
undefined method 'keys' for an instance of Hash (NoMethodError)
```

Swap `keys` for `values`, or make `Src.pick` return `[3, 1, 2]` and use `first` / `last`, for the same result.

## Repro — yielding (silent wrong answer)

```ruby
class Shadow
  def map; nil; end
end

class Src
  def self.pick(w)
    w ? [3, 1, 2] : Shadow.new
  end
end

a = Src.pick(true)            # an Array at runtime
n = 0
a.map { |x| n = n + 1 }
puts "ok " + n.to_s
```

```
$ ruby main.rb
ok 3
$ spinel main.rb -o m && ./m
ok 0
```

The block is never entered — consistent with `Shadow#map` (which takes no block and returns nil) being selected for an Array receiver.

## Matrix

Shadow a name with a user class, put both in one poly slot, call it on the builtin receiver:

| method | receiver | spinel | CRuby |
|---|---|---|---|
| `first` | Array | `NoMethodError` | `3` |
[...]

PR #4242, opened 2026-08-31
Reach a user #replace through an untyped slot

### LINKED ISSUE #4240: A user method named after a builtin, called with a block through an untyped slot, runs the builtin's semantics instead
At `1ef31373`. A class defines `replace` (a name `String`/`Array`/`Hash` also own) taking a block. Through a **typed** receiver the user method runs. Through an **untyped** slot (an element of a heterogeneous array), the call silently takes the builtin's path — with the builtin's semantics applied to a non-String object — rather than dispatching to the user candidate or raising.

```ruby
class Frag
  def initialize(html)
    @html = html
  end

  def replace(selector)
    r = yield
    Frag.new("user-replace:" + selector + ":" + r.to_s)
  end

  def to_s
    @html
  end
end

f1 = Frag.new("abc")
p f1.replace("sel") { nil }.to_s   # typed receiver

box = [Frag.new("abc"), "just a string"]
f2 = box[0]
p f2.replace("sel") { nil }.to_s   # untyped slot
```

| | typed | untyped slot |
|---|---|---|
| CRuby | `"user-replace:sel:"` | `"user-replace:sel:"` |
| spinel | `"user-replace:sel:"` | **`"sel"`** |

The untyped-slot answer is `String#replace`'s: the receiver's content became the argument — on a `Frag`. Same family as #4205 (there the user candidate was dropped by kwargs arity; here the discriminator seems to be the block) and adjacent to #4218's instance-table routing.

Where it bit: ONCE Campfire's `ContentFilters::SanitizeTags#apply` is `fragment.replace(selector) { nil }`, and the fragment arrives through an untyped filter-chain slot — so the app's tag-stripping filter has been a silent no-op on the compiled binary the whole time (each message body still passed a downstream safe-list sanitizer, which is why nothing unsafe rendered; the divergence surfaced as a byte-diff against Rails, not as an XSS). Found the day a Rails-vs-binary frame comparator first ran.

Repro: the file above, `spinel main.rb && ./main` vs `ruby main.rb`.

rubys:
Fixes #4240.

## What happened

A user method named after a builtin ran the **builtin** when the receiver arrived through an untyped slot. `f = box[0]; f.replace("sel")` on a `Frag` answered `"sel"` — `String#replace`'s semantics applied to a non-String: the receiver's contents became the argument.

The block in the issue's repro turns out to be incidental. The discriminator is the **name**: `replace`, `concat`, `prepend`, `clear` and the bang forms are the set spinel's string-handle (strbuf) machinery keys on. `upcase` and friends through the same slot already dispatched correctly.

## Why

Two independent paths, both the "a user class owns this name" rule the neighbouring code already follows (#3459, #4071):

1. **`src/analyze.c`** — the container-read alias rule (#3941) narrowed `f = box[0]` to a string handle on name-keyed evidence alone (`f.replace(…)` "must be" `String#replace`). The slot's static type became `sp_String *`, so the user candidate was never in play. Gate the narrowing on the slot still being poly *and* the mutator name being one a user class defines — placed **after** the container's own stores are demanded into handles, so a genuine `String` element still mutates in place.

2. **`src/codegen_call_recv.c`** — the poly-receiver `replace` arm emitted `sp_poly_replace(…)` unconditionally, unlike `join` / `pack` / `delete` beside it, which stand down via `user_defines_or_reads`. Give it the same guard.

Standing that arm down opens the `cls_id` dispatch — where a *genuine* `String` or `Array` receiver then had no arm of its own and landed on the raise default. So:

3. **`src/codegen_call.c`** — give the dispatch a builtin `replace` arm, next to the `IO#write` / `read_nonblock` arms and on the same terms already stated there: a user class owning a name must not change what a builtin receiver does.

## Test

`test/poly_slot_user_replace.rb` covers all three: the typed receiver, the untyped slot, and a genuine mutable `String` reached through the same slot (whose `#replace` must still mutate the array's element in place).

`make test`: **3378 pass, 0 fail, 0 error**.

https://claude.ai/code/session_011kwS4WizPEpFCJTmyvHARf

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect method dispatch for `replace` when untyped values can refer to user-defined classes.
  * Preserved built-in `String` and `Array` replacement behavior while correctly invoking matching user-defined methods.
  * Prevented user-defined `replace` methods from being mistaken for built-in string operations.

* **Tests**
  * Added coverage for replacement dispatch through untyped values and user-defined methods.

### DIFF
--- a/src/analyze.c
+++ b/src/analyze.c
@@ -9024,6 +9024,27 @@ static int strbuf_mut_kind(Compiler *c, const char *vn, Scope *vs) {
                                    vs ? (int)(vs - c->scopes) : -1, 0);
   return v ? *v : 0;
 }
+/* True when the mutator evidence for local `vn` rests on a name a USER class
+   also defines. The mutator table is keyed on the NAME alone, so `f.replace(x)`
+   reads as String#replace whatever `f` holds; on a slot whose type is still
+   poly/unknown -- an element read out of a heterogeneous container -- that is
+   not evidence the value is a string, and narrowing the slot to the handle
+   applied the builtin's semantics to a user object (#4240). */
+static int strbuf_mut_name_user_owned(Compiler *c, const char *vn, Scope *vs) {
+  const NodeTable *nt = c->nt;
+  if (!vn) return 0;
+  NT_FOREACH_KIND(nt, NK_CallNode, u) {
+    int ur = nt_ref(nt, u, "receiver");
+    if (ur < 0 || nt_kind(nt, ur) != NK_LocalVariableReadNode) continue;
+    const char *urn = nt_str(nt, ur, "name");
+    if (!urn || !sp_streq(urn, vn) || comp_scope_of(c, ur) != vs) continue;
+    const char *un = nt_str(nt, u, "name");
+    if (!un || !sp_str_mutator(un, SP_MUT_LOCAL)) continue;
+    if (an_user_defines_method(c, un)) return 1;
+  }
+  return 0;
+}
+
 /* In-place mutation status of ivar `nm` of class `cid` (same contract as
    strbuf_mut_kind). The supported set is narrower: the shadow-copy shim
    cannot rename an ivar, so []=/insert/slice!/setbyte disqualify. */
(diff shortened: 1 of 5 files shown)


rubys:
Superseded by c178fc15 — which landed before this went up. Same two mechanisms, independently found; closing.

One thing worth keeping from this branch. The analyzer gate differs: c178fc15 declines the promotion for the whole container when any store is provably non-string, while this branch keeps `strbuf_demand_container_stores` running and declines only the *local's* narrowing. That difference is visible in the mixed case:

```ruby
class Frag
  def initialize(h); @h = h; end
  def replace(x); "user:" + @h + ":" + x; end
end
box = [Frag.new("abc"), +"just a string"]
f = box[0]
p f.replace("sel")
s = box[1]      # a genuine String through the same untyped slot
s.replace("new")
p s
p box[1]
```

|  | CRuby | c178fc15 | this branch |
|---|---|---|---|
| `f.replace("sel")` | `"user:abc:sel"` | `"user:abc:sel"` | `"user:abc:sel"` |
| `s` after `replace` | `"new"` | **`"just a string"`** | `"new"` |

That is the neighbour called out as #2 in the issue comment, so it is known rather than a surprise — the String element never becomes a handle once the container is refused wholesale, so its `#replace` answers the receiver.

It also turns out to be a one-line reorder on top of c178fc15: run the heterogeneity gate *after* the demand rather than before, and the container's string stores stay handles while the local still keeps its poly type. Verified on master; follow-up PR to come.

PR #4246, opened 2026-09-01
Give Proc#curry CRuby's completion rules

blacktm:
## Why

Currying is how Ruby code builds specialized functions from general ones — `add.curry[1]` and friends show up in ordinary functional code. Spinel's curry deferred forever on the shapes CRuby completes: `->(*values) { values.length }.curry.call` answered a Proc where Ruby invokes and answers 0, so a program that curried anything variadic (or a Method, which curries through the same machinery) silently got an accumulator object where it expected a result. `curry(nil)` and any computed count refused to compile, and a curried proc stored in a hash came back unusable.

Concretely:

- A zero-required variadic base was special-cased to realize only after a first application, so `curry.call` with no arguments handed back another accumulator, typed Proc, where CRuby invokes at the minimum arity — zero — and answers the value.
- A *computed* count compiled but was silently ignored: the static chain traced the base's own arity instead, so `pr.curry(n)[1][2][3]` with n = 3 answered nil. And the compile-time count check mis-fired on optional-parameter lambdas: `->(x, y = 2) { [x, y] }.curry(2)[7][8]` raised ArgumentError where CRuby answers `[7, 8]`.
- `curry(nil)` (CRuby's no-count spelling) and `curry(obj)` with a `to_int`-answering object were unsupported-call compile errors.
- `h = { f: add.curry[1] }; h[:f][2]` answered nil — the poly index path had no idea a boxed curry applies rather than indexes.
- `curried.parameters` was a compile error (CRuby: `[[:rest]]`) and `curried.to_proc.call` a runtime NoMethodError.

## What

- A curry now completes exactly where CRuby's does: at `curry(n)`'s n when given, else the base's min arity — required parameters, trailing posts and required keywords included. The count is stamped into the accumulator and every application path defers or realizes by it: the static chain, the runtime saturation helper, `to_proc`'s wrapper, boxed callable calls, and container reads.
- A curried proc read out of a container applies its `[]` argument whatever the key kind — String, Symbol, Float and object keys used to fall into the key-typed hash reads and answer nil.
- A partial carried in a container or an untyped slot now converts under `&` — into user `&blk` methods, `yield` and `.call` — via a curry arm in `sp_poly_to_proc`, the `&obj` conversion point (partials bound to a local already drove builtin iterators through the block-arg desugar). And calling a boxed non-callable raises CRuby's exact NoMethodError — for every value kind, curry or not: a realized curry applied once more used to crash on a raw proc cast, and `h[:z].call` on a nil answered 0 where CRuby raises.
- `curry(nil)`, computed counts, boxed counts out of containers, and `to_int`-convertible counts all work and are honored; a boxed count is read at run time exactly as CRuby's `proc_curry` reads it — nil (an int-typed slot's nil sentinel included) is no count, and everything else converts through the runtime's Integer argument protocol, the user-object `to_int` bridge and CRuby's exact TypeErrors included, and the count validates against the lambda's min..max up front with CRuby's ArgumentError and its exact expected form (`1`, `1..2`, `1+`; a keyword hash counts as one more slot) — replacing the compile-time check that raised a countless message, mis-fired on optional-parameter lambdas CRuby accepts, and skipped computed counts (and their side effects) entirely.
- A curried proc read out of a container applies through `[]`, realizing when it completes.
- `parameters` answers `[[:rest]]` whatever the base; `to_proc` hands back the callable wrapper the runtime already had.

## How

- `sp_Curry` (lib/sp_proc.h) carries its completion count, stamped at creation: `sp_curry_new` derives the min from the target's arity (negatives encode it as `-req-1`); the new `sp_curry_new_n` validates an explicit count against min..max — the max passed in by the compiler, which alone can see optionals (a compile-time guess below the target's own min defers to the target) — and stores it; and `sp_curry_new_v` reads a boxed count at run time through `sp_poly_arg_int_chk`, the runtime's existing Integer argument protocol. `sp_curry_call_poly` reads the stored count, `sp_curry_apply`'s struct copy carries it, and the `to_proc` trampoline and `sp_poly_callable_call` both delegate to the same saturation decision instead of realizing count-blind. The `curry(n)` emission roots the receiver while the count expression runs — a freshly built proc was otherwise collectible under an allocating count.
- The static curry chain (src/analyze_pass.c) drops its variadic special case (arity from the required count now), reads `curry(nil)` as no count, and declines to trace a computed count — those chains fall to the runtime saturation path, which decides with the stored count. The compile-time count check and `curry_base_info`, its last caller, are deleted in favor of the constructor's runtime check.
- A new desugar (src/analyze_desugar.c) rewrites `proc.curry(obj)` to `proc.curry(obj.to_int)` only for a count of some settled non-Integer type (a `to_int` object, a Float) — Integer, nil, boxed and not-yet-settled counts are left for the arm, so the rewrite never fires on the wrong fixpoint round; the codegen curry arm accepts nil and any int-typed, nil-typed or boxed count expression.
- The poly index paths (lib/spinel_rt.h) gain curry arms — the boxed-key entry point ahead of its key-typed dispatch, the String- and Symbol-keyed reads, and the cold integer path — so one rule covers every key kind: a curried receiver applies, everything else indexes.
- Two small TY_CURRY codegen arms: `to_proc` emits the existing `sp_curry_to_proc`; `parameters` builds `[[:rest]]`, interning `:rest` at run time (the TU's symbol table is sealed before this arm runs).

## Validation

- `make test`: green, including the new `test/proc_curry_arity.rb` (40 expected lines, CRuby-4.0.6-generated, plain and under `SPINEL_GC_STRESS=1`).
- Generated-C sweep across all 3,394 pre-existing suite/benchmark programs: byte-identical except three curry tests whose diff is the counted constructor (receiver rooted, static max supplied) at their `curry(n)` sites — all three still match CRuby end-to-end.
- Differential corpus of ~2,100 minimized programs (same corpus behind #3999/#4003/#4029): 766 -> 774 match, 8 gained and 0 lost — the variadic curry invocations (Proc and Method), curry(nil), both to_int counts, the container-carried application, parameters, and to_proc. Nothing lost in any of the four full replays run during development.

## Left alone, on purpose

- A TYPED count with no `to_int` raises `NoMethodError: undefined method 'to_int'` where CRuby raises `TypeError: no implicit conversion into Integer` — the desugar spells CRuby's conversion, not its refusal message; a boxed count gets CRuby's TypeError through the argument protocol (whose existing family renderings differ from CRuby's on a `to_int` answering the wrong type or raising, and truncate a bignum count where CRuby raises RangeError). A `to_int` answering a statically non-nil non-Integer stays a compile error; one whose return infers nil — or returns nil at run time — reads as no count where CRuby raises, the price of the nil-sentinel rule. The conversion runs once, on both paths, where CRuby re-runs it per application.
- The count validates against the max only where the compiler can see the parameter list; a curry built from an untraceable optional-parameter base checks the min alone, with the `min+` message form.
- Nil counts of every provenance — literals, nil-typed expressions, nil-holding locals, boxed nils, and an int-typed slot's nil sentinel (`curry(arr[10])`) — read as no count.
- A curry counted above the 16-slot argument channel (`curry(17)`) can never realize — the accumulator caps at 16 applications, a family-wide limit that an explicit count now makes visible.
- An INLINE partial application as a block operand on a builtin iterator (`[1, 2].map(&partial[10])` with a run-time count) refuses at run time, and the refusal renders as the iterator's own NoMethodError rather than pointing at the operand; bind the partial to a local first. This is the one shape master answered better: ignoring the computed count gave right answers exactly when the count equaled the base's arity. A boxed partial has the same builtin-iterator gap (`map(&h[:p])`), identical on master; `find`/`detect`/`find_index` with any `&partial` fail the C build on master and branch alike; and with a user `#call` defined anywhere, block-passing any poly-carried callable into a builtin iterator misses — the pre-existing user-dispatch stand-down.
- A count expression that needs statement hoisting evaluates its hoisted part before the receiver — the emitter's pre-statement buffer, shared with other call shapes; the receiver-first order holds for inline counts, and master dropped such counts entirely.
- Reflection on a curry read out of a container (`h[:f].arity`, `.parameters`) still misses — the poly reflection arms know Proc and Method but not the curry, a pre-existing asymmetry; `class` answers Proc correctly. Likewise `curried.curry` (CRuby allows re-currying), explicit `inspect`/`to_s` and `puts` on a curry, `p` printing without CRuby's `(lambda)` marker, and `curry.to_proc.parameters` (`[]` vs `[[:rest]]`) predate this change.
- Under `--int-overflow=promote` the test now compiles (boxed counts fixed the former build refusal); the run still stops partway there on the pre-existing promote-mode arithmetic family shared with other suite tests (nothing in CI sets the mode, and the Makefile's mode filter is the precedent for excluding such tests if it ever does).
- The curry arms link ~1.3KB of text into programs whose poly index cold path is reachable, and nothing into programs where it is not (object files byte-identical); generated C stays byte-identical, and the hot inline path is untouched.

## Summary by CodeRabbit

- **New Features**
  - Improved `Proc#curry` and `Method#curry` arity handling, including explicit counts and `nil`.
  - Added support for computed and convertible curry counts.
  - Enhanced partial application, container calls, `to_proc`, and parameter reflection.
  - Added validation for optional, keyword, post, and variadic arguments.

- **Bug Fixes**
  - Corrected curry completion behavior across boxed and polymorphic values.
  - Added clearer errors for invalid curry counts and unsupported callable receivers.
  - Improved handling of keyword hashes and block arguments.

### DIFF
--- a/lib/sp_proc.c
+++ b/lib/sp_proc.c
@@ -61,6 +61,41 @@ sp_Curry *sp_curry_new(sp_Proc *p) {
   SP_GC_ROOT(p);  /* the target proc has no other root across this alloc */
   sp_Curry *c = (sp_Curry *)sp_gc_alloc(sizeof(sp_Curry), NULL, sp_curry_scan);
   c->target = p; c->nargs = 0;
+  /* no count given: realize at the target's required-parameter count (CRuby's
+     min arity; a negative arity encodes it as -req-1) */
+  c->arity = p ? (p->arity < 0 ? -p->arity - 1 : p->arity) : 0;
+  return c;
+}
+/* Proc#curry(n): the count overrides the default. CRuby validates n against
+   a lambda's min..max arity at the curry call, before anything is applied.
+   The min comes off the target's arity; the max only the compiler can see
+   (an optional widens it, a rest lifts it), so the call site passes it in --
+   max < 0 means unlimited-or-unknown, and the message then says "min+". */
+sp_Curry *sp_curry_new_n(sp_Proc *p, sp_int n, sp_int max) {
+  /* an int-typed slot's nil sentinel is CRuby's nil count: no count at all */
+  if (n == SP_INT_NIL) return sp_curry_new(p);
+  if (p && p->lambda_p) {
+    sp_int min = p->arity < 0 ? -p->arity - 1 : p->arity;
+    sp_int mx = max >= 0 ? max : (p->arity >= 0 ? p->arity : -1);
+    /* the max is the compiler's guess from a visible parameter list; the min
+       is the target's own truth. A guess below it names a different write of
+       the same name -- trust the target, drop the guess. */
+    if (mx >= 0 && mx < min) mx = -1;
+    if (n < min || (mx >= 0 && n > mx)) {
+      const char *want = mx < 0 ? sp_sprintf("%lld+", (long long)min)
+                       : mx == min ? sp_sprintf("%lld", (long long)min)
+                       : sp_sprintf("%lld..%lld", (long long)min, (long long)mx);
+      /* both strings live on the collected heap; the second sp_sprintf and
+         the raise path itself allocate, so root them across those */
+      SP_GC_ROOT_STR(want);
+      const char *msg = sp_sprintf("wrong number of arguments (given %lld, expected %s)",
+                                   (long long)n, want);
+      SP_GC_ROOT_STR(msg);
+      sp_raise_cls("ArgumentError", msg);
+    }
+  }
+  sp_Curry *c = sp_curry_new(p);
+  c->arity = n < 0 ? 0 : n;
   return c;
 }
 /* Each accumulated argument is stored boxed so a non-int arg (a String, an
--- a/lib/sp_proc.h
+++ b/lib/sp_proc.h
@@ -23,7 +23,11 @@
 #include "sp_alloc.h"   /* sp_PolyArray, sp_box_sym, sp_box_poly_array, sp_raise_cls */
 
 typedef struct sp_Proc { void *fn; void *cap; void (*cap_scan)(void *); sp_int arity; sp_bool lambda_p; sp_int param_count; const sp_sym *param_kinds; const sp_sym *param_names; sp_bool frozen; /* Object#freeze observed (sp_gc_alloc zero-fills) */ void *origin; /* dup/clone lineage root for Proc#== (NULL: self is the root) */ } sp_Proc;
-typedef struct { sp_Proc *target; sp_int nargs; sp_RbVal args[16]; } sp_Curry;
+/* arity: the count this accumulator realizes at -- Proc#curry(n)'s n, else
+   the target's required-parameter count (CRuby's min arity, so a variadic
+   base realizes on its first call). Carried here so a curry that travels
+   through a container or an untyped slot still knows when it is done. */
+typedef struct { sp_Proc *target; sp_int arity; sp_int nargs; sp_RbVal args[16]; } sp_Curry;
 
 sp_int sp_proc_call(sp_Proc *p, sp_int argc, sp_int *args);   /* defined in the generated TU */
 extern SP_TLS sp_RbVal _sp_proc_poly_args[16];                   /* defined in the generated TU */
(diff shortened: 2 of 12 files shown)


matz:
Merged. The test passes plain and under SPINEL_GC_STRESS=1, and the full gate is green.

Moving the count check from compile time into the constructor is the right direction and the writeup says why better than a diff could: the old check could not see optionals, could not see computed counts at all, and raised a message CRuby does not use. Stamping the completion count into the accumulator so the static chain, the runtime saturation helper, to_proc and the container reads all read one number is what makes the whole family agree instead of four places each guessing.

Rooting the receiver while the count expression runs is the kind of detail that only shows up when someone actually writes the allocating case, and it is good that it went in with the change rather than as a later GC report.

One note for later, not a change request: the gcc CI leg on this branch took 20 minutes where the other two took 5-6. Master's test leg is unwrapped from sccache now (970112c6) and your branch has it, so that looks like a slow runner rather than the cache-miss path that used to cause it. If you see it again on a fresh branch, say so and I will look at the workflow rather than assume weather.

PR #4249, opened 2026-09-01
Exception#set_backtrace + Exception#backtrace on builtin + user subcl…

rofl0r:
…ass roundtrip

Add an Exception#set_backtrace(bt) builtin that stores the array on the base sp_Exception struct, and make Exception#backtrace read it back (falling back to sp_backtrace_captured() when nothing was attached). CRuby already has both; AOT previously had neither, so the common 'rescue; err.set_backtrace(bt); raise err' pattern raised NoMethodError.

The codegen arm gates on 'TY_EXCEPTION or TY_OBJECT of an exception subclass' and stands down when a user class defines its own #backtrace (the chain check at the gate would have returned >= 0 in that case). The arm reads the base struct's backtrace field for a builtin receiver and, for a user subclass whose concrete class the analyze pass lost (a rescue variable typed as sp_Exception *), reads the first user ivar at the fixed offset every user exception subclass shares (96 bytes: 4 pointers + 4 sp_RbVal of the 8-field base prefix). The user class's own #set_backtrace stores there.

Three supporting changes were needed for the rescue path to actually work end-to-end:

* sp_StrArray_join and sp_StrArray_length tolerate NULL receivers. The legacy AOT backtrace returned NULL; the rescue path then segfaulted inside the join/length call. Both now degrade to empty.

* sp_raise_poly accepts user-class exception instances. Its first arm checked v.cls_id == SP_BUILTIN_EXCEPTION (the base id), so a boxed RequestTransportError (cls_id 46) was rejected with 'exception class/object expected'. The new arm additionally inspects v.v.p->parent_cls_name -- non-NULL means it inherits Exception and should be re-raised as itself.

* class_is_exc_subclass carve-out in the sp_box_nullable_obj_dyn decision. The _dyn variant reads the first sizeof(sp_int) bytes of the object as its cls_id; user exception subclasses start with cls_name (a pointer), so the read returned garbage and the poly dispatch's switch missed every case. Exception subclasses now use the static cls_id (the .new site stamps the requested class, so the static type always matches the runtime type).

The roundtrip test raises a StandardError, attaches a 3-element backtrace via set_backtrace, reads it back via backtrace, and checks the array contents and join output. It passes in CRuby and AOT.

## Summary by CodeRabbit

- **New Features**
  - Added support for reading and assigning exception backtraces.
  - Backtraces can be replaced and retrieved as arrays, including for custom exception types.
  - Exceptions can now be re-raised correctly across custom exception classes.

- **Bug Fixes**
  - Prevented crashes when querying or joining null string arrays.
  - Improved safe handling of exception objects and their backtrace data.
  - Non-exception objects passed to `raise` now produce the expected `TypeError`.
  - Exceptions without backtraces now receive an empty backtrace array.


matz:
Not merging this yet: it crashes an ordinary shape that works on master.

A user exception subclass whose first ivar is not the backtrace makes #backtrace read that ivar as an sp_StrArray * and dereference it. Eleven lines:

    class MyErr < StandardError
      def initialize(msg, code)
        super(msg)
        @code = code
      end
    end
    begin
      raise MyErr.new("bad", 42)
    rescue => e
      p e.backtrace.class
    end

CRuby and current master both print Array. With this branch the binary dumps core: the rescue variable is typed sp_Exception *, so the arm takes its user-subclass path and reads `*(sp_StrArray **)((char *)e + offsetof(sp_Exception, has_recv))`, which here holds the boxed 42.

The offsetof spelling is better than a literal 96, and I agree with the reasoning that every user exception subclass shares the base prefix. What does not hold is the step after it: the slot at that offset is the subclass's FIRST ivar, whichever ivar that happens to be, and only a class whose first ivar is the one its own #set_backtrace writes reads back a backtrace there. Your test passes because its subclass has no other ivar. An exception carrying an attribute is the common case, so this needs a way to know the slot is a backtrace before reading it: a discriminator in the base prefix that the setter writes, a per-class ivar index the codegen looks up (the concrete class is known where the call site has it, and the rescue variable's is exactly what is lost), or keeping the storage on the base struct for every subclass rather than in a user ivar.

The other three changes look independently right to me and I would happily take them on their own: the NULL-tolerant sp_StrArray_join/_length, sp_raise_poly accepting a user-class instance via parent_cls_name, and the static-cls_id carve-out for exception subclasses in the boxing decision. If you want to split those out, I will merge them while the backtrace storage question is settled.

Two smaller notes while you are in here:

- The gate calls comp_method_in_chain(c, ty_object_class(rrt), ...) unconditionally, and rrt can be TY_EXCEPTION, for which ty_object_class is a large negative number. It happens to be harmless because every loop that consumes a class id guards on cid >= 0, but it reads as an out-of-bounds index and the next person to touch those loops would be right to assume it cannot be negative.
- Worth running the generated-C sweep over the suite for the boxing carve-out specifically: sp_box_nullable_obj_dyn is reached from more than the exception path, and a change in which boxing form is chosen is exactly the kind that shows up as a byte diff somewhere unrelated.

matz:
Thank you for the six follow-ups: my eleven-line repro passes on the branch now, and mirroring the base field into the subclass struct is the shape I was hoping for.

The branch is red though, on all three legs, and it is not the offset question. 16 exception tests regress: bang_exception_object, begin_rescue_array_body, case_in_begin_value_position, exception_cause, exception_cause_ensure_reraise, exception_introspection_accessors, exception_real_objects, exception_value_flow_match, exception_value_parity, format_flags_mro_errinfo, format_raise_dollarbang_symproc, issue_2990, issue_3060, issue_3079, signal_module_surface, systemcallerror_hierarchy. (Tests: 3374 pass, 16 fail on ubuntu/gcc; I reproduced them locally on the branch too.)

`$!` stops being the rescued object. Seven lines:

    begin
      raise "boom"
    rescue => e
      p $!.equal?(e)
      p $!.class
      p $!.message
    end

CRuby and master: `true / RuntimeError / "boom"`. The branch: `false / NilClass / ""`.

`rescue => e` still binds correctly (`e.message` and `e.class` are right on their own), so it looks like the $! slot rather than the exception. My guess is the last commit, "rescue handler: backfill empty backtrace on freshly-caught exception" -- if the backfill rebuilds or replaces the caught object rather than writing into it, $! would point at the old one and `equal?` would be false. Worth checking whether the backfill can write through the object the handler already published instead.

Everything else about the update reads well to me, so this is the one thing between the branch and a merge.

matz:
An observation about the branch's CI history rather than about any one failure, because the pattern is the useful part.

Seven runs so far, six red:

| commit | result | what broke |
|---|---|---|
| 9d2631d2 | fail | raise_non_exception_typeerror |
| 8195e945 | fail | exc_stack_window_cleared |
| 6432c17a | fail | exc_stack_window_cleared |
| 0732e994 | **pass** | |
| 28d405eb | fail | exception_subclass_accessors |
| 4f39aebc | fail | exception_backtrace_empty |
| 9fee20b3 | fail | 16 tests, $! empty |

Every one of those is an EXISTING test, not one this branch adds, and until the last commit each run broke exactly one -- a different one each time. That is the signature of fixing forward against CI: each push repairs the previous failure and lands on the next, and the suite finds them one per round trip because that is the resolution CI gives you.

The suite will tell you all of them at once. `make test` is about 75 seconds here and needs no network:

    make test          # the full suite
    make gate          # test + bench + optcarrot + rubyspec, what I run before pushing

Running that locally before each push turns six round trips into one, and it is the same suite CI runs, so a green local gate is a green CI.

The exception area is unusually well covered -- 16 tests noticed the $! change, which is why the last push looks so much worse than the ones before it even though it is probably a smaller mistake. That coverage is an asset when you run it, and a slow oracle when CI runs it for you.

Nothing about the change's direction has changed my view: mirroring the base field into the subclass struct is right, and my eleven-line repro passes on the branch. It is the $! regression and the loop it arrived through.

rofl0r:
thanks for your review. the agent is currently churning away trying to find the root cause. i actually had my agent instructed to never call make test in the spinel repo, because at least in my musl-based environment it takes so much time that my impression was that it gets stuck in an endless loop.
```
FFFFFFFFFFFFFFFFFext-cruby-test: pass
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFrbs-seed-test: pass
FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF<- minutes and minutes adding more F's here
```
hopefully this PR will soon be in upstreamable shape. thanks for the patience.

rofl0r:
@coderabbitai review

matz:
Merged. Both of my reproducers pass: the attribute-carrying subclass answers Array where it used to dump core, and $! is the object the handler caught again (true / RuntimeError / "boom"). The 16 exception tests are green, and so is the full gate, with your two new tests under SPINEL_GC_STRESS=1.

Thank you for staying with it through nine commits. The two things that made the difference are worth naming, because they were both structural rather than incremental: mirroring the base field into the subclass struct, instead of reading whichever ivar happened to sit at a known offset, and publishing the rescue variable so $! and the handler's binding are the same object rather than two things kept in step.

The rest of the change carries its weight independently -- the NULL-tolerant join/length, sp_raise_poly accepting a user-class instance, the static-cls_id boxing carve-out. Any of those would have been worth taking alone.

One thing for next time, said once and then dropped: the suite is about 75 seconds locally and it is the same suite CI runs. Six of the nine pushes here were red, each on a different existing test, and `make test` before pushing would have shown you all of them at once instead of one per round trip. That is the only note I have; the work itself landed well.

PR #4255, opened 2026-09-01
packages/openssl: fix compatibility with libressl 3.1.x

rofl0r:
everything builds, except the new *_AEAD_* macros, which are just alias for the corresponding *_GCM_* macros in newer releases. work around those, and lower the bar to 3.1 as minimal supported.

## Summary by CodeRabbit

* **Compatibility**
  * Added support for LibreSSL versions 3.1 and newer.
  * Improved compatibility with LibreSSL releases that use legacy GCM controls for authenticated encryption operations.

### DIFF
--- a/packages/openssl/sp_openssl.c
+++ b/packages/openssl/sp_openssl.c
@@ -38,8 +38,8 @@
    #error says so rather than leaving a page of diagnostics about missing
    declarations. */
 #if defined(LIBRESSL_VERSION_NUMBER)
-# if LIBRESSL_VERSION_NUMBER < 0x3050000fL
-#  error "sp_openssl.c needs LibreSSL 3.5 or newer"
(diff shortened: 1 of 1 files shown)


matz:
Merged, and thank you for running the experiment rather than taking my floor at its word. Your report is the evidence: everything builds on 3.1.5 except the AEAD control macros, and those are aliases of the GCM ones for GCM, which e16f2165 had already added. So the 3.5 floor was more conservative than the file needs and 3.1 is the honest number.

Worth saying what the two changes do together, since they arrived from opposite directions. Yours lowers the floor so the package builds on your machine. Mine makes the probe compile the package's own source, so on a machine where it genuinely cannot build, the package drops out and `make` still succeeds instead of stopping at a #error. Neither replaces the other: the floor should be as low as it truthfully can be, and the probe should be the thing that decides, not a version number that ages.

PR #4261, opened 2026-09-01
Give the boxed division family its Bignum arms

blacktm:
## Why

Big integers are ordinary Ruby: `(10 ** 25).div(3)` is a number a program computes without thinking about machine words. In Spinel that was true only while the compiler could see the types — a Bignum that reaches `div`, `/`, `%`, `modulo`, `remainder` or `divmod` *boxed* (out of a poly container, an untyped accessor, a block parameter) was converted with the 64-bit truncating helper, so the quotient came out seven orders of magnitude wrong and the remainder came out 0, silently.

Concretely:

```ruby
big = [10 ** 25, 3]
p big[0].div(big[1])   # spinel: 530299326119804928   CRuby: 3333333333333333333333333
p big[0] % big[1]      # spinel: 0                    CRuby: 1
p big[0].divmod(big[1])# spinel: [530299326119804928, 0]
```

The typed receiver paths build these divisions inline per receiver kind and were already exact; only a receiver or operand arriving boxed took the truncation.

## What

The five boxed helpers behind `/`, `%`/`modulo`, `div`, `remainder` and `divmod` gain a Bignum arm: when either side is a Bignum, the pair goes through the existing bigint helpers instead of falling through to the machine-word division. `sp_bigint_div` floors toward −∞, which is what `Integer#div` and the operators do; `remainder` keeps the receiver's sign through `sp_bigint_remainder`. Zero divisors raise as they did, and a Float operand keeps the float path unchanged.

## How

One rule, five insertion points in `lib/spinel_rt.h`, each a single guarded arm placed above the final truncating fallback, plus three forward declarations the header needed. No compiler change.

## Validation

Tests 3396 pass, 0 fail, 0 error. Benchmarks 61 pass. Optcarrot OK on its usual checksum. ext-cruby, rbs-seed and the rubyspec gate pass; the macOS-only spin-e2e leg fails identically on untouched master. Generated C is byte-identical for all 3412 test and benchmark programs plus optcarrot — the change is runtime-only, so the new test compiles to the same C on master and diverges purely at run time. Inference fixpoint converges on every suite program, unchanged. The new test matches under SPINEL_GC_STRESS=1 and under ASAN+UBSAN with no diagnostics, and runs in well under a second. On the differential corpus of ~2,100 minimized programs that behave differently under CRuby 4.0.6 and Spinel (the same corpus behind #3999/#4003/#4029): one finding flipped to matching, none lost.

The new test is CRuby-generated and covers both operand sides, the four sign combinations, a quotient that fits back in 64 bits, and the Float path; every divergent line fails on master.

While validating this under GC stress we hit a pre-existing crash on master — a begin/rescue after a run of boxed Bignum operations corrupts the catch-side exception object under SPINEL_GC_STRESS=1, with or without this change (the identical binary linked against the untouched master runtime crashes the same way). The test therefore pins values only; the zero-divisor raises are unchanged by this rule and are noted, not pinned. Happy to share the reproducer.

## Left alone, on purpose

- **`Integer#quo` at a Bignum receiver.** Its exact answer is a BigRational, and that path loses the value under a collector for reasons outside this rule; it keeps today's truncating answer rather than gaining an unstable one.
- **A Float operand beside a Bignum** keeps the float path, including CRuby's own float rounding differences for `div` at that magnitude (`(10**25).div(2.5)` answers a Float-rounded integer in CRuby; Spinel's float path prints a Float — pre-existing, untouched).

Found preparing the numeric coerce protocol change (the follow-up PR), which routes a coerced Bignum operation through these same helpers; the corpus behind #3999/#4003/#4029 measures both.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed division, modulo, remainder, and `divmod` operations involving very large integers so values are calculated without truncation.
  * Preserved correct behavior for positive and negative large-number operands, including mixed integer types and boundary-sized results.

* **Tests**
  * Added regression coverage for boxed large-integer arithmetic and floating-point operand handling.


matz:
Merged. I reproduced your byte-identical claim (all test and benchmark programs, zero differing), checked the answers against CRuby directly including the sign combinations and a quotient that fits back in 64 bits, and ran the new test under SPINEL_GC_STRESS=1. Full gate green.

One arm per helper, above the truncating fallback, is the right shape for this: the rule is the same in all five places and reads the same in all five. Naming which one floors and which one keeps the receiver's sign in the comments matters more than usual here, because that is the part a reader cannot check locally.

On the GC-stress crash you hit while validating: I could not reproduce it from the shape you describe (a run of boxed Bignum operations followed by begin/rescue, twenty rounds, under stress) on either side of the merge. Please do file it with the reproducer -- if it needs a particular allocation rhythm to land, the exact program is the whole thing. Pinning values only and saying why is the right call in the meantime.

Leaving quo alone because its exact answer is a BigRational and that path is unstable under a collector is a good boundary. Worth its own issue eventually, since the truncating answer it keeps is wrong in the same way this PR just fixed.

PR #4262, opened 2026-09-01
A timed readiness wait parks the green thread, not its OS worker

rubys:
A timed readiness wait -- `IO#wait_readable(t)`, `IO#wait_writable(t)`, `IO.select([io], nil, nil, t)` -- ran on a blocking `select(2)` in the threaded runtime, so it held the OS worker for the whole timeout. Once every worker is pinned by a waiter, the thread that would make the fd ready cannot run, and each worker then serves its waiters one full timeout at a time.

**Repro** (fails on `62a5510f`):

```ruby
r, w = IO.pipe
n = 300
seen = Queue.new
threads = Array.new(n) do
  Thread.new do
    ready = r.wait_readable(3)
    seen << (ready.nil? ? 0 : 1)
  end
end
sleep 0.2
t0 = Process.clock_gettime(Process::CLOCK_MONOTONIC)
w.write "x"
w.flush
threads.each(&:join)
total = 0
n.times { total += seen.pop }
elapsed = Process.clock_gettime(Process::CLOCK_MONOTONIC) - t0
puts "#{total} of #{n} readable"
puts(elapsed < 1.0 ? "woke promptly" : "woke late")
```

```
$ ruby wait.rb
300 of 300 readable
woke promptly
$ spinel wait.rb -o wait && time ./wait
0 of 300 readable
woke late
./wait  0.02s user 0.17s system 0% cpu 57.313 total      # 16 workers: 300 waiters / 16 x 3s
$ SPINEL_WORKERS=1 ./wait                                # 300 x 3s -- I stopped it
```

**The change.** `sp_sched_wait_io_timeout(fd, events, seconds)` is `sp_sched_wait_io` with a deadline: the waiter carries a `wake_deadline` the way a sleeper does, and the monitor folds it into its poll timeout and wakes the thread with no revents once the clock passes it. `sp_io_wait_events` and the single-fd `IO.select` route through it under `SP_THREADS`. Priority (`exceptfds`) and multi-fd `IO.select` keep `select(2)` -- poll's `POLLPRI` is not the same thing, and the monitor parks a thread on one fd -- and so does a zero timeout, which is a peek. The open-ended park now zeroes the deadline so a stale one from an earlier `Kernel#sleep` cannot be read as this wait's.

After: 300 of 300 readable, woke promptly, 0.6s wall -- on one worker and on sixteen. `scripts/tsan-run.sh` on the new test is clean. `make test`: 3396 pass / 0 fail / 0 error. `docs/thread.md` and the M:N design notes say what is and is not scheduler-aware now.

One thing the repro surfaced that this PR does not touch: `IO.pipe`'s write end answers `sync == false` here where CRuby's is `true`, so `w.write "x"` sits in stdio until a flush (hence the `w.flush` above). Filed as #4263 with a threadless repro.

Context: this is the primitive a thread-per-connection server needs for keep-alive and idle timeouts; found while moving roundhouse's tep server from its own fiber scheduler onto `Thread`.

## Summary by CodeRabbit

* **New Features**
  * Timed single-IO readiness waits now pause without occupying worker threads.
  * Read and write waits wake promptly when the IO is ready or the timeout expires.
  * One-IO `IO.select` calls support the same efficient timed waiting behavior.

* **Bug Fixes**
  * Improved timeout handling ensures expired waits return correctly.

* **Documentation**
  * Updated documentation to explain timed IO waiting behavior.

### DIFF
--- a/docs/internals/thread-mn-design.md
+++ b/docs/internals/thread-mn-design.md
@@ -90,7 +90,11 @@ A dedicated monitor thread ("sysmon"), off the workers' backs, has three jobs:
 3. **Scheduler-aware I/O (design 8).** `sp_sched_wait_io(fd, events)` parks a
    green thread on an fd (POLLIN/POLLOUT), freeing its worker; the monitor
    rebuilds a `poll()` set (`g_pfds`) from the I/O-waiter list each tick and
-   wakes the thread when the fd is ready. A self-pipe (`g_sysmon_pipe`) wakes the
+   wakes the thread when the fd is ready. `sp_sched_wait_io_timeout(fd, events,
+   seconds)` is the same park with a `wake_deadline`: the monitor folds it into
+   its poll timeout like a sleeper's and wakes the thread with no revents when
+   the clock passes it, which is how `IO#wait_readable(t)` and a one-io
+   `IO.select` wait without pinning a worker. A self-pipe (`g_sysmon_pipe`) wakes the
    monitor out of `poll()` when the wait set changes.
 
 The monitor idles on `g_sysmon_cv` when there is nothing to watch and is woken
--- a/docs/thread.md
+++ b/docs/thread.md
@@ -58,7 +58,11 @@ is a user guarantee.
 `Kernel#sleep` and blocking I/O are **scheduler-aware**: a sleeping or
 I/O-blocked thread frees its OS worker for other green threads instead of
 holding it idle, and the monitor wakes it when the deadline passes or the fd
-becomes ready.
+becomes ready. A *timed* readiness wait on one IO -- `IO#wait_readable(t)`,
+`IO#wait_writable(t)`, `IO.select([io], nil, nil, t)` -- parks the same way,
+and wakes on whichever of the fd or the deadline comes first. `IO.select`
+over several IOs still runs on `select(2)` and holds its worker for the
+duration.
 
 ### Synchronization primitives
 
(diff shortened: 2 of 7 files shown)


matz:
Merged. Your 300-thread reproducer answers what CRuby answers -- 300 of 300 readable, woke promptly -- and the new test passes on repeated runs alongside the full gate.

Registering a deadline with the monitor rather than holding the worker is the right shape, and reusing the sleeper machinery for it (the poll set pulls the nearest deadline in, the monitor wakes the waiter with no revents once the clock passes) means the timed wait is not a second mechanism to keep in step with the first. Setting wake_deadline explicitly to 0 rather than leaving a stale one from an earlier Kernel#sleep is the kind of detail that decides whether this is correct or merely usually correct.

The single-threaded fallback keeping its own bounded poll loop is right too: it has no monitor to register with, and the old unbounded 1000ms loop would have ignored the timeout entirely.

#4263 landed first (0d9870cd), so your premise holds on master now: IO.pipe's write end is sync and a byte written to it is visible to the reader without a flush. Both spellings of the call, since the destructured form is emitted inline and never reaches sp_io_pipe.

matz:
Merged, including your two follow-ups. Sorry for the round trip: I merged your branch, and by the time it landed you had pushed the rebase, so GitHub kept the PR open against a head I had not taken. Both improvements are in now.

The deterministic wait is the better of the two. A fixed sleep did not prove all 300 threads had reached their timed wait, and a late starter would have found the pipe already readable and passed without ever parking -- so the test could have gone green on the bug it exists to catch. Announcing before parking and writing after the 300th arrival removes that.

Your 300-thread reproducer answers what CRuby answers, the test passes on repeated runs, and the full gate is green.

PR #4268, opened 2026-09-02
Keep an ensure's deferred-return slot for object-returning methods

ryanseys:
14474695 made the slot guard ask c_type_name whether the return kind can be declared, but c_type_name names only the builtin kinds: object kinds are declared by emit_ctype through their class. Every object-returning method with an ensure therefore lost its slot, and the ensure's exit became a bare return; out of a non-void C function, which does not build. The guard now admits object kinds alongside c_type_name's, and a by-value class's slot defaults to its zero struct rather than default_value's NULL.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue affecting methods that return object values when they contain `ensure` blocks.
  - Object return values are now preserved correctly when `ensure` logic exits early or is inlined.
  - Improved handling for both value-based objects and heap-allocated objects, preventing invalid generated code and compilation failures.


matz:
Merged. Thank you for catching my regression, and for the exact diagnosis: c_type_name names only the builtin kinds, so asking it alone was the wrong question for a kind emit_ctype declares through its class. Both ensure tests pass and the full gate is green.

There is a second shape the same guard does not yet cover, which I hit while checking your fix and which I could not close in this pass. A method returning a BY-VALUE object class through an ensure still fails the build:

    class Box
      attr_reader :v
      def initialize(v); @v = v; end
    end
    def make(v)
      begin
        Box.new(v)
      ensure
        $done = true
      end
    end
    p make(3).v

CRuby prints 3; master (with your fix) fails with `error: invalid initializer` on `sp_Box _t1 = NULL;`. Your ensure-slot arm gets this right -- it emits `(sp_Box){0}` for a value class -- but a slot emitted EARLIER in the same function, before the ensure machinery, still takes default_value's NULL. That is the same mismatch one layer out: emit_ctype declares a value class as a bare struct and default_value cannot know that, since it sees only the TyKind and the value-ness lives on the class.

I did not find the emitting site in the time I had, so I am not guessing at a patch. Filing it as its own issue rather than holding your merge; if you would rather take it, say so and it is yours.

matz:
The by-value half is fixed in 3428632f, so that follow-up is closed out. It was the begin's own result slot, one slot before the ensure machinery: is_scalar_ret admits a by-value object class, emit_ctype declares it as a bare struct, and default_value answered NULL for it -- the same mismatch your fix resolved for the deferred-return slot, at the site that runs first. Thanks again for the diagnosis that made both halves findable.

Issue #4272, opened 2026-09-02
A boxed proc call's channel (_sp_proc_poly_ret/_args) is a per-worker root the barrier does not publish: SIGSEGV in sp_StrArray_scan on N>1 workers


rubys:
Filed separately from #4266 as asked. It is a GC root gap, not a marking race, and the reduction and a fix are on branches at rubys/spinel (PRs follow).

**Symptom.** A threaded HTTP server (roundhouse's campfire binary, thread per connection) on a 16-core Linux box with the autodetected worker count: a browser's parallel load of one page kills it in about one run out of two -- `SIGSEGV in sp_StrArray_scan` under `sp_gc_mark_drain`, every other worker parked at the barrier inside its own `sp_str_alloc`; one run instead lost a value quietly. `SPINEL_WORKERS=1`: 9 of 9 green. macOS, same binary: 4 of 4 green (fewer collections landing in the window, presumably).

**How it was found.** `SPINEL_GC_VERIFY=1 SPINEL_GC_STRESS=1 SPINEL_WORKERS=4` made it fail on the first page load, every time:

```
  [phase=globals ctx=...]
*** SPINEL_GC_VERIFY: collector reached a non-heap/corrupt object ***
  obj    = 0x7fffd0044fa0     (a worker arena block, freed)
#5 sp_gc_mark  #6 sp_mark_user_globals  #7 sp_gc_mark_all  #8 sp_gc_collect ...
```

In the `globals` phase `ctx` is not updated per slot, so the slot came from the return address into `sp_mark_user_globals`: the call in flight was `sp_mark_rbval` on two words loaded from `%fs:` -- the marker's last line, `sp_mark_rbval(_sp_proc_poly_ret)`. Twice out of two faults.

**Mechanism.** `_sp_proc_poly_args` / `_sp_proc_poly_ret` are per-worker TLS, defined in the generated TU. The globals hook marks them every cycle -- the collecting worker's own copies -- so on one worker whatever a boxed call last left there stays alive until the next boxed call, which is the contract the generated code runs under. On N workers the other copies are reachable from nowhere: `sp_publish_worker_roots` pushes the pending exceptions, the proc-return homes and the break values at the barrier, not the channel. A collection on another worker frees the object a stale slot names; the owning worker's next collection marks a freed pointer. Verified, that is the abort above; unverified, whatever reused the block faults in its scan hook, which is the `sp_StrArray_scan` from #4266.

[...]

Issue #4266, opened 2026-09-02
A program-level cap on the worker pool, set before the first Thread.new?


rubys:
A question about the worker pool, not a bug: is there, or should there be, a way for a program to cap its own OS worker count before the first `Thread.new`?

**Where this comes from.** roundhouse's tep server (the campfire binary) is moving from its own fiber scheduler onto `Thread` -- thread-per-connection, with #4262's timed wait doing the idle timeouts. The app is single-writer SQLite, so its useful parallelism is bounded by the database rather than by cores, and the migration wants to land in two steps: threads first with one worker (the semantics the fiber scheduler had, every I/O parking instead of blocking), then real parallelism once its shared state is audited.

**What exists.** `SPINEL_WORKERS` is read in `sp_sched_start_workers` from `main()`'s prologue, before any Ruby runs, so `ENV["SPINEL_WORKERS"] = "1"` in the program is too late. Helpers are spawned lazily, though -- `sp_sched_maybe_grow` reads `g_worker_cap` when a thread becomes runnable -- so a setter that lowered the cap before the first `Thread.new` would take effect exactly as the env var does.

**What I'd use**, spelling yours to pick: something like `Thread.workers = 1` (or a `SPINEL_WORKERS` default the program can declare, with the env var still winning so an operator keeps the last word). Documented on the same row of `docs/thread.md` as the env var.

If the answer is "set the env var in the launcher", that works too; the binary's README would say so. I'd rather ask than add a wrapper script around a question spinel might prefer to answer itself.

rubys:
Evidence that the cap matters now, in case it helps decide the shape.

The campfire binary is on `Thread` since yesterday (a green thread per connection, every wait a park). With the autodetected worker count on a 16-core Linux box, a browser's parallel load of the signed-in room page took it down in 4 of 7 runs; with `SPINEL_WORKERS=1`, 9 of 9 green. gdb on the crashes, both identical:

```
Thread 9 received signal SIGSEGV
#0 sp_StrArray_scan
#1 sp_gc_mark_drain
#2 sp_gc_collect
#3 sp_gc_collect_retune_all
#4 sp_stw_collect_impl.part
#5 sp_str_alloc
#6 sp_str_sub_range
#7 sp_Threaded_s_handle_one
#8 _fiber_body_1
#9 sp_fiber_trampoline
```

[...]

PR #4273, opened 2026-09-02
The per-class pool recycle pushes with a CAS: the sweep that calls it is parallel

rubys:
`SP_POOL_DEFINE`'s recycle carries the comment "runs only from the sweep (stop-the-world in the threaded build), so the list mutation needs no CAS -- no pop can run concurrently". Both facts still hold; the conclusion stopped following when the sweep became parallel: `sp_sched_par_sweep` has every parked worker sweep its own slot, so several workers call `sp_<Class>_pool_recycle` at once and push onto one free list with plain stores. `sp_PolyArray_pool_recycle` already pushes with a CAS loop under `SP_THREADS` for exactly this reason; this makes the class pools do the same and count atomically. Single-threaded expansion unchanged.

**Reduction** (`test/pool_recycle_parallel_sweep.rb`, in the PR): eight workers hold 5,000 pooled instances across a collection and drop them together, forty times each. Under `scripts/tsan-run.sh` with `SPINEL_WORKERS=8`:

| | TSan reports | of which `sp_Node_pool_recycle` |
|---|---|---|
| master 3428632f | 106 (macOS) / 173 (Linux) | 102 / 165 |
| this branch | 4 | 0 |

The four that remain are the pop side's documented stale-`next` read (`sp_pool_try_pop` vs the header writes after a successful pop), the pattern the comment above `sp_pool_try_pop` already argues is benign. Plain `make test` runs the program for its deterministic total.

**Where it came from.** roundhouse's campfire binary moved to `Thread` this week; with the autodetected worker count on a 16-core Linux box a browser's parallel load of the room page killed it in 4 of 7 runs (SIGSEGV in `sp_StrArray_scan` under `sp_gc_mark_drain`; one run instead lost a cable subscribe with the server alive), 9 of 9 green on one worker. A TSan build of that binary (your `tsan-run.sh` recipe plus the package objects) reported 1,579 races and every hot site was a synthesized `pool_recycle`. This change removes them, but it is not the crash: the binary still died on the patched runtime, and `SPINEL_GC_VERIFY` then pointed at an unpublished per-worker root instead -- that is #4272. This one stands on its own: a lost push strands storage and the counters tear; nothing dangles.

`make test` on the branch: 3404 pass, 0 fail (3403 + the new one). Related: #4266 (the pin we ship meanwhile), #4272 (the crash).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when pooled objects are recycled during parallel garbage-collection sweeps.
  * Prevented pooled objects from being lost or pool statistics from becoming inconsistent under concurrent activity.

* **Tests**
  * Added coverage for concurrent pool recycling across multiple threads.
  * Added expected output validation for the parallel sweep test.


rubys:
Both review points taken in 803aa2ff.

The cap: it was a load beside an add, so every concurrent sweeper could see room and all of them push, overshooting `pool_max` by up to their number. The threaded arm now adds first and gives the slot back over the cap. Probe: `SP_POOL_MAX=100 SP_POOL_REPORT=1 SPINEL_WORKERS=8` on the reduction reports `hwm=100 retained=0` exactly. Same check-then-add shape in `sp_PolyArray_pool_recycle` (yours), left alone here since this PR is the class pools; happy to do it as a follow-up if you want the same bound there.

The block comment above the macros now says the sweep is parallel and why neither direction of the Treiber stack has an ABA window, so nobody reintroduces the plain push from the old sentence. Single-threaded arm unchanged; `make test` 3405/0; the reduction under TSan still reports no recycle-side race (the four pop-side stale-`next` reads remain, as documented).

matz:
Merged as 959fa6e8. The reasoning in the comment held individually and the conclusion stopped following when sp_sched_par_sweep gave every parked worker its own slot, which is exactly the shape sp_PolyArray_pool_recycle already guards with a CAS loop. test/pool_recycle_parallel_sweep.rb passes 3/3 under SPINEL_WORKERS=8 here, and make gate is green (3407 pass, 0 fail).

Thank you for the TSan numbers; 106/173 down to 4 with none in sp_Node_pool_recycle is the kind of before/after that makes a memory-ordering claim checkable.

PR #4274, opened 2026-09-02
Publish the boxed proc channel at the barrier: another worker's copy is a root too

### LINKED ISSUE #4272: A boxed proc call's channel (_sp_proc_poly_ret/_args) is a per-worker root the barrier does not publish: SIGSEGV in sp_StrArray_scan on N>1 workers

rubys:
A proc called through an untyped slot passes its arguments and result through `_sp_proc_poly_args` / `_sp_proc_poly_ret`: per-worker TLS, defined in the generated TU, marked every cycle by the globals hook -- the collecting worker's own copies. On one worker that is the contract the generated code runs under: whatever a boxed call last left in the slots stays alive until the next boxed call. On N workers the other copies were reachable from nowhere. `sp_publish_worker_roots` pushed the pending exceptions, the proc-return homes and the break values at the barrier, not the channel; a collection on another worker freed the object a stale slot named, and the owning worker's next collection marked a freed pointer.

Under `SPINEL_GC_VERIFY=1` it reports `collector reached a non-heap/corrupt object`, `phase=globals`, from `sp_mark_user_globals`' `sp_mark_rbval(_sp_proc_poly_ret)` -- the disassembly at the return address was the give-away: the two words came from `%fs:`. Unverified, whatever reused the block faults in its scan hook: this is the `SIGSEGV in sp_StrArray_scan under sp_gc_mark_drain` from #4266, in about one browser run out of two on a 16-core box and never on one worker.

The fix pushes the seventeen slots as `sp_RbVal` roots in the publish hook, the way it already pushes the proc-return homes.

**Reduction** (`test/poly_channel_root_survives_other_workers_gc.rb`, in the PR): one worker makes a single boxed call whose Array result nothing else names, idles through six other workers' collections, then allocates.

| | `SPINEL_WORKERS=8 SPINEL_GC_VERIFY=1` | `SPINEL_WORKERS=1` |
|---|---|---|
| master 3428632f | aborts, 4 of 4 | passes |
| this branch | passes, 4 of 4 | passes |

The result is an Array on purpose: a String would go through `sp_mark_string`, which does not consult the registry, so the verifier would not see it (and the unverified fault would be a silent one-byte read). Plain `make test` runs it for the deterministic line.

`make test` on the branch: 3405 pass, 0 fail. The full-application check, the campfire binary rebuilt with this and #4273 on the 16-core box, autodetected workers, the same browser loop that killed the unpatched one in 4 of 7 runs: 0 faults in 5 runs, under `SPINEL_GC_VERIFY=1` and plain. (Two of those five stalled on a cable subscribe with the server alive and no verifier report -- a separate symptom, seen on the unpatched runtime too, being traced now; it is not this fault.)

Fixes #4272. Found alongside #4273 (the parallel-sweep pool race TSan reported first); that one loses nodes and tears counters but cannot dangle, and it was fixing it and still crashing that pointed here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented garbage collection from prematurely freeing objects used by boxed procedure calls across worker threads.

* **Tests**
  * Added regression coverage for cross-worker garbage collection during boxed procedure calls.
  * Verified successful completion with the expected `ok 20000` result.


matz:
Merged as 464fe31f, and this closes #4272.

I reproduced the fault before merging: the branch's own test aborts 3/3 on master under SPINEL_WORKERS=8 SPINEL_GC_VERIFY=1, and passes 4/4 after the merge. make gate is green.

The diagnosis is the important part and it is right: the globals hook marks the collecting worker's copy of a thread-local, which says nothing about the other seventeen copies. Publishing them at the barrier is the same shape the proc-return homes already use, so the fix does not introduce a new rule, it applies the existing one to a surface that had been missed.

PR #4276, opened 2026-09-02
Convert a String comparison's operand through #to_str

blacktm:
## Why

This is the follow-up promised on #4265. Reviewing that PR, matz raised `"abc" < money`, and #4265 answered the loud half: an object operand that a String cannot be compared with is CRuby's `comparison of String with Money failed`, not a byte compare against the object's address. CodeRabbit asked the other half, which #4265 left open — an object that *does* answer `#to_str` is not a failed comparison at all. CRuby converts it and compares the strings.

Spinel asked nothing, at any of the comparison methods, and each one was wrong in its own way:

```ruby
class Wrap
  def initialize(s) = @s = s
  def to_str = @s
end

p("abc" <=> Wrap.new("abd"))        # CRuby -1
p("abc".casecmp(Wrap.new("ABD")))   # CRuby -1
p("abc" < Wrap.new("abd"))          # CRuby true
p("abc".between?(Wrap.new("aaa"), Wrap.new("abd")))  # CRuby true
```

`#casecmp` and `#casecmp?` discarded the operand and answered `nil` — folded at compile time, so the call was not even emitted. `#<=>` had no arm for a non-String operand at all: it fell through to the generic object dispatch and raised `undefined method '<=>' for an instance of String` at run time, on a method String certainly has. The four ordered operators and `#between?` raised #4265's comparison error. Through a boxed operand — an element of a mixed array, a poly local — `#<=>` and the ordered operators answered `nil` or raised the same comparison error, and `#casecmp` answered `nil`.

## What

An operand that is not a String but whose class answers `#to_str` is converted, and the strings are compared, at `#<=>`, `#casecmp`, `#casecmp?`, `<`, `<=`, `>`, `>=` and `#between?`. A boxed operand — one whose static type is only `poly` — asks the runtime's own `rb_check_string_type` wherever it lands, including as a `#between?` bound, so a slot holding a String or a converting object compares there too. An operand that answers no usable `#to_str` keeps exactly the answer it had, at every one of the eight. `#==` does not convert in CRuby (`"abc" == Wrap.new("abc")` is false) and does not here.

## How

No new conversion machinery. The conversion is the direct call to the compiled `#to_str` that `obj_conv_method` and `emit_obj_conv_call_inline` (src/codegen.c) make at the tree's other statically-typed String conversions — `sp_Wrap_to_str((sp_Wrap *)…)` — with one difference: it is rendered on a spilled temp rather than on the operand's own expression.

That spill is what the eight entry points now share, as one prologue (`emit_str_cmp_prologue`, src/codegen.c): the receiver bound to a rooted `const char *`, then the operand spilled into a rooted `sp_RbVal`, then the conversion, then the compare. Both spills are load-bearing, because `#to_str` is user code that allocates. The receiver of a comparison is often a fresh string held in nothing else — an interpolation, a `+` — and the operand's own *object* is reachable from nothing but the argument slot being converted, while its `#to_str` allocates before it reads `self`. The order is also Ruby's own: receiver, operand, then `#to_str`, once each. The conversion is deliberately not put in the call's conversion hold, which would hoist it in front of the receiver and out of `#between?`'s short circuit.

The conversion answers `NULL` for "no conversion", and that one answer is what collapses the arms. CRuby's `rb_check_string_type` treats a `nil` from `#to_str` as *not a string*, and the comparison then gives its own answer: `nil` from `#<=>`, `#casecmp` and `#casecmp?`, `comparison of String with X failed` from the ordered operators and `#between?`. So the refusal is the fallback of the arm that converts, not a separate arm in front of it. The ordered operators and `#between?` reach that fallback for a class with no `#to_str` at all as well, because there a failed comparison is exactly what CRuby answers; `#<=>` does not — its arm is entered only by a class that converts, so its `NULL` can only be a `#to_str` that answered nil, and a class with no usable `#to_str` never reaches it. That is what makes `#between?` exact: `"abc".between?(Wrap.new("abd"), Plain.new)` is `false` in CRuby, because the first `<=>` settles it and the second bound is never asked for anything, and the conversions and refusals now render inside the `&&` where each compare is. It is also what keeps a nil-valued object slot right: `"abc" < obj` where `obj` is `nil` boxes as nil, converts to `NULL`, and raises `comparison of String with nil failed`, as master does. A `#between?` bound whose static type is only `poly` is not refused for want of a type either: it asks `sp_poly_check_str`, the same runtime check a boxed operand asks anywhere else, so `"abc".between?(Wrap.new("aaa"), hi)` with `hi` a poly slot holding `"abz"` is `true`, one holding `5` is `comparison of String with 5 failed`, and one holding `nil` names `nil` — each the value the slot actually held, where master named the other bound.

A `#to_str` the analysis can pin no further than poly is checked at run time by a new `sp_str_cmp_conv` (lib/spinel_rt.h) rather than by the strict String argument slot: `nil` is `NULL` (no conversion), a String is itself, and any other class raises CRuby's own `can't convert Wrap to String (Wrap#to_str gives Integer)`.

A boxed operand converts in the runtime instead. Every ordered operator, `#between?` and `#<=>` reach `sp_poly_cmp` for a boxed pair — six entry points — so "a String against a user object that answers `#to_str` compares the strings" is written once there, using a new `sp_poly_check_str`: `rb_check_string_type`, where a String answers itself, a user object answers its conversion through the generated bridge, and everything else answers NULL *without* raising, because "not a string" is a real answer to a comparison where in an argument slot it would be a TypeError. `#casecmp` does not go through `sp_poly_cmp`, so its poly arm asks that helper directly. The operand is rooted inside `sp_poly_check_str_obj`, the one point both boxed callers pass through, for the same reason the typed side spills it.

Typing had to move with it. `#casecmp` and `#casecmp?` were typed `nil` for any statically non-String operand, so the comparison's answer had nowhere to land; both the typed-receiver rule (src/analyze_infer.c) and the boxed-receiver rule (src/analyze_infer_recv.c) now type a converting operand's call `poly` — not the Integer or boolean a String operand gives, because a `#to_str` that answers nil makes the call `nil`. The predicate those rules share with the emitter, `class_has_to_str_shape` (src/analyze_util.c), asks the *shape* of `#to_str` — no parameters, answering a String or poly — and not whether the conversion is emittable, because emittability depends on reachability and that is computed long after the type fixpoint that has to answer the same question. It is the same split `#coerce` already uses. The boxed side does not ask that predicate and cannot: it reaches `#to_str` through the generated conversion bridge, which carries only a `#to_str` whose static return is a String, so a poly-returning one converts typed and answers "no conversion" boxed. That asymmetry is named in a comment at both ends.

## Validation

- Tests 3409 pass, 0 fail, 0 error. Benchmarks 61 pass. Optcarrot OK 59662. ext-cruby, rbs-seed and the rubyspec gate pass; the macOS-only spin-e2e leg is red on this machine for the reason it was during #4261 and #4265, progress output leaking onto stdout, independent of this change. No suite program changes its inference-fixpoint behaviour: over 3363 programs there are 0 new non-convergences, the two that hit the round cap hit it on master too, and one program's round count moves by one (test/enumerator_block_returns_self.rb, 4 to 3 — the known noise on that file, which comes and goes between runs).
- Generated C is byte-identical for 3422 of the 3424 pre-existing test and benchmark programs, plus optcarrot. Both files that differ produce byte-identical output on master and on this branch, and both still match their expected output. test/string_slice_casecmp_succ.rb passes `casecmp` and `casecmp?` a poly argument through a helper — the arm this change rewrote. test/numeric_coerce_protocol.rb has three String comparisons against an object with no `#to_str` (`"abc" < money`, `"abc" >= money`, `"abc".between?(money, "b")`): their refusal is now the converting arm's fallback rather than a separate eager arm, which is the same raise from a different shape of C.
- The new test is CRuby-generated and matches plain, under `SPINEL_GC_STRESS=1`, under ASAN+UBSAN plain, and under ASAN+UBSAN with GC stress. It runs in a few milliseconds. It does not compile on master at all: the compiler there refuses `("aaaa" + "").casecmp(Churn.new(4)) == 0` as an unsupported equality, because the arm that discarded the operand had folded the call to `nil`.
- It pins the shapes a review would ask about: `#to_str` inherited from a superclass, an operand reached through a method return, a boxed operand at all eight methods — `#between?` through a boxed receiver, because a typed receiver with both bounds boxed does not compile, see below — a boxed receiver, a receiver picked by a ternary whose condition is itself a converting comparison, `#to_str` called exactly once per comparison, and `#between?` stopping at the first bound with the call counts to prove it. Two shapes of "no conversion" are pinned at all eight methods each — a `#to_str` typed String that answers the nil String, and a `#to_str` typed poly that answers `nil` — and a class with no `#to_str` at the seven where its answer is the comparison's own, `#<=>` excepted because there it keeps master's NoMethodError. A `#to_str` typed poly answering an Integer pins CRuby's TypeError message at `#<=>`, `<` and `#casecmp`. A boxed `#between?` bound is pinned holding a String (matching and not matching), a converting object, an Integer, `nil`, and a `#to_str` counter that is never asked when the typed bound already settled the answer.
- The guards are pinned too: `#==` and `#eql?` are still false against a converting operand, #4265's Symbol and Integer operands are unchanged at `#<=>` and at `<`, a class with no `#to_str` still answers `nil` from `#casecmp` and raises the comparison error from `<`, and an object-typed slot holding `nil` raises `comparison of String with nil failed`.
- Both roots were verified by counterfactual under ASAN+UBSAN with GC stress and `SP_POOL_MAX=0`, since the per-class free list masks a collected-while-live object at the default cap: without the receiver root, a churn whose receiver is an interpolation is a heap-use-after-free; without the operand root, a 300-round loop comparing against a freshly popped boxed operand whose `#to_str` allocates before reading its ivar is one too. As emitted, that loop and the four typed churns are clean.
- On the differential corpus of 2,131 minimized programs that behave differently under CRuby 4.0.6 and Spinel (the same corpus behind #3999/#4003/#4029): 2 findings flipped to matching, none lost — the `casecmp` one and the `<=>` one.

## Left alone, on purpose

Recorded so nobody has to rediscover them. Each is master's behaviour; where this rule moved one of them, the bullet says so.

- **`"abc" <=> obj` for any class with no usable `#to_str`.** It still raises NoMethodError, as on master. CRuby has an answer there — `nil`, or `-(obj <=> "abc")` when the class defines `#<=>` (`rb_invcmp`, so with `def <=>(o) = -1` its `"abc" < obj` is `false`) — but that is `Object#<=>`'s rule, an answer rather than a conversion, and it needs its own change. Giving it here would also have silenced the shapes this rule deliberately declines, each of which CRuby answers differently again. A class defining both `#<=>` and a usable `#to_str` converts, as CRuby does — it asks `#to_str` first.
- **A `#to_str` reached by `alias`.** `alias to_str raw` compiles one function, named after `raw`, and reachability seeds the implicit protocols by the method's own name, so the aliased body is not emitted at all. Master already fails the C build on `"x" + A.new("y")` for exactly that reason; this rule declines the shape rather than adding a second way in. Worth its own change.
- **A `#to_str` that takes parameters, and `attr_reader :to_str`.** CRuby converts through `def to_str(x = 1)`, `def to_str(*a)`, `def to_str(k: 1)` and an `attr_reader`, since it only needs the method to be callable with no arguments. None of the four converts here. The three parameter shapes fail the C build at the tree's existing conversion site (`"x" + A.new("y")`) on master and on this branch alike, and an `attr_reader` defines no method scope for the lookup to find. At the comparisons they keep master's answers everywhere: NoMethodError from `#<=>`, the comparison error from the ordered operators and `#between?`, `nil` from `#casecmp`.
- **A `#to_str` whose static return is an Integer.** Not this shape, so it keeps master's answers (NoMethodError, `nil`, the comparison error) rather than CRuby's `can't convert Bad to String (Bad#to_str gives Integer)`. The poly-typed version of the same mistake does raise that, with CRuby's wording.
- **A boxed operand whose `#to_str` is poly-returning.** The conversion bridge the runtime reaches carries only methods whose static return is a String, so such an operand answers `nil` or the comparison error as it does on master, while the same object converts when its slot is typed. Widening the bridge is its own change.
- **A converting operand under a ternary receiver at `String#+`.** `((("abc" < "abd") ? "abc" : "zzz") + W.new("ABC")).length` is a heap-use-after-free on master under ASAN with GC stress and `SP_POOL_MAX=0`: `emit_operands_in_order` declines to spill an operand when an observable one is not a call node, and the fresh object then goes into `sp_W_to_str(sp_W_new(...))` unrooted. It belongs to the conversion machinery, not to any one method. The comparison arms in this change do not depend on that spill — they spill the operand themselves — so the same shape is clean at all eight of them.
- **`String#clamp` with object bounds.** CRuby raises `comparison of Wrap with Wrap failed` (it compares the two bounds first, and Wrap has no `#<=>`). Master fails the C build there — `sp_Wrap *` into a `const char *` — and so does this branch. A separate finding.
- **`String#between?` with a typed receiver and BOTH bounds boxed.** One boxed bound now converts, because the other bound being an object is what puts the call on the converting arm; with both bounds boxed there is no object to put it there, so the call still lowers to the old byte compare, passes an `sp_RbVal` into a `const char *`, and fails the C build exactly as on master. Through a boxed receiver the same comparison works, and the test pins that; the typed-receiver arm for two boxed bounds is its own fix. Two literal non-String bounds are the same pre-existing gap: `"abc".between?(1, 2)` and `"abc".between?(:a, :b)` are CRuby's `comparison of String with 1 failed` and `... with :a failed`, and fail the C build on master and here alike.
- **The other `#to_str` sites** — `Regexp#=~` and `#===`, `Thread#name=`, `IO#printf`'s format, `Kernel#String` on a boxed object, `ENV[]=`, `Array#join`'s elements. Each is its own rule and its own change; the String methods that already convert (`#start_with?`, `#include?`, `#+`) are untouched.

## Summary by CodeRabbit

- **Bug Fixes**
  - String comparisons now support operands that provide a valid `to_str` conversion.
  - Updated `<=>`, `casecmp`, `casecmp?`, ordered comparisons, and `between?` to handle convertible objects consistently.
  - Improved handling of objects returning `nil` or invalid values from `to_str`, including appropriate comparison results and error messages.
  - Preserved Ruby-compatible evaluation order and conversion behavior for comparison bounds.

- **Tests**
  - Added comprehensive coverage for string comparison conversions, failures, return values, error messages, and conversion call counts.

### DIFF
--- a/lib/spinel_rt.h
+++ b/lib/spinel_rt.h
@@ -2321,6 +2321,55 @@ static SP_INLINE const char *sp_poly_arg_str(sp_RbVal v) {
   if (v.tag == SP_TAG_STR) return v.v.s;
   return sp_poly_arg_str_slow(v);
 }
+/* The object half of the check below, split out and kept off the inlined path
+   for the reason sp_poly_arg_str_obj is: the object case is the rare one.
+
+   The value is rooted across the bridge call. #to_str is user Ruby code and
+   allocates, and an object reaching a boxed comparison is often held in
+   nothing but the by-value sp_RbVal the caller passed -- a copy of the
+   caller's, but the collector is non-moving and marks through whatever slot it
+   is handed, so rooting this copy keeps that object alive. Both callers rely
+   on this one root: sp_poly_cmp_to_str below, and the poly casecmp arm the
+   emitter renders. */
+static SP_NOINLINE const char *sp_poly_check_str_obj(sp_RbVal v) {
+  SP_GC_ROOT_RBVAL(v);
+  return sp_obj_to_str_fn((int)v.cls_id, v.v.p);
+}
+/* CRuby's rb_check_string_type, which the String COMPARISONS ask where the
+   argument slots above ask the raising forms: a String answers itself, a user
+   object whose #to_str the conversion bridge carries answers its conversion,
+   and every other value answers NULL without raising. "Not a string" is a real
+   answer here -- it is `"abc" <=> obj`'s nil and `"abc" < obj`'s comparison
+   error -- where in an argument slot it is a TypeError.
+
+   The bridge carries only a #to_str whose static return is a String
+   (conv_bridge_callee), so an object whose #to_str the analysis could pin no
+   further than poly answers NULL here and converts on the TYPED side alone.
+   That asymmetry is deliberate for now, not an oversight to tidy away: giving
+   the bridge a boxed-answer entry is its own change. */
+static SP_INLINE const char *sp_poly_check_str(sp_RbVal v) {
+  if (v.tag == SP_TAG_STR) return v.v.s;
+  if (v.tag == SP_TAG_OBJ && v.cls_id >= 0 && v.v.p && sp_obj_to_str_fn)
+    return sp_poly_check_str_obj(v);
+  return NULL;
+}
+/* The ANSWER of a #to_str the analysis could pin no further than poly, on its
+   way into a String comparison. CRuby's rb_check_string_type checks the answer
+   as well as the method: nil is "no conversion", which is the comparison's own
+   nil or its ArgumentError and which the NULL here hands back to the caller's
+   refusal arm; any other non-String is a TypeError naming both classes. `obj`
+   is the operand itself, for that message -- a class name, so it is read
+   twice. A String answer that is the NULL string is nil to Ruby, and takes the
+   no-conversion path too. */
+static SP_NOINLINE const char *sp_str_cmp_conv(sp_RbVal r, sp_RbVal obj) {
+  if (r.tag == SP_TAG_STR) return r.v.s;
+  if (r.tag == SP_TAG_NIL) return NULL;
+  sp_raise_cls("TypeError",
+               sp_sprintf("can't convert %s to String (%s#to_str gives %s)",
+                          sp_poly_class_name(obj), sp_poly_class_name(obj),
+                          sp_poly_class_name(r)));
+  return NULL;
+}
 
 /* The strict argument forms: nil / true / false in a typed String or Integer
    slot are CRuby's TypeError (File.join("a", nil), [1].take(nil)), where the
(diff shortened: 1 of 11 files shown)


matz:
Merged as 25ca248e.

I checked the claim the whole change rests on rather than the diff: emitting the C for every suite program on master and on this branch, 3421 of 3423 are byte-identical, and the two that move are the two you named. Both produce identical output on both sides and still match their expected files; test/string_slice_casecmp_succ moves because the casecmp arm is rewritten (and then every later temp renumbers), test/numeric_coerce_protocol because its three refusals are now the converting arm's fallback. make gate is green with the merge in place: 3410 pass, optcarrot 59662.

Against CRuby, the eight entry points agree on the converting operand, on the boxed operand, on `#==`/`#eql?` staying false, on a `#to_str` typed poly that answers nil at run time (`-1` then `nil`), and on `"abc".between?(Wrap.new("abd"), Plain.new)` being false with the second bound never asked.

One shape is left where it was, and it is worth naming so nobody reads the merge as covering it: a `#to_str` whose static return type is exactly nil, `def to_str = nil`. `class_has_to_str_shape` asks for a String or poly return, so that class is not a converting one and `#<=>` keeps master's NoMethodError where CRuby answers nil. That is the same answer master gave, not a regression, and the class is one whose `#to_str` can never produce a String, so I would not widen the shape predicate for it without a reason to.

The reasoning about why the conversion is not put in the call's conversion hold, and why the refusal is the converting arm's fallback rather than an arm in front of it, is the part that makes `#between?` exact. Thank you for writing that down.

PR #4280, opened 2026-09-02
add built-in tmpdir module

rofl0r:
my agent constantly writes ruby code using this module, then when i need to run it under spinel i need to refactor the code again. adding this as a built-in module finally closes this gap.

## Summary by CodeRabbit

* **New Features**
  * Added bundled temporary-directory support through `Dir.tmpdir` and `Dir.mktmpdir`.
  * Supports custom prefixes, suffixes, parent directories, retry limits, and block-based cleanup.
  * Creates temporary directories with unique names and restricted permissions.

* **Tests**
  * Added coverage for temporary-directory creation, customization, parent paths, block usage, and cleanup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,7 +128,7 @@ export LD_LIBRARY_PATH := $(OPENSSL_PREFIX)/lib$(if $(LD_LIBRARY_PATH),:$(LD_LIB
 endif
 endif
 endif
-BUNDLED_NATIVE_OBJS = packages/json/sp_json.o packages/stringio/sp_stringio.o packages/strscan/sp_strscan.o packages/base64/sp_base64.o
+BUNDLED_NATIVE_OBJS = packages/json/sp_json.o packages/stringio/sp_stringio.o packages/strscan/sp_strscan.o packages/base64/sp_base64.o packages/tmpdir/sp_tmpdir.o
 ifeq ($(OPENSSL_AVAILABLE),yes)
 BUNDLED_NATIVE_OBJS += packages/openssl/sp_openssl.o
 endif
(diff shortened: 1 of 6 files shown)


matz:
Merged as 4e600b36, with two follow-ups.

**The package could not compile at its simplest call.** `require "tmpdir"` plus `Dir.mktmpdir("pfx")` and nothing else fails on the branch: one String argument narrows `prefix_suffix`, so the `is_a?(String)` guarding it is answered statically, the ternary folds to its live arm, and that arm is the parameter itself. The fold rendered the arm at its OWN type where the unfolded pair renders both at the result type, so a plain string landed in a poly local. Your own test passes the Array form, which keeps the parameter poly and hides it.

That is a compiler bug, not yours, and it is fixed in 86a2c98d (ahead of the merge). Only the widening conversion is taken there: forcing a narrower one would convert rather than carry, and where the result type is narrower than the arm the analysis has already gone wrong, so the C type error is the honest answer. Generated C is byte-identical for all 3427 suite programs.

**Two things in the package** are fixed in 1a8ea14a:

- `max_try` was read out of the options and dropped. The C entry point took no such parameter and always used its own 10000, so the documented option did nothing. It is passed through now, and a non-positive one is refused rather than silently meaning the default.
- `Dir.tmpdir` returned `$TMPDIR` unseen. CRuby checks that it exists, is a directory, and can be written and searched, falling back to /tmp; otherwise the failure only moves to the mkdir with an error naming a path the program never chose. With `TMPDIR=/nonexistent`, CRuby answers "/tmp" and this now does too.

The rest reads well. Carrying the name format in C with the Ruby surface on top is the right split for this, the parent-directory sticky-bit check before the cleanup is the right instinct, and mapping mkdir's errno to its Errno class rather than a bare RuntimeError is what makes a failure actionable.

make gate is green (3414 pass, optcarrot 59662), and the package test passes.

Issue #64, opened 2026-04-27
Hash initialized as {} infers integer values when assigning strings


companygardener:
### Summary

An instance variable initialized as `{}` can become `StrIntHash` even when later writes store string values.

### Repro

```ruby
class Sources
  def initialize
    @file_sources = {}
    @current_file = "a.rb"
  end

  def add(source)
    @file_sources[@current_file] = source if @current_file
  end
end

Sources.new.add("body")
puts "ok"
```

Commands:

```sh
./spinel_parse /tmp/string_hash_value.rb /tmp/string_hash_value.ast
ruby spinel_codegen.rb /tmp/string_hash_value.ast /tmp/string_hash_value.c
cc -O0 -Werror -ffunction-sections -fdata-sections -Ilib /tmp/string_hash_value.c lib/libspinel_rt.a -lm -o /tmp/string_hash_value
```

### Expected

The program compiles and prints:

```text
ok
```

### Actual

Compilation fails because the generated hash setter expects `mrb_int`:

```text
incompatible pointer to integer conversion passing 'const char *' to parameter of type 'mrb_int'
sp_StrIntHash_set(..., lv_source)
```

---
Reproduced against matz/spinel `7bfc13c` on 2026-04-27.

PR #4281, opened 2026-09-02
Drop the catch entries a raise unwinds past, and bound every handler-stack push

blacktm:
## Why

Two defects in the same machinery, both found by a reviewer's ASAN probes while reviewing the exception-roots PR (#4269). They are on master and on that branch alike, so neither is #4269's doing, and neither needs GC stress or threads to show.

The first: a raise that leaves a `catch` block never dropped that catch's entry.

```ruby
c = 0
200.times do
  begin
    catch(:x) { raise ArgumentError, "boom" }
  rescue ArgumentError
    c += 1
  end
end
puts c          # CRuby 200; master segfaults
```

Every round leaves one dead entry behind. `sp_catch_top` is an index into fixed 64-slot arrays, so round 65 writes past the end of `sp_catch_tag` — UBSAN says `index 64 out of bounds for type 'const char *[64]'` and ASAN reports a global-buffer-overflow WRITE 0 bytes after `sp_catch_tag`; an ordinary build corrupts whatever the linker put next and, here, dies with SIGSEGV before printing anything.

Rounds 1 to 64 are worse than a leak, because a dead entry still answers a `throw`:

```ruby
begin
  catch(:y) do
    begin
      catch(:x) { raise "e" }
    rescue RuntimeError
    end
    puts "before throw"
    throw :x
  end
rescue UncaughtThrowError => e
  puts "uncaught: #{e.message}"
end
puts "end"
```

CRuby prints `before throw`, `uncaught: uncaught throw :x`, `end`. Master prints `before throw` twice and then dies with an unrescued `uncaught throw :x (UncaughtThrowError)` on stderr, exit 1. The emitted C says why: the finished inner `catch` is still counted, its `setjmp` buffer belongs to a C frame that is still live (the same function), and `sp_catch_exc_top` for it is above the depth at the `throw`, so `sp_throw` jumps straight into it — landing in the `else` arm of a `catch` that completed. Execution resumes there, runs the `sp_exc_top--` that belongs to the finished `begin` (so the handler depth is now one below reality), prints `before throw` again, throws again, finds nothing this time, and raises `UncaughtThrowError` with `sp_exc_top` at 0 — which is why the enclosing `rescue` never sees it.

The second: `sp_exc_top++` and `sp_catch_top++` were unchecked, unlike `sp_brk_push` beside them. 65 lexically nested `begin` frames write past `sp_exc_rootmark`, `sp_exc_msg` and `sp_exc_obj`, and `setjmp` writes past `sp_exc_stack`. The depth is not a lexical property, which is what makes this reachable from ordinary code: a method whose body is a `begin`/`rescue` keeps its frame armed across the calls that body makes, so plain recursion 65 deep does it.

```ruby
def rec(n)
  begin
    return 0 if n <= 0
    rec(n - 1) + 1
  rescue ZeroDivisionError
    -1
  end
end
p rec(70)       # CRuby 70; master prints 70 too, having written 7 ints past sp_exc_rootmark
```

It arms one frame per level, so `rec(70)` arms 71 and writes indices 64 through 70 — seven `int` slots past the end of `sp_exc_rootmark`, and the same past `sp_exc_msg` and `sp_exc_obj`. Under ASAN+UBSAN that is `index 64 out of bounds for type 'int[64]'` and a global-buffer-overflow WRITE past `sp_exc_rootmark`. In a plain build it is silent.

The catch leak had a matching hole in the two other handler stacks. `sp_proc_homes_unwind` — which drops proc-return homes and break scopes an exception unwound past — was emitted at only six of the ten exception landings. Kernel#loop (both the statement and the expression form), the enumerator fold and the ext-host try never called it:

```ruby
200.times do
  loop do
    [1].each { |x| break 5 if x == 9; raise StopIteration }
  end
end
puts "end"      # CRuby prints end; master exits 1 with "stack level too deep (SystemStackError)"
```

The valued `break` in the block makes the call push a break scope; the `StopIteration` lands in the `loop` frame, which never pruned it. Round 65 hits `sp_brk_push`'s own bound check.

## What

Two rules, one commit each.

A raise now drops every catch entry, break scope and proc-return home that was opened inside the frame it lands in — at the raise, once, for every kind of landing.

Every fixed-depth handler-stack push now checks its bound: reaching 64 exception frames or 64 live `catch` blocks prints CRuby's `stack level too deep (SystemStackError)` on stderr and exits 1, which is what `sp_brk_push` has always done for break scopes. Inside a fiber or a thread, the frame the runtime arms around the body takes one of the 64 slots, so the body itself has 63.

## How

All three stacks already record the exception depth at their own entry — `sp_catch_exc_top[i]`, `sp_brk_exc_top[i]`, `sp_proc_home.exc_top` — so all three want the same test, and `sp_raise_cls` is the one place to apply it. Just before it longjmps to `sp_exc_stack[sp_exc_top-1]`, it calls `sp_handler_stacks_unwind`, which pops from each stack while the recorded depth is `>= sp_exc_top`. That handler is the frame the exception lands in; it was armed at `sp_exc_top-1`, so anything entered after it recorded `sp_exc_top` or more and anything entered before it recorded less. This is the same test the six landings were making as `> sp_exc_top` after their own `sp_exc_top--`, moved one step earlier where it can be written once. `sp_proc_homes_unwind` becomes that function (renamed, since it now prunes three stacks and is called from the raise), the six emitted calls go away, and where the `if (sp_unwind_kind == SP_UNWIND_NONE)` guard existed only to wrap the call, the guard goes with it.

A proc return, a `throw` and a valued `break` do not come through `sp_raise_cls` at all. Each of the other four `longjmp(sp_exc_stack…)` sites belongs to one of them — `sp_throw`, `sp_brk_throw`, `sp_proc_return`, and `sp_unwind_resume`, the step that carries them from one intervening `ensure` to the next — and each trims its own target stack before it jumps (`sp_throw` sets `sp_catch_top = i + 1`, `sp_brk_throw` sets `sp_brk_top = i + 1`, `sp_proc_return` walks to a live home). They keep managing themselves; only a real exception prunes here. `sp_raise_cls` already clears `sp_unwind_kind` on entry, so an exception raised from inside an `ensure` running during one of those unwinds prunes, correctly, as the exception it now is.

The bound check is a pair of one-line inlines, `sp_exc_check_depth` and `sp_catch_check_depth`, both calling a new cold `sp_stack_too_deep` that holds the one copy of the message `sp_brk_push` used to print inline. Placing them is the only fiddly part: every exception-frame arm stores into `sp_exc_rootmark[sp_exc_top]` (or `sp_exc_msg[sp_exc_top]`) *before* it bumps `sp_exc_top`, so the check has to precede the first indexed store, not the `++`. It opens the frame-arming sequence at all ten emitted arms — the two `begin` forms, both rescue-modifier forms, both Kernel#loop forms, the block-with-ensure filter loop, the enumerator fold, `Mutex#synchronize` and the ext-host try frame — and in the runtime's own `sp_exc_arm`, the fiber trampoline's. Most arms declare a temp or a landing flag before it, and `Mutex#synchronize` takes its lock first; what holds at all eleven is that nothing stores into a fixed-depth array ahead of the check. `sp_catch_check_depth` goes in front of the `catch` arm's first store, where it covers both the tagged and the `catch { |tag| }` autotag path. In a `begin` with `retry`, the check sits after the retry label, so a retry re-checks.

The failure is fatal rather than a Ruby `SystemStackError`: catching one needs a handler slot, which is exactly what has run out. `sp_rescue_push` keeps its own older, more specific wording (`rescue nesting too deep (> 64)`), which names the nesting that overflowed; that is its own rule and its own message.

## Validation

- Gate on the branch: tests 3411 pass, 0 fail, 0 error; benchmarks 61 pass; optcarrot OK 59662; ext-cruby, rbs-seed and the rubyspec gate pass; props (alloc-report, infer, collect-errors) pass; the macOS-only spin-e2e leg is red on this machine for the reason it has been on every run here, progress output leaking onto stdout, independent of this change. The inference fixpoint is unchanged: 0 new non-convergences over 3365 programs, the 2 that hit the round cap hit it on master too, and one program's round count moves by one (test/enumerator_block_returns_self.rb, 4 to 3 — the known noise on that file).
- The emitted C changes for 745 of the 3427 test, benchmark and optcarrot programs swept, and for every one of them the change is only these two lines. Proved mechanically rather than read: for each program, emit with both compilers, delete from master's C exactly the `sp_proc_homes_unwind();` calls (whole line, guarded line, or the inline one in `Mutex#synchronize`'s landing) and from the branch's C exactly the `sp_exc_check_depth();` / `sp_catch_check_depth();` calls, and require the two files to be byte-identical. All 745 are; 0 differ otherwise and 0 fail to compile. In total 2594 landing calls are removed and 2735 checks added.
- Optcarrot's whole diff is one line added and one line removed, in one `begin`/`rescue` frame: `sp_exc_check_depth();` added at line 2365 of the branch's C, in front of that arm's first indexed store, and `sp_proc_homes_unwind();` gone from line 2378 of master's, in that same frame's landing.
- Cost, measured on the arm rather than guessed: a `while` loop whose body is a `begin`/`rescue`, compiled `-O2` for arm64, grows the loop header by three instructions — a load of `sp_exc_top`, `cmp w8, #64`, and a never-taken `b.ge` to a cold `bl _sp_stack_too_deep` — and loses the landing's `bl _sp_proc_homes_unwind` plus the guard load in front of it. The whole function goes from 214 to 218 instructions; the rest of the diff is register allocation moving around. **Optcarrot not measured here** — no callgrind instruction count was taken, and the `make optcarrot` / `make bench` wall-clock numbers are not that count.
- The corpus of 2131 differential findings is unchanged: 804 matching before, 804 after, 0 lost, 0 gained. Neither rule is a compatibility fix in the corpus's sense; both are memory-safety and control-flow fixes underneath it.
- New test, `test/raise_drops_catch_entries.rb`, expected output generated by CRuby, runs in a few milliseconds. It pins the raise-out-of-a-catch count, the stale entry answering a `throw`, a `throw` from an `ensure` running during a raise (with and without a rescue around the `catch`), the same raise-out-of-a-catch inside Kernel#loop, a raise unwinding past a proc-return home followed by a `return` that must still reach a live one, the break scope inside a Kernel#loop landing, and the same break scope under Kernel#loop's expression form (`v = loop do … end`), which is a different landing again — that last shape on its own exits 1 on master with `stack level too deep (SystemStackError)` and prints 200 here. It passes plain, under `SPINEL_GC_STRESS=1`, and under ASAN+UBSAN with GC stress, with no sanitizer output. On master it segfaults with no output at all; the `throw` pin diverges on the first round, without any repetition.
- Clean under ASAN+UBSAN with `SPINEL_GC_STRESS=1`, stdout matching, on the new test and on test/catch_throw_ensure.rb, test/catch_throw_through_loop.rb, test/catch_in_fiber_body.rb, test/exc_frame_break_next_pops.rb, test/exc_frame_early_return_pops.rb, test/loop_stopiteration_result.rb and test/proc_return_ensure_raise.rb. Also clean on the four repro programs above and on a fiber raising out of a `catch` 200 times and then yielding and resuming normally.
- The bound is in the suite as two programs that print nothing: `test/exc_frame_depth_bounded.rb` (recursion 70 deep through a method whose body is a `begin`/`rescue`) and `test/catch_depth_bounded.rb` (recursion 70 deep through a `catch`). CRuby runs each to its closing `exit 1` with nothing on either stream; the branch stops at the bound with `stack level too deep (SystemStackError)` on stderr. The harness compares stdout with `test/x.rb.expected` and stderr with `test/x.rb.err.expected` and never reads the exit status (Makefile:1084-1092), so both `.expected` files are empty and both `.err.expected` files are hand-held. The closing `exit 1` is what protects them: the snapshot rule skips a test whose CRuby run exits non-zero (Makefile:1155-1171), so no regen target rewrites any of the four files, and a comment above each `exit 1` says so. Both fail on master, which runs the recursion silently; under ASAN+UBSAN master is `index 64 out of bounds` on `int[64]` and on `const char *[64]`. 64 frames of either kind still run on the main thread; 65 is the first that does not.
- Multi-threaded: the catch repro inside 4 threads answers `[200, 200, 200, 200]` on the branch and on master, plain. Under ASAN with GC stress the same pre-existing heap-use-after-free in a raise fires on both — the separately tracked MT message-lifetime defect, untouched here. Two of four master runs also showed the catch-stack overflow this PR removes, `index 64 out of bounds` on the catch arrays (`sp_catch_tag`, `sp_catch_tag_kind`, `sp_catch_val`, `sp_catch_exc_top`, `sp_catch_rootmark`, `sp_catch_stack`); master's sanitizer output is not stable run to run. The branch shows the use-after-free and nothing else.

## Left alone, on purpose

- **`SP_EXC_STACK_MAX` and `SP_CATCH_STACK_MAX` themselves.** Still 64. Whether the handler stacks should grow, become heap-backed like the break stack, or stay fixed is a sizing decision, not this rule.
- **`SystemStackError` as a rescuable exception.** The new bound exits with CRuby's message rather than raising, because raising needs the slot that has run out. Making it a real Ruby exception is a separate change with its own design.
- **`sp_rescue_push`'s wording.** It has its own bound and its own message, which says which nesting overflowed. Folding it into the shared one would lose that.
- **One of the two depth programs changes shape between the commits.** They are different programs and they behave differently, so both are worth spelling out. `rec(70)`, the recursion in the Why, prints `70` and exits 0 on master and with the first commit alone — silently overflowing each time — and is the fatal with both. The other, 70 lexically nested `begin` frames whose rescue bodies each re-raise, exits 1 on master with `rescue nesting too deep (> 64)`: loud, but that is `sp_rescue_push` catching a consequence long after the arms wrote past their arrays. With the first commit alone that one segfaults, because the pruning at the raise reads stacks the overflow has already corrupted; with both it is the clean fatal at the first arm that would have overflowed. That is why they are one PR.
- **A bare `break` or `next` out of a `catch` body.** The emitted catch keeps today's C-break behaviour, which may skip the `sp_catch_top--` on that path. It is a different rule, being probed separately; nothing here depends on it, and none of the pins above trip it.
- **`sp_exc_ctx_t` not saving `sp_exc_rootmark`, `sp_rescue_mark` or `sp_catch_rootmark` per fiber.** Noticed while reading the fiber context; unrelated to either rule here, and being looked at on its own.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved protection against excessively deep exception, catch, and loop-handler nesting.
  - Deeply nested handlers now terminate with a clear `SystemStackError` message instead of risking invalid runtime behavior.
  - Exception unwinding now consistently removes handler state that is no longer active, improving behavior for `raise`, `throw`, `break`, and proc returns.
  - Prevented stale handlers from affecting later control-flow operations after an exception is raised.


matz:
Merged as 42b1e8a5 (see the note below about the merge with #4282).

I reproduced both defects on master before merging. The 200-round catch loop dies with SIGSEGV and prints nothing, and after the merge it prints 200, as CRuby does. The reasoning about why rounds 1 to 64 are worse than a leak is the part that makes this more than a bounds fix: a dead entry still answers a throw, so the program takes a jump to a catch that has already returned.

make gate is green (3424 pass, optcarrot 59662).

**One thing to know if you rebase anything onto this.** #4282 rewrote the tail of `sp_raise_cls` too, so the two conflict. Kept naively, both tails survive and yours runs first, which terminates an uncaught SystemExit before the at_exit hooks #4282 adds ever run. I resolved it as one tail: #4282's structure with your `sp_handler_stacks_unwind()` in the longjmp branch.

matz:
Correction to the SHA above: the merge commit is dad16dd4, and the follow-up that resolves the shared tail with #4282 is c3a16a43.

PR #4282, opened 2026-09-02
Run at_exit hooks when the program ends by exit, abort or an uncaught exception

blacktm:
## Why

Spinel ran a program's `at_exit` hooks only when `main` fell off its end. Every other way a Ruby program can finish dropped them silently:

```ruby
at_exit { puts "atexit ran" }
begin
  nil.nope(7)
rescue NoMethodError => e
  puts e.message
  exit 3
end
```

CRuby prints the message, then `atexit ran`, and exits 3. Spinel printed the message and exited 3. The hook never ran, and nothing said so.

It is not one path. `exit` at the top level, `exit` inside a rescue arm, `abort`, and an uncaught exception all drop the hooks; so does an `exit` from a non-main thread. A hook that itself calls `exit` never got the chance to. The one shape Spinel got right, `exit!`, is right because CRuby skips the hooks there too.

What hid this is that `END { }` blocks were never affected: they are registered with libc `atexit`, so they run on every one of these paths. A program that used `END` saw correct behaviour and a program that used `at_exit` did not, which is a difference nobody would think to look for.

The reviewer who found this saw only the rescue-arm case. The rule is wider, and this PR is the whole of it.

## What

`at_exit` hooks now run on every termination CRuby runs them on: falling off the end of main (as before), a `Kernel#exit` that ends the program (whether it terminated directly or first travelled as a rescuable `SystemExit` nothing caught), `Kernel#abort`, and an uncaught exception of any other class. For an uncaught exception the hooks run *before* the error text is printed, which is the order CRuby prints them in — a hook writing to stderr appears above the error, not below it.

They keep running most-recently-registered first, and — this is CRuby's other rule about them — a hook that ends the program ends only itself. Whatever a hook does, the hooks registered before it still run:

```ruby
at_exit { STDERR.puts "h2" }
at_exit { exit 7 }
begin
  raise "x"
rescue
  exit 2
end
```

CRuby runs `exit 7`, then still runs the `h2` hook, and exits 7. So does this branch; master printed nothing and exited 2.

That is also where the exit status comes from. The path's own status is the starting point (0 falling off the end of main, N for `exit N`, 1 for `abort` or an uncaught exception), and each hook that ends the program replaces it — the last one to do so wins. A hook that raises anything other than a `SystemExit` makes the status 1, whatever the path was: `at_exit { raise "x" }; exit 3` is status 1 in CRuby and here.

A hook's own uncaught error is printed where it happens, before the hooks after it run, and the program's own uncaught error is printed once, after all of them:

```ruby
at_exit { STDERR.puts "second" }
at_exit { raise "hookerr" }
raise "boom"
```

CRuby's stderr is `hookerr (RuntimeError)`, then `second`, then `boom (RuntimeError)`, exiting 1. This branch prints the same three lines in the same order (with the message-only tail format an ordinary Spinel build uses). Master printed only `boom (RuntimeError)`.

`exit!` still terminates directly, with no hooks, as CRuby does.

This is two commits. The first, "The at_exit hook table is a GC root", stands on its own; the second is the rule itself: the runner, the call sites and the codegen line. The first had to come first: the hook table is now a GC root. The registering expression stores the Proc in the table and drops it, so between `at_exit { }` and the hook actually running the table was the only reference to the Proc and to the environment it captured — and nothing marked it. On master, `at_exit { puts "hook ran" }` in a program that allocates enough to collect is a segfault under `SPINEL_GC_STRESS=1` on the one path that did run the hooks. That had to be true before the hooks could run on paths where a collection has just happened, or where the hook itself allocates.

## How

The walk that codegen emitted inline at main's tail becomes one runtime function, `int sp_at_exit_run(int status)` (lib/spinel_rt.h). It takes the status the calling path would have used and answers the one the process should end with. There are four termination paths and five calls: the runtime holds four — `exit(sp_at_exit_run(status))` and `exit(sp_at_exit_run(1))` in `sp_exit_raise` and `sp_abort_raise`, and in `sp_raise_cls` the uncaught-`SystemExit` arm and the uncaught-exception path — and codegen emits the fifth at main's tail as `return sp_at_exit_run(0);`, or as a bare call in the `--ext-init` form, which has no status to return.

Each hook runs inside its own protect frame. That is the substance of the change and it is what CRuby does: every end proc runs under its own `rb_protect`, so whatever one hook does — `exit`, `abort`, an uncaught raise — ends that hook and no more. The frame is the same short sequence the generated ext-host `<init>_try` helper arms (src/codegen.c): record `sp_exc_rootmark`/`sp_rescue_mark`, clear the frame's message and object slots, `sp_exc_top++`, `setjmp`. On a landing it pops the frame, restores `sp_gc_nroots` and `sp_rescue_sp` from the marks, and reads the exception out of the slot it was stored in — a `SystemExit` gives the new status, anything else is printed there and then and leaves the status 1. Either way the loop continues with the next hook.

Because a frame is armed, `exit` inside a hook is an ordinary rescuable `SystemExit` raise that lands in the drain, not a call into libc `exit()`: the direct-exit branch in `sp_exit_raise` and `sp_abort_raise` is now reached only when nothing is armed. Nothing re-enters the drain, and — the reason this matters beyond tidiness — the uncaught error the drain was called in front of is still there to be printed when the drain returns.

Each hook is popped off the table before it is called, rather than the table being walked by index. That keeps LIFO order while letting a hook register another hook during the drain (CRuby runs those too), and it means no hook can run twice.

Before each hook the three non-exception handler stacks — the proc-return homes, the break scopes, the catch stack — are emptied. CRuby has unwound the whole program before it calls the first end proc, and a hook that ends by raising has unwound itself before the next one runs, so in CRuby a `break` or `throw` out of a hook, or a `return` through an escaped proc, has nothing to land on and raises `LocalJumpError` or `UncaughtThrowError`. Here those stacks name C frames that have already returned, so leaving them standing means a `longjmp` into a dead frame. The shape that shows it needs two hooks: one calls a method that leaks a `proc { return }` and then raises, and the next calls that escaped proc — CRuby raises `LocalJumpError`, and without the clearing the branch jumped into the raised-past method's frame and printed an invented error. The exception stack itself is deliberately not touched: the drain's own frame is on it.

Three things are read or held across the hook calls:

- The uncaught-`SystemExit` arm reads its status out of the pending exception object before the hooks run, because nothing roots that object once a hook starts allocating.
- The message the uncaught-exception path prints afterwards is safe as it stands: `sp_raise_cls` laundered it onto the string heap and rooted it at the top of the function, and the root is never popped, so a hook that allocates under GC stress cannot sweep it. The class name is a codegen-emitted literal the collector never touches. The backtrace substrate is not safe — a hook that raises overwrites `sp_bt_buf` — so a debug build, the only build where that substrate is live, keeps a copy of it across the drain.
- The pending-raise state is cleared before the first hook runs: `sp_pending_exc_obj` and the three cause fields. They hold the terminating exception, and `sp_raise_cls`' recovery guards read `sp_pending_exc_obj` and take it as the object being raised — so without the clear, the first `raise` inside a hook bound the exception that was killing the program. `begin; raise ArgumentError, "inner"; rescue => e` inside a hook bound a `SystemExit`.

The uncaught-exception printer is factored out as `sp_exc_print_uncaught` so the drain can print a hook's error in exactly the format the program's own error gets. It is the code that was inline at the end of `sp_raise_cls`, unchanged except that it returns instead of exiting.

The GC root is a marker beside the table, `sp_mark_at_exit_hooks`, called from `sp_re_mark_globals` alongside the regex registers, the exception stack, the proc-return homes, the break values and the fiber storage. The popped hook is also rooted for the duration of its own call, since the pop takes away the collector's only handle on it.

The codegen change is the one line at main's tail, which becomes `return sp_at_exit_run(0);` so a hook can decide `main`'s status (src/codegen.c). The `--ext-init` form has no status to return, so there the call stands alone. Two stale comments that said main's tail is where the hooks run were updated to name the runtime helper.

The helper is named after the table rather than after the verb on purpose. `mc_top` (src/codegen_util.c) reserves the first underscore-separated segment of every runtime symbol against a top-level `def`, from a list the Makefile generates by scanning the runtime sources, and gives a colliding Ruby method an `rb_` infix. The table already reserves `at`, so `sp_at_exit_run` costs nothing; a name built on the verb would have reserved that verb and renamed the emitted symbol for every program with a top-level method of that name — ten of the suite's own programs, in the first draft of this change. The printer is `sp_exc_print_uncaught` for the same reason: `exc` is reserved already and `print` is not.

## Validation

- Tests 3415 pass, 0 fail, 0 error. Benchmarks 61 pass. Optcarrot OK 59662, and optcarrot's generated C is byte-identical. ext-cruby, rbs-seed and the rubyspec gate pass; the macOS-only spin-e2e leg is red on this machine for the pre-existing reason it has been on the last several PRs, progress output leaking onto stdout, independent of this change.
- On the differential corpus of 2,131 minimized programs that behave differently under CRuby 4.0.6 and Spinel: 804 matched at the merge-base and 804 match on this branch. Nothing lost, nothing gained, which is what the corpus should say here: not one of its programs registers an `at_exit` hook.
- Generated C is byte-identical for 3420 of the 3431 test and benchmark programs, plus optcarrot. The eleven that differ are exactly the programs that register an `at_exit` hook — test/at_exit_lifo.rb, test/at_exit_returns_proc.rb, test/signal_module_surface.rb, test/warn_kernel.rb and the seven new tests — and each differs in the same two lines of main's tail, the inline loop and the bare `return 0` becoming `return sp_at_exit_run(0);`. All eleven still match their expected output.
- No suite program fails to reach its inference fixpoint that reached one before: over 3369 programs there are 0 new non-convergences, no round count differs from the merge-base's, and the two that hit the round cap hit it there too.
- test/at_exit_hook_survives_collection.rb, the first commit's own test, is the crash program: a hook that captures a local, a bare hook, and 200 string allocations. It matches CRuby plain, under `SPINEL_GC_STRESS=1`, and under ASAN+UBSAN both plain and stressed, and it runs in well under a second. On master it passes plain and segfaults under `SPINEL_GC_STRESS=1` with no output at all. The first commit was checked out on its own and builds and passes that test with the codegen tail still the old inline loop.
- test/at_exit_hook_nonlocal_jump.rb pins the handler-stack clearing: a hook whose escaped `proc { return }` outlives the method an earlier hook raised out of, a `throw` to a `catch` the program has already left, both answering `LocalJumpError` / `UncaughtThrowError` as CRuby does. On master the two hooks behind the raising one never run at all.
- The other five tests for the rule match CRuby plain, under `SPINEL_GC_STRESS=1`, and under ASAN+UBSAN both plain and with GC stress; each runs in a few milliseconds. The harness compares stdout and stderr separately and ignores the exit status (Makefile:1084-1092), so the tests whose point is an ordering put the ordering inside one stream: test/at_exit_runs_on_uncaught.rb has its hooks write to stderr, where the program's error also goes, so the sidecar pins hooks-before-error; test/at_exit_hook_raises_and_exits.rb is the raising-and-exiting shape, whose sidecar pins a hook's error printed between the hooks around it and the program's error printed last; test/at_exit_runs_on_exit.rb exits 0 through the `exit`-inside-a-rescue path and pins LIFO order plus a hook that calls `exit` itself. test/at_exit_skipped_by_exit_bang.rb pins the one termination where the hooks must NOT run; it passes on master too, and is there so that a fifth runtime call site cannot be added by mistake. On master the other four lose exactly the hook lines and nothing else. The sidecars hold Spinel's own tail format for an uncaught raise, the format test/uncaught_exception_message.rb already pins.
- Fourteen programs covering every shape in "Why" and "What" — the seven terminations, and seven more for the hook-ends-the-program rule (a hook that exits 0 or 5 under an uncaught raise, a hook that raises with another hook behind it, a hook that aborts, `exit 7` in a hook with a later hook still to run, `raise SystemExit.new(4)`, and a raising hook after `exit 3`) — were compared with CRuby 4.0.6 on stdout, stderr and exit status. All fourteen match on stdout and on status, and on stderr line for line; the only difference anywhere is the pre-existing one, that Spinel prints `boom (ArgumentError)` where CRuby prefixes `file:line:in '<main>':` and, for some raises, echoes the source line as well. Every hook line lands on CRuby's side of it.
- A hook that allocates 300 strings was run on all four paths (fall-off-the-end, `exit`, `abort`, uncaught raise), plain, under GC stress, and under ASAN+UBSAN with GC stress: output identical to CRuby's every time, no sanitizer reports.
- A hook that calls `abort` was checked against CRuby: byte-identical on stdout, on stderr and on the exit status (the abort message, then the remaining hooks, then status 1).
- A hook that raises, and a hook whose own `begin`/`rescue` rescues and re-raises, were checked against CRuby: same stdout, same stderr order, same status (the re-raise makes it 1, outranking the `exit 4` the program asked for).
- `Thread.new { exit 3 }.join` with a hook registered: the hook now runs, as it does in CRuby, and the status is 3 on both. (Spinel additionally prints a `terminated with exception` line for the thread, which it did on master too and which is a separate difference.)
- The GC-root half was verified by counterfactual: `at_exit { puts "hook ran" }` in a program that allocates 200 strings segfaults on master under `SPINEL_GC_STRESS=1` and is clean here, plain, stressed, and under ASAN+UBSAN with stress.
- Under ASAN+UBSAN with `SPINEL_GC_STRESS=1 SP_POOL_MAX=0`: the seven hook-ends-the-program programs, the allocating hook on all four paths, the raising hook, the aborting hook, and the hook that rescues and re-raises. No sanitizer reports, and each ends with the status CRuby gives it.

## Left alone, on purpose

- **`END { }` versus `at_exit` ordering when a program mixes them.** CRuby keeps one LIFO list, so `at_exit { A }; END { B }` runs B then A. Spinel registers `END` blocks with libc `atexit` and `at_exit` hooks on this table, and the table always drains first, so it runs A then B. That is what master did on its one working path as well; this change does not move it. Unifying the two lists is its own change.
- **`$!` inside a hook.** CRuby sets it to the exception ending the program: the raised one on an uncaught raise, the `SystemExit` after `exit` or `abort`. Here it is whatever was being handled where the program ended: after `exit` inside a rescue arm it is that arm's exception (CRuby: the `SystemExit`), and after a top-level `exit`, an `abort` or an uncaught raise it is `nil`. After a normal end it is `nil` in both. A bare `raise` inside a hook therefore raises a fresh `RuntimeError` here where CRuby re-raises the terminating exception — but that is a pre-existing property of a bare `raise` with no lexically enclosing rescue, visible on master in a helper method called from a rescue body, and it is its own fix. Setting `$!` for the drain would fix the first three cases and not that one, so it waits for that fix.
- **A `return` written directly in a hook.** CRuby raises `LocalJumpError` and exits 1. Spinel compiles a top-level `return` as the program's exit and a `return` inside a method's block as the block's own return, so the hook ends silently, and on the top-level form the status becomes 0. Both are pre-existing codegen of `return` in a block, visible on master's own normal-end path, and are parked for their own fix. A `return` through a proc that escaped a method is different, goes through the proc-return home, and is handled: it misses and raises `LocalJumpError`, as the new test pins.
- **Backtraces in a `--debug` build.** Nothing is left alone here: the drain keeps a copy of the backtrace substrate across the hooks so the program's own error keeps its own frames, and each hook's error is printed inside its own landing, so it keeps its own too.
- **The `terminated with exception` line Spinel prints when a non-main thread exits.** Pre-existing, unrelated to the hooks, and unchanged.
- **`exit!`** — no hooks, by design, matching CRuby. Note that libc `atexit` handlers, and therefore `END` blocks, still run there because `exit!` lowers to libc `exit()`; CRuby's `exit!` skips those too. Separate.
- **The drain's protect frame has no depth check.** It mirrors the ext-host `_try` helper, which has none either; on every path that reaches the drain `sp_exc_top` is 0, so the frame it pushes is the first. There is another PR in flight over this same handler stack that adds an `sp_exc_check_depth()` to each arm; whichever of the two lands second should give this frame the same one line.
- **The unbounded registration.** `SP_AT_EXIT_MAX` is 256 and the registering expression does not check the count, so a program with more hooks than that writes past the table and crashes. Pre-existing, and unchanged in kind: 300 hooks die at registration on master and here alike, before any of this code runs. Giving it a bound is its own change, and it needs a decision about what CRuby, which has no limit, should be matched with.

## Summary by CodeRabbit

- **New Features**
  - `at_exit` hooks now run reliably during normal exits, aborts, and uncaught exceptions.
  - Hooks execute in reverse registration order, and remaining hooks continue if one raises or exits.
  - Hook state is preserved during garbage collection.
  - `exit!` continues to bypass `at_exit` hooks.

- **Bug Fixes**
  - Improved handling of non-local jumps and error reporting within exit hooks.

- **Tests**
  - Added coverage for exit, abort, uncaught exceptions, garbage collection, and hook failure behavior.


matz:
Merged as 2085375b.

I checked the paths against CRuby rather than the diff. `exit` in a rescue arm, `abort`, an uncaught exception (hooks first, then the error text) and `exit!` (no hooks) all agree now, and so does the rule that makes this more than a list of paths: a hook that ends the program ends only itself, so `at_exit { STDERR.puts "h2" }; at_exit { exit 7 }` with a rescue arm calling `exit 2` prints h2 and exits 7 on both. The observation that `END` was never affected, because it rides libc atexit, explains why nobody found this earlier and was worth writing down.

**Two corrections.**

First, the merge with #4281. Both PRs rewrote the tail of `sp_raise_cls`, so they conflict. Kept naively both tails survive and #4281's runs first, terminating an uncaught SystemExit before your hooks run. It is one tail now (c3a16a43): your structure, with #4281's `sp_handler_stacks_unwind()` in the longjmp branch.

Second, one claim in the description does not hold. The list says an `exit` from a non-main thread now runs the hooks, and the validation section says `Thread.new { exit 3 }.join` runs the hook with status 3 on both. Measured here after the merge:

    at_exit { STDERR.puts "HOOK" }
    Thread.new { exit 3 }.join

CRuby prints HOOK and exits 3. Spinel prints the `terminated with exception` line and exits 3, with no HOOK. Without the `.join` the hook does run, and a thread raising an ordinary exception joins and runs the hooks correctly, so the gap is specifically a SystemExit propagating out of `Thread#join`: it does not reach `sp_raise_cls` at top level, which is where your termination path lives.

That is master's behaviour too, so it is not a regression and I merged on the strength of the four paths that are fixed. But the description should not claim the fifth. If you want to finish it, the thing to find is where the joined thread's stored SystemExit turns into the process's exit status without going through `sp_raise_cls` -- the status is right, so something reads it on a path of its own.

PR #4285, opened 2026-09-03
rbs-seed: align instantiate hash type with find_by caller

rofl0r:
The override_family_ret fixture had Hash[String, untyped] for instantiate but Hash[Symbol, untyped] for find_by. find_by calls instantiate(conditions) so the receiver hash type never matched the parameter hash type. CRuby permits this (untyped hashes are duck-typed) and the test was originally written against a GCC that did not promote -Wincompatible-pointer-types to an error under RBS_SEED_STRICT (-Werror=incompatible-pointer-types). GCC 14 does promote it, so the parameter mismatch blocks the syntax-only check before the return-seed path (#3203) is ever reached.

Align the instantiate parameter to Hash[Symbol, untyped] so it matches the caller's hash type. The test still exercises the return seed scenario: sp_OfrBase_s_instantiate returns sp_RbVal (family/boxed) while sp_OfrBase_s_find_by returns sp_OfrBase * (concrete class pointer), which is the exact #3203 surface.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the `OfrBase.instantiate` method signature to accept hashes with symbol keys, aligning the declared interface with supported input.

### DIFF
--- a/test/rbs-seed/sig/override_family_ret.rbs
+++ b/test/rbs-seed/sig/override_family_ret.rbs
@@ -1,4 +1,4 @@
 class OfrBase
-  def self.instantiate: (Hash[String, untyped] _row) -> OfrBase
+  def self.instantiate: (Hash[Symbol, untyped] _row) -> OfrBase
   def self.find_by: (Hash[Symbol, untyped] conditions) -> OfrBase?
 end


matz:
Merged as bb1cda12.

I verified the claim rather than the reasoning, since this machine has GCC 13 and cannot reproduce the failure directly. Compiling the fixture's generated C with the diagnostic forced:

    gcc -fsyntax-only -Ilib -Werror=incompatible-pointer-types ofr.c

master's fixture fails on exactly the line you name, passing `sp_SymPolyHash *` to a `sp_StrPolyHash *` parameter of `sp_OfrBase_s_instantiate`; with your signature it is clean. That is the same construct GCC 14 rejects by default.

One correction for the record: this fixture's check is `$(CC) -fsyntax-only -Ilib` with no `RBS_SEED_STRICT` (Makefile line 941), so it is GCC 14 promoting the warning by default rather than the strict flag that breaks it. Same conclusion, different route.

The important thing is that the change keeps what the fixture is for. After it, `instantiate` still returns `sp_RbVal` (the family view) while `find_by` returns `sp_OfrBase *`, and the call site still unboxes through the family:

    return (sp_OfrBase *)(SP_RBS_CHECK_TAG(sp_OfrBase_s_instantiate(lv_conditions), ...)).v.p;

which is the #3203 surface, intact. make gate is green, rbs-seed-test passes.

PR #4335, opened 2026-09-04
codegen: accept TY_UNKNOWN and TY_POLY in the user-to_io dispatch

rofl0r:
A user class that wraps an IO -- a TLS socket, a passthrough, a process pipe -- exposes the underlying handle through #to_io. The codegen gate for adding the class to sp_user_to_io_dispatch required m->ret == TY_IO, which a poly body fails: the common shape is 'def to_io; @sock; end', and the constructor's parameter is poly, so @sock is poly and the return is TY_POLY (or TY_UNKNOWN when the analysis could not narrow further). The class was left out of the dispatch, and IO.select reached the runtime helper for a class the dispatch did not know, which raised:

    no implicit conversion of Wrap into IO (TypeError)

The fix accepts TY_UNKNOWN and TY_POLY alongside TY_IO in both gates (emit_user_to_io_dispatch and the g_has_user_to_io check in codegen_program). The runtime gets a small sp_poly_to_file inline in spinel_rt.h that turns a boxed IO into an sp_File * and returns NULL for anything else, so a wrong return is still a TypeError -- just at the moment of select, not at class discovery. The poly arm in the dispatch boxes the #to_io answer and asks sp_poly_to_file to unwrap it; the TY_IO arm keeps its direct cast because that function still returns sp_File * directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `IO.select` compatibility with custom `#to_io` methods returning dynamically typed IO values.
  * Added support for wrapper objects that return raw sockets through `#to_io`.
  * Corrected handling of non-yielding `#to_io` methods that accept a block parameter.
  * Excluded yielding `#to_io` methods from unsupported dispatch paths.
  * Preserved direct handling for methods with explicitly typed IO returns.

### DIFF
--- a/lib/spinel_rt.h
+++ b/lib/spinel_rt.h
@@ -1591,6 +1591,7 @@ static inline sp_bool sp_class_eq(sp_Class a, sp_Class b) {
   return (an && bn) ? strcmp(an, bn) == 0 : an == bn;
 }
 static inline const char *sp_poly_to_s(sp_RbVal v);   /* defined below; used by the user-object arm */
+static inline sp_File *sp_poly_to_file(sp_RbVal v); /* defined below; IO.select's user-#to_io unwrap */
 static inline void sp_poly_puts(sp_RbVal v) {
   switch (v.tag) {
     case SP_TAG_INT: printf("%lld\n", (long long)v.v.i); break;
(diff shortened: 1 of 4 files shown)


matz:
Merged as b658debb. Thank you -- the diagnosis is right and the repro reproduces exactly as you describe: `def to_io; @sock; end` over a constructor parameter types poly, the gate required TY_IO, and the class never entered the dispatch.

What I checked before taking it, since it changes a codegen gate:

- The reported failure on master, and the fix: `no implicit conversion of Wrap into IO` becomes a select, matching CRuby.
- Six neighbouring shapes against CRuby: a wrong return (Integer), a nil return, a typed IO return, a TY_UNKNOWN return reached through a helper method, an `attr_reader` return, and a wrapper alongside a raw socket in one select. All agree.
- The generated C for 140 programs with no user `#to_io` is byte-identical to master, so nothing else moved.
- `make gate` green here, on top of your green CI.

The two mismatches you found on the way -- a yielding `#to_io` having no standalone symbol, and a non-yielding `&block` one taking an `sp_Proc *` the dispatch was not passing -- are the more interesting half of the change, and I confirmed the first is load-bearing: dropping `m->yields` from the gate breaks the C build for a program with a reachable yielding `#to_io`.

That is the one thing I added a commit for. The `YieldWrap` case in your test is never instantiated, so its `#to_io` is unreachable and never reaches the dispatch at all -- the case passed whether or not the gate skipped yielding scopes. It is instantiated and called now, with a wrapper select beside it, and I verified it fails the build with the exclusion removed.

One known limit, worth stating rather than leaving implicit: a yielding `#to_io` is now excluded rather than supported, so `IO.select` on such a class raises TypeError where CRuby selects. That is strictly better than master (which failed the C build for the typed-return version of it) and the shape is unusual, but it is a gap. If you hit it in real code, open an issue and I will make the dispatch force a non-inlined copy for that method.

PR #4337, opened 2026-09-04
codegen: read_nonblock with exception: false returns nil at EOF

### LINKED ISSUE #4336: IO#read_nonblock with exception: false returns :wait_readable at EOF
When a socket's peer closes the connection, the next read returns
zero bytes. With `exception: false`, CRuby maps that to `nil` so
the caller can distinguish EOF from "would block". Spinel maps
zero-byte reads on a socket to `:wait_readable` instead, which
loses the distinction and turns every IO.select + read loop into
a busy spin: select says readable, read says would block, forever.

Reproducer:

```ruby
    require "socket"
    a, b = Socket.pair(Socket::AF_UNIX, Socket::SOCK_STREAM, 0)
    b.write "x"
    b.close
    p a.read_nonblock(64, exception: false)  # => "x"
    p a.read_nonblock(64, exception: false)  # CRuby: nil; Spinel: :wait_readable
```

Root cause is in `sp_sock_read_nb` (lib/sp_io.c). The zero-byte
read branch checks `exc` and raises `EOFError` when true, but on
the false branch it falls through into the EAGAIN/EWOULDBLOCK
arm that returns `:wait_readable`. The kernel returned 0 (peer
closed), not -1 with EAGAIN -- the two paths are separate and
need separate returns.

rofl0r:
sp_sock_read_nb folded EOF and would-block into the same NULL return. The codegen for exception: false then mapped NULL to :wait_readable, so the call could not tell the peer closed from the fd simply not ready yet. Every IO.select + read_nonblock loop turned into a busy spin: select said readable, read said would block, forever.

The fix adds an out-parameter so sp_sock_read_nb can signal EOF to the codegen. The exception: false path returns nil at EOF and :wait_readable when the fd is not ready. The raising form still raises EOFError at EOF. recv_nonblock shares the same runtime function and gets the same fix.

closes #4336

## Summary by CodeRabbit

- **Bug Fixes**
  - Non-blocking reads now distinguish end-of-file from temporarily unavailable data.
  - `read_nonblock` returns `nil` at end-of-file instead of incorrectly reporting `:wait_readable`, preventing unnecessary busy-waiting.
  - When exceptions are enabled, reaching end-of-file raises `EOFError` as expected.
  - Reads that would block continue to report `:wait_readable` when exceptions are disabled.

### DIFF
--- a/lib/sp_io.c
+++ b/lib/sp_io.c
@@ -851,7 +851,8 @@ const char *sp_File_readpartial(sp_File *f, sp_int n) {SP_GC_ROOT(f);
   return r;
 }
 
-const char *sp_sock_read_nb(sp_File *f, sp_int len, sp_bool exc, sp_bool is_recv) {SP_GC_ROOT(f);
+const char *sp_sock_read_nb(sp_File *f, sp_int len, sp_bool exc, sp_bool is_recv, sp_bool *eof) {SP_GC_ROOT(f);
+  if (eof) *eof = 0;
   if (is_recv) sp_sock_nb_prepare(f, "recv_nonblock");
   else if (!f || !f->fp) sp_raise_cls("IOError", "closed stream");
   if (len <= 0) return sp_str_from_bytes("", 0);
@@ -873,6 +874,7 @@ const char *sp_sock_read_nb(sp_File *f, sp_int len, sp_bool exc, sp_bool is_recv
   if (n > 0) { const char *s = sp_str_from_bytes(buf, (size_t)n); free(buf); return s; }
   if (n == 0) {
     free(buf);
+    if (eof) *eof = 1;
     if (!exc) return NULL;                     /* CRuby: nil at EOF */
     sp_raise_cls("EOFError", "end of file reached");
   }
--- a/lib/sp_io.h
+++ b/lib/sp_io.h
@@ -76,7 +76,7 @@ const char *sp_File_inspect(sp_File *f);
 const char *sp_io_kind_name(sp_File *f);
 sp_File *sp_sock_accept(sp_File *f);
 sp_File *sp_sock_accept_nb(sp_File *f, sp_bool exc);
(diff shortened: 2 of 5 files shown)


matz:
Merged as ed2fad0e. The diagnosis is right and the busy-spin it describes is real: NULL meant both "peer closed" and "not ready yet", so `IO.select` + `read_nonblock` could never make progress once the peer closed. The out-parameter is the right shape, and read_nonblock now matches CRuby exactly -- nil at EOF with `exception: false`, `:wait_readable` when not ready, EOFError in the raising form.

Two commits on top, both about the `recv_nonblock` arm the change also touched. Neither was covered by a test, which is why CI was green.

**The arm lost the opening `({` of its statement expression while keeping the closing `; })`.** `sock.recv_nonblock(64, exception: false)` on a TY_IO receiver stopped compiling:

    error: expected expression before 'sp_bool'

Master builds it, so that was a regression rather than a gap. It takes a socket from `TCPServer#accept` to reach -- `recv_nonblock` goes through a different emitter than `read_nonblock`, gated on a TY_IO receiver, which a `Socket.pair` socket does not satisfy today. That is also why your test file could not have caught it.

**And `recv_nonblock` is not `read_nonblock` at EOF.** CRuby:

    recv_nonblock(64, exception: false) at EOF  ->  ""
    recv_nonblock(64)                  at EOF  ->  ""
    read_nonblock(64, exception: false) at EOF ->  nil
    read_nonblock(64)                  at EOF  ->  EOFError

`recv_nonblock` neither answers nil nor raises; only `read_nonblock` tells the two apart. Since the runtime already knows which one it is serving (`is_recv`), the empty string is returned there. That leaves NULL meaning would-block and nothing else for the recv arm, so it goes back to the shape it had, with no EOF branch -- which incidentally is why the missing `({` was easy to miss: the arm did not need to change at all.

The test gained the three recv rows. Whole matrix verified against CRuby, `make gate` green.

Thank you -- the read_nonblock half is a good catch and a shape that would have bitten every select loop.

Issue #4334, opened 2026-09-04
Two threads that allocate are slower than one; eight spend 5.9x the CPU for 1.18x the work


rubys:
Two threads that allocate are slower than one, and eight spend 5.9x the CPU
to deliver 1.18x the work. Integer work on the same runtime, same box, same
run scales 5.68x, so this is not thread scheduling — it is what allocation
costs when more than one thread does it.

## Two standalones, identical apart from the loop body

`cpu_scale.rb` — the control. No allocation, no I/O.

```ruby
n     = (ARGV[0] || "1").to_i
iters = (ARGV[1] || "30000000").to_i
threads = []
t0 = Process.clock_gettime(Process::CLOCK_MONOTONIC)
i = 0
while i < n
  threads << Thread.new do
    x = 0
    j = 0
    while j < iters
      x += j % 7
      j += 1
    end
    x
  end
  i += 1
end
threads.each { |t| t.join }
dt = Process.clock_gettime(Process::CLOCK_MONOTONIC) - t0
puts n.to_s + " threads  " + dt.round(3).to_s + " s wall"
```

`alloc_scale.rb` — the same program with `x += j % 7` replaced by a string
build. One live reference is kept per 500k iterations so the work cannot be
optimised away entirely.

```ruby
n     = (ARGV[0] || "1").to_i
iters = (ARGV[1] || "2000000").to_i
threads = []
t0 = Process.clock_gettime(Process::CLOCK_MONOTONIC)
i = 0
while i < n
  threads << Thread.new do
    keep = ""
    j = 0
    while j < iters
      s = "row-" + j.to_s
      keep = s if j % 500000 == 0
      j += 1
    end
    keep
  end
  i += 1
end
threads.each { |t| t.join }
dt = Process.clock_gettime(Process::CLOCK_MONOTONIC) - t0
puts n.to_s + " threads  " + dt.round(3).to_s + " s wall"
```

```
$ spinel cpu_scale.rb   -o cpu_scale
$ spinel alloc_scale.rb -o alloc_scale
$ for n in 1 2 4 8; do /usr/bin/time -f "%e %U %S" \
    env SPINEL_WORKERS=12 ./cpu_scale $n 200000000; done
$ for n in 1 2 4 8; do /usr/bin/time -f "%e %U %S" \
    env SPINEL_WORKERS=12 ./alloc_scale $n 4000000; done
```

[...]

PR #4344, opened 2026-09-05
An application can name the allocator it wants

rubys:
Follows #4334. You closed that with "allocation is no longer what limits
that... open a fresh issue if you hit a specific shape" — this is a shape,
and it turned out to be small enough to be a patch instead.

## What the profile says

`6fb02909` did what it claimed: on the campfire binary under load,
`try_to_wake_up` is now **0.21%** of the profile where the contended path
used to be 23.6%. What is left in its place is the allocator itself. A
60-second `perf record` at 12 OS workers, 66,690 samples:

| bucket | share |
|---|--:|
| glibc `_int_malloc`, `cfree`, `_int_free`, `malloc_consolidate`, `malloc`, `_int_free_merge_chunk`, `free@plt`, `__libc_calloc`, `unlink_chunk` | **~35%** |
| `sp_gc_collect`, `sp_gc_mark`, `sp_gc_sweep_slot`, `sp_gc_mark_drain`, `sp_str_sweep_young` | ~13% |

Your tcmalloc probe in #4334 found nothing, and I think that reading was
correct at the time: the ask storm dominated, so the allocator underneath
it could not show. With the storm gone it does.

## The measurement

`LD_PRELOAD=libjemalloc.so.2` against an otherwise identical build, two
applications with very different shapes, median of repeats:

| | glibc | jemalloc |
|---|--:|--:|
| 420 KB page, 1 worker | 23.8 req/s | **37.5** (+58%) |
| 420 KB page, 12 workers | 44.8 | **54.6** (+25%) |
| 6 KB page, 1 worker | 7435 | **9041** (+22%) |
| 6 KB page, 12 workers | 4613 | **7336** (+59%) |

CPU per request on the 420 KB page falls 42.5 ms -> 27.1 ms at one worker.
Note the 6 KB row: glibc *regresses* from 1 worker to 12 (7435 -> 4613)
where jemalloc largely holds.

A binary **linked** this way rather than preloaded lands in the same place
— 9297 req/s on the 6 KB page at one worker, against the preload arm's
9041 — so the manifest field buys what the preload measured.

## The change

```toml
[package]
allocator = "jemalloc"
```

Two lines of behaviour: `Project` reads the key, `spin_flags` appends
`--link -l<name>` unless it is `""` or `"system"`. It goes in `spin_flags`
because that is the one place `spin build` and `spin flags` both read, so
the two cannot disagree about which allocator a build got.

**Deliberately not a default.** spinel compiles optcarrot and batch
programs as readily as servers, and a program that runs for a second and
exits spends no meaningful time in `malloc`. The right allocator is a
property of the program, not of the language, and only the manifest knows
which kind of program this is. The absent key and `"system"` both keep the
platform allocator, so nothing changes for anyone who does not ask.

**Declared and missing is a hard link error**, not a fallback:

```console
$ spin build
/usr/bin/ld: cannot find -ljemalloc: No such file or directory
```

I first wrote this with a probe and a warn-and-continue fallback, and it
was worse. A compiled program already fails this way when `-lsqlite3` has
no `libsqlite3-dev` behind it; special-casing one library would be the
inconsistency. And quietly building something slower than what was asked
for is how one machine's binary comes to differ from another's without
anyone noticing. Note this needs the development package
(`libjemalloc-dev`) — the runtime `libjemalloc2` that `LD_PRELOAD` uses is
not enough to link against.

## Verified

`spin.rb` is itself spinel-compiled, so the patch stays in the subset;
`make bin/spin` builds it. Then, on Ubuntu 24.04 / gcc 13.3:

- absent key -> no `-l` in `spin flags`; `"system"` -> no `-l`
- `"jemalloc"` -> `--link -ljemalloc`, and a real `spin build` produces a
  binary whose `ldd` shows `libjemalloc.so.2`
- a name that does not exist -> the link error above

Unrelated, but you may want to know: `make bin/spin` fails at `b80f6302`
on macOS/arm64 with `_sp_io_raise_closed` undefined, referenced from
`sp_File_gets`. It reproduces with this patch stashed, so it is not from
here. The symbol is defined at `lib/sp_io.c:246` and is present in
`libspinel_rt.a`, so it looks like a link-path issue rather than a missing
definition. Linux is fine. Happy to file it separately.

https://claude.ai/code/session_015GntgyveNAjNS98MjuchFJ

## Summary by CodeRabbit

* **New Features**
  * Projects can now select a custom memory allocator through the package manifest.
  * Allocator libraries are linked automatically during builds when configured.
  * The platform’s default allocator remains available when no custom allocator is specified.

* **Documentation**
  * Added guidance on configuring allocators, build requirements, failure behavior, and benchmark results.

### DIFF
--- a/docs/spin.md
+++ b/docs/spin.md
@@ -221,6 +221,46 @@ compiles it unless something requires it, and an excluded `.h` is still on the
 include path for the C that is compiled. An application scaffolded by
 `spin new` has no `[package]` table; add one to use the field.
 
+## Choosing an allocator
+
+A program that runs for a second and exits spends no meaningful time in
+`malloc`. A server does: it allocates for every request for as long as it is
+up, and on that shape the allocator is a measurable fraction of the whole
+profile. `allocator` names the one this program wants:
+
+```toml
+[package]
+allocator = "jemalloc"
+```
+
+Anything `-l<name>` can link is accepted, and it becomes an ordinary link
+input. `"system"` and the absent key both mean the platform's own allocator,
+which is the default and stays the default: spinel compiles batch programs and
+benchmarks as readily as servers, and the right allocator is a property of the
+program rather than of the language. Only the manifest knows which kind of
+program this is. An application scaffolded by `spin new` has no `[package]`
+table; add one to use the field.
+
+The library has to be linkable at build time, which on Debian and Ubuntu means
+the development package (`libjemalloc-dev`) and not just the runtime one
+(`libjemalloc2`). Asking for an allocator that is not installed fails the
+build:
+
+```console
+$ spin build
+/usr/bin/ld: cannot find -ljemalloc: No such file or directory
+```
+
+That is deliberate. The manifest states what the program needs, and an unmet
+statement should fail the way an unresolvable dependency does — quietly
+building something slower than what was asked for is how one machine's binary
+comes to differ from another's without anyone noticing.
+
+On two Rails-derived applications compiled by spinel, `"jemalloc"` was worth
++58% on one OS worker and +25% on twelve for a 420 KB page, and +22% and +59%
+for a 6 KB one. It is the same reason Rails ships jemalloc in its production
+image.
+
 ## Building outside spin
 
 `spin build` owns the tree it sits in. When the build is driven from somewhere
--- a/tools/spin.rb
+++ b/tools/spin.rb
@@ -939,7 +939,7 @@ end
 # --- manifest ----------------------------------------------------------------
 
 class Project
-  attr_reader :root, :name, :dep_paths
+  attr_reader :root, :name, :dep_paths, :allocator
 
   def initialize(root)
     @root = root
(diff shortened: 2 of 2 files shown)


matz:
Merged as 53117e02, with one sentence added rather than asked for.

What I checked here, since the mechanism is what we are taking on and the jemalloc numbers are the motivation for it: the whole matrix of the key against a scaffolded app. Absent, `""` and `"system"` all emit no `-l`; a linkable name emits `--link -l<name>` and `spin build` produces a binary that really links it; a name with nothing behind it fails the link and `spin build` exits 1 rather than 0. So the field does exactly what it says, and the failure is loud in the way you argued for. I could not run the jemalloc A/B, since this box has no libjemalloc at all, but that is a property of jemalloc rather than of the patch.

The sentence I added says the field is the *building* manifest's. A dependency declaring `allocator` has no effect on a dependent's program, which I confirmed. That is right, and for a reason worth stating: a process has one allocator, and a library does not get to choose it for everyone who depends on it. But it followed only from R2's rule that a dependent never compiles a dependency's `bin/`, and a reader of that docs section has no reason to have that in mind.

Worth recording that the shape of the key was questioned before it was taken. The objection was that a root-only setting does not belong in `[package]`, which is the identity table and whose every other key is meaningful in any package's manifest. It does not survive: `allocator` is not root-only. It is "the allocator for the executables this package produces", which is per-package like everything else, and looks root-only only because a dependent never builds your `bin/`. `[package]` is right, and no new table is needed. That is what the added sentence is there to make visible.

#4345 is in too. Thank you for both, and for the profile that made the case: `try_to_wake_up` at 0.21% where it was 23.6% is a satisfying thing to read.

PR #4350, opened 2026-09-05
The object under construction is rooted across its inlined initializer

blacktm:
## Why

A class whose `initialize` yields, and `Set` is one, has no standalone C symbol for that method: the constructor only allocates, and the initializer's body is inlined at each `new` site with `self` bound to the fresh object. The constructor roots that object for its own extent, which ends when it returns. So the inlined body ran with the object unrooted, and any allocation in it, an ivar's `[]` or a push into it, could collect the object out from under its own initializer:

```ruby
require "set"
sets = (0..1).map { |k| Set[Set[k], Set[k + 10], Set[k + 20]] }
p sets.map(&:inspect)
# CRuby: ["Set[Set[0], Set[10], Set[20]]", "Set[Set[1], Set[11], Set[21]]"]
# Spinel under SPINEL_GC_STRESS=1: never returns
```

That is three runs out of three. In a longer program the same miss turned `Set[1, 2]` into `Set[]`, and a user class whose yielding `initialize` fills an array ivar answered `[]` for it under the switch. Without the stress switch the swept object usually keeps its bytes long enough to look alive, which is why this went unseen.

## What

The temporary the inlined body writes through is rooted for the extent of the construction, so the object survives every allocation its own initializer makes. A value-type class is unaffected: its constructor answers the value itself. Nothing else changes.

## How

One emitted line in `emit_ctor_yield_inline` (`src/codegen_call.c`): `SP_GC_ROOT(_tN);` right after `sp_X *_tN = sp_X_new(...)`, for a heap class. Every `new` of a class whose `initialize` yields gains it, with or without a block at the site. A plain initializer runs inside the constructor, which roots `self` already, and is untouched.

## Validation

`.compat/gate.sh` GREEN: 3472 tests, 61 benchmarks, optcarrot checksum 59662, ext-cruby, rbs-seed and rubyspec all as on master; the inference fixpoint leg reports no new non-convergence. Compatibility corpus 812 to 812 of 2131: this is a memory-safety rule, not a corpus family.

New test `test/ctor_inline_root.rb`, expectations generated by Ruby 4.0.6: the shape above, a user class whose yielding initialize allocates two ivars, built with and without a block, and `Set.new(enum) { }`. Clean plain, under `SPINEL_GC_STRESS=1`, and under ASAN and UBSAN. Compiled on master it passes plain and never returns under the stress switch.

Generated-C sweep against master over 3487 programs: 3430 SAME, 57 CHANGED, 0 FAIL; every changed file gains exactly the one line per `new` site of a class whose `initialize` yields, 286 lines in all, and nothing is removed. optcarrot's generated C is byte-identical. Run under `SPINEL_GC_STRESS=1`, those 57 programs go from 40 passing on master to 50 here; the other 7 also fail on both trees from other sites (below), six the same way and one, which master cannot finish at all, here completing with one boolean wrong.

## Left alone, on purpose

Comparing two freshly constructed objects, `Set[Set[3]] == Set[Set[3]]`, leaves the receiver unrooted while the operand is built, and under the stress switch it answers false. A different site and a different rule, and the family is not exhausted: seven of the fifty-seven programs above still tear under the switch here as on master.


matz:
Merged as 3041feb9. Every claim in the report checked out here.

What I verified rather than took on trust:

- The nested-Set shape fails on master under `SPINEL_GC_STRESS=1` and passes on this branch, three runs of three. My master run tore differently from yours: `can't modify frozen Array (FrozenError)` rather than a hang, which is the same freed object read back through whatever the sweep put there next.
- `make gate` ALL GREEN, 3488 tests, checksum 59662.
- The generated-C sweep. **One caution worth passing on:** compared raw, a worktree build differs from the main tree in every `#line` directive, since those carry the compiler's own path to the package sources. That alone showed 78 changed files with hundreds of differing lines before I normalised it. With the paths normalised: 3384 same, 56 changed, **0 lines deleted, 279 added, and every one of them `SP_GC_ROOT(_tN);`**. optcarrot's generated C is identical.

The reasoning in the PR is the part I would keep in the record. A constructor that only allocates roots the object for an extent that ends before the body which fills it runs, and the body is exactly where the allocations are. That the plain-initializer case is safe for the opposite reason -- the body runs inside the constructor, whose root is still live -- is what makes the yielding case easy to miss.

Your "left alone, on purpose" section is the right call too. `Set[Set[3]] == Set[Set[3]]` leaving the receiver unrooted while the operand is built is a different site with a different rule, and folding it in here would have made a memory-safety fix into a survey. Please do open it separately if you take it on.

PR #4351, opened 2026-09-05
A container walk that meets an object already on its path stops there

blacktm:
## Why

`a = []; a << a` is legal Ruby, and so is a Struct that holds itself or a Hash that is its own value. Every walk that descends into a container followed such a cycle until the C stack ran out:

```ruby
a = []; a << a
p a                              # CRuby: [[...]]                                      Spinel: SIGSEGV
p a == [a]                       # CRuby: true                                         Spinel: SIGSEGV
p a.flatten                      # CRuby: ArgumentError "tried to flatten recursive array"   Spinel: SIGSEGV
p a.join(",")                    # CRuby: ArgumentError "recursive array join"         Spinel: SIGSEGV
h = {}; h[:h] = h; p h           # CRuby: {h: {...}}                                   Spinel: SIGSEGV
S = Struct.new(:a); s = S.new; s.a = s
p s                              # CRuby: #<struct S a=#<struct S:...>>                Spinel: SIGSEGV
t = S.new; t.a = t; p s == t     # CRuby: true                                         Spinel: never returns
b = []; b << b; p({ a => 1 }[b]) # CRuby: 1                                            Spinel: SIGSEGV
require "set"; st = Set.new; st.add(st); p st       # CRuby: Set[Set[...]]              Spinel: SIGSEGV
require "json"; JSON.generate(a)                    # CRuby: JSON::NestingError         Spinel: SIGSEGV
```

`#hash` did terminate, but only by giving up: a global depth counter answered 0 for the whole container once the walk was 24 deep, so `a.hash == [].hash` was true where CRuby says false.

A second crash sits in the same code and is fixed here because the first fix walks into it. The default `#inspect` for an object with instance variables builds its text in an `sp_String` that nothing roots, so a collection during the walk sweeps the half-built string and the appends write into freed memory:

```ruby
class N
  def initialize(k); @a = "x" * 200; @b = [1, 2, 3]; @c = k; end
end
200.times { |i| N.new(i).inspect }    # CRuby: fine    Spinel: aborts, rc 134
```

That is three runs out of three, with no GC stress and no recursion; fifty iterations still pass. The Struct and Data `#inspect` beside it has always rooted its builder.

## What

Each of those now answers what CRuby answers. `a.inspect` is `[[...]]` and `h.inspect` is `{h: {...}}`; the object the walk is standing inside is the only one elided, so `[a].inspect` is `[[[...]]]` and a sibling copy of the same object still renders in full. `==`, `eql?` and `<=>` call a repeated pair equal, so `a == b` is true and `a <=> [a, 1]` is -1 (the pair counts as equal and the lengths decide). The `eql?` a Hash lookup runs takes the same guard, so `{ a => 1 }[b]` is 1 and `{ a => 1 }.key?(b)` is true. `#hash` gives a repeated object a fixed value and mixes in the container's length and a per-type seed, so `a.hash == b.hash` is true while `a.hash == [].hash`, `a.hash == [nil].hash` and `{}.hash == [].hash` are false. Unlimited `#flatten` and `#join` raise CRuby's ArgumentError; a counted `flatten(1)` still prints `[[[...]]]`, because it terminates by counting down and CRuby does not raise there. `puts a` prints the single line `[...]`.

A Struct, a Data and a plain object render the same way: `#<struct S a=#<struct S:...>>`, `#<data D x=[#<data D:...>]>`, and `#<Node:0x... @kids=[#<Node:0x... @kids=[], @parent=#<Node:0x... ...>>], @parent=nil>`. Set gets `Set[Set[...]]`, `ArgumentError "tried to flatten recursive Set"`, and a hash that differs from `Set.new.hash`. `JSON.generate` raises `JSON::NestingError` with CRuby's message.

## How

One rule, one facility. `lib/sp_inspect.h` adds a per-fiber stack of frames `{a, b, kind}` holding the objects, or object pairs, the current walk is already inside, with `sp_poly_recur_seen` / `sp_poly_recur_push` / `sp_poly_recur_pop`. That is CRuby's `rb_exec_recursive` and `rb_exec_recursive_paired`. Frames hold raw pointers that are only ever compared, never dereferenced; the collector does not move objects, and pushing a frame cannot trigger a collection (the frames live in malloc'd storage, not on the GC heap), so a walk cannot collect the object it is remembering. The state lives in the archive (`lib/sp_inspect.c`) so the generated translation unit, the runtime and a carried package all share one path. The frames are malloc'd on the first push and doubled from 64 as a walk descends, so a cycle whose repeated object sits deeper than any fixed budget still answers; only a pointer and two counts live in thread-local storage.

A lookup scans the path. A real program's path is under ten frames, and a scan of that beats any table; but a scan at every level of a deep walk is quadratic, a 2000-deep nest would compare 2000 frames at each of its 2000 levels, where CRuby's hash stays linear. So past 32 frames the path is also indexed by an open-addressing table in `sp_inspect.c`. A slot names a frame by index and repeats its key, and is live only while that frame is still on the path with that key, so a pop lowers the count and leaves the slots to go stale on their own; a lookup checks the frame, an insert may take a stale slot, and the table is rebuilt from the live frames once half of it is in use. Nothing in it runs until a walk is deeper than the scan budget; on the hot path, push and pop each gain one compare.

Only the walks that can meet themselves carry it. In `sp_poly_eq` the recursion-capable arms (two arrays of any storage kinds, two hashes of different variants or of a poly-valued one, and two user objects compared field-wise through the generated hook) move into `sp_poly_eq_deep`; the value types above them, Time, Rational, Range, Complex and IO, and two hashes of the same primitive-keyed variant, push nothing, because none of them can contain a Ruby object. `sp_poly_eql`, `sp_poly_arr_cmp` and `sp_rbval_eql_key` share the pair frames with `==`, since all of them answer "equal" for a pair they are already inside. `sp_PolyArray_eq` gains CRuby's identity fast path and carries the guard itself, because generated code reaches it directly. In `sp_rbval_hash_key` the depth counter is gone and the array, hash and user-object arms take the path instead; the array and hash accumulators start from the length mixed with a per-type seed, CRuby's `rb_hash_start`. Only the three unlimited `flatten` walks check, and `sp_PolyArray_flatten_n` now carries a negative depth as "no limit" rather than counting down from `INT64_MAX`, so the walk that carries the path is the one the code says it is.

Any non-local exit out of a guarded walk longjmps past the pops, so every stack that can be landed on records the path depth where its arm was armed, and every longjmp into one restores it: `sp_poly_recur_mark` beside the exception stack, `sp_catch_recur_mark` beside the catch stack, `sp_brk_recur_mark` beside the break scopes, and a field on the proc-return home node. The exception, catch and break arms are each opened by exactly one runtime call (`sp_exc_check_depth`, `sp_catch_check_depth`, `sp_brk_push`), so only the proc-return home, a struct the method builds on its own C stack, needs a line in generated code. The flatten and join raise sites additionally drop their own frames first, and the at_exit hooks start from an empty path, because a raise that reached the top unhandled, or an exit from inside a walk, abandoned whatever walk it was inside; so nothing is left behind even where no handler restores the depth. A green thread carries its path and those marks across a suspension in `sp_exc_ctx_save` / `sp_exc_ctx_load`, beside the handler stacks it already carries and sized the same way, from the live depth; there is normally nothing to copy, because a yield lands between walks rather than inside one. A fiber that does yield from inside a walk pays a copy of its frames each way and a re-index on its next deep lookup: about 3 percent of a switch with a 10-deep walk parked, 12 percent with a 300-deep one, nothing with an empty path.

Two generated inspects need the guard emitted, because they recurse in generated code rather than through the runtime: the Struct and Data `#inspect`, which renders a member of its own class by calling straight back into itself, and the plain-object ivar walk in `sp_obj_inspect_sw`. Both take it only when the class has members; one with none cannot be reached from inside itself. The guard sits in each arm rather than once at the head of the dispatcher because two kinds of arm must not take it, a class with a user-defined `#inspect` (CRuby does not guard those either) and a memberless class, and the arms already encode both exclusions. The plain-object arm also gains the `SP_GC_ROOT(_s)` the Why describes. The generated Struct and Data `#hash` accumulates in `uint64_t` rather than `sp_int`: the rolling `h * 31 + x` is meant to wrap, on a signed type that is undefined behavior rather than wraparound, and a member that now contributes a large fixed constant overflows it on the next multiply.

Set does its walking in Ruby (it is Array-backed, and `e.eql?(x)` on an element compiles to a direct call that never passes through the runtime), so `packages/set/set.rb` joins the same path through three `native_func` bindings in a nested `Set::RecursionGuard` module: `enter_inspect`, `enter_eq` and `enter_flatten` answer the mark for `leave`, or -1 when the walk is already inside that Set or pair, which is the walk's cue to stop. Their frames are kinds of Set's own, so a frame the runtime pushed for the same pair is not mistaken for the Set walk's. Nothing else changes in Set: no `ensure`, because a raise, throw or break out of a walk lands on a handler that restores the depth it was armed at; no thread-local state; no method added to Set; and `require "set"` still compiles to the single-threaded runtime. The module is reachable by name from Ruby like any constant, so `leave` ignores a mark that is not a Set walk's own rather than moving the depth under a walk that is still running. `Set#hash` needs no change: it hashes each element through `sp_rbval_hash_key`, which is guarded.

JSON answers a cycle the way CRuby's json does, by depth rather than by identity: `packages/json/sp_json.c` threads a depth through `sp_json_val` and raises `JSON::NestingError` on entering the 101st nested array or object, with CRuby's message. `JSON.pretty_generate` already threaded a depth and now checks it. `lib/sp_exc.c` gains one row, `JSON::NestingError` under `JSON::ParserError`, so `e.is_a?(JSON::ParserError)` answers true as it does in CRuby.

## Validation

`.compat/gate.sh` GREEN: 3491 tests, 61 benchmarks, optcarrot checksum 59662, ext-cruby, rbs-seed and rubyspec all as on master. The inference fixpoint leg reports no new non-convergence over 3444 programs, and no round count moved.

Compatibility corpus: 847 to 863 of 2131, 16 gained, 0 lost, out of the 17 programs in the corpus that build a self-containing container. The one that does not flip compares two distinct self-containing Sets and is in Left alone below.

New tests: `test/recursive_containers.rb` and `test/recursive_set_json.rb`, expectations generated by Ruby 4.0.6, each under 10 ms compiled. Compiled on master, both die before their first line of output. Both are clean under `SPINEL_GC_STRESS=1` and under ASAN and UBSAN, plain and stressed. Between them they pin the ellipsis and equality and hash and flatten and join shapes, the Struct and Data and plain-object renders, a raise and a throw and a break and a proc's return out of a walk, a Hash lookup by an equal-but-distinct recursive key, `{}.hash != [].hash`, a two-member recursive Struct's hash, a cycle 600 levels deep, a raise rescued inside an outer Set walk and a `retry` through a nested one, a fiber parked inside a Set walk and dropped, a cycle that runs through an Array and a Set, a 2000-deep nest compared, hashed and ordered, a cycle closed 2000 levels down onto an element near the top, and fifty 40-deep nests side by side, none of which may show an ellipsis.

The threaded shapes are checked by hand rather than pinned: four threads inspecting ordinary nested Sets, four comparing them, two flattening them, and the same over recursive Sets, at 1, 4 and 16 workers, twice each, all matching CRuby line for line.

The Set test's GC-stress run is what exposed the constructor root miss fixed in #4350; this branch sits on that merge.

Generated-C sweep against master over 3506 programs: 2648 SAME, 858 CHANGED, 0 FAIL. The 806 files outside the Set package were recompiled with both binaries and every added and removed line matched against the insertions below: zero unexplained lines. The 52 that load `set` were read rather than line-matched; what they gain is listed in their own bullet.

- 683 files: a class with instance variables and no user `#inspect` gets, in its arm of `sp_obj_inspect_sw` (which the compiler already marks cold and noinline), a two-line prologue that answers `#<Name:0x... ...>` for the object the render is inside, the `sp_poly_recur_push` that follows it, the `SP_GC_ROOT(_s)` on the builder, and one `sp_poly_recur_pop` before the return. Five lines per class.
- 160 files: a generated Struct or Data `#inspect` gets the same prologue in its `"#<struct Name:...>"` form and the same one-line epilogue. Three lines per struct.
- The same 160 files: the generated Struct and Data `#hash` changes its accumulator from `sp_int` to `uint64_t` and casts the result back. Three lines per struct, changed rather than added.
- 52 files: loading the Set package, by `require "set"` in 42 of them and through the compiler's automatic require of it in the rest, re-emits the Set class body with the three walks rewritten, adds the four `extern` prototypes of the bindings, and adds the scaffold the compiler emits for any module, here `Set::RecursionGuard`: its struct and pool, an allocator whose only job is to raise NoMethodError for `new`, its arm in the object inspect and to_s dispatchers, and a row in the class tables, about twenty lines of which only the pool's start-up and shutdown hooks ever run. The binding declarations intern seven names into the symbol table (`enter_inspect`, `enter_eq`, `enter_flatten`, `leave`, `any`, `int`, `nil`, as every binding package's declarations do), and the changed body renumbers the temporaries and symbol ids in the user's own functions, which the diff shows as deleted and re-added lines. The program still links the single-threaded runtime.
- 13 files: a method that owns a proc-return frame gains `_h.recur_mark = sp_poly_recur_save();` beside the `_h.exc_top` it already sets. One line per method.

optcarrot is one of the 683 and carries nothing else. The same source through both compilers with `-S`: 13676 to 13726 lines, 50 added, 0 removed, and every one of them lies between the opening and closing brace of `sp_obj_inspect_sw` (its ten classes that have ivars and no user `#inspect`, five lines each). Nothing outside that function moved, and the emulator never calls it. Frames per second, five alternating runs of each tree's optcarrot binary, twice: the medians land within four percent of each other with the branch nominally ahead both times, which is the run-to-run spread.

A walk's cost against master, per call, medians of three (both trees built at -O2): `==` on two eight-element arrays 6.4 to 6.3 ns; `==` on `[[1, 2], [3, 4], [5, 6], [7, 8]]` 118 to 125 ns; `#hash` on the flat array 28 to 29 ns and on the nested one 59 to 78 ns; `#inspect` on the nested one 880 to 894 ns; `==` on two three-key Hashes with array values 125 to 150 ns; `==` on a two-member Struct 11.2 to 10.7 ns and on one holding an array and another Struct 47 to 50 ns; a Hash lookup keyed by `[1, [2]]` 51 to 61 ns. That is the frame push and the scan at each level a walk descends. Deep walks, where the index takes over, compared with master (seconds per twenty walks, medians of three): `==` on a 1000 / 2000 / 4000 / 8000-deep nest 0.0009 / 0.0017 / 0.0028 / 0.0069 against 0.0003 / 0.0006 / 0.0013 / 0.0026, and `#hash` on the same 0.0007 / 0.0012 / 0.0028 / 0.0047 against master's constant time, which was the depth cap answering 0. Linear on both sides; a scan alone was quadratic, 0.207 s at 8000 deep.

Set, branch against master, medians of three: 300k comparisons of `Set[1, 2, 3, 4, 5, 6, 7, 8]` with an equal Set 0.18 s on both; 100k comparisons of `Set[Set[1], Set[2], Set[3]]` with an equal Set 0.024 to 0.027 s, the three bindings and a frame per nested comparison; 200k `inspect` of the eight-element Set 0.088 to 0.090 s.

Hash VALUES change for arrays, hashes and Structs, because the accumulator now starts from the length mixed with a per-type seed. Only bucket placement depends on those numbers; iteration is insertion-ordered, and the whole suite and corpus are unchanged by it. Three collisions end as a side effect: `[0].hash == [].hash`, `[nil].hash == [].hash` and `{}.hash == [].hash` were true on master and are now false as in CRuby. Both array storage kinds take the same seed, so an IntArray `[0, 0]` still collides with the same numbers rebuilt as a PolyArray, and a Hash keyed by one still finds the other.

## Left alone, on purpose

`Set#==` on two distinct recursive Sets answers true here and false in CRuby, and everything that reaches `==` or `eql?` follows: `include?`, `subset?`, `superset?` and `intersect?` answer true where CRuby answers false, `disjoint?` false where CRuby answers true, `<=>` 0 where CRuby answers nil (even against itself, since CRuby's `<=>` goes through `subset?`), and a Hash keyed by one recursive Set finds a distinct equal-shaped one, as do `Array#uniq`, `Array#-` and `Array#include?`. CRuby's Set is a hash table, and the element's hash was stored while the Set was still empty, so its membership probe misses; Spinel's Set is Array-backed with a linear `eql?` scan and cannot reproduce that without storing per-element hashes, which is a different rule. A pair guard answering true is what makes the walk terminate at all. This is the one corpus program of the seventeen that does not flip.

Set carries a nested `RecursionGuard` module for its three bindings, which CRuby's Set does not have: `Set.constants` lists it, and `Set::RecursionGuard.respond_to?(:enter_eq)` answers true here and NameError there. It is not an API: `leave` with a mark that is not a Set walk's is ignored, and an `enter` without its `leave` leaves that object on the path for the rest of the walk it was called from, or, with no walk around it, until the next thing that restores the depth, which at top level is the start of the at_exit hooks and inside a hook is nothing at all. Nothing else about Set's surface changes.

`JSON.generate(x, max_nesting: n)`: the package's `native_func` binding takes no options, so CRuby's default of 100 is what Spinel enforces.

`rescue JSON::ParserError` does not catch a `JSON::NestingError`. A rescue arm naming a class the compiler does not know compiles to a name comparison rather than a hierarchy walk, which is how `JSON::ParserError` itself already behaves; `rescue => e` and `rescue JSON::NestingError` both catch, and `e.is_a?(JSON::ParserError)` is true.

`JSON.dump` on a recursive array raises `JSON::NestingError` here; CRuby raises `SystemStackError` for that one entry point while `JSON.generate` on the same value raises `NestingError`.

JSON of a Struct reflects it into a hash (deliberate, `native_obj_reflect`) where CRuby serializes its `to_s` as a string. Not this rule.

An anonymous Struct that holds itself prints `#<struct a=#<struct :...>>`. CRuby prints the anonymous class's own inspect, which carries an address: `#<struct a=#<struct #<Class:0x...>:...>>`. One deterministic form is worth more here than an address.

A user `#inspect`, `#==` or `#hash` that recurses through itself still exhausts the C stack. CRuby raises SystemStackError there; Spinel has no SystemStackError for C recursion. CRuby does not guard those either, so this is the same shape of program, not a new gap. `Array#hash` on a nest hundreds of thousands deep now runs off the C stack the way `==` and `inspect` always did; master answered a constant there, which was the wrong number.

A user `#hash` on an element of a recursive array runs once here and twice in CRuby: CRuby's `rb_exec_recursive_outer` restarts the whole computation from the outer object when it detects the cycle. A user `#inspect` runs exactly once on both.

The inline `Struct#hash` at a call site does not push the outer struct; the first member that recurses does, one level down. The answer is deterministic and differs from `S.new(nil).hash`, which is what the rule asks for.

The memberless-class arm of the plain-object `#inspect` keeps its unrooted builder: it appends one byte, which reallocs off the GC heap and cannot collect, so the crash in Why cannot reach it.

Comparing two freshly constructed objects, `Set[Set[3]] == Set[Set[3]]`, leaves the receiver unrooted while the operand is built, and under `SPINEL_GC_STRESS=1` it answers false on master too. A different site and a different rule; the test holds both sides in locals. Seven of the suite's programs that load Set also fail under the stress switch, six of them the same way on master and here, and one that master cannot finish at all completes here with one wrong boolean; none of those sites is this rule's, and this PR claims stress-cleanliness for its own tests only.

`spinel -E` maps a child killed by a signal to exit 1 with no message, which is why the crashes above showed as a silent exit rather than a signal. Pre-existing and separate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Recursive arrays, hashes, sets, structs, and objects now terminate safely during inspection, comparison, hashing, flattening, and joining.
  * Recursive containers display ellipses or raise the appropriate `ArgumentError` instead of overflowing the stack.
  * JSON serialization now enforces a 100-level nesting limit and raises `JSON::NestingError` when exceeded.
  * `JSON::NestingError` is now catchable with `JSON::ParserError`.

* **Tests**
  * Added comprehensive coverage for recursive container behavior, Set handling, JSON nesting, fibers, exceptions, and interrupted operations.

### DIFF
--- a/lib/sp_exc.c
+++ b/lib/sp_exc.c
@@ -333,6 +333,10 @@ const char *sp_exc_parent_of_name(const char *cls) {
     {"LoadError",             "StandardError"},
     {"RegexpError",           "StandardError"},
     {"StringScanner_Error",   "StandardError"},
+    /* the json package raises these by name; NestingError is what a document
+       too deep to serialize raises, and `rescue JSON::ParserError` catches it
+       in CRuby because it is a ParserError */
+    {"JSON::NestingError",    "JSON::ParserError"},
     {"FiberError",            "StandardError"},
     {"UncaughtThrowError",    "ArgumentError"},
     {"SyntaxError",           "ScriptError"},
--- a/lib/sp_inspect.c
+++ b/lib/sp_inspect.c
@@ -8,6 +8,145 @@
 #include "sp_inspect.h"
 #include "sp_string.h"   /* sp_String, sp_alloc.h, SP_GC_ROOT via sp_gc.h */
 #include "sp_str.h"      /* sp_sym_inspect_key for the symbol hash-key short form */
+#include <stdlib.h>   /* realloc for the growing path below */
+
+/* The walk guard's per-worker path (see sp_inspect.h). It lives here because
+   this archive unit is the one every side of the runtime already links for
+   #inspect, and the guard has to be one set of frames shared by the generated
+   translation unit, the archive and the carried packages -- a per-TU copy would
+   let a walk that crosses the boundary forget where it had been. */
+SP_TLS sp_poly_recur_frame *sp_poly_recur_stack = NULL;
+SP_TLS int sp_poly_recur_top = 0;
+SP_TLS int sp_poly_recur_cap = 0;
+/* Make room for at least `want` frames, doubling from 64. Off the hot path:
+   a walk deep enough to reach the end of the buffer has already paid far more
+   in its own recursion than this call costs. */
+#define SP_POLY_RECUR_FIRST_CAP 64   /* frames; 1.5 KB, more than any real walk */
+SP_COLD void sp_poly_recur_grow(int want) {
+  int cap = sp_poly_recur_cap ? sp_poly_recur_cap : SP_POLY_RECUR_FIRST_CAP;
+  while (cap < want) cap *= 2;
+  sp_poly_recur_frame *f = (sp_poly_recur_frame *)
+      realloc(sp_poly_recur_stack, sizeof(sp_poly_recur_frame) * (size_t)cap);
+  if (!f) sp_oom_die();
+  sp_poly_recur_stack = f;
+  sp_poly_recur_cap = cap;
+}
+
+/* ---- the deep-walk index (SP_POLY_RECUR_SCAN, sp_inspect.h) ----
+   Open addressing over the frames themselves: a slot names a frame by index
+   and repeats its key, and is live only while that frame is still on the path
+   with that key. A pop lowers sp_poly_recur_top and sp_poly_recur_ixtop and
+   leaves the slots to go stale on their own; a lookup checks the frame, an
+   insert may take a stale slot, and once the slots in use (live or stale) pass
+   half the table it is rebuilt from the live frames at a quarter load. Nothing
+   here runs until a walk is deeper than the scan budget. */
+typedef struct { const void *a, *b; int kind, frame; } sp_poly_recur_slot;
+static SP_TLS sp_poly_recur_slot *sp_poly_recur_ix = NULL;
+static SP_TLS int sp_poly_recur_ixcap = 0;    /* a power of two, or 0 before first use */
+static SP_TLS int sp_poly_recur_ixused = 0;   /* slots holding a frame, live or stale */
+SP_TLS int sp_poly_recur_ixtop = 0;
+
+/* Multipliers for the slot hash: 2^64 / phi and two more odd constants of the
+   same shape (xxHash's), so a pointer's bits spread across the word before the
+   fold and the mask. Any odd multipliers would do; these are the usual ones. */
+#define SP_POLY_RECUR_HASH_M1 0x9E3779B97F4A7C15ull
+#define SP_POLY_RECUR_HASH_M2 0xC2B2AE3D27D4EB4Full
+#define SP_POLY_RECUR_HASH_M3 0x165667B19E3779F9ull
+#define SP_POLY_RECUR_IX_MIN  256   /* slots; an eighth full at the scan budget */
+/* Callers guarantee a built table: sp_poly_recur_index rebuilds before any probe. */
+static inline int sp_poly_recur_slot_of(int kind, const void *a, const void *b) {
+  uint64_t h = ((uint64_t)(uintptr_t)a >> 4) * SP_POLY_RECUR_HASH_M1
+             ^ ((uint64_t)(uintptr_t)b >> 4) * SP_POLY_RECUR_HASH_M2
+             ^ (uint64_t)kind * SP_POLY_RECUR_HASH_M3;
+  h ^= h >> 32;
+  return (int)(h & (uint64_t)(sp_poly_recur_ixcap - 1));
+}
+static inline int sp_poly_recur_slot_live(const sp_poly_recur_slot *s) {
+  if (s->frame >= sp_poly_recur_top) return 0;
+  const sp_poly_recur_frame *f = &sp_poly_recur_stack[s->frame];
+  return f->a == s->a && f->b == s->b && f->kind == s->kind;
+}
+static void sp_poly_recur_ix_insert(int frame) {
+  const sp_poly_recur_frame *f = &sp_poly_recur_stack[frame];
+  int mask = sp_poly_recur_ixcap - 1;
+  int i = sp_poly_recur_slot_of(f->kind, f->a, f->b);
+  for (;;) {
+    sp_poly_recur_slot *s = &sp_poly_recur_ix[i];
+    if (s->frame < 0) { sp_poly_recur_ixused++; break; }
+    if (!sp_poly_recur_slot_live(s)) break;   /* a stale slot is free */
+    i = (i + 1) & mask;
+  }
+  sp_poly_recur_ix[i].a = f->a;
+  sp_poly_recur_ix[i].b = f->b;
+  sp_poly_recur_ix[i].kind = f->kind;
+  sp_poly_recur_ix[i].frame = frame;
+}
+SP_COLD static void sp_poly_recur_ix_rebuild(int frames) {
+  int cap = SP_POLY_RECUR_IX_MIN;
+  while (cap < frames * 4) cap *= 2;
+  if (cap != sp_poly_recur_ixcap) {
+    free(sp_poly_recur_ix);
+    sp_poly_recur_ix = (sp_poly_recur_slot *)malloc(sizeof(sp_poly_recur_slot) * (size_t)cap);
+    if (!sp_poly_recur_ix) sp_oom_die();
+    sp_poly_recur_ixcap = cap;
+  }
+  for (int i = 0; i < cap; i++) sp_poly_recur_ix[i].frame = -1;
+  sp_poly_recur_ixused = 0;
+  sp_poly_recur_ixtop = 0;
+}
+void sp_poly_recur_index(void) {
+  int top = sp_poly_recur_top;
+  if (sp_poly_recur_ixused + (top - sp_poly_recur_ixtop) > sp_poly_recur_ixcap / 2)
+    sp_poly_recur_ix_rebuild(top);
+  for (int i = sp_poly_recur_ixtop; i < top; i++) sp_poly_recur_ix_insert(i);
+  sp_poly_recur_ixtop = top;
+}
+int sp_poly_recur_seen_deep(int kind, const void *a, const void *b) {
+  /* a fiber switch (sp_exc_ctx_load) puts other frames under the index, and
+     the first deep lookup of a worker finds no table at all */
+  if (sp_poly_recur_ixcap == 0 || sp_poly_recur_ixtop < sp_poly_recur_top) sp_poly_recur_index();
+  int mask = sp_poly_recur_ixcap - 1;
+  int i = sp_poly_recur_slot_of(kind, a, b);
+  for (;;) {
+    const sp_poly_recur_slot *s = &sp_poly_recur_ix[i];
+    if (s->frame < 0) return 0;
+    if (s->a == a && s->b == b && s->kind == kind && sp_poly_recur_slot_live(s)) return 1;
+    i = (i + 1) & mask;
+  }
+}
+
+/* ---- the Set package's walks (prototypes and kinds in sp_inspect.h) ---- */
+static sp_int sp_poly_recur_enter(int kind, const void *a, const void *b) {
+  if (sp_poly_recur_seen(kind, a, b)) return -1;
+  return sp_poly_recur_push(kind, a, b);
+}
+sp_int sp_poly_recur_enter_inspect(sp_RbVal set) {
+  return sp_poly_recur_enter(SP_POLY_RECUR_SET_INSPECT, set.v.p, NULL);
+}
+sp_int sp_poly_recur_enter_eq(sp_RbVal set, sp_RbVal other) {
+  return sp_poly_recur_enter(SP_POLY_RECUR_SET_EQ, set.v.p, other.v.p);
+}
+sp_int sp_poly_recur_enter_flatten(sp_RbVal set) {
+  /* the Ruby raises on -1; as in the runtime's flatten, the walk's own frames
+     go before the raise (sp_poly_recur_drop_kind) */
+  if (sp_poly_recur_seen(SP_POLY_RECUR_SET_FLATTEN, set.v.p, NULL)) {
+    sp_poly_recur_drop_kind(SP_POLY_RECUR_SET_FLATTEN);
+    return -1;
+  }
+  return sp_poly_recur_push(SP_POLY_RECUR_SET_FLATTEN, set.v.p, NULL);
+}
+/* The mark comes back from Ruby, where anyone can spell the module's name. One
+   this path never handed out to a Set walk -- outside the frames, or naming a
+   frame of the runtime's own -- is ignored rather than moving the depth under
+   a walk that is still running. A Set walk's own mark whose frame a handler
+   has already dropped is beyond top, and ignored the same way. */
+void sp_poly_recur_leave(sp_int mark) {
+  if (mark < 0 || mark >= sp_poly_recur_top) return;
+  int kind = sp_poly_recur_stack[mark].kind;
+  if (kind != SP_POLY_RECUR_SET_INSPECT && kind != SP_POLY_RECUR_SET_EQ &&
+      kind != SP_POLY_RECUR_SET_FLATTEN) return;
+  sp_poly_recur_pop((int)mark);
+}
 
 const char *sp_inspect_container(sp_RbVal v) {
   /* The container is live across the sp_String allocations below; root it so a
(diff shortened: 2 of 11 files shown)


matz:
Merged as cbfe83b4, with one commit of ours on top and one fix that had to land in master first.

## The rubyspec regression, and what it turned out to be

`make gate` here failed on `hash_spec__009`, `[1, [1, []]].hash.should_not == [2, [2, []]].hash`, which passes on master. So the report's "rubyspec all as on master" does not hold as written -- but the defect is not this branch's.

Compiled directly, this branch answers CRuby on all ten of the hash shapes I checked, master gets three of them wrong, and the two array hashes here are genuinely different. What fails is the spec's own comparison: `SpecExpectation` holds the hash in an ivar, so `@v == other` goes through `sp_poly_eq`, which reached its numeric fallback **before** the tag switch and converted two `SP_TAG_INT` values to double. Above 2^53 that is lossy, and your new hash values put those two 62 apart -- `-3145750702635002333` and `-3145750702635002395` land on the same double.

Master has that bug too, reachable by any poly-slotted comparison of two large integers, a Hash lookup and `Array#include?` included. `sp_poly_cmp` has had the integer fast path in front of the same conversion all along. Fixed in 39bbfd15 with its own test; your branch merged onto it gates ALL GREEN, 3492 tests, checksum 59662.

Worth naming as method: your change did not create that failure, it *aimed* a value at it. A hash function whose outputs move is a good way to find comparisons that were only accidentally right.

## What I checked rather than took on trust

- The six crash shapes: `p a`, `h[:h] = h`, and the plain-object inspect all segfault on master here (rc 139) and answer correctly on the branch. Twenty behaviours -- ellipsis placement, `==`, `<=>`, `hash` collisions and non-collisions, `flatten` and `flatten(1)`, `join`, `puts`, the Struct and plain-object renders -- diff **identical** against Ruby 4.0.4.
- Set and JSON: `Set[Set[...]]`, the flatten message, `JSON::NestingError` with CRuby's text and `is_a?(JSON::ParserError)` true, the 100-level boundary. All match.
- The declared Set divergence reproduces exactly as you describe: two distinct recursive Sets are `true` here, `false` in CRuby.
- Both new tests: clean plain and three of three under `SPINEL_GC_STRESS=1`.
- optcarrot: 50 lines added, 0 removed, and I checked every added line number against the brace extent of `sp_obj_inspect_sw` -- all 50 inside it. That settles the fps question better than a measurement would on this box, which drifts 25 percent run to run.

One measurement caution, since your sweep and mine both had to deal with it: compared raw, a worktree build differs from the main tree in every `#line` directive, because those carry the compiler's own path to the package sources. Normalising them is what makes the sweep readable.

## The commit I added

`docs/limitations.md` now carries the two Set departures under "Intentional incompatibilities with CRuby": two distinct self-containing Sets comparing equal, with your hash-table-versus-linear-scan reasoning, and `Set::RecursionGuard` being visible through `Set.constants`. Both were argued carefully in the PR, and that file is where a user goes looking. `Set.constants` already differed from CRuby's, which lists `CoreSet` where master lists nothing, so the new name is a third answer rather than a first divergence -- but it is one a reader should be able to find.

If you ever want that constant back out of reflection, the compiler already hides synthesized *methods* from `respond_to?` and `instance_methods` by name (`name_is_synth_method`); the same rule for a synthesized constant would do it. Not a request, just where I would start.

Thank you for the "Left alone, on purpose" section. Knowing which of the seventeen corpus programs does not flip, and why, is what made this reviewable at this size.

PR #4354, opened 2026-09-06
A closed IO or Dir handle raises IOError for every operation

blacktm:
## Why

Using a handle after closing it is a bug in the program, and Ruby reports it: every read, write, positioning call, flag query and descriptor query on a closed File or IO raises IOError "closed stream", and the same calls on a closed Dir raise IOError "closed directory". Only `closed?`, `close`, `inspect` and `path` still answer. In Spinel the write side and the position queries already did that, but most of the rest answered something plausible instead: `getc` and `getbyte` answered nil, `readchar`, `readbyte`, `readpartial` and `readline` raised EOFError, `each_line`, `each_byte` and `each_char` silently ran zero iterations, `lineno`, `sync`, `autoclose?`, `close_on_exec?`, `tty?`, `binmode?` answered their defaults, `stat` answered a File::Stat, `sysseek`, `flock` and `fsync` answered 0, and a closed Dir's `read` answered nil, `tell` 0, `fileno` -1 and `rewind`/`seek` the handle. A program that kept using a handle it had closed ran on as if nothing had happened, and a `rescue IOError` around it never ran.

Concretely: sp_io.h already had the check, `SP_IO_OPEN(f)`, raising through `sp_io_raise_closed`, with a comment saying every operation should use it; about twenty entry points did not. Seven of them (`fcntl`, `pwrite`, `pread`, `stat` and the three readiness waits) carried their own hand-written copy of the same test, as did the socket entries, `IO.select` and `addr`/`peeraddr`, which answered a zero address for a closed socket. Four (`lineno`, `sync`, `autoclose?` and their setters) were not runtime calls at all: codegen read the struct field inline, so there was nowhere to put the check. Dir had no counterpart at all except in `children`/`entries`.

## What

Every IO handle operation starts with `SP_IO_OPEN`: `getc`, `readchar`, `getbyte`, `readbyte`, `ungetc`, `ungetbyte`, `readpartial` (and so `sysread`), `gets` with a separator (and so `readline`, `readlines` with a separator, and `each_line`), `each_byte`, `each_char`, `each_codepoint`, `binmode`, `binmode?`, `close_on_exec?` and `close_on_exec=`, `tty?`/`isatty`, `fileno`/`to_i` (the check now comes before the `for_fd` descriptor answer, so a closed autoclose-false wrapper raises as it does in CRuby), `stat`, `sysseek`, `flock`, `fsync`/`fdatasync`, `putc`, `advise`, `winsize`, and `wait_readable`/`wait_writable`/`wait`. `fcntl`, `pwrite`, `pread`, the waits, `stat`, the socket entries (`read_nonblock`, `write_nonblock`, `send`, `recv`, `shutdown`, `listen`, `connect_nonblock`) and `IO.select` drop their hand-rolled copies of the check in favour of the macro, and a socket's `addr` and `peeraddr` get it, so a closed socket raises instead of answering `["AF_INET", 0, "", ""]`.

The loops that read a handle in the generated code, `each_line`/`each`, `each_byte`/`each_char`/`each_codepoint` and `Dir#each`/`each_child`, now root the handle for their whole run, as the block form of `File.open` already rooted its handle. With a temporary receiver, `File.open(p).each_line { }`, `Dir.open(d).each { }`, the loop's C local was the only reference, and a body that allocates let the collector finalize the handle mid-loop: master silently read 28 of 200 lines, or 2 of 22 entries, and with the readers now raising the same program would have stopped with "closed stream". They read every line now, as CRuby does.

The typed `sync` and `sync=` emitters used to spell the receiver expression twice, so `mk.sync` evaluated `mk` twice on master; the accessor call evaluates it once, as CRuby does.

`lineno`, `lineno=`, `sync`, `sync=`, `autoclose?` and `autoclose=` get runtime accessors (`sp_File_lineno`, `sp_File_set_lineno`, `sp_File_sync_p`, `sp_File_set_sync`, `sp_File_autoclose_p`, `sp_File_set_autoclose`) that make the check and then read or write the field; codegen emits calls to those in both the typed-receiver arm and the poly-carried-handle arm instead of the inline field reads. `sync=` keeps the flush a truthy value implies, once, inside the accessor.

Dir gets the counterpart, `SP_DIR_OPEN(d)` raising through `sp_dir_raise_closed` ("closed directory"), used by `read`, `rewind`, `tell`/`pos`, `seek`/`pos=`, `fileno`, `children` and `entries`; `each` and `each_child` read through `sp_Dir_read` and so raise with it. `Dir#close` on a closed handle still answers nil, and `path` and `inspect` still work, as in CRuby.

## How

One rule, one macro; the closed-check functions are declared `SP_NORETURN SP_COLD` in sp_io.h as `sp_raise_cls` is, so the generated code past a check is not kept for a return that cannot happen. `gets` with a separator checks again after its readiness park, where another thread could have closed the handle meanwhile (a close does not yet wake a parked reader here, so that second check is a guard for the day it does). The runtime's own callers of the changed readers (`readline_sep` and `readlines_sep` through `gets_sep`, `readchar` through `getc`) want the raise, and nothing in the standard streams, ARGF or the packages calls them on a closed handle (the standard streams are never closed here; `close` on them is a no-op, unchanged). `sp_poly_as_io`, which the poly arm unboxes through, never hands back NULL (it raises NoMethodError for a non-handle), so the `_t && ...` guard the old inline `sync` read carried was dead and is gone with it.

## Validation

`make gate` (the local mirror of CI): 3490 tests, 61 benchmarks, the optcarrot checksum (59662), ext-cruby, rbs-seed, rubyspec 6/6 and the property checks all pass; the macOS-only spin-e2e progress line is the known red it is on master, and one corpus program timed out under the replay's load as it does on master. Compatibility corpus (a differential corpus of ~2,100 minimized programs that behave differently under CRuby 4.0.6 and Spinel, the same corpus behind #3999/#4003/#4029): 847 to 860 of 2131, 13 gained, 0 lost. The thirteen are the twelve closed-handle programs of this family (the readers, the flag accessors, `stat`, `tty?`, `sysseek`/`flock`/`advise`, `readline`, `each_line`, the byte and character iterators, and a closed Dir's `read` and positioning calls) plus a thirteenth that calls `each` on a closed Dir. The one program of the family that does not flip compares `io.rewind.equal?(io)` on a closed handle: its first three lines now match, and the fourth still answers false because a statically-false `equal?` between an Integer receiver and a handle folds away the receiver call, so `rewind` is never called; that is a different rule and is left for its own change.

Generated-C sweep over the suite and benchmarks against 3041feb9 with the `#line` paths normalised: 3492 same, 13 changed, 0 failed of 3505. The thirteen are the new test, eight IO tests (io_for_fd_autoclose, io_handle_value_surface, io_instance_read_surface, io_pipe_sync_write, io_poly_sync_assign, issue_3038, issue_3131, socket_handle_kind) whose inline `lineno`, `sync`, `sync=`, `autoclose?` and `autoclose=` field reads became the accessor calls, 1 to 7 lines each, and four (bm_io_wordcount, bundle_io_sys, dir_handle_objects, each_entry_returns_receiver) that iterate a temporary handle and gain the one root line at the loop's entry. Every changed line is one of those two substitutions and nothing else. optcarrot's generated C is identical, and its checksum unchanged.

New test `test/io_closed_stream.rb`, expectation generated by Ruby 4.0.6, 0.01 s compiled at -O2: every reader, writer, positioning call, flag accessor and descriptor query on a closed File, `wait` and `winsize` included; the four exempt methods; a closed pipe end read back out of its pair (the poly arm); the open-handle answers the new accessors replace (`lineno` counting across `gets`, `sync=`, `autoclose=`, `getbyte`, the EOF answers of `getc`/`getbyte`/`readbyte`); `each_line`, `each_byte`, `each_char`, `each_codepoint`, `Dir#each` and `Dir#each_child` on a temporary receiver with an allocating block, counting every line, byte, character, codepoint and entry of a twenty-file directory; `sync` and `sync=` evaluating their receiver once; `addr` and `peeraddr` on a closed TCP socket and a closed server; and every read and positioning call on a closed Dir. Compiled on 3041feb9 it prints 117 lines of which 63 are identical to the expectation; the temporary-receiver loops there lose their handle at a point the collector picks, so which of those six counts fall short varies run to run, and the one-file-and-three-entries fixture of an earlier version sometimes fell short nowhere at all, which is why the directory has twenty entries now. Clean under `SPINEL_GC_STRESS=1` and under ASAN and UBSAN, plain and stressed.

Checked by hand against CRuby and not pinned: closing the handle inside its own `each_line`, `each_byte` or `each_char` block raises "closed stream" on the next read; a pipe read end whose writer closed reads to EOF without error; `IO.select` on a closed end raises; a handle closed inside a Thread raises; `ensure` runs and sees `closed?` true. (Closing a Dir inside its own `each` segfaults CRuby 4.0.6 itself, so that shape is not compared; here it raises "closed directory".)

## Left alone, on purpose

The metadata a File answers by its path (`size`, `mtime`, `atime`, `ctime`, `birthtime`, `lstat`, `chmod`, `chown`) still answers on a closed handle here, where CRuby raises. A File::Stat is carried as the same `sp_File` struct with no descriptor, so `SP_IO_OPEN` in those accessors would refuse every File::Stat; that check needs to tell the two apart and is a different rule. The comment in sp_io.h says so.

The write family (`write`, `print`, `puts`) on a handle opened read-only answers here and raises IOError "not opened for writing" in CRuby, `wait_writable` on one answers the handle, `lineno` on a write-only handle answers 0, and `read(-1)` answers the rest of the file where CRuby raises ArgumentError: mode and argument checks, not the closed check. `pid` on a closed handle answers nil (its emitter makes no runtime call that could hold the check). `each_line` and `each` with no block are not supported on IO here (CRuby answers an Enumerator without raising).

The string those same loops yield to the block is not rooted either: `f.each_line { |l| ... }` with an allocating body yields "" for some lines (19 of 300 plain, every line under GC stress), and `each_char` likewise; on master the same, once the handle survives long enough for the string to be the thing swept. That is the yielded value's root, not the handle's, and Matz asked in #4350 that memory-safety fixes stay one rule each; it is the next change after this one, in the same three loops.

Also as on master: a close does not wake a reader parked on the handle in another thread (CRuby raises "stream closed in another thread"; both sides wait); `$stdin`/`$stdout` answer after `close`, which is a no-op on the standard streams here, and `Kernel#gets` on a closed `$stdin` is a NameError; `close_read`/`close_write` on a closed non-duplex File raise where CRuby answers nil; `pread(0, 0)` on a closed handle raises where CRuby answers ""; `io.sync = "x"` answers true, not "x"; a pipe end's `path` renders as `[]`. Two readers in spinel_rt.h that no emitter uses (`sp_File_gets_buf`, `sp_File_gets_into`) keep answering NULL, and `reopen` on a closed handle answers the handle where CRuby raises: its rebinding of the descriptor is a different shape from the check and is left for its own line.

## Summary by CodeRabbit

* **Bug Fixes**
  * Operations on closed files, pipes, sockets, and directories now consistently raise `IOError` instead of returning misleading values or silently doing nothing.
  * Closed-handle behavior now more closely matches CRuby, including appropriate `EOFError` responses.
  * File iteration and directory traversal are more reliable during garbage collection.
  * File settings such as line number, synchronization, and autoclose are handled consistently.

* **Tests**
  * Added comprehensive coverage for closed-handle behavior, iteration safety, and receiver evaluation.

### DIFF
--- a/lib/sp_cold.c
+++ b/lib/sp_cold.c
@@ -1572,8 +1572,9 @@ const char *sp_dir_home_user(const char *user);
 sp_StrArray *sp_dir_children(const char *path);
 
 const char *sp_File_gets_sep(sp_File *f, const char *sep, sp_int limit, sp_bool chomp) {SP_GC_ROOT(f);SP_GC_ROOT_STR(sep);
+  SP_IO_OPEN(f);
   sp_io_wait_readable(f);
-  if (!f || !f->fp) return NULL;
+  SP_IO_OPEN(f);   /* the park can return after another thread closed the handle */
   size_t sl = sep ? strlen(sep) : 0;
   /* fast path: the default "\n" separator with no limit reads via fgets
      (the byte-wise loop below costs a call per character) */
@@ -1630,7 +1631,7 @@ const char *sp_File_readline_sep(sp_File *f, const char *sep, sp_int limit, sp_b
   return r;
 }
 const char *sp_File_getc(sp_File *f) {SP_GC_ROOT(f);
-  if (!f || !f->fp) return NULL;
+  SP_IO_OPEN(f);
   int ch = fgetc(f->fp);
   if (ch == EOF) return NULL;
   int extra = ((ch & 0xE0) == 0xC0) ? 1 : ((ch & 0xF0) == 0xE0) ? 2 : ((ch & 0xF8) == 0xF0) ? 3 : 0;
@@ -1652,40 +1653,38 @@ const char *sp_File_readchar(sp_File *f) {SP_GC_ROOT(f);
   return r;
 }
 sp_int sp_File_getbyte(sp_File *f) {
-  if (!f || !f->fp) return SP_INT_NIL;
+  SP_IO_OPEN(f);
   int ch = fgetc(f->fp);
   return ch == EOF ? SP_INT_NIL : (sp_int)ch;
 }
 sp_RbVal sp_File_ungetc(sp_File *f, sp_RbVal v) {
-  if (f && f->fp) {
-    if (v.tag == SP_TAG_STR && v.v.s && v.v.s[0]) {
-      size_t n = sp_str_byte_len(v.v.s);
-      for (size_t i = n; i > 0; i--) ungetc((unsigned char)v.v.s[i - 1], f->fp);
-    }
-    else if (v.tag == SP_TAG_INT) ungetc((int)v.v.i, f->fp);
+  SP_IO_OPEN(f);
+  if (v.tag == SP_TAG_STR && v.v.s && v.v.s[0]) {
+    size_t n = sp_str_byte_len(v.v.s);
+    for (size_t i = n; i > 0; i--) ungetc((unsigned char)v.v.s[i - 1], f->fp);
   }
+  else if (v.tag == SP_TAG_INT) ungetc((int)v.v.i, f->fp);
   return sp_box_nil();
 }
 sp_int sp_File_sysseek(sp_File *f, sp_int off, sp_int whence) {
-  if (!f || !f->fp) return 0;
+  SP_IO_OPEN(f);
   fseek(f->fp, (long)off, whence == 1 ? SEEK_CUR : whence == 2 ? SEEK_END : SEEK_SET);
   return (sp_int)ftell(f->fp);
 }
 sp_int sp_File_flock(sp_File *f, sp_int op) {
-  if (!f || !f->fp) return 0;
+  SP_IO_OPEN(f);
   return flock(fileno(f->fp), (int)op) == 0 ? 0 : 1;
 }
 sp_int sp_File_fsync(sp_File *f) {
-  if (!f || !f->fp) return 0;
+  SP_IO_OPEN(f);
   fflush(f->fp);
   fsync(fileno(f->fp));
   return 0;
 }
 sp_RbVal sp_File_putc(sp_File *f, sp_RbVal v) {
-  if (f && f->fp) {
-    if (v.tag == SP_TAG_INT) fputc((int)(v.v.i & 0xff), f->fp);
-    else if (v.tag == SP_TAG_STR && v.v.s && v.v.s[0]) fputc(v.v.s[0], f->fp);
-  }
+  SP_IO_OPEN(f);
+  if (v.tag == SP_TAG_INT) fputc((int)(v.v.i & 0xff), f->fp);
+  else if (v.tag == SP_TAG_STR && v.v.s && v.v.s[0]) fputc(v.v.s[0], f->fp);
   return v;
 }
 const char *sp_file_ftype(const char *path) {SP_GC_ROOT_STR(path);
@@ -1942,10 +1941,10 @@ static int sp_stat_pathless(sp_File *f) {
   return f && f->fp && (!f->path || f->path[0] == 0);
 }
 sp_File *sp_io_stat_handle(sp_File *f) {SP_GC_ROOT(f);
-  if (!f) sp_raise_cls("IOError", "closed stream");
+  SP_IO_OPEN(f);
   if (f->path && f->path[0]) return sp_file_stat_handle(f->path);
   struct stat st;
-  if (!f->fp || fstat(fileno(f->fp), &st) != 0)
+  if (fstat(fileno(f->fp), &st) != 0)
     sp_raise_cls("Errno::EBADF", "Bad file descriptor");
   sp_File *h = (sp_File *)sp_gc_alloc(sizeof(sp_File), NULL, sp_file_stat_scan);
   h->fp = f->fp;
@@ -2114,7 +2113,7 @@ sp_Dir *sp_Dir_for_fd(sp_int fd) {
    to re-open, and re-opening by path would miss a directory that has since been
    renamed. Rewinds first so the listing is complete however far #read got. */
 sp_StrArray *sp_Dir_entries_h(sp_Dir *d, sp_int children) {SP_GC_ROOT(d);
-  if (!d || !d->dp) sp_raise_cls("IOError", "closed directory");
+  SP_DIR_OPEN(d);
   rewinddir(d->dp);
   sp_StrArray *a = sp_StrArray_new();
   SP_GC_ROOT(a);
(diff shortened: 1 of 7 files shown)


matz:
Merged as 40dbcbba. Every claim held here, and two of them are worth restating because they are the parts a reader would skim past.

**The matrix.** I built the same 26 calls on a closed File plus four on a closed Dir and diffed against Ruby 4.0.4. Master is wrong on 23 of the 26; this branch matches on 25. The one that differs is `winsize`, where CRuby answers NoMethodError because `IO#winsize` lives in `io/console` and spinel has it on File unconditionally. That is a pre-existing surface difference, not the closed check: master answers `[0, 0]` there, which is wrong in a different way.

**The GC bug is the bigger half of this PR and the title does not say so.** Under `SPINEL_GC_STRESS=1` on master, `Dir.open(d).each` over a 23-entry directory with an allocating block reads **5 entries** and reports success. Three runs, same answer. On this branch it reads all 23. Without the stress switch my fixture happened to survive, which is exactly what makes that class of bug expensive: the program does not crash, it quietly does less work. If you write this one up anywhere, the sentence I would lead with is that the readers raising is what turns a silent truncation into a stopped program, and that is why the root had to land in the same change rather than after it.

`mk.sync` evaluating `mk` twice on master and once here reproduces exactly (`2 4` against `1 2`, CRuby `1 2`).

**The sweep.** 11 of 3446 changed in my run against 40dbcbba's parent, and I categorised every added line mechanically: 20 accessor calls replacing inline field reads, 10 loop-entry roots, and zero of anything else. optcarrot identical. Your count of 13 is the same set plus the benchmark programs my sweep does not cover.

One note on your validation text: you recorded the macOS-only spin-e2e progress line as "the known red it is on master". It is green here on both, and `make gate` on this branch is ALL GREEN with 3494 tests. If you are seeing a red spin-e2e locally, that is worth its own look rather than being carried as expected.

**On what you left alone.** Keeping the yielded string's root out of this change is the right call and matches what was asked in #4350. `f.each_line { }` yielding `""` for some lines with an allocating body is a real bug and I would take it as its own PR whenever you get to it. The File::Stat-versus-File distinction blocking the path metadata is also the right place to stop; that check needs to tell two things apart that currently share a struct, which is a rule about the type, not about the handle.

PR #4355, opened 2026-09-06
JSON: the walk's buffer is released before the nesting error unwinds

blacktm:
## Why

A JSON document that nests too deep, or one that contains itself, ends in a `JSON::NestingError` after #4351. Raising is right, but both serializers got there holding a growing buffer, the pretty one its single buffer for the document and the compact one a separate buffer for every container it was inside, and the raise jumped over all of them. A program that keeps hitting the limit, say a server that serializes whatever it is handed and rescues the error, leaked the partial document every time and never got it back: twenty thousand rescued raises of each kind in a loop grew the process to 312 MB, most of it the pretty form's indented partial documents. CodeRabbit flagged this on #4351 after it was merged; this is the fix.

Concretely: `sp_json_pretty_val` appends into one `jbuf` owned by `sp_json_pretty`, and `sp_json_val_d` built each array and hash in a `jbuf` of its own and copied the finished text into the parent's, so at the moment the depth check fired at level 100 there was one live `malloc` buffer in the pretty walk and a hundred in the compact one, and `sp_raise_cls` longjmps past every `jb_finish` that would have freed them.

## What

The nesting error frees the walk's buffer before it raises, and the compact serializer appends into one buffer for the whole document, the way `JSON.pretty_generate` already did, so there is exactly one buffer to free in either form. The text, the 100-level limit and the message are as before. One byte-level difference, on purpose: a user `#to_json` answer is appended by its own length rather than as C text, so a NUL inside it now survives nested as well as at top level, where the old code handed the string through untouched; a new test pins both.

## How

`sp_json_val_d` becomes `sp_json_val_b(jbuf *b, v, depth)`, which appends: containers write their brackets, commas and keys straight into `b` and recurse with the same buffer; everything else goes through `sp_json_scalar`, the old leaf arms of the switch, and its string is appended. `sp_json_val` is the entry point: it answers a scalar directly and otherwise owns the one buffer, so its callers and `JSON.pretty_generate`'s leaf case see no difference. `sp_json_too_deep` takes the buffer and frees it before `sp_raise_cls`; a walk that a user `#to_json` runs inside another one owns a buffer of its own, which is not the outer one's to free. The compact form no longer copies each level into its parent; on the documents measured that is neither faster nor slower to speak of.

## Validation

Forty thousand rescued over-deep `JSON.generate` calls in a loop, compiled at the default -O2: 63 MB max RSS before, 2.2 MB after. Forty thousand `pretty_generate` calls the same way: 560 MB before, 2.4 MB after, the pretty buffer holding the indented partial document being the larger leak by far. The two interleaved, twenty thousand each: 312 MB to 2.4 MB. The review's own variants agree: a 100-deep nest whose innermost array is wide, so the buffer being freed is large, 5.4 GB to 12 MB; the loop across four threads, 122 MB to 10 MB; across fibers with a `Fiber.yield` out of the middle of a walk, 24 MB to 2.4 MB. The loop's other lines, a mixed nested document in both forms and three top-level scalars, print the same text on both builds and match Ruby 4.0.6.

`make gate` ALL GREEN on this branch on master cbfe83b4: 3493 tests, 61 benchmarks, optcarrot checksum 59662, ext-cruby, rbs-seed, rubyspec 6/6, the property checks. The differential corpus of ~2,100 minimized programs that behave differently under CRuby 4.0.6 and Spinel (the same corpus behind #3999/#4003/#4029) stays at 863 of 2131 matching, nothing gained and nothing lost, as expected for a change that alters no output. The three JSON tests in the suite (`json_parse_temp_source`, `recursive_set_json`, `user_to_json_nested`) pass plain, under `SPINEL_GC_STRESS=1`, and under ASAN and UBSAN plain and stressed; the parse test exceeds the stress-run watchdog on master and here alike, which is the parser under a collection per allocation, not a change. Generated C is untouched, since the change is inside the JSON package: 3508 suite and benchmark programs compiled with both compilers, 3508 identical, optcarrot included.

The leak itself is not observable from Ruby, and the existing recursive-Set-and-JSON test already pins the raise, its message and the 100-level boundary. The new test `test/json_user_to_json_bytes.rb`, expectation generated by Ruby 4.0.6, pins the one byte-level change: a user `#to_json` answer with a NUL inside it comes through whole, at top level and nested, as the old code did only at top level; it is clean under `SPINEL_GC_STRESS=1` and under ASAN and UBSAN, plain and stressed.

## Left alone, on purpose

A raise out of a user `#to_json` in the middle of a walk still unwinds past that walk's one buffer, as any raise from user code did before; it is one buffer now rather than one per level, and freeing it would need the raise site to know about the walk it is inside. The same holds when the nesting error fires inside a `JSON.generate` that a user `#to_json` runs: the inner walk frees its own buffer, the outer walk's is left behind. Two use-after-frees the review found on master, in the Struct and Data reflection and in the pretty form's re-parse of a user `#to_json` answer, both walking a hash nothing roots, and a GC-root imbalance when the nesting error is raised inside a green thread, are separate memory-safety rules and come next on their own. `JSON.parse` has its own depth limit and its own buffer, unchanged here. A plain String with a NUL inside it is cut at the NUL by the escaper, `sp_json_str`, here as on master, where CRuby writes `\u0000`; that is the escaper's rule, not the walk's.

## Summary by CodeRabbit

- **Bug Fixes**
  - JSON generation now preserves embedded NUL bytes returned by custom `to_json` implementations.
  - This behavior works consistently for top-level values and values nested within arrays or hashes.
  - Improved handling of deeply nested JSON structures prevents unnecessary memory retention when nesting limits are exceeded.

- **Tests**
  - Added regression coverage for NUL-byte preservation and serialized byte output.


matz:
Merged as ec2b62d6. Measured the leak here rather than taking the numbers: 20,000 rescued over-deep `JSON.generate` calls in a loop went 65 MB peak RSS on master to 2.8 MB, and `pretty_generate` 243 MB to 2.8 MB. Gate ALL GREEN, and the generated C is identical for all 3447 programs swept, which is what a change inside the package should look like.

The framing is the part I would keep: a raise that jumps over an owned buffer is not a JSON bug, it is what every `sp_raise_cls` does, and the answer here was to have one buffer rather than to teach the raise about them. That the compact form now matches the pretty form's shape is the real fix; freeing before the raise is the small half.

PR #4359, opened 2026-09-06
An IO, File, File::Stat or Dir handle answers Object's to_s, <=>, hash and object_id

blacktm:
## Why

Every Ruby value answers a handful of methods just for being an object: `to_s` renders it as `#<Class:0x...>`, `<=>` says 0 for the same object and nil for anything else, `hash` and `object_id` give it an identity. Programs use these on file handles without thinking about it: a handle as a Hash key, a log line that interpolates the stream, a `sort` over stats. In Spinel a File, IO, File::Stat or Dir handle is a typed C struct in the generated program, and for these four methods the generated C did not compile: `f.hash`, `f.object_id`, `f.to_s` and `f <=> f` each stopped the build with a C type error, and on a Dir, where the value happened to be typed, `d.to_s` answered `""` and `d <=> d` nil.

Concretely: the inference's IO block ends in "poly" for any method it does not list, so the slot holding `f.hash` was an `sp_RbVal` while the arm that emits `hash` produced an `sp_int` (the same pointer hash the Hash container buckets a handle by); `object_id` met the same slot with an `sp_int`. `to_s` and `<=>` had no arm for a handle at all and reached the end-of-dispatch placeholders, a `""` and an int-or-nil sentinel, which the poly slot refused too. The fix types `hash` and `object_id` as their arms emit them, and gives `to_s` and `<=>` a real arm through the mechanism the handles already use for `==`, `equal?`, `eql?`, `frozen?` and `freeze`: one decision function says which names Object's protocol claims on which kinds, the inference types from it and the emitter emits from it, so the two cannot drift. A boxed handle, one read out of an Array or a Hash, answers the same way through the runtime's poly `to_s` and `<=>`, so `puts [f, d]` renders and an Array of stats sorts.

## What

- `hash`, `object_id` and `__id__` on a File, IO or File::Stat handle are Integers, the handle's identity, stable across calls and agreeing with how a Hash keyed by the handle finds it.
- `to_s` on any handle of the family is Object's `#<Class:0xADDR>` under the class the handle presents as: `#<IO:0x...>` for a pipe end or a standard stream, `#<File:0x...>` for a path-backed handle, `#<File::Stat:0x...>` for a stat, `#<Dir:0x...>` for a Dir. `inspect` is untouched and keeps each handle's own render (`#<File:PATH>`, `#<Dir:PATH>`). String interpolation and `puts f`, which was a front-end rejection ("unsupported puts argument"), render the same way.
- `<=>` on any handle of the family is 0 against the same object and nil against anything else, whatever its static type, with one exception CRuby has too: two File::Stat handles order by modification time, -1, 0 or 1, the same reading `==` already takes for them (the file's mtime is read when the two are compared, as `==` reads it here; CRuby compares the values the stats captured). The same answers come back through a boxed handle, so `[sb, sa].sort`, `max` and `min` over stats work and `[f, r].sort` raises CRuby's `comparison of File with IO failed`.
- Nothing changes for `==`, `equal?`, `eql?`, `frozen?`, `freeze`, `class`, `inspect` or any IO, File, File::Stat or Dir method of its own.

## How

`ty_object_protocol_answers` (src/types.c), the one decision point for Object's protocol on the native kinds, now also claims `to_s` (no argument) and `<=>` (one argument) when the receiver is a File/IO/File::Stat handle or a Dir handle; a small predicate names that family, and the header comment says so. `infer_call` (src/analyze_infer.c), which already types from that decision, maps `to_s` to String and `<=>` to a boxed value (an Integer or nil), and its TY_IO block types `hash`, `object_id` and `__id__` as Integer, mirroring how the block already lists `nil?`, `is_a?` and `inspect`, since the catch-all `return TY_POLY` at its end is what made those slots poly. `respond_to?` is deliberately not typed there: the compile-time fold answers it from an analyze-time probe of `recv.name`, which that same catch-all answers for every name, so a typed slot would have turned today's compile refusal into a wrong `true` for `f.respond_to?(:nope)` (the review round caught exactly that).

`emit_native_object_protocol_text` (src/codegen_call_recv.c) renders `to_s` as `sp_io_to_s(f)` or `sp_Dir_to_s(d)`, and `<=>` as the receiver first, into a temporary rooted across the operand, which may allocate, then one runtime call, `sp_io_cmp` or `sp_Dir_cmp`, that reads the operand as a boxed value: a same-kind operand and nil are boxed, a poly one already is. A value of any other static kind is never the same object and cannot be a stat, and a nil handle is not it either, so that shape answers nil outright with both sides evaluated for their effects and nothing boxed (a Random has no boxed form at all, and boxing it would have stopped the compiler). The early call site in `emit_call` that already routed a File/IO handle's `==`, `equal?` and `eql?` to the arm now also routes `to_s` and `<=>` on a File/IO or Dir receiver, ahead of the generic spaceship and to_s fallbacks, and exactly those names: a first draft routed every protocol name there, which shadowed the dedicated arm for `handle != nil` (a NULL handle is nil in this backend) and made `io.wait_readable(0) != nil` answer true; the review caught it, and the test now pins that idiom. The string interpolation emitter (src/codegen_expr.c), which rendered a handle through its inspect because Object's to_s "could not be reproduced", now calls the same two helpers for the IO family, `puts` (src/codegen_stmt.c) gets one arm for a handle argument, and `sp_poly_to_s` and `sp_poly_cmp` (lib/spinel_rt.h) get one arm for a boxed File/IO or Dir each, calling the same helpers, where a boxed handle's `to_s` answered "" before (`puts [f, d]`, a handle read out of an Array) and its `<=>` answered nil even for two stats, so an Array of stats could not sort.

The runtime side is four functions in lib/sp_cold.c next to `sp_io_eq`: `sp_io_cmp` answers 0 for the same handle, orders two stat handles by mtime seconds then nanoseconds through the same `sp_stat_handle_p` / `sp_stat_handle_mtime` helpers `sp_io_eq` uses, and nil for anything else; `sp_Dir_cmp` is the identity half of that; `sp_io_to_s` formats `#<Class:0xADDR>` with the class name from the stat test and `sp_io_kind_name`, the same authority `.class` and the NoMethodError texts use, so a socket names its socket class; `sp_Dir_to_s` does the same for a Dir. A NULL handle is nil in this backend: its `to_s` is the empty string, and it is 0 against nil and nil against everything else.

## Validation

Suite, on this branch rebased onto master 40dbcbba (the merge of #4354, which touched the same header): 3495 tests, 61 benchmarks, optcarrot checksum 59662, ext-cruby, rbs-seed, rubyspec and the property checks all pass, as on master (the macOS spin-e2e "progress leaked" red is master's).

Compatibility corpus, a differential corpus of ~2,100 minimized programs that behave differently under CRuby 4.0.6 and Spinel (the same corpus behind #3999/#4003/#4029): 876 to 885 of 2131, 9 gained, 0 lost. The nine are the programs that hash, print or compare a pipe end, a File or a File::Stat handle, and the two that compare and print a Dir handle. Two more programs of the same shape stat a README.md the replay directory does not have, so they die on the missing file on both sides and stay out of the count; run where the file exists they answer like CRuby. The program that asks a pipe end `respond_to?(:path)` stays a compile refusal, for the reason under Left alone.

Generated-C sweep over the suite and the benchmarks against that master: 3508 files identical, 2 changed, none newly failing, optcarrot identical. The two: the new test, and `test/interpolate_handles.rb`, where `"#{r} #{w}"` on two pipe ends now calls `sp_io_to_s` instead of boxing the handle into the inspect renderer (the output still starts with `#<IO:`, which is what that test pins). Every other program, `==` on a Dir against a poly value included, emits the same C as master. No hot path changes: `to_s` and `<=>` on a handle were compile errors before. The fixpoint round counts show no new non-convergence.

New test: `test/io_object_protocol.rb`, expectations generated by Ruby 4.0.6, about 20 ms compiled at the default -O2. It pins the `to_s` render of a pipe end, a File, a Dir, a File::Stat and both standard streams (addresses masked), that `to_s` differs from `inspect` for each, the same render through string interpolation, `puts`, a boxed handle's own `to_s` and `Array#join`; `<=>` against the same object, another handle of the same kind, a handle of another kind, nil, an Integer, a String, a Random, and values read out of a poly Array, receiver and operand both ways; two stats of files whose mtimes were set 1,000,000 seconds apart ordered both ways, a stat against a fresh stat of the same file and against the open File of that file, two files written one after the other ordered both ways, `sort`, `max` and `min` over stats, and `[f, r].sort` raising CRuby's ArgumentError with its message; `==`, `eql?` and `equal?` unchanged next to `<=>`; `hash` and `object_id` as Integers, stable across calls, distinct between handles, `__id__` equal to `object_id`, and a Hash keyed by four handles finding each of them and neither a different handle nor a fresh stat of the same file; every answer again on the handles after `close`; `$stdout` and `STDOUT` as one object; and a handle-typed slot holding nil (an empty pipe's `wait_readable(0)`), which is 0 against nil and itself, nil against a handle, an Integer and a poly Integer, renders as the empty string, and still answers `!= nil` and `nil?` as before. The test is clean under `SPINEL_GC_STRESS=1` and under ASAN and UBSAN, plain and stressed.

## Left alone, on purpose

`respond_to?` on a File, IO or File::Stat handle still refuses to compile, as on master. Its compile-time fold reads an analyze-time probe of `recv.name`, and the IO inference block's catch-all answers that probe for every name, so typing the slot would have made `f.respond_to?(:nope)` answer true where CRuby says false; the first draft of this change did exactly that, the review caught it, and it is out. An answer by the handle's run-time kind (a pipe end, a File, a stat and a socket each have their own surface) is its own rule. A Dir handle, whose inference has no catch-all, already answers `respond_to?` correctly.

`dup` and `clone` on a handle answer the handle itself, so `f.dup.equal?(f)` is true here and false in CRuby, and the corpus program that compares a stat's `object_id` with its dup's stays where it was. Duplicating a descriptor (dup(2), the path and mode, a Dir's fresh handle) is its own rule.

`File::Stat#inspect` prints the File render, and `File.stat(p).is_a?(File::Stat)` answers false, because a stat is carried as a path-carrying File handle; `to_s` and `.class` already read the stat marker and this change follows them, but a real File::Stat render and ancestry is a separate change. Comparable's `<`, `between?` and `clamp` on two stats, which CRuby answers through `<=>`, are not wired here either, and an Array of stats that now sorts hands its elements back untyped, so `stats.sort.first.mtime` raises NoMethodError one step past where master stopped; typed elements out of `sort` are a different rule.

A reopened `class IO` or `class Dir` that defines its own `to_s`, `<=>` or `hash` is not consulted: the handle answers Object's version, as the rest of the IO surface ignores such a reopening today (`class IO; def ==(o)` is ignored on master too). A program that reopens Dir with `def to_s` fails to build with a symbol collision on the new `sp_Dir_to_s`, exactly as `def path` collides with `sp_Dir_path` on master; `Dir` is not in the compiler's reserved-prefix list (only lowercase runtime prefixes are), and fixing that is its own change. `handle == nil` on a handle-typed slot holding nil folds to false on master and here (CRuby true) through the equality arm this change leaves alone; `handle != nil` answers correctly on both.

The other heap handles are untouched: `Queue.new.to_s` and `Mutex.new.to_s` still answer `""` through the placeholder this change retires for the IO family, and a boxed Dir's `inspect` still prints `#<Object>`. Spinel's Dir implements `closed?`, which CRuby 4.0.6's Dir does not have, so `d.respond_to?(:closed?)` stays true; whether the method should exist is a separate decision. `Dir#path` still answers the same String each call.


matz:
Merged as b367b399. The probe I wrote before reading the test stops the build on master with three C type errors and matches Ruby 4.0.4 exactly here, stat ordering by mtime and `[f, r].sort` raising CRuby's message included. Sweep: 1 of 3450 changed, the `interpolate_handles` test you name. Gate ALL GREEN, 3498 tests.

I confirmed the two things the review rounds caught, since they are the interesting part. `respond_to?` on a handle is still a compile refusal, on master and here alike, so the typed slot did not turn it into a wrong `true`. And `io.wait_readable(0) != nil` still answers `false`, which is what the first draft broke.

**One pre-existing bug I found while probing that, and it is not yours.** On a handle-typed slot holding nil, `x == nil` answers `false` where CRuby answers `true`, on master and on this branch alike. `x != nil` is also `false`, so the two are inconsistent with each other, and `nil == x` is `true`. Same shape as the idiom you pinned, opposite operator. I will take it separately; flagging it here so it is on the record next to the test that pins the other half.

The decision-function approach is what makes this reviewable: one place says which names the protocol claims on which kinds, and inference and codegen both read it. That is the structure the earlier `hash`/`object_id` mismatch came from not having.

PR #4364, opened 2026-09-06
A read loop roots the String it hands the block

blacktm:
## Why

`IO#each_line` (and its alias `IO#each`), `IO#each_char`, `Dir#each` (with `#each_child` and `#each_entry`) and `ARGF.each_line` hand the block a freshly allocated String on every turn, and each of them held that string in nothing but C temporaries: the one the `while` condition assigns to, and a second one for the block's parameter. Neither is a GC root, so a block that allocates could collect the string the loop had just handed it, and read it back as an empty String.

```ruby
File.open(path).each_line do |line|
  200.times { "q" * 64 }   # any allocation will do
  puts line                # empty, for one line in fifteen here
end
```

Nothing else goes wrong, which is what makes it hard to see: the loop visits every line, stops where it should, and the program exits 0. Only the value in the block's hands is gone.

On a release build of master, with the new test's inputs — a 300-line file, a directory of 40 files, a 200-line ARGF file — 20 of the 300 lines, 48 of the 1392 characters, 2 of the 42 directory entries and 7 of the 200 ARGF lines came back empty, the same on three runs. Those are timing counts rather than a rate — the plain-run numbers shift by one or two as anything around them changes. With `SPINEL_GC_STRESS=1` every yield of all four loops came back empty, which is the answer that does not move.

A block that rebinds its parameter loses the string the same way, in the parameter's own slot — `line = line.upcase`, or `line << "x"`, which lowers to the same rebinding. So does a block that suspends its Fiber in the middle of the loop: 10 of 300 lines on master, and all 300 under stress.

## What

Four emitters in `src/codegen_call.c`, one per loop:

- `ARGF#each_line`, `#each` and `#each_string`, over `sp_argf_gets`
- `Dir#each`, `#each_child` and `#each_entry`, over `sp_Dir_read` — this one already rooted the Dir handle, but not the entry name it yields
- `IO#each_char`, over `sp_File_getc`
- `IO#each_line` and `#each`, over `sp_File_gets_sep`

Each of them now roots both slots: the loop's temporary, declared `NULL` ahead of the loop, and the block's parameter where it is bound.

## How

`SP_GC_ROOT_STR` records the address of a slot rather than the value in it, so one push before the loop covers every iteration of the loop's own temporary: each turn's fresh string lands in a slot the collector already knows to mark. That slot is set to `NULL` first because the mark walker reads it as it stands — `sp_mark_string` returns immediately on a NULL, but on an uninitialized pointer it would read the marker byte at `s[-1]`.

The block's parameter is a separate slot, bound inside the loop, so its root is pushed and popped once per iteration where the parameter is declared. That is not measurable: 2 million `each_line` yields followed by 7 million `each_char` yields answer in 0.53–0.54 s on both trees, three runs each, and optcarrot's generated C does not change at all.

This is the shape `emit_hash_filter_loop` already emits for the String keys and values it hands `Hash#delete_if` and its neighbours (`src/codegen_iter.c`), and for the same two reasons its comment gives: the block can drop the pair it was handed, or rebind the name.

`IO#each_byte` and `IO#each_codepoint` are left as they were: both hand the block an `sp_int`. `each_codepoint` does read a character string, in its loop condition, but never reads it again, so nothing can observe its collection.

## Validation

Gate GREEN on 53e77d76: the suite is 3501 pass / 0 fail (the new test is the extra one, and the spin-e2e leg is red on macOS on master too), optcarrot answers 59662 from byte-identical generated C, the 2131-program compatibility corpus stays at 905 with nothing gained and nothing lost, the inference fixpoint gains no non-convergences (its one "round count differs" line, on `test/enumerator_block_returns_self.rb`, is pre-existing: that program's generated C is byte-identical on both trees, and running the leg with master on both sides reproduces the same flip), and ASAN and UBSAN are clean on the new test both plain and under `SPINEL_GC_STRESS=1`.

The generated-C sweep over all 3516 programs in the tree is byte-identical except for eight, and in those eight every single differing line is one of the two additions — the loop slot's `= NULL` and its `SP_GC_ROOT_STR`, or the parameter's `SP_GC_ROOT_STR` — with nothing else changed: `benchmark/bm_io_wordcount.rb` (2 lines), `test/argf_reads_args.rb` (1), `test/bundle_io_sys.rb` (1), `test/dir_handle_objects.rb` (2), `test/each_entry_returns_receiver.rb` (2), `test/io_instance_read_surface.rb` (3), `test/io_closed_stream.rb` (8), and the new test (7).

`test/loop_yield_root.rb` has seven arms: one per loop, plus three that rebind the block parameter before looking at it. Every block allocates before it reads what it was handed, and counts what came back wrong. The iteration counts are printed too, so an arm that ended early would show as a changed answer rather than as a passing test. The ARGF arm reads a file named in `test/loop_yield_root.rb.args`, which the test writes itself, so no fixture joins the tree. On master the four plain arms answer `bad=20`, `bad=48`, `bad=2` and `bad=7`, the three rebinding arms answer `bad=20`, `bad=45` and `bad=2`, and under `SPINEL_GC_STRESS=1` every count is bad.

## Left alone, on purpose

`String#each_line` and `String#each_char` with a block have a related defect and a different symptom: those loops end EARLY rather than yielding an empty string. `s.each_line` delivers 14 of its 300 lines here and 1 of 300 under GC stress, and `each_char` truncates the same way when its receiver is a temporary (`array.join.each_char { ... }`) — with an ordinary receiver it is fine. What goes uncollected there is the lines array and the receiver, not the yield, so it is a different value, a different macro and a different place; it belongs in its own change. `Hash#each` has the same shape — its hoisted receiver is unrooted where its own `delete_if` twin roots one — and belongs with it.

`File.foreach`, `IO.foreach`, `Dir.foreach` and the `Dir.each_child` class method answer correctly on master, plain and under GC stress: they do not route through these four emitters and are untouched here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed built-in file, directory, character, and argument-based iteration so yielded values remain valid while the loop block runs, including when the block allocates objects or reassigns its parameter.

* **Tests**
  * Added coverage for line, character, directory, and argument-file iteration scenarios, including parameter reassignment and garbage-collection pressure.

### DIFF
--- a/src/codegen_call.c
+++ b/src/codegen_call.c
@@ -18042,8 +18042,12 @@ static void emit_call_body(Compiler *c, int id, Buf *b) {
       int bbn = 0; const int *bbb = bdy >= 0 ? nt_arr(nt, bdy, "body", &bbn) : NULL;
       int lt = ++g_tmp;
       buf_puts(b, "({ ");
-      buf_printf(b, "const char *_t%d; while ((_t%d = sp_argf_gets()) != NULL) {", lt, lt);
-      if (bpn) buf_printf(b, " const char *lv_%s = _t%d;", bpn, lt);
+      /* the line ARGF yields is a fresh string held in this C temporary and
+         nowhere else: root the slot, and the block's parameter with it, the
+         way IO#each_line below does */
+      buf_printf(b, "const char *_t%d = NULL; SP_GC_ROOT_STR(_t%d);"
+                    " while ((_t%d = sp_argf_gets()) != NULL) {", lt, lt, lt);
+      if (bpn) buf_printf(b, " const char *lv_%s = _t%d; SP_GC_ROOT_STR(lv_%s);", bpn, lt, bpn);
       for (int k = 0; k < bbn; k++) emit_stmt(c, bbb[k], b, 0);
       buf_puts(b, " } (&sp_argf_obj); })");
       return;
@@ -18158,13 +18162,17 @@ static void emit_call_body(Compiler *c, int id, Buf *b) {
       buf_puts(b, "({ ");
       /* rooted for the block's duration: a temporary receiver (Dir.open(d).each)
          has no other reference, and a body that allocates would let the
-         collector close it mid-loop (the File loops below root theirs too) */
-      buf_printf(b, "sp_Dir *_t%d = %s; SP_GC_ROOT(_t%d); const char *_t%d; ", tdh, dr, tdh, tdn);
+         collector close it mid-loop (the File loops below root theirs too).
+         The entry string the loop yields is a fresh allocation held in a C
+         temporary just the same, so it and the block's parameter take roots
+         of their own. */
+      buf_printf(b, "sp_Dir *_t%d = %s; SP_GC_ROOT(_t%d); const char *_t%d = NULL; SP_GC_ROOT_STR(_t%d); ",
+                 tdh, dr, tdh, tdn, tdn);
       buf_printf(b, "while ((_t%d = sp_Dir_read(_t%d)) != NULL) {", tdn, tdh);
       if (skip_dots)
         buf_printf(b, " if (sp_str_eq(_t%d, (&(\"\\xff\" \".\")[1])) ||"
                       " sp_str_eq(_t%d, (&(\"\\xff\" \"..\")[1]))) continue;", tdn, tdn);
-      if (dbpn) buf_printf(b, " const char *lv_%s = _t%d;", dbpn, tdn);
+      if (dbpn) buf_printf(b, " const char *lv_%s = _t%d; SP_GC_ROOT_STR(lv_%s);", dbpn, tdn, dbpn);
       for (int k = 0; k < dbbn; k++) emit_stmt(c, dbbb[k], b, 0);
       buf_printf(b, " } (sp_Dir *)_t%d; })", tdh);
       return;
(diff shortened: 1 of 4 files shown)


matz:
Merged as 2b7443b5. Thank you, and the reasoning about slot lifetime here is right.

One correction worth having on the record, since it changes what a future reader would conclude from this thread. I ran your test against **current master** rather than the branch's base, and only two of the four loops still reproduce:

| loop | master, plain | master, `SPINEL_GC_STRESS=1` |
|---|--:|--:|
| `IO#each_char` | 48 of 1392 empty | **1392 of 1392** |
| `Dir#each` | 0 | 2 of 42 |
| `IO#each_line` | 0 | 0 |
| `ARGF#each_line` | 0 | 0 |

You measured on `53e77d76`, and the GC-root wave that landed after it (#4355-#4361) moved the allocation timing out from under two of the four. That does not make those two fixes wrong, and I want to be explicit about why I took them anyway: master emits the loop slot as an uninitialized, unrooted `const char *`, and the block parameter shadows an outer, rooted `lv_line` with an unrooted one. Nothing roots either. What changed is not the defect, only whether today's allocator happens to land a collection inside the window. A fix that depends on that is not a fix.

So the useful lesson from this one is the opposite of the usual: **the count that did not move is the one worth quoting.** Your stress numbers were right for all four; the plain numbers had already gone stale by the time the branch was opened, through no fault of the branch.

Verified before merging: my own probe first, then your test, then the generated C. The sweep over the test tree changes six programs, and after removing the two additions -- the loop slot's NULL initializer with its root, and the parameter's root -- the residual diff is **zero lines**, so this is exactly additive. `make gate` ALL GREEN at 3508 tests, and every arm answers correctly plain and under stress on the merged tree.

Your "left alone, on purpose" section is the right call. `String#each_line` and `String#each_char` truncating rather than yielding empties is a different value going uncollected, and splitting it out keeps both changes readable. Please do file it.

PR #4368, opened 2026-09-06
A fresh File::Stat handle is rooted before it builds its own path string

blacktm:
## Why

`File.stat` allocates the handle it is about to answer, and then fills that handle's path with a copy of the path it was given. Making the copy allocates, so it can collect — and the only thing holding the new handle across it was a C local, which is not a GC root. So the collector swept the object the constructor was still filling in. The constructor then wrote the path and the mode into freed memory and handed the freed block back as a `File::Stat`.

Nothing looks wrong until that block is handed out again. Then the handle's mode no longer reads `"stat"`, so `File::Stat#==` stops recognising it as a stat handle and answers false for two stats of one unchanged file and `#<=>` answers non-zero for the same pair, the accessors answer for whatever owns the memory now, and a program that allocates between taking a stat and reading it either dies on a fatal signal or raises about a path it has read back short.

```ruby
sa = File.stat(path)
300.times do
  big = "x" * 200_000      # puts the string heap over its collection trigger
  sb = File.stat(path)     # ... so the copy inside this call collects
  20.times { [1, 2, 3] }   # ... and something else takes the freed block
  p sa == sb               # CRuby: true
end
```

Under `SPINEL_GC_STRESS=1` master dies on that program with SIGBUS in 6 runs of 6; without it, with SIGSEGV in 6 runs of 6. Shrunk to a shape that survives long enough to answer — 200 turns, no big string — master under GC stress answers `sa == sb` false 5 times in 200 and `sb.size == 5` false the same 5 times, identically in 6 runs of 6, where CRuby 4.0.6 answers true on all 200; written as `(sa <=> sb) == 0` it fails on the same 5 turns of 200, again 6 runs of 6, where CRuby answers 0 every time. Plain, that shrunk shape is clean in 3 runs of 3. The stressed count is the one that holds still — the same 5 of 200 on this merge-base as on the previous one — while what the fault looks like, once the collection is aimed at the constructor deliberately, moves with the allocator: the same program was SIGSEGV rather than SIGBUS under stress one merge ago.

`File.lstat` is the same constructor for the link itself, and it fails the same way: SIGSEGV plain, and under GC stress `Errno::ENOENT` for `/tmp/sp` — the path read back short — which is the same corruption from the other end. `File::Stat.new(path)` is rewritten to `File.stat(path)` before codegen, and `IO#stat` on a stream that carries a path calls `sp_file_stat_handle` directly, so both reach it too.

## What

Two constructors in `lib/sp_cold.c`, `sp_file_stat_handle` and `sp_file_lstat_handle`, each gain one line: `SP_GC_ROOT(f)` between the allocation and the `sp_sprintf` that fills `f->path`.

Nothing else changes. No generated C changes — the sweep over every program in the tree is byte-identical, optcarrot included.

## How

`SP_GC_ROOT` registers the address of a stack slot and pops it when the scope ends, so pushing it on the line after the allocation covers exactly the window the bug is in: the handle is a root from the moment it exists until the constructor returns it into a caller slot that roots it in turn.

The root is safe to push before the fields are filled because `sp_gc_alloc` zeroes the object: `sp_file_stat_scan`, which the root exposes to the collector, reads a NULL path and a NULL mode and marks neither until `sp_sprintf` lands.

This is the shape the rest of the runtime already uses for the same situation. `sp_Dir_new`, further down the same file, allocates a `sp_Dir`, sets `d->path = NULL` "set below: the sprintf may GC, and the scan reads path", pushes `SP_GC_ROOT(d)`, and only then calls `sp_sprintf`. `sp_addrinfo_new` (`lib/sp_cold.c`) and `sp_exc_new` (`lib/sp_exc.c`) do the same before building their address strings and laundering their message, and `sp_proc_compose` (`lib/spinel_rt.h`) roots its fresh `sp_ProcCompose` before handing it to `sp_proc_new_meta`, which allocates.

Of the 45 `sp_gc_alloc` calls in the runtime's out-of-line sources (`lib/*.c`), these two are the only ones where a GC allocation stands between a fresh object and its first root — with one exception, `sp_Enumerator_with_index`, described below.

## Validation

Gate GREEN on 1b185bd, against a worktree built at the merge-base: the suite is 3509 pass / 0 fail / 0 error with this test among them and the benchmarks 61 pass / 0 fail (the spin-e2e leg is red on macOS at the merge-base too), optcarrot answers 59662, the generated-C sweep over all 3524 programs in the tree is byte-identical — 3524 SAME, 0 CHANGED, 0 FAIL, optcarrot included — a differential corpus of ~2,100 minimized programs that behave differently under CRuby 4.0.6 and Spinel stays at 905 matching with nothing gained, nothing lost and no timeouts, and the inference fixpoint gains no non-convergence (the three programs that hit the round cap hit it at the merge-base too). Its one "round count differs" line is `test/enumerator_block_returns_self.rb`, 3 rounds against 4: that program's generated C is byte-identical on both trees, per the sweep above, and it is the same program that splits 3/4 run to run on master.

The gate's sanitizer leg runs clean on the new test, plain and under `SPINEL_GC_STRESS=1`, but it is not evidence for this fix and I am not offering it as such: `--cc=.compat/sancc` instruments the generated translation unit and the headers compiled into it, while `lib/sp_cold.c` is prebuilt into `lib/libspinel_rt.a` without the sanitizer, so a bad access inside these constructors has no instrumented frame — and the quarantine that holds a freed block back is exactly what stops this bug becoming visible. The differential against CRuby above is the evidence, and `SPINEL_GC_VERIFY=1` is the sharper form of it: master dies on the verifier's own mark path (`*** SPINEL_GC_VERIFY: fault on the GC mark path (signal 11)`, rc 139) in 3 runs of 3, and this branch answers `[0, 0]`, `[0, 0]`, `0` under `VERIFY` together with `STRESS`, and under `VERIFY` alone, in 3 runs of 3 each.

The root stack is per-worker thread-local, so the fix had to hold on the threaded runtime too. Four threads taking 200 `File.stat`s each, with a 200 KB string and twenty small arrays per turn: under `SPINEL_GC_STRESS=1` master answers 3 wrong of 800, then faults (rc 139), then answers 2 wrong, across three runs, and plain it faults (rc 139) in 3 runs of 3; this branch answers 0 wrong in 3 runs of 3 in both modes, which is CRuby's answer.

The two pushes cost nothing measurable in the constructor they are in: 2,000,000 `File.stat` calls answer in 5.50 s, 5.50 s and 5.51 s at the merge-base and 5.50 s, 5.52 s and 5.54 s on this branch, release builds of both, interleaved, under the machine-wide lock.

`test/stat_handle_root.rb` has three sections, one per entry point: `File.stat`, `File.lstat` and `IO#stat`. Each takes a handle before its loop, then in the loop allocates a 200 KB string (so the string heap is over its collection trigger when the constructor runs), takes a fresh handle, allocates twenty small arrays (so the freed block is handed out before the handle is read), and then checks the fresh handle against the one from before the loop and against the file it describes. Each section prints its bad counts, so a section that stopped short would show as a changed answer rather than as a passing test. Compiled on master and run one section at a time, all three fail in 3 runs of 3 in both modes: `File.stat` and `IO#stat` die with SIGSEGV under `SPINEL_GC_STRESS=1` and plain, and `File.lstat` under stress raises `Errno::ENOENT` for the truncated path `/tmp/sp` — the freed path string read back short, the same corruption from the other end — and dies with SIGSEGV plain. The whole test, master, is SIGSEGV in 6 runs of 6, 3 in each mode. On this branch it answers `[0, 0]`, `[0, 0]`, `0` in 3 runs of 3 in both modes, and under `SPINEL_GC_VERIFY=1` as well, in 0.32 s.

## Left alone, on purpose

Two things inside the two constructors I edited are wrong and stay wrong, because each is its own rule. Both answer identically on master and on this branch, byte for byte, plain and under GC stress. They raise `Errno::ENOENT` whatever `errno` says — `File.stat("/" + "a" * 5000)` raises `Errno::ENAMETOOLONG` in CRuby and `Errno::ENOENT` here — while `sp_file_open_raise`, one screen up in the same file, already reads `errno` and picks the right class. And a path with an embedded NUL is truncated rather than refused, so the constructor answers about a different file: `File.stat(path + "\0x").size` raises `ArgumentError` in CRuby and answers `5`, the size of `path`, here.

`sp_io_stat_handle`'s other branch — the one for a stream with no path, from `IO.for_fd` or a pipe end — is untouched, and not because it is sound. This rule does not reach it: it allocates its handle and then stores `sp_str_empty` and a static mode, with nothing allocating in between, so there is no window to root across. What is wrong there is a different rule. That handle takes `h->fp = f->fp` with a NULL finalizer and a scan that marks only `path` and `mode`, so nothing keeps the source stream alive; when the source is collected, `sp_File_fin` `fclose`s the `FILE` the stat handle goes on reading through `fileno`. Thirty turns that each make a fresh `IO.pipe`, take `#stat` of the reader and then drop the reader answer 30 of 30 wrong under `SPINEL_GC_STRESS=1` and 6 of 30 plain, 3 runs each, identically on master and on this branch, where CRuby answers 0; the same loop holding both ends alive answers 0 on both, in both modes. Giving a borrowed stream an owning reference means a new field on `sp_File` and a scan that follows it, which is a much wider change than this one.

The standard streams reach the *path* branch rather than that one, because they carry a placeholder path: `$stdout.stat` hands `"<STDOUT>"` to `sp_file_stat_handle`, which `stat(2)`s a name that does not exist, so `$stdout.stat.ftype` raises `Errno::ENOENT` where CRuby answers the stream's type. Same on master and here.

The same rule this PR states is broken in three more places, none of them fixed here. `sp_Enumerator_with_index` (`lib/sp_cold.c`) allocates an `sp_WiCap` capture, stores the source enumerator in it, and calls `sp_Enumerator_new_gen`, which allocates — with neither the capture nor the source held in anything but a C local, since the `SP_GC_ROOT(e)` a few lines below is in the other branch. `100.times { en = loop.with_index; 20.times { [1, 2, 3] }; en.first(3) }` dies with SIGSEGV (rc 139) on master under `SPINEL_GC_STRESS=1` in 3 runs of 3 and is clean plain, while the same loop over `loop` alone and over `[1, 2, 3].each.with_index` (the non-generator branch) is clean in 3 runs of 3 in both modes. Two unrooted candidates in one window is a diagnosis I have not done, so it wants its own change and its own test. `sp_Enumerator_new_from` in `lib/spinel_rt.h` has the identical shape for an endless range, and `200.times { en = (1..Float::INFINITY).each; 20.times { [1, 2, 3] }; en.first(3) }` dies with SIGBUS (rc 138) on master under GC stress in 3 runs of 3, clean plain. `sp_OpenStruct_new` and `sp_OpenStruct_new_from` in the same header fill `o->tbl` from `sp_SymPolyHash_new()`, which allocates, with the OpenStruct unrooted; the equivalent loop over `OpenStruct.new` answers correctly on master in 3 runs of 3 under GC stress and 3 of 3 plain, so there the window is in the source and I have no program that shows it. The last two live in the header compiled into every generated program, so a change there rewrites the generated C of every program in the tree and wants its own sweep evidence.

Three older `File::Stat` divergences sit next to all of this and belong to the shape of the type rather than to this rule — a stat handle is an `sp_File` carrying a path, not a frozen `struct stat`. It is a live view: rewrite the file and `s.size` follows it, 5 in CRuby and 12 here. The mode predicates re-resolve the path and ignore the handle's own kind, so `File.stat(link).symlink?` and `File.lstat(link).file?` are both true here and both false in CRuby. And once a handle goes through a container the accessors are gone: `[File.stat(p)][0].size` answers 5, `[File.stat(p)][0].ftype` raises `NoMethodError`, because the accessors are emitted only for a receiver whose static type is the handle. All three answer the same on master and on this branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where file metadata handles could become invalid during garbage collection while being created.
  * File and symlink status checks now remain reliable under heavy memory allocation.


matz:
Merged as 4a1645a4. Thank you.

Reproduced independently before reading your validation: 200 turns taking a second stat of an unchanged file, with allocation on either side, SIGSEGVs on master under `SPINEL_GC_STRESS=1` and answers cleanly on the branch both plain and stressed, where CRuby answers true throughout. `make gate` ALL GREEN at 3509 tests, and since the diff is two lines in `lib/sp_cold.c` no generated C can move.

The framing in your report is the part I want to endorse, because it is the one that generalises: the stressed count is what holds still, and what the fault *looks* like -- SIGBUS here, SIGSEGV a merge ago -- moves with the allocator. Two of us have now had to re-measure a plain count that had gone stale between a branch being opened and being read. The signal is the shape, not the signal number.

PR #4369, opened 2026-09-06
A fold or collect loop roots the receiver it reads on every turn

blacktm:
## Why

Seven of the fold and collect emitters begin by copying the receiver they are about to walk into a C temporary, and then read that temporary again on every turn: its entry count or its length is the loop bound, and the entry the block is handed comes out of it. None of the ten hoists they make was a GC root, and the block between two turns allocates, so a receiver that nothing else holds — the hash or the array a method call answered — could be collected while its own fold was still walking it.

```ruby
def make_hash
  h = {}
  (1..40).each { |i| h["k#{i}"] = "v#{i}" }
  h
end

make_hash.map { |k, _v| 100.times { "q" * 64 }; k }   # 29 of 40 entries on master
```

This is the rule of PR #4367 at a second family of emitters. That PR covered the builtin block loops that walk a container — `String#each_line`, `Hash#each`, `Array#each_cons` and their neighbours. These are the loops that fold a container into an answer, and they hoist the same way: `Hash#map` and `#collect`, `#select`, `#filter` and `#reject`, `#transform_keys` and `#transform_values`; `Array#min_by`, `#max_by`, `#minmax_by`, `#group_by`, `#partition`, `#grep` and `#grep_v` with a block; and `#each_with_object` on both a Hash and an Array receiver.

The symptom is usually a short walk rather than a wrong answer, which is what makes it easy to miss: the loop reads a count out of freed memory and stops, and the program exits 0 having quietly skipped the rest of the collection.

Over a 40-entry receiver whose block allocates 100 short strings a turn, three runs each on an idle machine: a plain release build of master delivers 29 of the 40 turns, and under `SPINEL_GC_STRESS=1` it delivers 1, or 2 for the Hash loops. The plain count is worth reading for what it is — it says when a collection lands inside the window, not whether the receiver can be collected, and it moves with the block: the same arms answer 15 of 40 when the block allocates twice as much. The stress column is the one that says the hoist is unrooted.

Four of the loops do not answer short at all. `Hash#select`, `#filter`, `#reject` and `#transform_keys` segfault on a plain release build of master over that same receiver, every run.

And where the answer is picked from the walk rather than accumulated, a short walk is a wrong answer rather than a partial one. `max_by` over a fresh `["a1", ..., "a40"]` answers `"a1"` under `SPINEL_GC_STRESS=1` on master, where CRuby answers `"a9"`, and `minmax_by` answers `["a1", "a1"]`: the walk ended inside the first turn, so the best-so-far it reports is the first element it saw.

## What

One file, `src/codegen_fold.c`, and ten hoists in seven emitters.

`emit_hash_collect_expr` covers `Hash#map`, `#collect`, `#select`, `#filter` and `#reject`. `emit_transform_hash_expr` covers `#transform_keys` and `#transform_values`, and hoists the receiver twice — once in the empty-block arm and once in the general one. The empty-block arm has no probe that misbehaves, since a block with no body allocates nothing of its own; it hoists the same way as its sibling and is rooted with it. `emit_minmax_by_expr` hoists twice as well: once in the `minmax_by` branch, and once in the `min_by`/`max_by` branch that takes an array receiver — the branch beside it, which materializes a Range into an array first, already rooted its hoist and says why in a comment. `emit_group_by_expr`, `emit_grep_expr` (`#grep` and `#grep_v` in their block form) and `emit_partition_expr` hoist once each. `emit_each_with_object_expr` hoists once for a Hash receiver and once for an Array receiver.

Each of the ten becomes a GC root, ahead of the loop that reads it.

## How

The root goes on the temporary itself. `SP_GC_ROOT` records the address of the slot rather than the value in it, so one push ahead of the loop covers every turn, and the pop is the cleanup attribute's, which runs on every way out of the enclosing scope — off the end, a `break`, a `break` with a value, a `return` from inside the block, a `redo`, and a raise, which restores the root stack from the mark the rescue emitters take. All ten are plain declarations in the pre-statement buffer and take the macro directly, which is how the receivers already rooted in this file are written.

Every one of these emitters has already decided the receiver is a hash or an array before it hoists — `ty_hash_cname` or `ty_is_array` gates each of them, and returns 0 otherwise — so the temporary is always a collectable pointer and the plain macro is the right one.

The roots are plain declarations in whatever C block encloses the statement, so they pop at the end of that block rather than at the end of the loop: for a fold written at method-body level the receiver stays reachable until the method returns. That is a retention window, not a leak — every `while`, `if` and `when` body is braced in the emitted C, so a fold inside one pops per turn rather than accumulating. Measured on an anonymous 200,000-element receiver, the window holds 2.6 MB for the rest of the method; writing the receiver into a local first, which any Ruby program may do, costs exactly the same.

A root also costs a slot. `_sp_gc_root_push` silently returns 0 once the root stack reaches its 65,536-entry cap, so a program deep enough to fill it stops rooting rather than saying anything, and this change puts one more slot on a fold's frame for the rest of the enclosing method. Measured on a recursion that folds a fresh hash in every frame: master answers correctly to about 27,000 frames deep and segfaults past that, and this branch reaches about 13,500. Both are well past the roughly 9,000 frames at which CRuby raises SystemStackError for the same program, so no CRuby-compatible program reaches either — but the cap is a silent one, and rooting more moves any program that does reach it nearer.

The idiom is the file's own. `Hash#sort_by`, `#sum`, `#find`, `#group_by` and `#partition`, and `Array#map`, `#filter_map` and `#each_with_index`, all root the receiver they hoist — measured by reading the C each of them emits, not by reading the emitter, because a branch that roots the same C variable makes a neighbouring branch look rooted when it is not. What the ten changed here had in common is that they hoisted and did not root.

## Validation

Gate GREEN on `d9d85ff1`: the suite is 3512 pass / 0 fail / 0 error (the new test is the extra one; the spin-e2e leg is red on macOS on master too), 61 benchmarks pass, optcarrot answers 59662, and a differential corpus of ~2,100 minimized programs that behave differently under CRuby 4.0.6 and Spinel stays at 905 matching with nothing gained and nothing lost — nothing in that corpus lands on these folds, so this change does not move it. The inference fixpoint gains no non-convergences over 3465 programs and reports no differing round counts at all.

ASAN and UBSAN are clean on the new test, plain and under `SPINEL_GC_STRESS=1`. What that covers: `--cc` instruments the generated translation unit and the collector headers inlined into it, not the prebuilt runtime archive — so it covers these loops' own reads of the temporaries this change roots, and not the collector's later walk of the root stack.

The generated-C sweep over all 3527 programs in the tree is byte-identical except for 125, and optcarrot's is unchanged — the emitters this touches are not on its path at all. Those 125 files differ by 432 lines, and those lines carry exactly 432 added roots, one apiece. Deleting `SP_GC_ROOT(_t<n>);` from the new side of each of them makes all 125 compare equal to master with nothing left over: no line moved, no line was removed, and no other text changed.

`test/fold_receiver_root.rb` has eighteen arms, one per emitter and receiver shape. Every block allocates before it looks at what it was handed, and every arm prints its turn count as well as its answer, so an arm that ends early reads as a changed line rather than as a passing test; the `min_by`/`max_by`/`minmax_by` arms print the element they picked, which a short walk gets wrong. Two arms hold the receiver in a local first — those were always safe, because the local is a root, and they are there to say so. The whole program segfaults on master.

## Left alone, on purpose

The same rule reaches another family, in `src/codegen_call_recv.c` and `src/codegen_iter.c`, and wants its own PR rather than a quiet extension of this one: `Array#find` and `#detect`, `#take_while`, `#drop_while`, `#count` with a block, `#find_index` and `#index`, and `#map!` and `#collect!` all hoist their receiver unrooted. Over the same 40-entry receiver they deliver 29 of the 40 turns on a plain build and 1 under `SPINEL_GC_STRESS=1`; `map!` segfaults on a plain build, and `take_while` answers a 1-element array where CRuby answers 40.

Seven more hoists in this same file are left for a following change, because they are a different shape rather than the same one in a new place — the loop bound is hoisted separately, or the temporary is aliased, or the walk is over a chained enumerator — and each wants its own probe and its own line in a test. All are pre-existing and none is touched here: `arr.max { |a, b| }` and `#min` with a comparator block answer `""` where CRuby answers `"a9"` and `"a1"`; `arr.sort! { |a, b| }` and `arr.sum("") { }` on the String seed arm segfault on a plain build; `each_cons(n).map` walks 9 of 39 turns and `each_slice(n).map` 1 under stress; `arr.each.with_index(n).inject { }` walks 29 of 40; and `slice_when { }.to_a.inspect` answers 176 characters where CRuby answers 236.

`Array#reduce` and `#inject` hoist their receiver the same way and are deliberately not in this PR. Rooting that hoist is not enough to make them right: the accumulator they carry across the block is itself unrooted unless the seed is an empty `[]` or `{}`, so `arr.reduce("") { |m, x| m + x }` with an allocating block answers a 0-length string on a plain build of master where CRuby answers 111, with an ASAN use-after-free in `sp_str_concat`. Rooting only the receiver moves that fault around rather than fixing it — one arm that master happens to get right answers wrong with the receiver rooted and the accumulator not. The two belong together, with their own evidence, in their own change.

`Array#flat_map`'s inner flatten loop reads its bound from an unrooted temporary too — the array the block just answered — while the push it performs can allocate. It is left alone because no probe makes it misbehave: it answers 40 of 40 on a plain build and under stress, three runs each. The window is recorded rather than closed on speculation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of hash and array iteration operations when their receivers are temporary values and the iteration block performs memory allocations.
  - Prevented iteration results from being prematurely truncated during operations such as collect, transform, grouping, filtering, partitioning, and each-with-object.

- **Tests**
  - Added coverage for temporary receivers across supported hash and array operations, including expected invocation counts and result sizes.

### DIFF
--- a/src/codegen_fold.c
+++ b/src/codegen_fold.c
@@ -303,6 +303,12 @@ int emit_hash_collect_expr(Compiler *c, int id, Buf *b) {
   { Buf rb; memset(&rb, 0, sizeof rb); emit_expr(c, recv, &rb);
     emit_indent(g_pre, g_indent); emit_ctype(c, rt, g_pre);
     buf_printf(g_pre, " _t%d = ", trecv); buf_puts(g_pre, rb.p ? rb.p : ""); buf_puts(g_pre, ";\n"); free(rb.p); }
+  /* The loop below reads this temp's `len` as its bound on every turn and
+     takes the entry out of it on every yield, and the block between two
+     turns allocates. A receiver with no other holder -- the hash a method
+     call returned -- is rooted here, as the Hash sort_by, sum and group_by
+     hoists in this file already were. */
+  emit_indent(g_pre, g_indent); buf_printf(g_pre, "SP_GC_ROOT(_t%d);\n", trecv);
 
   if (is_sel || is_rej) {
     /* select/reject: produce a same-type hash with matching pairs */
@@ -859,6 +865,9 @@ int emit_transform_hash_expr(Compiler *c, int id, Buf *b) {
     Buf rb2; memset(&rb2, 0, sizeof rb2); emit_expr(c, recv, &rb2);
     emit_indent(g_pre, g_indent); emit_ctype(c, rt, g_pre);
     buf_printf(g_pre, " _t%d = ", ts2); buf_puts(g_pre, rb2.p ? rb2.p : ""); buf_puts(g_pre, ";\n"); free(rb2.p);
+    /* rooted for the walk: the entry count below is re-read from this temp
+       on every turn and the block can allocate */
+    emit_indent(g_pre, g_indent); buf_printf(g_pre, "SP_GC_ROOT(_t%d);\n", ts2);
     emit_indent(g_pre, g_indent); emit_ctype(c, rt, g_pre);
     buf_printf(g_pre, " _t%d = sp_%sHash_new(); SP_GC_ROOT(_t%d);\n", td2, shn, td2);
     emit_indent(g_pre, g_indent);
@@ -888,6 +897,7 @@ int emit_transform_hash_expr(Compiler *c, int id, Buf *b) {
   int ts = ++g_tmp, td = ++g_tmp, ti = ++g_tmp, tk = ++g_tmp;
   Buf rb; memset(&rb, 0, sizeof rb); emit_expr(c, recv, &rb);  /* recv preludes flush to g_pre first */
   emit_indent(g_pre, g_indent); emit_ctype(c, rt, g_pre); buf_printf(g_pre, " _t%d = ", ts); buf_puts(g_pre, rb.p ? rb.p : ""); buf_puts(g_pre, ";\n"); free(rb.p);
+  emit_indent(g_pre, g_indent); buf_printf(g_pre, "SP_GC_ROOT(_t%d);\n", ts);
   emit_indent(g_pre, g_indent); emit_ctype(c, dt, g_pre); buf_printf(g_pre, " _t%d = sp_%sHash_new(); SP_GC_ROOT(_t%d);\n", td, dhn, td);
   emit_indent(g_pre, g_indent); buf_printf(g_pre, "for (sp_int _t%d = 0; _t%d < _t%d->len; _t%d++) {\n", ti, ti, ts, ti);
   emit_indent(g_pre, g_indent + 1); emit_ctype(c, skt, g_pre);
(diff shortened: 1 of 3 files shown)


matz:
Merged as b5445f21. Thank you.

Reproduced independently first: `Hash#map`, `Hash#select`, `Array#max_by` and `Array#group_by` over receivers a method answered, each with an allocating block. Master SIGSEGVs both plain and under stress; the branch answers CRuby's values both ways, including `max_by` giving `a19` where a truncated walk would report the first element.

The sweep is worth a note, because it nearly read as a much bigger change than it is. 124 of 3464 test programs move, and my first classification left 414 residual lines after removing the added roots. They were **blank lines** -- where a root is emitted on its own line, stripping the text leaves the line behind. Dropping empty lines too gives **zero residual in all 124**. Same conclusion as #4367, one normalisation step further; worth stating so the next reader of a sweep this wide does not stop at the first number. optcarrot's generated C does not move at all here.

Your framing of the two symptoms is the part I would keep: a fold that accumulates answers short and exits 0, while a fold that PICKS from the walk answers wrong -- `max_by` reporting the first element because the walk ended inside the first turn. Those look like different bugs from the outside and are one hoist.

Issue #4376, opened 2026-09-07
[GC] Thread#value reads a collected string: run_thread_once's write barrier is consumed by the body it then stores the result from


bartleusink:
## What happens

`Thread#value` can return a string the collector already took, under the
generational minor mark.

```ruby
t = Thread.new do
  acc = 0
  i = 0
  while i < 100_000
    acc += ("fill-" + i.to_s).length
    i += 1
  end
  "answer-" + acc.to_s
end
t.join

i = 0
sink = 0
while i < 200_000
  sink += ("churn-" + i.to_s).length
  i += 1
end

p t.value
p sink > 0
```

| | answer |
|---|---|
| CRuby | `"answer-988890"` |
| spinel, default build | `"answer-988890"` |
| spinel, `SPINEL_GC_MINOR=1` | `"churn-34436"` |

The value read back is a *different, later* string — the slot was recycled. It
is deterministic in this shape. `test/thread_alloc_scaling` hits the same bug
intermittently (3 runs in 10 on this machine, reading `""` or a wrong row), and
is the second of the two failures in `SPINEL_GC_MINOR=1 make test` on master.

The default build is not affected: the minor mark is off by default.

## Why

`run_thread_once` (`lib/sp_sched.c`) records the thread for the barrier on the
way **in**:

```c
static void run_thread_once(sp_thread *t) { sp_gc_wb((void*)t);
  ...
  sp_Fiber_transfer_catch(t->fiber, in, &raised, &ec, &em, &eo);   /* runs the body */
  SCHED_LOCK();
  ...
  if (t->fiber->state == 3) {
    t->retval = t->fiber->yielded_value;     /* young value, no live record */
```

The transfer runs the thread's body, which allocates. Any collection in there
clears every old object's dirty bit and empties the remembered set:

```c
/* lib/sp_gc.c */
else{ for(sp_gc_hdr*h=sp_gc_old_heap;h;h=h->next)h->dirty=0; }
sp_gc_nremembered=0; sp_gc_rem_overflow=0;
```

So by the time the retval store runs, `t` is old and the record made on entry
is gone. The value stored is allocated by that same body, and a minor mark
never walks the old list — so it is swept while the thread is holding it. The
exception fields set just below (`exc_msg`, `exc_obj`) are in the same position.

**A barrier before a call that can collect only covers the stores before the
call.** The codegen side already knows this — `gc_wb_cells` runs its barrier
after the store, and says why in its own comment — but the hand-placed barriers
[...]

PR #4377, opened 2026-09-07
Parenthesise the proc cell store so the barrier post-pass sees it

### LINKED ISSUE #4375: [GC] A captured Proc's cell store is invisible to the write-barrier post-pass, so the minor mark frees a live Proc
## What happens

A captured `Proc` stored into its capture cell gets no write barrier, so the
generational minor mark frees it while it is still live.

```ruby
i = 0
while i < 200
  Array.new(2_000, i * 1.0)      # churn, so the collector actually runs
  GC.start if (i % 50).zero?
  i += 1
end
a1 = proc { |x| x * 2 }
a2 = proc { |x| a1.call(x) + 1 }
a3 = proc { |x| a2.call(x) * 10 }
GC.start
GC.start
p a3.call(4)
```

| | answer |
|---|---|
| CRuby | `90` |
| spinel, default build | `90` |
| spinel, `SPINEL_GC_MINOR=1` | `0` |

`SPINEL_GC_MINOR=1 make test` on master is **3505 pass, 2 fail** — the existing
`test/proc_cell_capture_marked.rb` is one of the two.

The default build is not affected: the minor mark is off by default.

## Why

Codegen hoists a capture cell to scope entry, so the cell promotes on the first
sweep that reaches it — while the `Proc` stored into it can be allocated much
later. Nothing records the store, so an old cell holds a young `Proc` that is on
no remembered set, and a minor mark does not walk the old list.

There is already a mechanism for exactly this: `gc_wb_cells` (`src/codegen.c`)
finds `(*X) = v` where `X` names a cell whose declaration carries an
`sp_cell_scan_` hook, and wraps the store so `sp_gc_wb` lands on the cell. The
captured `TY_PROC` local is the one cell store it never matches, because that
path emits a **bare** deref:

```c
*_cell_a1 = (sp_int)(uintptr_t)(sp_proc_new_meta(...));   /* not (*_cell_a1) */
```

so the generated TU for the repro above contains **zero** `sp_gc_wb` calls.

## Confirmation

Three independent readings agree, so this is not inferred from the wrong answer
alone:

- **Causality.** minor off → `90`; minor on → `0`; minor on with
  `SPINEL_GC_FULL_INTERVAL=1`, so no minor cycle ever runs → `90`.
- **The collector names the holder.** `SPINEL_GC_VERIFY_GEN=1` reports
  `2 young object(s) reachable only through an old one the barrier did not
  record`, `holder scan=0x...`. ASLR slides by whole pages, so the low 12 bits
  survive, and exactly one symbol in the binary has that page offset:
  `sp_cell_scan_procint`. The count of 2 matches the `a1` and `a2` cells.
- **The generated C**, above.

#4077 was the other end of this same slot — the cell's scan function rather
than the store into it.

bartleusink:
Fixes #4375.

`gc_wb_cells` already finds `(*X) = v` where `X` names a cell the collector
scans, and wraps the store so `sp_gc_wb` lands on the cell. A captured
`TY_PROC` local is the one cell store it never matched — that path emits a
**bare** deref, `*_cell_x = (sp_int)(uintptr_t)(...)`, and the scan looks for
the parenthesised form. The generated TU carried no barrier at all.

Since codegen hoists a capture cell to scope entry, the cell promotes on the
first sweep that reaches it while the `Proc` stored into it is allocated much
later. An old cell then holds a young `Proc` on no remembered set, and a minor
mark does not walk the old list, so it is swept while live.

## Why parenthesise rather than emit the barrier at each store site

I wrote it the other way first — a `cell_needs_wb` predicate plus explicit
`sp_gc_wb` emission at the five places codegen writes a cell — and it was
worse in two ways.

`gc_wb_cells` already decides from the cell's *declaration* whether it carries
an `sp_cell_scan_` hook, which is the same question, so a second mechanism has
to re-derive it and the two can then drift apart.

More importantly, it runs the barrier **after** the store, and says why in its
own comment: the value expression allocates, and a collection between an
earlier barrier and the store clears the record that barrier just made. My
hand-placed version emitted the barrier before the store and so reintroduced a
narrower version of this very bug — narrow enough that the whole suite passed.
Reusing the existing pass gets the ordering for free.

The diff is two lines of emitted syntax as a result.

## Test

`gc-minor-test` ran one hardcoded program, which is part of why this shipped.
It takes a list now, and `test/proc_cell_capture_marked.rb` is the second leg.
I checked the leg actually discriminates: reverting `src/` and rebuilding gives

```
gc-minor-test: FAIL (proc_cell_capture_marked: SPINEL_GC_MINOR=1 output mismatch)
-90
+0
```

and it passes with the change. No new test file: the repro in #4375 is the
`a1`/`a2`/`a3` tail of that existing test, so adding it would duplicate it.

I also drafted a test for the other scanned cell kinds (celled String, Array,
Hash, poly) and **dropped it** — it passes without the fix, because those
stores already go through `gc_wb_cells`, and a test that cannot fail is not
worth the runtime.

## Gate

```
Tests:     3519 pass, 0 fail, 0 error
gc-minor-test: pass
```

`make gate` also runs `spin-check`, which fails with
`spin flags: progress leaked onto stdout`. That reproduces on unmodified
master, so it is not from this change.

## Scope

The default build is unchanged — the minor mark stays off. This only closes a
barrier gap that `SPINEL_GC_MINOR=1` exposes. Whether that default should
change is a separate question with its own measurements, and one I'd leave to
you: on a 70 MB live heap the minor mark cuts mark time 4.8x and total
collection time 3.1x, but it was enabled once and reverted, and there is a
second barrier gap behind it (#4376, fixed separately).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where Proc values assigned to captured local variables could be incorrectly reclaimed during minor garbage collection.
  * Improved reliability when running minor garbage collection with Proc and thread-local variable scenarios.

* **Tests**
  * Expanded garbage-collection coverage to include multiple barrier reproduction cases.
  * Tests now continue through remaining scenarios when one compilation fails, with separate results for each case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -782,18 +782,26 @@ reject-test: $(SPINEL)
 # only by the map between the write and the read -- under SPINEL_GC_MINOR=1.
 # It costs a fifth of a second, and it is the leg that was missing when a
 # barrier pointed at the wrong object shipped.
+# One leg per barrier gap that shipped. A single program was what this target
+# ran when a store into a capture cell shipped with no barrier at all, so a
+# fix here adds its reproducer to the list rather than testing by hand.
+GC_MINOR_TESTS := test/gc_minor_thread_local_slot.rb \
+                  test/proc_cell_capture_marked.rb
+
 gc-minor-test: $(SPINEL) $(SP_RT_LIB) $(SP_RT_MT_LIB) $(SPINEL_TIMEOUT)
 	@tmp=$$(mktemp -d /tmp/spinel-gcminor.XXXXXX); ok=1; \
-	src=test/gc_minor_thread_local_slot.rb; \
-	$(SPINEL) "$$src" -o "$$tmp/m" >/dev/null 2>&1 || \
-	  { echo "gc-minor-test: FAIL (compile)"; rm -rf "$$tmp"; exit 1; }; \
-	for mode in 0 1; do \
-	  SPINEL_GC_MINOR=$$mode $(TIMEOUT60) "$$tmp/m" > "$$tmp/out.$$mode" 2>&1; \
-	  rc=$$?; \
-	  if [ $$rc -ne 0 ]; then echo "gc-minor-test: FAIL (SPINEL_GC_MINOR=$$mode exited $$rc)"; tail -3 "$$tmp/out.$$mode"; ok=0; \
-	  elif ! cmp -s "$$tmp/out.$$mode" "$$src.expected"; then \
-	    echo "gc-minor-test: FAIL (SPINEL_GC_MINOR=$$mode output mismatch)"; \
-	    diff -u "$$src.expected" "$$tmp/out.$$mode" | head -10; ok=0; fi; \
+	for src in $(GC_MINOR_TESTS); do \
+	  bn=$$(basename "$$src" .rb); \
+	  $(SPINEL) "$$src" -o "$$tmp/$$bn" >/dev/null 2>&1 || \
+	    { echo "gc-minor-test: FAIL ($$bn: compile)"; ok=0; continue; }; \
+	  for mode in 0 1; do \
+	    SPINEL_GC_MINOR=$$mode $(TIMEOUT60) "$$tmp/$$bn" > "$$tmp/$$bn.$$mode" 2>&1; \
+	    rc=$$?; \
+	    if [ $$rc -ne 0 ]; then echo "gc-minor-test: FAIL ($$bn: SPINEL_GC_MINOR=$$mode exited $$rc)"; tail -3 "$$tmp/$$bn.$$mode"; ok=0; \
+	    elif ! cmp -s "$$tmp/$$bn.$$mode" "$$src.expected"; then \
+	      echo "gc-minor-test: FAIL ($$bn: SPINEL_GC_MINOR=$$mode output mismatch)"; \
+	      diff -u "$$src.expected" "$$tmp/$$bn.$$mode" | head -10; ok=0; fi; \
+	  done; \
 	done; \
 	rm -rf "$$tmp"; \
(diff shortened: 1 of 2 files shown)


matz:
Merged as eb95a8eb; #4375 closed with it.

I verified this in the **generated C** rather than by the answer, because the answer does not move on this box -- the reproducer prints 90 on master in both modes here. The evidence is unambiguous anyway:

| | `sp_gc_wb` in the TU | the store |
|---|--:|---|
| master | **0** | `*_cell_a1 = ...` |
| branch | **2** | `(*_cell_a1) = ...` |

A translation unit that carries no barrier at all is a defect whatever today's allocator does with it, and that is the same call I made on two of the four loops in #4364. The sweep moves seven test programs, all proc-cell shapes, and the residual after removing the added parentheses and barriers is zero lines.

Two things in how you did this are worth naming, because they are the reason it is a two-line diff and not a new mechanism.

**Reusing `gc_wb_cells` rather than emitting the barrier yourself.** Your own account of writing the hand-placed version first, putting the barrier before the store, and reintroducing a narrower form of the same bug -- narrow enough that the whole suite passed -- is the strongest argument for the version you shipped. The pass already runs the barrier after the store and already says in its comment why. Joining it gets the ordering for free; deriving "does this cell carry a scan hook" a second time invites the two answers to drift.

**Widening the target that missed it.** `gc-minor-test` running one hardcoded program is part of why this shipped, and you said so and fixed it in the same change rather than adding a leg by hand. That is the right instinct: a gap that a gate could not see is two defects, and only one of them is in the emitter.

And dropping the test for the other cell kinds because it passes without the fix is the right call. A test that cannot fail costs runtime and buys nothing.

Issue #4311, opened 2026-09-03
minor-GC-only: under `SPINEL_GC_MINOR=1` the campfire binary's fiber lane segfaults in `sp_PolyPolyHash_get` on its first room-page render (a `Thread.current[...]` Hash read inside a Fiber on the main thread)


rubys:
Filed the way #4260 asked: a divergence that appears only under `SPINEL_GC_MINOR=1`, from the campfire vehicle. It is on the binary's **fiber lane** — `TEP_SERVER=fiber`, added today in roundhouse c9ed5e42 so the two server designs measure on one binary (#4306) — where every request is a `Fiber` on the main thread. The threaded server, same binary, same variable, serves the same page.

## Reproduction

```
curl -sL https://rubys.github.io/roundhouse/campfire/spinel.tgz | tar xz && cd campfire
spin build
SPINEL_GC_MINOR=1 TEP_SERVER=fiber SPINEL_WORKERS=1 PORT=3000 ./build/bin/campfire &
curl -s -c jar -o /dev/null -X POST http://localhost:3000/first_run \
  --data-urlencode 'user[name]=Walker' --data-urlencode 'user[email_address]=w@example.com' --data-urlencode 'user[password]=secret123'
curl -s -b jar -o /dev/null -w '%{http_code}\n' http://localhost:3000/rooms/1
```

The first account is created (302), and the first room-page render kills the process. Deterministic: 5 of 5 on Linux/gcc (12-core Ubuntu 24.04), spinel `790db4cb`, archive built from roundhouse `c9ed5e42`. The same commands without `SPINEL_GC_MINOR=1` answer 200; the same commands with `SPINEL_GC_MINOR=1` but without `TEP_SERVER=fiber` (the threaded server) answer 200, and that server serves 8 seconds of 16-way load under the variable without a fault.

## Where it dies

```
Thread 1 "campfire" received signal SIGSEGV
#0  sp_PolyPolyHash_get
#1  sp_ViewHelpers_s_get_slot
#2  sp_ViewHelpers_s_content_for_set
#3  sp_Views__Rooms_s_show
#4  sp_RoomsController_show
#5  sp_Main_s_dispatch
#6  sp_App_dispatch.isra.0
#7  sp_Scheduled_s_handle_one
#8  _fiber_body_4
```
The other two OS threads are a parked helper worker and the monitor in `poll`.

[...]

PR #4378, opened 2026-09-07
Record the thread again after the body it takes the result from

### LINKED ISSUE #4376: [GC] Thread#value reads a collected string: run_thread_once's write barrier is consumed by the body it then stores the result from

bartleusink:
Fixes #4376.

`run_thread_once` records `t` for the write barrier on the way **in**, then
transfers to the fiber that runs the thread's body. The body allocates, and any
collection in there clears every old object's dirty bit and empties the
remembered set — so by the time the stores after the transfer run, the record
made on entry is gone. `retval` and the exception fields are all allocated by
that body, so an old `t` is left holding young values nothing recorded, and a
minor mark does not walk the old list.

`Thread#value` then reads a *different, later* string — the slot was recycled:

```
p t.value    # "churn-34436" under SPINEL_GC_MINOR=1; CRuby says "answer-988890"
```

`test/thread_alloc_scaling` hit this intermittently, 3 runs in 10 on this
machine, reading `""` or a wrong row. It is 20 in 20 clean now.

## The rule was already written down

A barrier before a call that can collect only covers the stores before the
call. `gc_wb_cells` makes exactly this choice on the codegen side and explains
it in its own comment — it just wasn't checked against the barriers placed by
hand in the runtime. Moving the record to *after* the stores rather than adding
a second one keeps that single rule.

## Not gated on `SP_THREADS`

Green threads run on the plain archive too, and there the cycle clears the
dirty bit of everything the remembered set names, `t` included — so this is not
a threads-only bug and the fix should not be conditional. It does mean the
non-threaded runtime object changes.

## Test

`gc-minor-test` ran one hardcoded program, which is part of why this shipped.
It takes a list now, with `test/gc_minor_thread_retval.rb` as its second leg —
a small deterministic program rather than `thread_alloc_scaling`, which is
heavy (4 threads x 20k plus a 400k loop) and only intermittently wrong, so it
would make a slow flaky gate leg.

I checked the leg discriminates: reverting `lib/` and rebuilding gives

```
gc-minor-test: FAIL (gc_minor_thread_retval: SPINEL_GC_MINOR=1 output mismatch)
-"answer-988890"
+"churn-34436"
```

and it passes with the change.

## Gate

```
Tests:     3520 pass, 0 fail, 0 error
gc-minor-test: pass
```

`make gate` also runs `spin-check`, which fails with
`spin flags: progress leaked onto stdout`. That reproduces on unmodified
master, so it is not from this change.

## Relationship to #4377

Independent — different file, different mechanism, different test. They only
overlap on the `gc-minor-test` list generalisation, so whichever lands second
rebases to a one-line addition.

The default build is unchanged: the minor mark stays off. Whether that should
change is a separate question I'd leave to you — on a 70 MB live heap it cuts
mark time 4.8x and total collection time 3.1x, but it was enabled once and
reverted, and these two gaps are what `SPINEL_GC_MINOR=1 make test` was
failing on.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where values returned by threads could become invalid after minor garbage collection.
  * Thread results now remain readable even after substantial allocation and garbage-collection activity.

* **Tests**
  * Added regression coverage for thread return values involving newly allocated strings.
  * Expanded minor garbage-collection testing to run multiple fixtures under different collection modes, with clearer failure reporting.


matz:
Merged as 6b6841ce.

I reproduced it before reading past your diagnosis. On this machine your
program does not answer the wrong string under `SPINEL_GC_MINOR=1`, it
segfaults: master is 5 for 5 SIGSEGV, the branch is 10 for 10 correct and
matches CRuby 4.0.4 byte for byte. Same defect, further along.

`make gate` on the merge is green: 3520 tests, bench 61, optcarrot checksum
59662 at 2258.9 fps, rubyspec 6 legs, spin-e2e. `src/` is untouched, so the
generated C is identical by construction and there was nothing to sweep.

Two notes.

**The conflict.** #4377 landed first and turned `gc-minor-test`'s single
program into a list, with the same comment yours has, for the same reason.
The resolution keeps one list with both legs plus a third (below).

**The record on entry stays.** Your text says you moved the record rather
than adding a second one, but the diff leaves the `sp_gc_wb` at the head of
`run_thread_once` and adds yours after the stores. I merged it as written:
the extra record is idempotent (`sp_gc_wb_slow` skips an already-dirty
object) and free with the mode off, and the head barrier is on a path I did
not want to reason about under a merge.

Worth recording where that head barrier came from, because it explains the
class rather than the instance. 09558de1 placed it, and its audit was
field-granular: it took each type's scan function as the authority on which
fields hold a reference, asked which functions write those fields, and put a
barrier in those functions. A barrier at the head of a function passes that
audit and still misses a store made after a call that can collect. Your
sentence is the rule that audit was missing.

Running that rule over the rest of `lib/sp_sched.c` found one more, and it is
in the same commit's list: `sp_Thread_tls_set`. The first
`Thread.current[...]` write allocates the TLS map and stores it into the
thread. The barrier there is on the map, which covers writes into an existing
map (that is #4311's leg), but nothing recorded the thread for `t->tls = m`,
and the map's own allocation is the point that can collect. A thread that had
already survived a cycle then held a young map nothing had recorded, and the
next read answered `nil` -- the entire map, not one slot.

It hides behind your fix for a thread that terminates, since your new record
covers the store on the way out. It is live for a thread that is still
running, which is the ordinary case for `Thread.current[:buf]` in a worker.
8 runs in 10 before, 10 in 10 after, `test/gc_minor_thread_tls_first_write.rb`
as a third leg. Fixed on master.

PR #4382, opened 2026-09-07
Say which phase of a collection the time went to

bartleusink:
As offered on #4380. `sp_gc_stat_seconds` says how much a collection cost and
never which part of one, so "the collector is half of wall time" gives no way to
choose between attacking the mark and attacking the sweep.

```
[gcph] mark 0.178s  old sweep 0.006s  slot sweep 0.560s  remembered clear 0.000s  string sweep 0.001s  trim 0.000s  of 0.745s total
```

Phases named as the collector's own comments name them, per your request — a
number leads to the code that spent it without a translation step. The doc table
in `docs/internals/gc.md` uses the same words.

## What it is for

It is what found #4380. Before that landed, `remembered clear` was **5.7s of
13.1s** of collector time on the campfire server — the largest single phase of
the stopped window, and completely invisible in a total. On current master it
reads `0.000s`.

It also answers a question a total cannot, because the answer is not the same
everywhere:

| workload | mark | sweep |
|---|--:|--:|
| threaded churn, small live set | 1% | 99% |
| large live heap, c=8 | 39% | 61% |
| large live heap, c=64 | **54%** | 46% |

Which half is worth attacking depends on the workload and inverts with
concurrency. #4352 had to reason about "the cost of a collection" from totals
and noted that spinel exposed no counter for it; this is that counter. Running
it against that application, collections per second *fall* 4.6x from c=8 to c=64
while cost per collection *rises* 4.7x — which is what that report deduced and
could not check.

The remembered clear gets a bucket of its own rather than being folded into a
sweep, and earned it: nothing attributed a whole-old-heap walk that ran on every
non-full cycle.

## Cost when off

One not-taken branch per phase boundary — six per collection, against a
collection measured in microseconds. optcarrot is 0.11s CPU either way, which is
arithmetic rather than a measurement: nine collections make 54 branches.

`SPINEL_GC_PHASES` arms the reporter on its own, so the breakdown does not also
require `SPINEL_GC_STATS`.

## Test

`gc-phases-test` holds the two halves of the contract that the ordinary harness
cannot reach, since it cannot vary the environment per test: the flag must not
change what a program computes, and must say nothing at all when unset. I
checked it discriminates — forcing the flag on unconditionally makes it fail
with `wrote to stderr with the flag unset`.

## Gate

```
Tests:     3523 pass, 0 fail, 0 error
gc-minor-test: pass
gc-phases-test: pass
```

Two notes on the gate, both unrelated to this change:

- One `make gate` run showed `catch_throw_tags_values` and
  `catch_throw_through_loop` failing with **no output at all**, on a machine at
  ~10 load average. Both pass standalone on this branch and on master, and a
  clean `make test` re-run is 3523/0/0. I read it as those two hitting the
  harness timeout under load rather than anything here, but flagging it in case
  it is a known flake worth a longer timeout.
- `spin-check` fails with `spin flags: progress leaked onto stdout`, which
  reproduces on unmodified master.

## Summary by CodeRabbit

- **New Features**
  - Added the `SPINEL_GC_PHASES=1` option to report garbage-collection timing by phase, including marking, sweeping, remembered-set clearing, string sweeping, trimming, and total collection time.

- **Documentation**
  - Documented the new garbage-collection phase timing option and its diagnostic output.

- **Tests**
  - Added automated coverage to verify output remains consistent while phase timing diagnostics appear only when enabled.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ RBS_LIB      = build/librbs.a
 
 .PHONY: all regexp rbs_extract rbs-test rbs-seed-test re-lit-test reject-test backtrace-test gc-minor-test ext-test ext-cruby-test alloc-report-test rubyspec rubyspec-gate spin-check \
         test test-run clean-test-results regen-rbs-expected \
-        regen-expected regen-expected-err bench optcarrot gate check gate-legs gate-test gate-bench \
+        regen-expected regen-expected-err bench optcarrot gate check gate-legs gate-test gate-bench gc-phases-test \
         gate-optcarrot clean install uninstall deps tools
 
 # `make all` includes the RBS extractor when vendor/rbs has been fetched
@@ -671,7 +671,7 @@ test: $(SPINEL_TIMEOUT)
 # The actual run. rbs-test golden-checks the RBS extractor (cheap, C-only).
 # rbs-seed-test checks the seeds actually reach the analyzer (incl. nested
 # classes, #1417).
-test-run: rbs-test rbs-seed-test re-lit-test reject-test backtrace-test gc-minor-test ext-test ext-cruby-test $(TEST_TARGETS) $(PKG_TEST_TARGETS)
+test-run: rbs-test rbs-seed-test re-lit-test reject-test backtrace-test gc-minor-test gc-phases-test ext-test ext-cruby-test $(TEST_TARGETS) $(PKG_TEST_TARGETS)
 	@if [ -z "$(TIMEOUT_BIN)" ]; then echo "Note: no 'timeout' command found; running without time limits."; fi
 	@if [ -t 1 ]; then printf '\n'; fi
 	@pass=$$(grep -l '^PASS' build/test-results/*.ok 2>/dev/null | wc -l); \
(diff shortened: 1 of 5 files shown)


matz:
Merged as 613a39d8. Thank you for sending it.

Verified here on a 440 MB live heap with four churning threads:

```
[gcph] mark 0.091s  old sweep 0.092s  slot sweep 0.170s
       remembered clear 0.000s  string sweep 0.000s  trim 0.000s  of 0.353s total
```

The parts add to the total, and `remembered clear` at zero is #4380 showing
through -- which is a nice thing for an instrument to do on its first run:
report that the thing it was built to find is gone.

Both halves of the contract hold as written: zero bytes on stderr with the flag
unset, byte-identical stdout with it set. optcarrot 2283.5 fps at checksum
59662.

One thing I changed, in 0bd2605e. The phase comment pointed at
`docs/internals/gc-threaded-pause-plan.md`, which is not in this tree -- design
documents stay in their author's work tree here, so that reference would have
been dead for every reader. The same comment put the remembered clear at 8% of
the stopped window, which was true of the master you branched from and is not
true of the one it merged into. I kept both facts and stated them inline: the
5.7s of 13.1s figure, and that this is what found it.

On the two tests you flagged: `catch_throw_tags_values` and
`catch_throw_through_loop` pass standalone on the merge, and the gate run was
clean. Failing with no output at all is what the harness timeout looks like
here, so a machine at ~10 load average explains it. Worth knowing that we have
seen the same signature from a different cause -- a run where a whole
alphabetical tail of tests errored at once, which turned out to be a compiler
binary being rebuilt underneath a running gate. If it recurs on an idle
machine, that is worth an issue.

The phase split answering differently per workload is the part I find most
useful, and it is the argument for keeping it rather than measuring by hand:
"which half is worth attacking" inverting between c=8 and c=64 is not something
a total can express, and it is exactly the question #4352 could not settle.

PR #4383, opened 2026-09-07
Predicate loops follow changes to a boxed array's length

k-peek:
## Why

Predicate loops over boxed arrays cache the receiver's length before running the block. If the block shrinks the array, the loop visits removed positions, which return `nil`. If it grows the array, appended elements are skipped.

The extra block calls can expose exceptions:

```ruby
holder = { "values" => [1, 2, 3], "label" => "array" }
values = holder["values"]

result = values.all? do |x|
  raise "visited a removed element" if x.nil?
  values.clear
  true
end
p result
```

CRuby prints `true`. Before this change, Spinel raises `RuntimeError` after clearing the array.

Skipping appended elements can also produce incorrect predicate results.

## Changes

The boxed predicate loop reads an Array's current length on each iteration, matching the concrete-array loop. This applies to `all?`, `any?`, `none?`, and `one?`.

Other boxed receiver types retain their saved traversal bound. Hash iteration has separate mutation rules, so the length refresh applies only to Arrays.

This builds on #4379, which added predicate short-circuiting and made `all?` answer from a flag rather than comparing a match count with the receiver's length.

## Tests

`test/boxed_predicate_mutation.rb` covers clearing and growth across all four predicates, partial shrinking, mutation through an alias, different array storage kinds, empty arrays, `next`, and the exception example above.

Its expected output matches CRuby 4.0.6. The regression fails on the unmodified base and passes with this change under Clang and GCC, both normally and with `SPINEL_GC_STRESS=1 SPINEL_GC_VERIFY=1`.

Additional probes cover Ruby `break` and `return`, nested predicates, receiver rebinding, temporary receivers, and growth requiring reallocation.

## Validation

On macOS arm64, against base `6b8ddcd8`, `make gate` passes:

- Feature tests: 3,524 pass, 0 fail, 0 error.
- RubySpec: all 2,109 expected-PASS examples pass.
- Benchmarks: 61 pass, 0 fail, 0 error, 0 skip.
- Optcarrot: OK, checksum `59662`.
- Remaining gate checks pass.

Generated C changes for 56 of 3,476 pre-existing test programs: exactly 88 loop-condition substitutions, with no unexplained differences. The comparison excludes the new test and uses `--no-line-map`, with both compiler binaries in the same directory.

Optcarrot's generated C is unchanged.

## Performance

The loop now checks the array's current length on each iteration, matching the existing concrete-array loop and CRuby's array predicates with blocks. Local microbenchmarks show added overhead for very inexpensive blocks. I'd welcome suggestions for a cheaper implementation that still observes mutations made by the block.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed boxed array predicates (`all?`, `any?`, `none?`, and `one?`) to correctly reflect array mutations made during predicate evaluation.
  * Updated traversal behavior so appended elements are processed, removed elements are skipped, and shrinking arrays can end iteration correctly.

* **Tests**
  * Added regression coverage for array mutations across supported array types and predicate control-flow scenarios.


matz:
Merged as 0e24b04a.

I wrote my own probe before reading the diff. 13 shapes: the four predicates
shrinking to empty, partial shrink, growth, growth past capacity (forcing a
realloc mid-scan), mutation through an alias, a string array, a Hash receiver,
and Ruby `break`/`return` through the block. Master differs from CRuby 4.0.4 on
10 of them; your branch matches all 13.

Worth stating plainly in the issue text, because it is stronger than "extra
block calls": this changes **answers**.

```ruby
a.one? { |x| a.clear; true }   # false before, true in CRuby
```

`grow` was wrong the same way -- the appended element was never tested, so
`all?` answered true where CRuby answers false. Your `raise` example is the
sharpest form, but the wrong answers stand on their own.

I also checked receiver rebinding inside the block (`a = something_else`): the
loop keeps scanning the receiver it was handed, which is what CRuby does. That
one matters for this change specifically, since the condition now dereferences
the receiver temp on every turn.

## On the cost you flagged

I could not reproduce a slowdown, so I would not complicate it yet.

The narrowest shape I could build -- a boxed 2000-element array, the cheapest
block that can exist (`{ |x| x > 0 }`), 3000 scans, with the emitted C confirmed
to take the new path -- runs **0.048s on the branch against 0.050s on master**,
interleaved five times. bench is 61/61 and optcarrot is 2281.4 fps at checksum
59662.

That is Linux x86-64 with gcc 13 and yours was macOS arm64, so it does not
contradict your measurement. But it suggests where the cost is not: the kind
test is loop-invariant and inline, so a compiler hoists it out of the loop on
its own. What remains is the `->len` load, and that one cannot be hoisted --
being re-read is precisely the property the fix buys. So the cheaper
implementation you were looking for may already be what the C compiler emits.

If it does show up on arm64, the lever is to hoist the kind test explicitly
(`int is_arr = sp_rbval_is_array(recv);` before the loop) and leave only the
length read inside. I would want a measurement showing it before adding it,
since it costs another emitted line in all 56 files.

## On the sweep

Your generated-C count reproduces exactly here: 56 files, 88 substitutions, and
every changed token is the loop condition. Using `--no-line-map` and putting
both binaries in the same directory is the right call -- thank you for picking
that up from the last one.

PR #4388, opened 2026-09-08
Dump the allocation report on a signal, because a server never reaches atexit

rubys:
`SPINEL_ALLOC_REPORT` is written from `atexit`, which is a mode a long-lived
program cannot use: a server is stopped by a signal, so the counters it spent
the whole run filling are lost at exactly the moment they are worth reading.
Nothing about the table needed changing — it is complete at every instant — it
just had no way out before the process ended.

`SIGUSR1` now asks for a dump in flight, and the program carries on.
`SPINEL_ALLOC_REPORT_SIGNAL` moves it, by number or by name.

## Shape

The handler cannot write the report: `fopen`, `malloc` and
`backtrace_symbols` are none of them async-signal-safe. What it *can* do is
`write` one byte to a pipe, and a thread parked on the other end does the
writing. **Cost when the report is off: nothing at all — no branch anywhere.**
The thread, the pipe and the handler are installed only when it is on.

Every dump rewrites the whole cumulative table, the same one `atexit` writes,
so a window is two dumps subtracted rather than a mode of its own — which is
also what keeps a server's boot out of the profile. `docs/profiling.md` carries
the recipe.

**The thread is the difference between working and nearly working.** The first
version set a flag and let the next counted allocation write the file. That is
fine on a busy server and useless on an idle one: I signalled a warm Campfire
server between load runs and the dump waited for the next request. A
single-threaded build still uses that path, because a single-threaded program
that is idle is inside a syscall with nothing else to run — that limit is
documented rather than papered over.

## Tests

`alloc-report-test` gains two programs that never exit, and each was checked
against the version it is meant to catch:

- **busy**: signal, assert it is still running and has written a table, signal
  again, assert the second table is later than the first, then `SIGKILL` — so a
  report that exists at the end can only have come from a signal. Against
  master's runtime that program dies at the first signal (`SIGUSR1`'s default
  disposition is terminate) and writes nothing.
- **idle**: a threaded program that allocates a little, then parks. Against the
  flag-only version above it reports nothing at all.

`make gate`: 3536 tests pass, 0 fail, 0 error. One unrelated step fails,
`spin-e2e FAIL: spin flags: progress leaked onto stdout` — it fails identically
on master at `2d160ed2` here, so it is not from this change; happy to file it
separately if it is not already known.

## What it was built for, and what it found immediately

Campfire, warm, `/rooms/1`: dump, 100 requests, dump, subtract. Boot excluded
because it is in both dumps. 1,408 rows, no `(unattributed)` row, so the
8,192-entry table held.

```
TOTAL per request: 11,530 KB in 10,094 allocations
  KB/req  allocs/req  share  site
  9591.3        2608  83.2%  sp_str_concat
   774.4           6   6.7%  sp_str_append_grow
   386.8           7   3.4%  sp_Views__Rooms_s_show
   383.9           9   3.3%  _proc_38
```

83% of the bytes a request allocates are 2,608 calls to `sp_str_concat`
averaging 3.7 KB each, building a 423 KB page — an accumulation that copies
what it has so far, which is ours to fix in the emitter, and which is the kind
of thing that was simply not observable before: the process it happens in never
exits.

One thing that surfaced with it, not part of this PR: an allocation made
*inside* a runtime helper attributes to the helper, since the captured frame is
the allocator's caller. `sp_str_concat` is the answer to "what allocates" but
not to "which Ruby code asked". The doc already warns that one frame is not
always the frame you want; worth knowing that runtime helpers are the case
where it always bites.

https://claude.ai/code/session_013vQWvnWEJ6qSqtFKaLHvFX

## Summary by CodeRabbit

- **New Features**
  - Added on-demand allocation report generation triggered by a configurable signal, with `SIGUSR1` used by default.
  - Reports can be captured while applications are active or idle, including cumulative snapshots for comparing allocation changes.
  - Supports both threaded and single-threaded applications, as well as forked processes.

- **Documentation**
  - Added guidance for configuring signals and interpreting live allocation reports.

- **Tests**
  - Added coverage for repeated signal requests and idle-process reporting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1584,6 +1584,22 @@ alloc-report-test: $(SPINEL) $(SP_RT_LIB)
 	full=$$(awk '/;\(no-scan\) /{print $$NF}' "$$tmp/s.folded" | head -1); \
 	sat=$$(awk '/;\(no-scan\) /{print $$NF}' "$$tmp/sm.folded" | head -1); \
 	[ -n "$$full" ] && [ "$$full" = "$$sat" ] || { echo "alloc-report-test: FAIL (a saturated run changed a surviving row: $$full vs $$sat)"; ok=0; }; \
+	$(SPINEL) test/alloc-report/signal_dump.rb -o "$$tmp/sig" >/dev/null 2>&1 || { echo "alloc-report-test: FAIL (compile signal_dump)"; exit 1; }; \
+	SPINEL_ALLOC_REPORT="$$tmp/sig.folded" "$$tmp/sig" >/dev/null 2>&1 & \
+	sigpid=$$!; sleep 1; kill -USR1 $$sigpid 2>/dev/null; sleep 1; \
+	kill -0 $$sigpid 2>/dev/null || { echo "alloc-report-test: FAIL (the signal ended the program instead of dumping)"; ok=0; }; \
+	first=$$(awk '/^alloc;.*String /{print $$NF; exit}' "$$tmp/sig.folded" 2>/dev/null); \
+	[ -n "$$first" ] || { echo "alloc-report-test: FAIL (no report from a running program)"; ok=0; }; \
+	sleep 1; kill -USR1 $$sigpid 2>/dev/null; sleep 1; \
+	second=$$(awk '/^alloc;.*String /{print $$NF; exit}' "$$tmp/sig.folded" 2>/dev/null); \
+	kill -9 $$sigpid 2>/dev/null; wait $$sigpid 2>/dev/null; \
+	[ -n "$$second" ] && [ "$$second" -gt "$${first:-0}" ] || { echo "alloc-report-test: FAIL (the second signal did not re-dump a later table: $$first then $$second)"; ok=0; }; \
+	$(SPINEL) test/alloc-report/signal_dump_idle.rb -o "$$tmp/idle" >/dev/null 2>&1 || { echo "alloc-report-test: FAIL (compile signal_dump_idle)"; exit 1; }; \
+	SPINEL_ALLOC_REPORT="$$tmp/idle.folded" "$$tmp/idle" >/dev/null 2>&1 & \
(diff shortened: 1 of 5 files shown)


rubys:
Correcting one sentence in the description above. Nothing about the patch
changes; the reading of what it found does.

I wrote that the 83% was "an accumulation that copies what it has so far" and
ours to fix in the emitter. **The accumulation is not the problem — our views
already build with `String.new` + `<<`, which lowers exactly as intended:**

```c
lv_io = sp_String_new_shared(...);
sp_String_append_bin(lv_io, ...);       // in place, amortised
```

What the concats are is the **materialisation at the strbuf boundary**. Of the
131 static `sp_str_concat` sites in this program, **129 are
`sp_str_concat(sp_String_cstr(buf), "")`** — the copy that turns a growable
buffer into a string value that cannot alias it:

```c
return (_sp_ret_strbuf = (void *)lv_out, sp_str_concat(sp_String_cstr(lv_out), ""));
```

That is a value-semantics requirement, not a bug — `(s << "x").equal?(s)` is
the comment next to one of the emit sites. But it prices every nesting level of
a template: each view function copies its whole buffer on the way out, its
caller appends that copy into its own buffer, and a 423 KB page comes out
having been copied about twenty times over. Hence 9.6 MB per request in 2,608
copies averaging 3.7 KB.

The fix on our side is to hand the parent's buffer down rather than return a
string up, which is a change to how Roundhouse emits nested templates and needs
no runtime change.

There may be one on your side too, and I would rather ask than assume: when the
buffer is dead after the materialisation — which a `return buf` at the end of a
builder always is — the copy is pure waste, and `_sp_ret_strbuf` suggests the
returning buffer is already tracked for something. If a last-use transfer is
plausible there, this workload prices it: 83% of every request's allocated
bytes. Happy to file it as an issue with the numbers rather than leave it in a
PR comment.

PR #4398, opened 2026-09-09
openssl: add connect_nonblock and CRuby-compatible aliases

rofl0r:
Add SSLSocket#connect_nonblock(exception: true) matching CRuby's behavior: returns :wait_readable or :wait_writable when the handshake needs the fd in that state, or self on completion.

Add attr_accessor :sync_close for proper cleanup when the SSLSocket owns the underlying IO (mirrors CRuby's behavior where `@sync_close` closes the wrapped IO on SSLSocket#close).

Add aliases read_nonblock -> sysread_nonblock, read -> sysread, write -> syswrite, close -> sysclose so callers that bypass OpenSSL::Buffering (e.g. HTTP clients driving the socket directly for chunked/streaming) get the plain names.

These changes enable HTTP clients with deadline-based streaming to work with the Spinel OpenSSL package without code changes.

## Summary by CodeRabbit

* **New Features**
  * Added nonblocking SSL connection support, including readable and writable wait states during handshakes.
  * Added a `sync_close` option for SSL sockets, defaulting to disabled.
  * SSL socket closure can now also close the underlying connection when `sync_close` is enabled.

### DIFF
--- a/packages/openssl/openssl.rb
+++ b/packages/openssl/openssl.rb
@@ -119,6 +119,7 @@ module OpenSSL
       include OpenSSL::Buffering
 
       attr_accessor :hostname
+      attr_accessor :sync_close
 
       def initialize(io, context = nil)
         super()
@@ -126,6 +127,7 @@ module OpenSSL
         @context = context.nil? ? SSLContext.new : context
         @handle = -1
         @hostname = ""
+        @sync_close = false
       end
 
       def context
@@ -146,6 +148,23 @@ module OpenSSL
           raise SSLError, "SSL_connect returned an error: #{Native.last_error}"
         end
         @handle = h
+        self
+      end
+
+      # Non-blocking connect: returns the SSLSocket on completion, or
+      # :wait_readable / :wait_writable if the handshake needs the fd in
+      # that state. Matches CRuby's SSLSocket#connect_nonblock(exception:
+      # false) used by HTTP clients with a deadline.
+      def connect_nonblock(exception: true)
+        begin
+          connect
+        rescue SSLErrorWaitReadable
+          return :wait_readable unless exception
+          raise
+        rescue SSLErrorWaitWritable
+          return :wait_writable unless exception
+          raise
+        end
         self
       end
 
@@ -226,8 +245,18 @@ module OpenSSL
         return nil if @handle < 0
         Native.close(@handle)
         @handle = -1
+        @io.close if @sync_close && @io && !@io.closed?
         nil
       end
+
+      # CRuby-compatible aliases. The native methods are named sys* to leave
+      # the unprefixed names free for the buffering wrappers in
+      # OpenSSL::Buffering, but callers that bypass buffering (e.g. an HTTP
+      # client driving the socket directly) want the plain names.
+      alias_method :read_nonblock, :sysread_nonblock
+      alias_method :read, :sysread
+      alias_method :write, :syswrite
+      alias_method :close, :sysclose
 
       def peer_subject
         @handle < 0 ? "" : Native.peer_subject(@handle)
--- a/packages/openssl/openssl.rb
+++ b/packages/openssl/openssl.rb
@@ -249,15 +249,6 @@ module OpenSSL
         nil
       end
 
-      # CRuby-compatible aliases. The native methods are named sys* to leave
-      # the unprefixed names free for the buffering wrappers in
-      # OpenSSL::Buffering, but callers that bypass buffering (e.g. an HTTP
-      # client driving the socket directly) want the plain names.
-      alias_method :read_nonblock, :sysread_nonblock
-      alias_method :read, :sysread
-      alias_method :write, :syswrite
-      alias_method :close, :sysclose
-
       def peer_subject
         @handle < 0 ? "" : Native.peer_subject(@handle)
       end


matz (changes requested):
Two thirds of this I want, and the four aliases I cannot take -- they undo the
thing the `sys*` naming exists to protect.

## The good half

`connect_nonblock` is genuinely missing and your shape matches CRuby's. So is
`sync_close`, and closing `@io` inside `sysclose` when it is set is right.
I would merge those today.

## The aliases

```ruby
alias_method :read_nonblock, :sysread_nonblock
alias_method :read, :sysread
alias_method :write, :syswrite
alias_method :close, :sysclose
```

`SSLSocket` does `include OpenSSL::Buffering`, and Buffering defines **all
four of those names already** -- `read` at buffering.rb:75, `read_nonblock`
at :99, `write` at :200, `close` at :246. A method defined in the class wins
over one from an included module, so each alias replaces the buffered version
rather than filling a gap. I checked that resolution holds in spinel and not
just in CRuby before writing this; it does, in both.

Three consequences, in increasing order of how much they would cost:

- **`read` stops taking no arguments.** `Buffering#read(nil)` reads to EOF;
  `sysread` requires a maxlen, so `sock.read` becomes an ArgumentError.
- **`gets` and `read` stop sharing a buffer.** `gets` fills `@rbuffer` through
  `fill_rbuff`; an unbuffered `read` afterwards skips it and the buffered
  bytes are lost. `packages/net` reads exactly this way.
- **`close` stops flushing.** `Buffering#close` is `flush; sysclose`. Aliased
  to `sysclose`, a socket with anything pending in `@wbuffer` drops it on
  close.

And the naming is CRuby's the way it already stands: in CRuby, `SSLSocket#read`
IS the buffered one from Buffering and `sysread` IS the unbuffered one. So
"CRuby-compatible aliases" is the opposite of what these do.

## What I think you actually want

An HTTP client driving the socket directly wants the unbuffered path, and it
already has it under the CRuby names: call `sysread` / `syswrite` /
`sysclose`. If your client insists on the plain names, wrapping the socket is
the way to say "this one is unbuffered" without changing what the name means
for `packages/net` and every other caller.

If there is a case where the buffered version is genuinely wrong for you and
`sys*` will not do, say which and I will look again -- that would be a
Buffering bug and I would rather fix it there.

Drop the four `alias_method` lines and I will merge the rest as it stands.

rofl0r:
thanks for the review. addressed, and when you said `What I think you actually want` you were correct.

rofl0r:
hmm. although it seems i cant actually use the sys* functions: `spinel: http_client.rb:530: unsupported call: node 4903 (CallNode `sysread_nonblock`) recv=LocalVariableReadNode/ty46 argc=2 arg0ty3`. and my agent suddenly insists the connect_nonblock ruby function is bogus, since it calls the blocking connect method?! but everything worked fine with the aliases. seems some more research is due.

matz:
Your agent is right, and it is a better catch than the aliases. Two separate
things, and I could not reproduce the third.

## `connect_nonblock` cannot work as written

`connect` never raises `SSLErrorWaitReadable` or `SSLErrorWaitWritable`. Those
two are raised in exactly four places in the package, all of them inside
`sysread_nonblock` (openssl.rb:179, :182) and `syswrite_nonblock` (:204,
:207). `connect` either returns a handle or raises plain `SSLError`.

So both rescue arms in your `connect_nonblock` are dead code, and it is a
blocking `connect` that returns `self`. Not cosmetically wrong -- a client
that waits for `:wait_readable` before selecting on the fd will instead block
inside the handshake, which is the thing it was trying to avoid.

Making it real is a runtime change rather than a Ruby one: `sp_ssl_connect`
does the whole handshake, so a non-blocking form needs the C side to run
`SSL_connect` against a non-blocking fd and map `SSL_get_error`'s
`WANT_READ` / `WANT_WRITE` into those two exception classes. That is worth
doing and I would review it, but it is not this patch.

`sync_close` and closing `@io` in `sysclose` are good and I would take them on
their own.

## Why the aliases seemed to work

They routed you somewhere that already works. `Buffering#read_nonblock`
(:99) calls `sysread_nonblock`, which does raise the wait errors -- so with
the alias your client reached the working non-blocking read, while
`connect_nonblock` was quietly blocking. The aliases were not fixing anything;
they were changing which of two paths you were on.

## The compile error is the real blocker and I need one more thing

I could not reproduce it. Both of these build on master here:

```ruby
def pump(ssl)  = ssl.sysread_nonblock(4096, exception: false)   # untyped param
ssl = OpenSSL::SSL::SSLSocket.new(sock)
r = ssl.sysread_nonblock(4096, exception: false)                # typed local
```

Your message says `recv=LocalVariableReadNode/ty46`, so the receiver has a
concrete class and the compiler could not find `sysread_nonblock` on it. What
would settle it: **what is that local at http_client.rb:530** -- an
`SSLSocket`, or a wrapper of your own, or something that may be either a
plain socket or an SSL one? That last shape is my first guess, because a
receiver that is sometimes a TCPSocket has no `sysread_nonblock` on one of its
arms, and `read_nonblock` exists on both -- which would explain the aliases
working and the plain names not, without either being a coincidence.

If that is it, the bug is ours: the two arms should not have to agree on a
method name that only one of them needs.

Leaving this open. Drop the four aliases and I will take `sync_close` now; the
non-blocking handshake is worth its own PR, and if the receiver turns out to
be a union I will fix that end.

rofl0r:
> what is that local at http_client.rb:530

sorry, that was a red herring. i confounded the tls_sock variable (which was a SocketPair endpoint) with the actual ssl socket. so no compile error when applied to the right variable.

matz:
Merged as `d7b391fd`, and `connect_nonblock` is real as of `823f1fd5` rather than dead.

I said the non-blocking handshake was worth doing and not this patch. It turned out small enough to do straight away, and leaving your method in the tree with two dead rescue arms was the worse of the two options.

## What the runtime does now

`sp_ssl_connect_nb` builds the connection, arms O_NONBLOCK on the descriptor and takes one handshake step. `sp_ssl_connect_cont` takes the next. Which of the two waits it needs, or a real failure, rides in `sp_ssl_want`, the slot `sp_ssl_read_nb` and `sp_ssl_write_nb` already answer through. The handle comes back as soon as the connection exists, unfinished handshake and all, because OpenSSL's half-done state has nowhere else to live.

Your Ruby method keeps its signature and its two answers. The body now starts the handshake on the first call and resumes it on every call after.

One difference from `sysread_nonblock` worth knowing if you are writing against it: this one arms O_NONBLOCK itself. The handshake is the first thing on a connection, so there is no earlier call to have done it, and a blocking descriptor would put the wait back in the kernel.

## What proves it

`packages/openssl/test/openssl_connect_nonblock.rb`, and it needs no certificate and no peer that speaks TLS, which is what makes it a test of the mechanism rather than of a network. The peer accepts and then says nothing, so the ClientHello goes out and the reply never comes:

```
:wait_readable
:wait_readable
raised OpenSSL::SSL::SSLErrorWaitReadable
caught as IO::WaitReadable
true
end
```

Those are CRuby 4.0.4's own lines, matched three runs of three. The wrapped version could not produce any of them: nothing to raise, and it never returns.

The last two lines are your `sync_close`: the handle is live because the handshake is unfinished rather than absent, so `sysclose` releases it and takes the socket with it.

One divergence I wrote down at the accessor rather than leaving to be found. CRuby's unset `sync_close` reads back nil, this one false: a statically typed ivar has no third state to start in. Setting it behaves the same.

## The compile error

Thank you for coming back on the red herring. Nothing to chase there, so the union-receiver theory stays a theory. If `read_nonblock` on a receiver that is sometimes a plain socket ever does fail to compile for you, that is the shape to send and the bug would be ours.

Gate green on the pair: 3553 tests, 61 benchmarks, optcarrot 59662, spin-e2e, six rubyspec legs. The blocking `connect` still completes a real handshake against a public endpoint, which is worth saying because the verify and SNI setup is now shared between the two rather than copied.

PR #4405, opened 2026-09-09
A spawn opens its input file read-only and closes the files it opened

blacktm:
## Why

`Process.spawn("/bin/cat", in: "input.txt", out: "output.txt")` should hand the child the input file to read and leave that file as it was. On master it empties the input file before the child starts, then gives the child a write-only descriptor as its standard input. In the test program, with `input.txt` holding `"valuable input\n"`, the spawn exits 1, `cat` prints `stdin: Bad file descriptor`, and both files read `""` afterwards. CRuby exits 0 and both files read the input. A program that uses the documented filename form of `in:` loses the file it meant to read.

The same spawn also keeps the descriptor it opened. The child closes its copy after the `dup2`, but nothing closes the parent's, so a program that spawns and waits in a loop grows its descriptor table by one per filename redirection. In the test program, twenty spawn-and-wait turns with `in:` and `out:` filenames grow `/dev/fd` by 40 on master and by 0 under CRuby. In a probe that runs twenty turns over one shape at a time, master grows `/dev/fd` by 20 for `out:` alone, 20 for `err:` alone, 40 for `out:` with `err:`, 60 for all three, 20 for a failed exec with `out:` a filename and 40 with `in:` and `out:`, every one of them 0 under CRuby and on this branch. A spawn that raised before forking, because a later redirection failed to open, kept the descriptor too: twenty such turns grow `/dev/fd` by 20 on master in the test program.

Both come from one line of the emitter, repeated for `in:`, `out:` and `err:`: every filename was opened with `O_WRONLY|O_CREAT|O_TRUNC` in the parent, with no record of which descriptors the spawn had opened itself. A third face of that line: a missing `in:` path was created empty instead of raising, so the spawn went on to run `cat` on a write-only descriptor to the empty file it had just made.

## What

The open moves out of the generated program into the runtime, next to the code that already closes the child's copies. `sp_process_open_redirect` in `lib/sp_process.c` opens an `in:` path `O_RDONLY`, an `out:` or `err:` path created and truncated as before, and records the descriptor in a three-slot array the generated program keeps for the duration of the call, one slot per redirection, `-1` wherever the spawn did not open the file itself. A failed open raises after closing the descriptors already opened for that spawn, with CRuby's message, `strerror - path`, built in a buffer sized for a path, and with the Errno class CRuby raises when the errno is one of `ENOENT`, `EACCES`, `EEXIST`, `EISDIR`, `ENOTDIR`, `ENAMETOOLONG`, `ELOOP`, `EROFS`, `EMFILE` or `ENFILE`, the four `sp_file_open_raise` maps plus six more that `open(2)` can answer for a path. Any other errno is a `SystemCallError`.

The flat options array `sp_process_spawn` reads gains an eighth slot: a bit per redirection the spawn owns. Once the fork has given the child its copies, the parent closes its own, before it reads the exec-failure pipe, so the descriptors go on the failure path too. Every raise in the runtime after the owned descriptors are known and before the fork goes through `sp_process_spawn_fail`, which closes the owned descriptors and then raises with errno intact, and the generated program's own argument-shape raises inside the spawn call go through the same function. A descriptor that came from a caller's IO is never recorded, so it is never closed. The file's header comment documents the eight-slot contract and the runtime-side open.

## How

Each face of the line has its own test arm, and each fails on master in its own way. `in:` a file holding `"valuable input\n"` with `out:` a file holding stale text: master answers exit 1 and `""` for both files, this branch exit 0 and the input text for both, as CRuby does. `in:` a missing path: master prints `no raise` and the path now exists, this branch raises `Errno::ENOENT` with `No such file or directory - DIR/absent.txt` and the path does not exist. Twenty spawn-and-wait turns with `in:` and `out:` filenames: master grows `/dev/fd` by 40, this branch by 0. Twenty spawns of a missing command with both redirections, each rescued: master 40, this branch 0. Twenty spawns whose `out:` path is in a missing directory, so the raise comes after `in:` was opened: master 20, this branch 0. A path through a regular file: master raises `Errno::ENOENT` with the bare path for `in:` and for `out:`, this branch `Errno::ENOTDIR` with `Not a directory - DIR/notdir/x`. A 600-character name: master `Errno::ENOENT`, this branch `Errno::ENAMETOOLONG`. A message of 634 characters once the temporary directory's name is replaced: this branch carries it whole, and prints its last twelve characters as CRuby does. A `File` the caller opened and handed as `out:`, written to by the child and then by the caller: `closed?` is false on both trees, and the file holds both lines. `out:` and `err:` both filenames, a caller's IO for `out:` with a filename for `err:`, and a spawn with no options answer the same on master, this branch and CRuby. They are there because those are the shapes where two owned descriptors close in one call, where the mask has a single high bit, and where every spawning program's generated C changes.

## Validation

`test/process_spawn_file_redirect.rb.expected` is the output of `ruby test/process_spawn_file_redirect.rb` under CRuby 4.0.6, and it regenerates byte for byte. Every count in the test is `/dev/fd` after minus before, so the expected value is 0 on any machine and the arm is portable to Linux, where `/dev/fd` is the same view. Compiled, the test runs in under a tenth of a second on this branch, and matches the expected file in three of three plain runs and in two of two under `SPINEL_GC_STRESS=1`, with nothing on stderr. On master it compiles and runs to the end in three identical runs. The `err:` arm, the caller's-IO arm and the three added shapes match CRuby, and every other arm answers differently. The existing `test/process_spawn.rb`, `test/file_null_constant.rb` and `test/waitpid_thread_drains_pipe.rb`, which redirect to IOs, to `File::NULL` by name and to a pipe end, match their expected output on this branch.

On `d1cac2ec` the 3542-test suite and the 61 benchmarks pass, optcarrot is unchanged at 59662, the 2131-program compatibility corpus holds at 906 passing with none gained and none lost, and the generated C changes only for the three existing programs that call `Process.spawn` and for this test, by the owned-descriptor array, the runtime open and the eighth options element. The test is clean under GC stress and under AddressSanitizer over its generated C.

## Left alone, on purpose

The three redirections are opened in `in:`, `out:`, `err:` order whatever order the caller wrote them, so `out: created, in: absent` raises with the output file not created here and created under CRuby. The order predates this change and master never raised on `in:` at all. Three raises in the runtime before the descriptors are known, two on a malformed options array and one on a non-Integer slot, stay bare. None is reachable from Ruby. A spawn whose exec fails still raises without reaping the child, and its message says `cannot execute` where CRuby names the command. `[:out, :err] => path` is ignored, `out: [path, "a"]` and `out: [path, "w", 0600]` raise, Integer keys such as `1 => path` are read as arguments, `in: nil` and `in: false` are ignored where CRuby raises, and `in: :foo` reports `bad redirect type`. All the same on master. The `[executable, argv0]` command form still raises `TypeError`. A non-String `chdir:` is swallowed by the emitter, so the runtime's `TypeError` never fires. `File.open` on a path through a regular file raises a bare `SystemCallError`, because `sp_file_open_raise` maps four errnos. This change does not touch it, and a spawn whose open fails with an errno outside its own ten raises a bare `SystemCallError` the same way, where CRuby names an Errno class for every errno. A closed IO in a later slot raises from the generated program before the runtime is entered, so a filename opened for an earlier slot is still leaked, as before: twenty turns of `out:` a filename with `err:` a closed IO grow `/dev/fd` by 20 on master and on this branch, and by 0 under CRuby. The two element-type raises after the argument vector is allocated still leak that vector, as before. The options hash is read through an unrooted temporary across three pushes, as before, and this change adds no allocation to that window. The redirection descriptors are opened without `O_CLOEXEC`, as before. `Dir.mktmpdir` leaves a non-empty directory behind on Spinel, which is why the test deletes its own files before the block ends.

## Summary by CodeRabbit

* **New Features**
  * `Process.spawn` now supports redirecting standard input, output, and error streams directly to filenames.
  * Input files are opened for reading, while output and error files are created or truncated as needed.
  * Existing file handles remain open and can be mixed with filename-based redirects.

* **Bug Fixes**
  * Improved error reporting for missing, invalid, and excessively long redirect paths.
  * Prevented file descriptor leaks when process launching or redirection fails.

### DIFF
--- a/lib/sp_process.c
+++ b/lib/sp_process.c
@@ -2,8 +2,8 @@
  *
  * Lives in its own TU because the codegen calls sp_process_spawn
  * with pre-resolved positional args (in_fd, out_fd, err_fd, pgroup,
- * rlimit_cpu, rlimit_as, chdir). All opts-hash unpacking happens at
- * compile time in the codegen; the runtime just needs primitive int
+ * rlimit_cpu, rlimit_as, chdir, owned). All opts-hash unpacking happens
+ * at compile time in the codegen; the runtime just needs primitive int
  * / int-fd / int-pgroup / int-rlimit values. The cmd is either a
  * String or an Array (boxed as a PolyArray). args is a PolyArray of
  * extra String args appended after cmd.
@@ -18,21 +18,29 @@
  * Process.spawn(cmd, *args, opts) -> child pid
  * Process.waitpid2(pid) -> [pid, raw_status]
  *
- * opts is a PolyArray of 7 elements in this order:
- *   [0] in_fd        - Integer fd (>= 0), -1 for false/nil, IO was
- *                      resolved to its fd by the codegen, String path
- *                      was opened by the codegen
+ * opts is a PolyArray of 8 elements in this order:
+ *   [0] in_fd        - Integer fd (>= 0), -1 for false/nil/absent. An IO
+ *                      was resolved to its fd by the codegen; a String
+ *                      path was opened here, by sp_process_open_redirect,
+ *                      which the generated program calls before this
  *   [1] out_fd       - same conventions
  *   [2] err_fd       - same conventions
  *   [3] pgroup       - Integer (0=inherit, 1=new, >1=specific pgid)
  *   [4] rlimit_cpu   - Integer (seconds), nil = no limit
  *   [5] rlimit_as    - Integer (bytes), nil = no limit
  *   [6] chdir        - String path or nil
+ *   [7] owned        - Integer bit mask, bit 0/1/2 set when [0]/[1]/[2]
+ *                      is an fd sp_process_open_redirect opened for this
+ *                      spawn; those are closed here once the child has
+ *                      its copies, and on every raise this file makes
+ *                      before the fork.
+ *                      A caller's IO or Integer fd is never in the mask.
  *
- * IO and String values in opts[0..2] are resolved to fds by the
- * codegen (which has access to sp_File_fileno and open(2)). The
- * [:child, :out|:err|Integer] form is also resolved by the codegen
- * (it sees the literal array at parse time).
+ * An IO value in opts[0..2] is resolved to its fd by the codegen (which
+ * has access to sp_File_fileno); a String value reaches the runtime
+ * through sp_process_open_redirect. The [:child, :out|:err|Integer]
+ * form is resolved by the codegen (it sees the literal array at parse
+ * time).
  */
 
 #include <fcntl.h>
@@ -47,6 +55,7 @@
 #include <sys/wait.h>
 #include <unistd.h>
 #include <errno.h>
+#include <limits.h>
 
 #include "sp_alloc.h"   /* sp_PolyArray, sp_RbVal, sp_box_*, sp_raise_cls */
 #include "sp_process_status.h"   /* sp_ProcessStatus, sp_box_process_status */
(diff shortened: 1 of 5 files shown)


matz:

Issue #4394, opened 2026-09-08
spinel program using threads does not terminate when it reaches the last statement


rofl0r:
## Summary
<strike>
`Thread#join(timeout)` blocks forever instead of returning after the specified timeout duration. This affects all blocking thread states (I/O, sleep, etc.).
</strike>

the below ruby program does **not terminate** when it finishes executing the last statement.

## Environment
- Spinel AOT compiler (279398012c1dc77731d2766cb0266d09c135ae9e)
- Linux

## Reproducer

```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true
# Minimal repro for Thread#join(timeout) on spinel

# Test 1: Thread blocked on FIFO read
puts "=== Test 1: Thread blocked on FIFO read ==="
dir = "/tmp/test_join_#{Process.pid}"
Dir.mkdir(dir)
fifo = File.join(dir, "test.fifo")
system("mkfifo #{fifo}") or raise "mkfifo failed"

reader = Thread.new do
  puts "[reader] opening FIFO for read"
  File.open(fifo, "r") do |f|
    puts "[reader] reading..."
    f.each_line { |l| puts "[reader] got: #{l}" }
  end
  puts "[reader] done"
rescue => e
  puts "[reader] error: #{e.class}: #{e.message}"
end

sleep 0.1  # ensure reader is blocked on open/read
puts "[main] joining with 0.5s timeout..."
start = Time.now
result = reader.join(0.5)
elapsed = Time.now - start
puts "[main] join returned after #{elapsed.round(3)}s: #{result.inspect} (thread=#{reader.inspect})"
puts "[main] thread status: #{reader.status}"

reader.kill rescue nil
Dir.rmdir(dir) rescue nil

# Test 2: Thread blocked on pipe read
puts "\n=== Test 2: Thread blocked on pipe read ==="
r, w = IO.pipe
reader2 = Thread.new do
  puts "[reader2] reading from pipe..."
  data = r.read
  puts "[reader2] got: #{data.inspect}"
rescue => e
  puts "[reader2] error: #{e.class}: #{e.message}"
end

sleep 0.1
puts "[main] joining with 0.5s timeout..."
start = Time.now
result2 = reader2.join(0.5)
elapsed = Time.now - start
puts "[main] join returned after #{elapsed.round(3)}s: #{result2.inspect} (thread=#{reader2.inspect})"
puts "[main] thread status: #{reader2.status}"

reader2.kill rescue nil
r.close
w.close

[...]

PR #4406, opened 2026-09-09
DO NOT MERGE: macOS FIFO pair probe for #4394

matz:
Isolation probe, to be closed as soon as macOS reports.

The FIFO open fix for #4394 was reverted in `bdfc8096` because the test's reader-and-writer arm never completed on macOS while the reported bug's own arm passed. This branch carries **that arm alone, with no runtime change**, so macOS answers one question: was passing lines through a FIFO between two green threads already broken there, or did the reverted change break it?

Linux answers `read ["line0", "line1", "line2"]` on this branch.

https://claude.ai/code/session_01BQxF2sMafrf1WPJ6qQBLLs

### DIFF
--- a/test/fifo_pair_probe.rb
+++ b/test/fifo_pair_probe.rb
@@ -1,0 +1,28 @@
+# ISOLATION PROBE, not a fix and not meant to stay.
+#
+# #4394 is that File.open on a FIFO waits inside open(2) with no descriptor to
+# wait on, so a green thread pinned to its worker stalls every green thread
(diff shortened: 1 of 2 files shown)


matz:
Answered, and it is the answer that matters: **macOS fails this probe with no runtime change at all.** Passing lines through a FIFO between two green threads was already broken there, so the revert in `bdfc8096` was aimed at the wrong commit.

- ubuntu gcc: pass
- ubuntu clang: pass
- macos clang: FAIL, `fifo_pair_probe`, the reader gets nothing

Re-landing the #4394 fix without the arms that pass data, and writing the macOS gap down where it can be found.

Issue #4407, opened 2026-09-09
GC budget `fixed`: a 16 MB floor costs more RSS and less throughput than a 64 MB one


rubys:
`SPINEL_GC_STR_BUDGET=fixed` / `SPINEL_GC_OBJ_BUDGET=fixed` (`ec4b8742`) exist so a measurement can hold the pacing still, and the natural second use is to bound a server's footprint. On this application a **16 MB floor is strictly worse than a 64 MB one on both axes** — more resident memory and less throughput — which is the opposite of what someone reaching for the knob would expect.

Not claiming a defect. The knob is documented as a diagnostic and this may be exactly what it should do. But the shape is reproducible and I cannot account for it, so it seems worth your eyes.

## The sweep

campfire, `/rooms/1`, one worker, `wrk -t4 -c16`, 10s warmup then 20s measured, sqlite pool held at 8 (it does not matter — see below). Both heaps pinned to the same floor, never one alone. PSS read from `smaps_rollup` immediately after the measured window.

| floor | aggregate trigger | PSS | req/s | collections | full | live str |
|---|--:|--:|--:|--:|--:|--:|
| 4 MB | 16 MB | 107 MB | 33 | 28,171 | 1.5% | 96.2 MB |
| **16 MB** | 64 MB | **279 MB** | **346** | 1,376 | 26% | **172 MB** |
| **64 MB** | 256 MB | **250 MB** | **391** | 397 | 49% | **13.8 MB** |
| 256 MB | 1024 MB | 829 MB | 405 | 105 | 49% | 40.9 MB |
| adaptive | 188 MB | 240 MB | 381 | 885 | 42% | 45.1 MB |

## It is not noise

Three runs of each of the two cells that invert:

```
budget   run   loadPss   req/s   liveStr   colls
16384    1       280     345.00   157.6     1368
16384    2       291     346.81   186.6     1378
16384    3       267     347.16   172.1     1376
65536    1       251     391.50    13.8      397
65536    2       250     391.32    13.8      397
65536    3       249     390.44    13.8      397
```

The 64 MB cell repeats to within 2 MB of PSS and reports the same 13.8 MB live string every time. The 16 MB cell sits 29 MB higher and 12% slower, with a live string set twelve times larger.

## Why I am not offering a mechanism

[...]

PR #4408, opened 2026-09-09
A benchmark can use threads, and one of them measures the collector

rubys:
This is the `benchmark/` extraction from #4384. @bartleusink noted that nothing
in `benchmark/` exercises the collector under concurrency and that the campfire
port was the only threaded-GC workload he could find; this is that shape as a
program in the tree, with the assertion you asked for.

## `bench` could not have hosted one

The recipe compiles and links by hand and always links `$(SP_RT_LIB)`, so a
benchmark using `Thread` would have failed to link — the archive selection in
`src/main.c` never runs on that path. It now reads the same
`/* SPINEL_USES_THREADS */` marker the test harness reads, and picks the
archive, `-DSP_THREADS` and `-lpthread` the same way. That is three lines, and
it is what makes the rest of this possible at all.

## The workload

`benchmark/bm_threaded_render.rb`: 32 green threads, each a request that
RETURNS its markup up a nest of views — page → room → message → body → span —
holding every fragment it has produced live while it produces the next one,
across a `Thread.pass`. Nothing is shared between requests: every string
interpolates the request id, so the live set scales with the number in flight
rather than sitting flat in a fixture.

Four properties are load-bearing and the header says which and why, so a later
change that makes it faster by removing one can see what it removed: the
nesting with the caller retaining, the yield, the freshly allocated text, and
string bytes over object bytes.

CRuby 2.9s, spinel 1.4s here.

## What it reproduces

The pacing has to be held still first. The adaptive budget scales per worker,
so a wider budget collects less often with more live at each collection, and
the mark then moves for a reason that is not the mark's. Pinning the aggregate
across the ladder (object floor pool-wide, string floor per worker, so both
bases divide by the worker count):

```
SPINEL_GC_OBJ_BUDGET=fixed SPINEL_GC_STR_BUDGET=fixed \
SPINEL_GC_THRESHOLD_OBJ_KB=$((65536/W)) SPINEL_GC_THRESHOLD_STR_KB=$((65536/W)) \
SPINEL_WORKERS=$W SPINEL_GC_PHASES=1
```

Apple M-series, 6 performance cores:

| W | colls | objs/coll | mark/coll | root walk | scan | ns/obj | slot sweep/coll |
|--:|--:|--:|--:|--:|--:|--:|--:|
| 1 | 37 | 182,158 | 2.81 ms | 0.000 ms | 2.81 ms | 15.4 | 5.89 ms |
| 2 | 38 | 183,220 | 3.34 ms | 0.000 ms | 3.34 ms | 18.2 | 4.24 ms |
| 4 | 37 | 182,923 | 3.51 ms | 0.000 ms | 3.49 ms | 19.2 | 3.14 ms |
| 8 | 38 | 182,717 | 3.66 ms | 0.000 ms | 3.63 ms | 20.0 | 2.95 ms |
| 12 | 36 | 177,930 | 3.58 ms | 0.000 ms | 3.56 ms | 20.1 | 2.39 ms |

Collections and objects-marked are flat — that is the control, and it is what
lets the other columns be read. The slot sweep falls to 0.41x; it is the phase
that gets the parked workers and it behaves. The mark rises 1.27x over the same
range with no more graph to walk. Same signs as the application ladder in #4384
(mark 2.04x, slot sweep 0.51x), on a smaller machine and a smaller graph.

**The mark split from `e62d2f75` says where that rise is, and it is not where I
guessed in #4384.** The root walk — roots, fibers and globals — is 0.000 ms at
every worker count. The whole of the mark, and the whole of the rise, is the
trace: 15.4 → 20.1 ns per object marked over a graph that did not grow. So
"more worker threads means more stacks to scan as roots" is falsified on this
shape, and slicing the fiber list to the parked workers would reach a column
that is already zero. Detail on #4384.

## The gate leg

`threaded-render-test` runs it for its ANSWER, not its time. The digests are
position-sensitive and fixed by each thread's index, so they do not depend on
the interleaving — a collection that frees a live fragment or hands one thread
another's buffer changes stdout rather than only a timing. The matrix is
worker count {1, 2, 8} crossed with the budget policies {default, obj, walk},
plus `SPINEL_GC_MINOR=1`: those are what move WHEN a collection lands relative
to a half-built page, which is the state the last three defects found on this
workload were in. `SPINEL_WORKERS=1` is in there deliberately — a green thread
parked at a yield with the collector running on its own worker is a different
root set from one parked while another worker collects.

Two shape assertions keep it discriminating rather than merely green, since a
workload that stops collecting still prints the right digests: at least 8
collections, and more than 1,000 objects marked per collection. Both confirmed
failing rather than assumed — a perturbed digest is caught at W=1, and pointing
the leg at a small threaded program trips the live-graph floor:

```
threaded-render-test: FAIL (W=1 default changed the answer)
threaded-render-test: FAIL (mark walks 91 objs over 11 collections: the live graph is gone)
```

Wired into `test-run`. `make bench` is 62 pass / 0 fail; `gc-phases-test`,
`gc-threshold-test`, `gc-obj-budget-test` and `gc-minor-test` all still pass.

Sized for the gate rather than for a paper — one leg is about 11s wall here. If
you would rather it were larger for ladder work, or smaller for the gate, say
which and I will resize it; the constants are all at the top.

https://claude.ai/code/session_015W6Dvd4ZtddD6TfzP9B5NE

## Summary by CodeRabbit

* **Tests**
  * Added a threaded rendering benchmark that exercises concurrent page rendering and garbage collection across multiple worker and budget configurations.
  * Added deterministic output validation, including per-thread render results and aggregate byte totals.
  * Included the threaded rendering test in the standard test run.

* **Chores**
  * Updated benchmark execution to automatically select the appropriate threaded runtime and compiler settings.


matz:
Merged as `5e52a905`. This is the third thing your application has had to answer for us, and it is the one that stops it having to.

## Verified here

* `threaded-render-test` passes, and catches a one-digit perturbation of the expected file: I changed `digest=249049704` to `...705` and the leg failed. A passing leg that cannot fail is the thing I would have checked anyway, so thank you for having checked the shape assertions the same way and said so.
* `make bench` is 62 pass / 0 fail. The marker-based archive selection works on this box.
* Gate green in 124s, of which the leg is 32s. On your M-series it was 11s, so it is about 3x that here; I am leaving it at your sizing. If the gate wall becomes a problem it is a knob we can turn later, and the correctness matrix is worth more than the seconds.
* Expected file regenerates byte for byte from CRuby 4.0.4.

## The `bench` recipe

You are right that it could not have hosted one, and the reason is worth restating because it is the general shape of that class of gap: the archive selection lives in `src/main.c` and the recipe bypassed it by driving `cc` itself. So the suite was structurally unable to contain the workload, and "there is no threaded GC benchmark" read as an omission when it was a wall. Three lines, as you say.

## The finding

I reproduced your split on this machine and it says what yours says:

```
[gcph] marked 2179545 objs  swept 5144609 slots
[gcph] mark: roots 0.000s  fibers 0.000s  globals 0.000s  scan 0.104s
```

The root walk is zero, all of it, and the whole mark is the trace. That falsifies my own reading as much as your original one: I wrote on #4384 that slicing the fiber list to the parked workers reaches the root walk and that its share grows with parked fibers, and offered the 2%-at-1-fiber / 42%-at-128 table as the reason to measure before pulling the lever. On this shape the column it reaches is 0.000 s at every worker count, so measuring first was right and the lever I had in mind is the wrong one.

What is left is 15.4 to 20.1 ns per object marked over a graph that did not grow, which is a per-object cost that rises with worker count and no more objects to walk. That is a different question from the one #4384 opened with and I have said so there.

One number from this box, since it also exercises the gate from `b64b1204`: the workload reports `mark share 0.55`, right in the 0.45-0.72 band you measured on the application. The synthetic reproduces the policy choice as well as the phase split.

The `SPINEL_GC_PHASES` line also now splits the string live set by generation (`fdeee693`), which is what #4407 needed:

```
[gcph] string live 170.9 MB young + 122.1 MB old  (major at 109.9 MB old)
```

PR #4422, opened 2026-09-10
String#slice!(start, len) in value position roots the part it hands back

ideasasylum:
The issue was identified and diagnosed by Claude in a spinel project I was working on, as was the suggested fix.

## What goes wrong

When `str.slice!(start, len)` is used as a value and `str` is a local or an instance variable, the string it
returns is sometimes one the garbage collector has already freed. CRuby is correct on the same program.
What the caller then sees depends on what malloc did with the memory next: an empty string, a string with
a wrong length, or a crash (SIGSEGV) in whatever reads the string next.

## Reproducer

```ruby
buf = +""
bad = 0
i = 0
while i < 200000
  want = (i + 1) % 50 + 1
  buf << ("x" * want)
  line = buf.slice!(0, want)
  bad += 1 if line.empty?
  i += 1
end
p bad
```

CRuby prints `0`. Spinel `6ada5aee` prints `26` on macOS arm64 and segfaults on Linux arm64 (gcc 13.3).
With `SPINEL_GC_STRESS=1` and 20000 rounds it prints `532` on macOS. The width has to vary: with a constant
`want`, malloc hands the freed block straight back for an identical string and the bug is hidden.

## Why

For `buf.slice!(0, want)` the compiler emits, in this order:

1. build the removed part as a new string and keep it in a plain C local;
2. rebuild the receiver: two `sp_str_sub_range` calls and one `sp_str_concat`, three allocations;
3. hand back the local from step 1.

Any allocation in step 2 can start a collection. The collector is precise: it only keeps what is reachable
from a registered root, and the local from step 1 is not registered anywhere. So the collector frees it, and
step 3 hands the caller a pointer into freed memory.

Two core dumps showed exactly this: the returned string's memory block was already on malloc's free list, and
the length field that sent the caller off the end of the heap was half of a free-list pointer.

## The fix

Register the removed part as a root (`SP_GC_ROOT_STR`) for the rest of the block, so it survives the
allocations in step 2. This is the same idiom the generator already uses in some forty other places for a
temporary that has to live across an allocation. Two arms of the `slice!` emitter build the removed part with
`sp_str_sub_range` and get the same line: `slice!(start, len)` and `slice!(range)` / `slice!(i)`.
`sp_box_str` does not allocate, so nothing else can collect between the end of that block and the caller's
variable.

The other arms were checked and are not affected: `slice!(str)` hands back its argument, which the runtime
helper roots itself, and `slice!(/re/)` and `slice!(/re/, n)` hand back match-register strings the collector
already sees as globals. Under `SPINEL_GC_STRESS=1` they stay at 0.

## Checks

- New `test/string_slice_bang_value_rooted.rb`: `slice!(start, len)` on a local, the same on an instance
  variable, and `slice!(range)`, 200000 rounds each. Before the fix it prints `26`, `31` and `26` (macOS
  arm64) and segfaults on Linux arm64; after, `0`, `0`, `0`.
- `make test` (3570 pass) and `make bench` (62 pass) on macOS arm64 (Apple clang 21).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `String#slice!` to reliably return the removed substring, preventing corrupted or invalid string values during repeated operations.
  - Improved handling when slicing strings directly, through instance variables, or with ranges.

- **Tests**
  - Added coverage for repeated `String#slice!` operations and verified that all returned values have the expected size.


matz:
Merged. Thank you -- the diagnosis is right and the write-up made it quick to check.

The symptom does not appear on this box (Linux x86_64, glibc): your program prints 0 here, plain and under `SPINEL_GC_STRESS=1`. So I checked the claim the way it can be checked independently of the allocator, in the generated C:

```c
_t4 = sp_str_sub_range(lv_buf, _t1, _t2);
lv_buf = sp_str_concat(sp_str_sub_range(lv_buf, 0, _t1),
                       sp_str_sub_range(lv_buf, _t1 + _t2, _t3 - _t1 - _t2));
} _t4;
```

`_t4` holds the removed part, three allocating calls follow, and nothing roots it. That is the defect whether or not a given malloc happens to expose it, and your note about the width having to vary explains exactly why one allocator shows it and another does not.

The root placement is right too: it sits inside the `if` block, so it covers the three allocations and pops before the value is handed back, which is fine because nothing between there and the caller allocates. You said so in the PR and I checked it rather than taking it.

`make gate` is green here (3570 tests, optcarrot checksum 59662), and CI is green on all three legs including macOS arm64 -- the platform where you measured 26, so that leg is the one that would have caught a regression.

One thing worth knowing about the test you added, and it is not a criticism: it cannot fail on this machine, in either direction. It guards the property on the platforms where the allocator exposes it, which is the same situation as the FIFO test in `docs/limitations.md`. That is worth having; it just means a green run here says less than a green run on your box does.

Issue #4420, opened 2026-09-10
packages/net: `Net::HTTP#request` ignores its block; `Net::HTTP.start` has no `ipaddr:`


rubys:
Verified on origin/master 9c15f27b. Serve with this listener (`ruby listener.rb &`):

```ruby
require "socket"
s = TCPServer.new("127.0.0.1", 8792)
3.times do
  c = s.accept
  while (l = c.gets) && l.strip != ""; end
  c.write "HTTP/1.1 200 OK\r\nContent-Type: text/html\r\nContent-Length: 2\r\n\r\nhi"
  c.close
end
```

then:

```ruby
require "net/http"
require "uri"
url = URI.parse("http://127.0.0.1:8792/")
Net::HTTP.start(url.host, url.port, use_ssl: false) do |http|
  http.request(Net::HTTP::Get.new(url)) do |response|
    puts "BLOCK code=#{response.code}"
  end
end
puts "done"
```

CRuby prints `BLOCK code=200` then `done`; the spinel binary prints only `done`. The package's `#request` takes no block (`net/http.rb`'s header says so under "no streaming body block"), and the caller's block is dropped without a diagnostic — see also the undeclared-keyword report for the same silence on `ipaddr:`.

Both spellings are how a Rails app that pins DNS is written:

```ruby
Net::HTTP.start(url.host, url.port, ipaddr: ip, use_ssl: url.scheme == "https") do |http|
  http.request(Net::HTTP::Get.new(url)) do |response|
    ...
  end
end
```

Since the body is already read whole, the block form is `yield res if block_given?; res` at the end of `#request`; `ipaddr:` is the connect target with `Host:` and the TLS hostname still `address`.

matz:
Both package gaps are closed on master (`6ada5aee`), and one of your three spellings still does not work -- for a reason that is not in this file.

## What works now

Verified against your listener.

- `ipaddr:` on `Net::HTTP.start`. It is the connect target only: `Host:` and the TLS hostname stay `address`, which is the point of separating them. Held as a String with empty meaning unset rather than nil-or-String, so CRuby's reader answers nil where this one answers `""`.
- A block on `#request`, outside a `start` block.

[...]

Issue #4417, opened 2026-09-10
`k.new(x)` on a class object passed as a value answers nil when the class is nested


rubys:
Verified on origin/master 9c15f27b.

```ruby
require "net/http"
require "uri"
class Req
  def initialize(u)
    @u = u
  end
  def method
    "M"
  end
end
def mk(k, u)
  k.new(u)
end
u = URI.parse("https://www.example.com/")
a = mk(Req, u)
puts "top-level: nil?=#{a.nil?} method=#{a.method}"
b = mk(Net::HTTP::Get, u)
puts "nested: nil?=#{b.nil?}"
puts "nested method=#{b.method}"
puts "done"
```

Expected (CRuby):

```
top-level: nil?=false method=M
nested: nil?=false
nested method=GET
done
```

Observed:

```
top-level: nil?=false method=M
nested: nil?=true
undefined method 'method' for nil (NoMethodError)
```

No compile-time diagnostic. A top-level class passed the same way constructs correctly; `Net::HTTP::Get` (any nested class — the package is incidental) comes back nil. campfire's `Opengraph::Fetch` is written on exactly this: `request(url, Net::HTTP::Get, ip: ip)` then `request_class.new(url)`.

matz:
Fixed on master (`9d271fe7`).

**The trigger is not nesting.** A top-level class with the same signature does it too, and a nested class without an optional parameter does not. What decides it is `initialize` having an OPTIONAL parameter: the class-value `new` dispatch built an arm only when `argc == nparams == nrequired`, so `initialize(path, initheader = nil)` -- which is exactly `Net::HTTP::Get`'s -- matched no class at all. `Net::HTTP::Get` was the class that bit you because of its signature, not its scope.

The switch then came out EMPTY and the nil seed fell straight through it. That is the second half, and it is the reason there was no diagnostic: an empty switch answers its seed. Both emitters now carry a `default: sp_raise_nomethod`, which the sibling emitter for a poly receiver has had all along.

An arm is now built for any argc between nrequired and nparams, with the rest filled from their defaults -- what a statically known `Klass.new` has always done.

[...]

PR #4423, opened 2026-09-10
A declined inline arm must not leave the call silently unemitted

rubys:
`emit_poly_recv_block_dispatch` (src/codegen_iter.c) opens one switch arm per
candidate class and fills it with `emit_inline_call`, **whose return value it
discarded**. A candidate that consumes its block through a declared `&blk`
rather than `yield` is not an inlining target — it has a standalone C function
instead, which is the whole point of that spelling — so the inline declined,
the arm came out as `case N: { break; }`, and the call disappeared with no
diagnostic.

That is the live half of #4420. The two spellings emitted differently:

```ruby
res = http.request(req)                    # case 29: _t230 = sp_Net__HTTP_request(..., NULL); break;
http.request(req) { |res| ... }            # case 29: { break; }
```

So it is not that the block is dropped from a call that still happens — the
call is not emitted. Not net/http-specific either: any mix where one candidate
yields and another declares `&blk` does it, which is the first test here.

**Fix:** build the switch into a scratch buffer and commit it only once every
arm has emitted a body. If one declines, discard and `return 0` so the ordinary
poly dispatch emits the call and threads the block, the way it already does for
a typed receiver. That is the same policy as the `|x|` gate a few lines above —
hand it to a path that emits something, never leave a silent empty body.

**Second hole, same switch:** it closed with no `default:` arm outside the
`map!`/`collect!` special case, so a receiver whose class is not in the
candidate set fell through it and the call silently did nothing where CRuby
raises `NoMethodError`. #3234 is that hole found from the builtin-array side,
patched then only for `map!`/`collect!`; this closes it generally, with the same
`sp_raise_nomethod` every other poly dispatch already emits.

## Verification

- Both tests **fail before** the change and pass after (checked by stashing the
  `src/` hunk and rebuilding, not by inspection).
- `make test`: **3570 pass, 0 fail, 0 error**.
- #4420's own repro against your listener now prints `BLOCK code=200` / `done`,
  matching CRuby; the no-block spelling is unchanged.

## One reviewer note

`Buf *b_sv = b; b = &sw;` rebinds the parameter so the `map!`/`collect!` block
below needs no edit. It keeps the diff to the part that changed, but if you would
rather see that block take `&sw` explicitly, say so and I will widen it.

https://claude.ai/code/session_01JK9T1ceGx3kPYXkSENZWia

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed block calls on mixed receiver types so methods using declared block parameters execute correctly instead of being skipped.
  - Calls on receiver types that do not support the method now raise `NoMethodError`, matching expected Ruby behavior.

- **Tests**
  - Added regression coverage for declared block parameters and unsupported receiver types in polymorphic dispatch.


matz:
Merged. This is the live half of #4420 and your reading of it corrects mine, so let me say which part.

I reported the mechanism as "the dispatch arm passes NULL where the callee's `&blk` goes", and quoted this line:

```c
case 29: _t260 = sp_Net__HTTP_request((sp_Net__HTTP *)_t259.v.p, ..., NULL);
```

That line is from the NO-BLOCK spelling in my own test program, where NULL is correct. The block spelling emits `case 29: { break; }` and no `default:`, exactly as you say -- the call is not made at all, rather than made without its block. I read the wrong line of my own output.

Verified rather than taken:

- Both tests fail with master's `codegen_iter.c` and pass with yours. (My first attempt at that check was `git stash push` on a branch where your change is committed, which stashed nothing and told me the tests passed either way. `git checkout master -- src/codegen_iter.c` is the check.)
- #4420's three spellings all match CRuby against a real listener now: `DIRECT-BLOCK`, `IPADDR`, `NESTED-BLOCK`.
- `make gate` green here, and CI green on all three legs.

## On your reviewer note

`Buf *b_sv = b; b = &sw;` is correct and I would take it as it stands -- everything below writes to the scratch and the swap back is unconditional. But I would rather the region were explicit, for the reason you half-anticipated by asking: the `map!`/`collect!` block is a long way from the rebinding, and the next person to add a `buf_puts(b, ...)` in there gets the scratch buffer without being told. That is a hazard the comment cannot reach, because it will be read after the edit rather than before.

You offered to widen it; I will do it myself rather than spend a round trip on five mechanical lines. The fix is yours either way.

The second hole -- no `default:` outside the `map!`/`collect!` case -- is the better catch of the two, and I would not have gone looking for it. A receiver whose class is outside the candidate set silently doing nothing where CRuby raises is the same shape as the empty switch in #4417 last night: an empty arm and a missing default both answer by not answering.

PR #4442, opened 2026-09-12
fix: invoke boxed callables in a user-shadowed poly .call/[] (#4395)

DmitryBochkarev:
A `poly` receiver routes `.call`/`[]` through `emit_poly_method_dispatch`, which switches on the boxed value's runtime class and enumerates only user-defined classes. That switch is reached only when NO user class defines `call`: `emit_call_body` gates its dedicated poly `.call` fast path on exactly that condition. So in a program that defines one -- e.g. `Defer#call` iterating an untyped `@blocks` array and calling `@blocks[i].call` -- the boxed Proc in the slot matches no `case` and hits the raising default:

    undefined method 'call' for an instance of Proc

instead of being invoked. The fix adds a runtime callable pre-arm ahead of the class switch and types the result dynamically, and hardens the bound-Method legacy ABI that the new dynamic path made reachable.

Poly dispatch (src/codegen_call.c, src/analyze_infer.c)
* `emit_poly_callable_prearm` runs ahead of the class switch in both the `argc == 0` and `argc > 0` branches of `emit_poly_method_dispatch`, so a slot shadowed by a user `call`/`[]` still reaches a boxed Proc/Curry/Method. Procs and Curries go through `sp_poly_callable_call`; a bound Method uses the same self-ful/self-less cast the unshadowed fast path already emitted.
* The analyzer now types every `<poly>.call`/`()` result as `TY_POLY` unconditionally. The result is genuinely dynamic (a boxed slot may hold a Proc regardless of user classes), so the user methods' return type must not constrain the slot; the old `has_user_call` scan is gone. This also fixes a mixed slot (Proc returning a String, user class returning an Integer) being coerced to int.

Bound-Method legacy ABI (src/codegen_call.c, lib/sp_proc.h, lib/sp_proc.c) The dynamic cast `sp_int (*)(void *, sp_int...)` reads raw C types from registers supplied by a call site that no longer knows the target, so the pre-arm exposed several pre-existing unsoundnesses. The classifier is hardened and its answers are persisted on the Method:
* `method_legacy_int_abi` stamps `legacy_int_abi`, an eight-char per-position `legacy_sig` (each scalar kind gets its own token, `TY_UNKNOWN` is the `0` wildcard, and every other kind uses `100000 + TyKind` so pointers never collide with the scalar range), `legacy_fixed`, `legacy_rest` and `legacy_ret`. `sp_bm_legacy_abi_ok` / `sp_bm_sig_pos_match` compare position by position: a bool is never handed an int, a `String*` is never handed an `IntArray*`, an `sp_Foo*` is never handed an `sp_Bar*`.
* A regular method may ride the cast only with a `TY_INT` return; a synthesized `__bam_` wrapper and the typed-array adapters record their real Ruby return (`SP_BM_RET_INT/STR/INT_ARRAY/STR_ARRAY`) and box through `sp_bm_box_ret`. The `TY_INT` default goes through `sp_box_int_or_nil` so the nullable `SP_INT_NIL` sentinel (an out-of-range `IntArray#[]`) is nil rather than a truthy Integer.
* Declined (raising `NoMethodError`, matching CRuby) instead of reading garbage: any rest parameter (the callee prologue roots its trailing `sp_PolyArray*` unconditionally, so even an unread rest segfaulted under GC), keyword and short optional calls, a class method whose C signature needs its receiving class, a float/poly/by-value-struct parameter, a per-position class mismatch, an `UnboundMethod` read out of a container, and `argc > 16`.
* `recv_bound` records whether the target's C ABI leads with the receiver. It replaces the `m->self != NULL` test, which cannot tell a self-less top-level method from a receiver whose VALUE is 0/false/nil and so dropped the receiver (`[0.method(:+)][0].to_proc.call(5)` returned garbage, `[false.method(:to_s)][0].to_proc.call` answered "true").
* `sp_method_proc_tramp` (the generic `Method#to_proc`) cannot see call-site types, so it now declines any target that is not legacy-callable, is unbound, has a rest/optional parameter, exceeds 16 fixed slots, mismatches `argc`, or carries a pointer parameter (it can only forward raw `sp_int`), and selects the self-less cast by `recv_bound`. Its dispatch is extended from 4 to 16 arguments; the old default silently dropped arguments 5 and up.
* The runtime `[]`/slice Method arms (`sp_poly_slice`, `sp_poly_arr_get_hash_cold`, `sp_poly_index_int`) validate with `sp_bm_legacy_abi_ok` and fall back to `sp_poly_callable_call`. `Proc#[]` on an int index and a 2-arg slice on a callable now reach the callable instead of answering nil.

Receiverless Kernel wrappers (src/codegen_fold.c, src/codegen_call.c)
* A receiverless wrapper's `__bam_r` is a real argument, not the bound receiver. `emit_args_filled` now counts it, so `method(:String).call` raises ArgumentError instead of invoking the wrapper with an undefined register; the upper-bound check is skipped for the variadic/two-arg builtins (puts/print/p/pp/Rational/Complex), which keep the previous truncation.
* `Method#arity` reports the real CRuby values from a `{"Kernel", ...}` section of `sp_builtin_arity_tbl`.
* `method(:sym)` on a `TY_POLY` receiver raises `NoMethodError` at run time instead of emitting `(void *)(sp_RbVal)`, which did not compile; the emitted fn was already 0 for that receiver.

Fast-path guards
* The no-user `.call`/`[]` fast paths gained the same `argc > 16` decline as the pre-arm, so a 17-argument proc call no longer emits `(sp_int[16]){...17...}` and writes `_sp_proc_poly_args[16]` out of bounds.

Tests
* New: `poly_call_user_defined_call` (the repro), `poly_call_legacy_abi_gate` and `poly_call_fast_abi_gate` (the ABI classification on the shadowed and unshadowed paths, including the GC-crash rest regression under an allocating body), `poly_call_result_method`, `bound_method_value_recv` (value-0/false receivers), `kernel_method_obj`, `poly_proc_aref`.
* `issue_3231` updated: a rest-target Method read out of a container now declines instead of being called through the cast.
* Makefile: the two ABI-gate tests are filtered under `SPINEL_INT_OVERFLOW=promote`, which boxes int parameters and does not consult the raise/wrap legacy classification.

Verification
* `make check`: 3594 pass, 0 fail, 0 error; alloc-report-test, infer-test and spin-e2e green.
* The reported program compiles and runs with output identical to CRuby (hello, world! / main on exit / config on exit, exit 0).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved polymorphic calls for `Proc`, `Curry`, and `Method` values, including `Proc#[]` behavior.
  - Corrected receiver handling for bound, unbound, class, top-level, and zero-valued receivers.
  - Added safer compatibility checks so unsupported signatures, arities, and parameter types raise `NoMethodError` instead of crashing.
  - Corrected kernel method wrapper arity reporting and argument validation.
  - Preserved dynamic results when callable values share slots with user-defined `call` or `[]` methods.
  - Updated rest-parameter call behavior to report declined calls.


matz:
Merged with one commit on top (d879c8b0). CI was red on optcarrot: the classifier let a regular Method ride the legacy cast only with a TY_INT return, and optcarrot's CPU store table is mostly not that (`Pulse#poke_0` ends in an IntArray assignment, `Triangle#poke_0` in a comparison, `APU#poke_4017` answers a poly, `CPU.poke_nop` is void), so `@store[addr][addr, value]` raised on the first such poke and the checksum was 0.

The bind site now stamps a kind for every return the sp_int cast can read (nil, bool, Symbol, the two typed arrays, a user-class pointer) and a poly return takes an sp_RbVal cast in every arm, including the trampoline. Checksum and fps are back to master's. `poly_call_legacy_abi_gate` changed accordingly: the bool and Symbol cases now answer the value instead of declining. Thanks for the careful work on the ABI gate.

PR #4453, opened 2026-09-13
Call `Method` objects that flow through poly slots: boxing, splat dispatch, argument binding, and GC safety

DmitryBochkarev:
Spinel lowers `.method(:m)` to a bound-`Method` object, and a call through it (`m.call(…)`, `m[…]`, `m.()`, `m === x`, `m.to_proc`, `&m`) is dispatched at compile time when the target is statically known. But a `Method` that escapes the local static view — stored in an array/hash alongside other values, returned from a method, passed as a parameter — is *poly*, and the call site can no longer see the target's C signature. That path previously worked only for a narrow slice of shapes.

This branch makes the poly-slot `Method` call path behave like the direct one:

- boxes every C return kind the target can have (Integer, String, Bigint, nil, `true`/`false`, Symbol, typed array, user object),
- dispatches a **splatted** call on the value's actual class (Proc vs. Curry vs. Method) instead of casting a `Method` to `sp_Proc`,
- types and boxes typed-array adapter Methods (`arr.method(:push)`, `arr.method(:[]=)`) and stamps their real Ruby arity,
- binds arguments to the target the way CRuby does (optional defaults, rest, keywords/`**kwrest`, splats), including defaults that read earlier parameters or create callee locals,
- roots fresh receivers across the allocating `Method` constructor, and
- refuses the shapes it cannot model with a clean `NoMethodError` / `ArgumentError` instead of reading the wrong C type or jumping through a NULL function pointer.

## Motivation: optcarrot's memory-mapping dispatch tables

Spinel's headline workload is optcarrot. Its CPU builds a store table of captured `Method`s and invokes them later through a poly container:

```ruby
# optcarrot's CPU memory-mapping shape
add_mappings(0x0000..0x07ff, @ram, @ram.method(:[]=))
# ... later, out of @store, which holds Methods alongside other values:
@store[addr][addr, val]
```

The captured `Method` leaves the compile-time view at `add_mappings`, so the whole call goes through the poly path. Anything that path does wrong — wrong return kind, wrong argument binding, a stray allocation under GC — shows up as a crash or a wrong value in the hottest loop of the benchmark.

The same pattern is common in ordinary Ruby too:

```ruby
handlers = [model.method(:on_save), model.method(:on_delete)]
handlers.each { |h| h.call(record) }

# passing a Method where a block is expected
%w[a b].map(&obj.method(:transform))
```

### 1. Return kinds other than Integer

A `Method` read out of a container used to be callable only when its target returned a value that rides the `sp_int` register as an Integer. String, Bigint, nil, and reference-object returns all declined with `NoMethodError`.

```ruby
class Resolver
  def resolve(name) = name.upcase      # String
  def version      = 2 ** 64           # Bigint (sp_Bigint *)
  def lookup(id)   = Node.new(id)      # user object
  def nothing      = nil               # C return is void
end

r = Resolver.new

# poly slot: the compiler can no longer see the target
table = [r.method(:resolve)]
puts table[0].call("a")     # was: NoMethodError;  now: "A"
puts table[0][1]            # was: NoMethodError;  now: "A"
puts table[0].to_proc.call("a")

puts [r.method(:version)][0].call          # was: NoMethodError;  now: 18446744073709551616
puts [r.method(:nothing)][0].call.inspect  # was: NoMethodError;  now: "nil"
puts [r.method(:lookup)][0].call(7).class  # was: NoMethodError;  now: Node
```

Each return kind is boxed through the `sp_bm_box_ret` arm that matches the target's real C return, so the register is never reinterpreted as the wrong Ruby value.

### 2. Typed-array adapter `Method`s (arity + boxing)

For a typed array, `arr.method(:op)` is a synthesized adapter. Its `.call` used to hand the raw register back as an Integer instead of the array/string it stands for, and its `.arity` was `nil`.

```ruby
ram = [0] * 0x800
set = ram.method(:[]=)

set.call(0x100, 7)      # was: wrote a garbage element / C type error
p ram[0x100]            # was: wrong;  now: 7

p set.arity             # was: nil;    now: -1  (CRuby's Array#[]= arity)

sa = ["x"]
p sa.method(:push).call("y", "z").inspect   # variadic: pushes BOTH now
p sa.inspect                                 # => ["x", "y", "z"]
```

Runtime splats are expanded against the adapter's fixed arity and each element is laundered by its **absolute** adapter position, so a fixed index before the splat doesn't get mistaken for the value slot:

```ruby
sa = ["x", "y"]
p sa.method(:[]=).call(0, *["z"]).inspect   # was: TypeError (String laundered as index)
p sa.inspect                                 # => ["z", "y"]
```

### 3. Splatted call on a poly-slot `Method` (was a segfault)

```ruby
class Calc
  def sum2(a, b) = a + b
end

slot = [Calc.new.method(:sum2)]
args = [1, 2]
puts slot[0].call(*args)   # was: SIGSEGV (Method read as sp_Proc);  now: 3
```

The spread now dispatches on the value's class — `Proc` through the proc ABI, `Curry` through the accumulator, `Method` through the generic trampoline with its boxed arguments laundered into raw `sp_int` slots.

### 4. Argument binding: defaults, callee locals, keywords

A default that reads an earlier parameter is evaluated at the call site, but the bound-`.call` frame previously emitted the callee's local (`lv_a`), which doesn't exist there — a **C compile failure**:

```ruby
class Point
  def initialize(x) = @x = x
  def shift(a, b = a + 5) = @x + a + b
end

p Point.new(100).method(:shift).call(1)         # was: C compile error;  now: 107
p Point.new(100).method(:shift).to_proc.call(1) # now: 107
```

Defaults that create locals, inline a block, destructure, or close over a captured local now declare those names in the call-site/trampoline frame:

```ruby
class Stats
  def m(a, c = [1, 2].map { |x| x + a }) = [a, c]
  def q(a, c = (t = a + 5; t * 2))        = [a, c]
end
p Stats.new.method(:m).to_proc.call(1).inspect  # => [1, [2, 3]]
p Stats.new.method(:q).call(1).inspect          # => [1, 12]
```

Keyword handling now matches CRuby's counting rules (a declared keyword or `**kwrest` binds by name and is not a positional; an undeclared keyword hash is a positional Hash):

```ruby
class Router
  def route(path, c: 3) = [path, c]
  def opts(path, **kw)   = [path, kw]
end
rt = Router.new

p rt.method(:route).call("/x", c: 9)      # => ["/x", 9]
p rt.method(:opts).call("/x", z: 2)       # => ["/x", {z: 2}]
p rt.method(:opts).call("/x").inspect     # omitted **kwrest is {}, not NULL
```

Incorrect counts raise CRuby's `ArgumentError` (with `given …` and the `required keyword:` suffix where CRuby has one), for both direct and splatted calls.

### 5. GC safety for fresh receivers

`obj.method(:m)` allocates, and a freshly constructed receiver has no other reference. Under GC stress the receiver could be swept before it was stored as the `Method`'s `self`, leaving a dangling `self`:

```ruby
# was: corrupt object / dangling self under GC stress
100.times { FreshRecv.new(7).method(:plus).call(1) }
```

The bind site now holds the receiver in a rooted C temporary across the constructor — and the same fix applies to `UnboundMethod#bind`, `Method#super_method`, and the generic `Method#to_proc` fallback, plus the poly-receiver `#to_proc` / `#unbind` / `#parameters` arms.

### 6. `Method` bound to a non-constant class value

```ruby
class Target
  def self.cm(a) = a
  def go = self.class.method(:cm).call(3)
end
Target.new.go   # was: jump through a NULL fn (crash);  now: NoMethodError
```

This shape can't be resolved statically; it now declines cleanly on every route (direct, poly slot, `#to_proc`) instead of dereferencing `NULL`.

### 7. Sharp edges now fail safely instead of misreading memory

- A pointer argument to an int-inferred parameter declines rather than reading the pointer as an Integer.
- A `false` argument to an int parameter declines rather than becoming a truthy `0`.
- A target with more than 16 positional parameters, or a rest target given more than 16 arguments, declines through `#to_proc` instead of reading `args[16]` past the argument array.
- An omitted `**kwrest` is CRuby's empty hash, not a NULL the callee dereferences.
- Missing adapter operands raise `ArgumentError` instead of reading an uninitialized register.

## Tests

New/updated fixtures, all CRuby-equal (snapshots from reference Ruby unless the declining cases necessarily differ):

| Fixture | Covers |
|---|---|
| `test/poly_method_return_kinds.rb` | String/Bigint/nil/object returns, adapter arity and boxing, adapter splats, `#to_proc` defaults and callee locals |
| `test/poly_call_fast_abi_gate.rb` | per-position ABI gating, rest/splat/keyword counts, class-value NULL targets, over-arity |
| `test/method_call_shapes.rb` | argument binding for `call`/`[]`/`.()`/`===`/`&`/`to_proc` |
| `test/bound_method_fresh_receiver.rb` | GC rooting (run under GC stress by `gc-minor-test`) |
| `test/class_value_method_call.rb` | NULL-fn class values raise on every route, both int-overflow modes |

Results:

- **Default mode, full suite: 3604 pass, 0 fail, 0 error** (includes `gc-minor-test`).
- **Promote mode** (`--int-overflow=promote`): 3478 pass / 35 fail / 91 error, vs. the `origin/master` baseline of 3379 / 33 / 190. No new errors; the only two fixtures whose failure mode changes (`builtin_arity_reach`, `thread_gc_start`) are already failing (as compile errors) on `master`, and the underlying case reproduces identically there.

## Known limitations

The remaining unsupported shapes are documented in `docs/limitations.md` ("Bound Methods read out of poly slots"), e.g.:

- a pointer argument to a parameter the analyzer seeded as `Integer`,
- any rest parameter, or a rest followed by a post-rest/keyword/`**kwrest`,
- splats wider than the 16-slot callable ABI,
- a value a typed array cannot hold,
- keyword arguments through a receiver-bound `Method#to_proc` (the proc ABI is positional),
- over-arity on a bound array operator (extra operands are ignored),
- `self.class.method(:m)` when `self.class` isn't a statically-known constant.

Each of these declines with `NoMethodError` (or CRuby's `ArgumentError` where the count is wrong) rather than producing a wrong value or crashing.

## Summary by CodeRabbit

- **New Features**
  - Improved calls through `Method`, `Proc`, and `to_proc`, including splats, defaults, keyword arguments, rest parameters, and typed-array adapters.
  - Added support for additional return types, including BigInt, strings, nil, and reference values.
  - Added clearer errors for invalid arguments, unsupported call shapes, missing targets, and excessive argument counts.

- **Bug Fixes**
  - Prevented crashes, invalid dispatch, incorrect argument handling, and unsafe value conversions during dynamic method calls.
  - Fixed handling of fresh receivers, captured locals, and keyword-rest parameters.

- **Documentation**
  - Documented limitations and supported behaviors for bound methods stored in polymorphic values.


matz:
Merged as 77e333c2, with one fix on top (c4d3c547): the direct `obj.method(:m).call(x)` form bound a fresh Method and then evaluated the target's allocating defaults in the same expression without rooting it, so poly_method_return_kinds' two_defaults loop jumped through a collected Method on every CI leg. The Method is now rooted for the length of the expression. Gate green, optcarrot checksum and fps unchanged.

PR #4472, opened 2026-09-14
A table of float rows gets the layout a table of int rows already had

bartleusink:
`TY_INT_ARRAY_ARRAY` stores an array of int-arrays as an `sp_PtrArray` of `sp_IntArray*`, so indexing yields a typed row and the inner loop is a machine operation on unboxed scalars. There was no float counterpart, so `Array.new(rows) { Array.new(cols, 0.0) }` — an ordinary numeric table, and the same source text that compiles well with an `Integer` element — fell to `TY_POLY_ARRAY`:

```c
/* Integer elements -- TY_INT_ARRAY_ARRAY */
lv_acc = sp_int_add(lv_acc, sp_IntArray_get(lv_row, lv_j));

/* Float elements -- no layout */
sp_RbVal _t11 = lv_acc; SP_GC_ROOT_RBVAL(_t11);
lv_acc = sp_poly_add(_t11, sp_poly_arr_get_hash(lv_row, lv_j));
```

Three costs in the float loop, none in the int one: a container-kind dispatch (the compiler cannot tell the row from a Hash), a tagged add, and a GC root registration — all three innermost. And no counted loop over contiguous `sp_float` for clang to vectorize.

## The change

The container half was already generic — `ty_is_ptr_array` existed and the runtime stores unboxed `void*` either way — so this is almost entirely an element cast. `TY_FLOAT_ARRAY_ARRAY` joins the kinds `ty_is_ptr_array` answers for, the `is_ia` flag in the ptr-array emitter becomes `ty_ptr_array_elem_ctype` naming `sp_IntArray` or `sp_FloatArray`, and the `OA_CLS_IA` element sentinel gains a sibling `OA_CLS_FA`. The two sentinels are distinct, so a table mixing int rows and float rows joins to a conflict and stays boxed, exactly as two user classes do.

**Nothing is added to the runtime:** `sp_FloatArrayPtrArray_inspect` was already in `lib/spinel_rt.h`, unreachable.

Like `TY_INT_ARRAY_ARRAY` and `TY_OBJ_ARRAY` the kind is produced only by the post-fixpoint `narrow_object_arrays` pass, never by forward inference, so it cannot cascade through the fixpoint; `ty_array_of` still declines to map a float array to it. The admitted op set is unchanged and `sort`/`min`/`max` still keep a component on the poly path, so the pass and the emitter stay in lockstep.

### Why it is appended to `TyKind` rather than placed beside its sibling

`abi_sig_token` encodes a kind as `100000 + (int)TyKind` and `sp_bm_sig_pos_match` reads that width and encoding. Inserting mid-enum renumbers every kind after it, which rewrites those tokens in the emitted C of programs this type has nothing to do with — two tests changed for that reason alone before the move, and stopped changing after it. It is self-consistent within a build either way (there are no range comparisons over the built-in kinds), so this is about blast radius, not correctness. The inline comment states the constraint in one clause; this is the reasoning behind it.

## Measurement

The trigger's inner loop becomes `lv_acc + sp_FloatArray_get(lv_row, lv_j)`. Same program, same output (`400000000.0`), best of three on one box: **2.60 s → 0.33 s, 7.9x.**

In the four-variant benchmark this came from, the float pair now matches its flat twin the way the int pair always did — nesting costs nothing once the layout exists:

| variant | outer ivar C type | before | after |
|---|---|--:|--:|
| `f_nested` | `sp_PolyArray *` → **`sp_PtrArray *`** | 2.34 s | **0.31 s** |
| `f_flat` | `sp_FloatArray *` | 0.26 s | 0.30 s |
| `i_nested` | `sp_PtrArray *` | 0.18 s | 0.21 s |
| `i_flat` | `sp_IntArray *` | 0.18 s | 0.20 s |

## Verification

Emitted C hashed for all 3,565 tests, baseline against patched: **byte-identical for 3,557**. Of the eight that move, seven lose boxed reads and none changes output:

| test | poly reads |
|---|--:|
| `float_array_array` (new) | 29 → 3 |
| `issue_2913` | 5 → 1 |
| `nested_container_array_surface` | 6 → 4 |
| `gc_first_thread_threshold`, `issue_3159`, `nested_array_element_narrows` | 2 → 1 |
| `gc_full_cycle_space_bound`, `thread_gc_start` | 2 → 2 (container type only) |

All 15 non-`.rb` targets pass — `ext-test`, `ext-cruby-test`, the six `gc-*`, `rbs-test`, `rbs-seed-test`, `reject-test`, `re-lit-test`, `backtrace-test`, `byref-capture-test`, `threaded-render-test`.

`test/float_array_array.rb` covers index / `[]=` / push / `first` / `last` / `length` / `empty?`, literals, a row bound into a local, the ivar-through-`attr_reader` path, object `inspect`, and a mixed int/float table that must **not** narrow.

## Two limits, both at exact parity with the int kind

Neither is introduced here; both behave identically for `Integer` and were verified to do so:

* a table built in one method and passed to another does not narrow;
* `p <the whole table>` keeps it boxed — passing the container to a builtin is not an op the pass models.

## Not addressed

An RBS seed still cannot request either nested kind: `docs/rbs-extract.md` maps `Array[<other>]` to `poly_array`, and because nested generics are not in the "dropped" list such a seed is silently honoured as poly, with no warning that it did nothing. That is separable from this change and left alone.

## New Features
- Added support for nested arrays of floating-point values.
- Nested float arrays support indexing, updates, appending, arithmetic, inspection, and mixed numeric rows.
- Added handling for null nested float-array values.

## Bug Fixes
- Improved type handling so nested float arrays retain their floating-point element type.
- Fixed assignments of polymorphic values to nested numeric arrays.

## Tests
- Added comprehensive coverage and expected results for nested float-array operations.


bartleusink:
Thanks — I checked both findings against the code rather than taking them as read. One is real and is now fixed in 251095f9; the other does not hold.

## Unbox poly row values before nested `[]=` — real, fixed

Confirmed, and it is worse than "the generated C fails for a polymorphic row assignment": it fails to compile *at all*, and it is not specific to float tables. Every pointer-array kind shares the arm, so on master today:

```ruby
def echo(x) = x
p echo(9)                      # widen echo's return to poly

boxes = [Box.new(1), Box.new(2)]
boxes[0] = echo(Box.new(3))    # error: initializing 'sp_Box *' ... from 'sp_RbVal'

ints = [[1, 2], [3, 4]]
ints[1] = echo([7, 8])         # error: initializing 'sp_IntArray *' ... from 'sp_RbVal'
```

So it predates this PR; the float kind inherits it by joining the same arm. The fix is the one suggested — mirror the push arm's `sp_poly_obj_ptr` unboxing, with a cast since this initializes a typed element pointer rather than passing a `void *`.

`test/ptr_array_set_poly_value.rb` covers all three kinds plus a directly typed value after each, so the non-poly path stays covered. It fails to build on master and passes here, which is what makes it a regression test rather than a description. Emitted C across the suite is unchanged by this fix — no test could have covered it, because the programs that reach it do not build.

## Preserve Ruby evaluation order for nested `[]=` — does not hold

The arm does emit the value before the receiver and index, but its arguments are already side-effect-free by then: an outer layer hoists the index and the value into temporaries **in Ruby order** before this emitter runs.

```ruby
tbl = [[1.0, 2.0], [3.0, 4.0]]
tbl[pick_idx] = pick_val
```

```c
({ sp_int _t5 = sp_pick_idx();                          /* index FIRST  */
   sp_FloatArray *_t6 = sp_pick_val(); SP_GC_ROOT(_t6); /* value SECOND */
   ({ sp_FloatArray *_t7 = _t6; sp_PtrArray_set(lv_tbl, _t5, _t7); _t7; }); });
```

Observed order is `["idx", "val"]` for both the int and float tables, matching CRuby exactly.

The receiver cannot carry side effects either: this arm is only reached when the receiver is a modeled slot read (a local, an `@ivar`, or its `attr_reader`), which the narrowing pass's admitted op set enforces. Writing a side-effecting receiver — `pick_tbl(tbl)[i] = v` — passes the table to a method, which kills the slot, and the whole thing emits `sp_PolyArray_set` instead. That was my first attempt at a repro and it proved nothing until I checked which arm had actually been emitted.

So there is no reachable ordering hazard here, and I have left that one alone.

PR #4483, opened 2026-09-14
The nested-seed build step says so when the binary does not build

bartleusink:
Follow-up to #4482, which merged before this landed on the branch.

The nested-array full-build leg I added to `rbs-seed-test` was chained with `&&`:

```make
$(SPINEL) test/rbs-seed/nested_array_ivar.rb --rbs test/rbs-seed/sig -o "$$tmp/na" 2>/dev/null && \
  { "$$tmp/na" > "$$tmp/na.out" 2>/dev/null; cmp -s ... || { ... ok=0; }; }; \
```

so a compile failure short-circuited the whole thing: no comparison, no message, and the test passed. A test that goes quiet exactly when the program stops building is worse than no test. Its neighbours in the same recipe all use an explicit `if ... else echo FAIL; ok=0; fi`, and this one now does too.

Verified rather than assumed — appending an undefined call to `test/rbs-seed/nested_array_ivar.rb` makes the leg report:

```
rbs-seed-test: FAIL (Array[Array[Integer]] seed did not reach the table)
rbs-seed-test: FAIL (Array[Array[Float]] seed did not reach the table)
rbs-seed-test: FAIL (nested array seed binary did not build)
```

and it passes again once reverted. Makefile only; no compiler change.

Found by CodeRabbit on #4482.

## Summary by CodeRabbit

* **Tests**
  * Expanded automated validation for nested array instance-variable handling.
  * Checks now verify that generated programs build, exit successfully, and produce the expected output.
  * Improved failure reporting identifies non-zero program exits and provides relevant output when results differ or the build fails.


bartleusink:
Both addressed in 1e564dbf, and checking the first one turned up a second, worse problem.

## The run's exit status — valid, fixed

Confirmed rather than assumed: appending `exit 3` after the final `p` left the test passing, because the run's status was discarded. It is checked now, with its own message.

## A temp-file collision I introduced

While verifying that I noticed the leg was writing `$tmp/na.c`, `$tmp/na.err`, `$tmp/na` and `$tmp/na.out` — every one already used by the `nilable_scalar_arg` leg further down the same recipe.

The happy path survives it, since each leg compares immediately after it writes. A failure does not: if `nilable_scalar_arg`'s own emission step failed, its build would pick up the stale `na.c` left by this leg, compile a different program, and diff it against `nilable_scalar_arg.expected` — a failure reported against the wrong test, which is exactly the kind of thing that burns an hour. The prefix is `nai` now, which nothing else in the recipe uses.

## All three branches verified by breaking the program

| injected | reported |
|---|---|
| `exit 3` after the last `p` | `FAIL (nested array seed binary exited non-zero)` |
| an undefined call | `FAIL (... seed did not reach the table)` + `FAIL (... did not build)` |
| a changed `p` | `FAIL (nested array seed output mismatch)` |

and `pass` again once reverted.

## One thing I deliberately did not change

The other **28** legs in this recipe discard their run's exit status the same way — the pattern is `"$$tmp/x" > "$$tmp/x.out" 2>/dev/null;` throughout. So this is the recipe's convention rather than an oversight in my leg, and sweeping all 29 seemed wrong to slip into a follow-up PR about one of them. Happy to do it as its own change if you want it.

Issue #4145, opened 2026-08-27
[Compiler] A method that recurses with a block of its own is inferred as answering `Integer`, and a block reaching it through another method's block parameter is built with no capture


elct9620:
`walk` calls itself with a block it writes, and is handed the block on line 14
through `hop`'s block parameter. That block runs — `$ran` reads 1, as it does
under CRuby — and what it assigns to `seen` does not reach `seen`. `walk` is
also inferred as answering `Integer`, where it answers `nil`.

Removing `hop` and writing the block at the call to `walk` prints what CRuby
prints, and leaves the inferred return type `Integer`. Having `walk` pass on the
block it was given rather than writing one prints what CRuby prints and infers
`nil`.

## Reproduction

`repro.rb`, entire, no requires:

```ruby
$ran = 0

def walk(n)
  yield
  walk(n - 1) { nil } if n > 0
  nil
end

def hop(n, &block)
  walk(n, &block)
end

seen = nil
hop(1) { $ran += 1; seen = 1 }
p [seen, $ran]
```

```
$ ruby repro.rb
[1, 1]

$ spinel repro.rb -o repro && ./repro
[nil, 1]

$ spinel --emit-rbs repro.rb && cat repro.rbs
class Object
  def walk: (Integer) -> Integer
  def hop: (Integer) -> untyped # spinel: widened to untyped (slow path)
end
```

## Inferred return type

The last expression of `walk` replaced, everything else as written. The right
column is the same file with `walk` passing on the block it was given, which is
the third variant below.

| `walk` ends with | as written | passing the block on |
|---|---|---|
| `nil` | `Integer` | `nil` |
| `7` | `Integer` | `Integer` |
| `:a` | `Integer` | `Symbol` |
| `"x"` | `Integer` | `String` |
| `[1]` | `Integer` | `Array[Integer]` |

As written, the first three build; `"x"` and `[1]` produce no binary:

```
repro.rb:6: error: incompatible pointer to integer conversion returning 'char *' from a function with result type 'sp_int' (aka 'long') [-Wint-conversion]
repro.rb:9: error: incompatible pointer to integer conversion returning 'sp_IntArray *' from a function with result type 'sp_int' (aka 'long') [-Wint-conversion]
```

Passing the block on, all five build and print what CRuby prints.

## Generated C

`spinel repro.rb -c -o repro.c`. The `// <--` markers are added; the C is
otherwise as emitted.

No capture struct is emitted. `seen` stays a plain local of `main`, and the proc
[...]

PR #4489, opened 2026-09-15
A lowered yielder answers what its block answered, whatever type that is

bartleusink:
A self-recursive method whose value comes out of a `yield` is lowered to a real `&block` (`sp_Proc`) function, and its return was forced to Integer — the **raw carrier**: the slot's bits, to be cast back to the real type at the call site. Only some consumer paths emitted that cast.

```c
puts countdown(2) { "done" }   →  const char *_ps = (const char *)(sp_countdown(...));   // cast
p    walk(2)     { |i| "v" }   →  const char *_t6 = sp_walk(2LL, _t7);                   // none
```

So by what the block answers:

| block answers | before | after |
|---|---|---|
| Integer, Float, Symbol, nil | ok | ok |
| **String, Array, Hash** | **no binary** — `sp_int` into a `const char *` slot | ok |
| **poly** (differs by path) | **no binary** — raw bits cannot carry a tag | ok |

```ruby
def walk(n)
  return yield(n) if n <= 0
  walk(n - 1) { |x| yield x }
end
p walk(2) { |i| "v#{i}" }        # CRuby: "v0"
```

```
error: incompatible integer to pointer conversion initializing 'const char *' ...
```

The word-sized answers fit the carrier, which is why the shape survived this long.

## The fix

One C function serves every call site — the same `countdown` is driven by a String block and an Integer block — so no per-site type can sit in its signature. Type it poly and have `method_call_ret` agree.

Everywhere else in the compiler a function's declared return and the call node's type are the same thing, which is exactly why every other consumer path needs no cast of its own. This shape was the exception, and the mismatch stayed invisible only while the carrier happened to fit.

## What I tried first, in case it looks tempting

**Typing the function from the block** (`yield_value_type`) is wrong, and `recursive_implicit_yield_value.rb` already covers why: it calls one `countdown` with `{ "done" }` and with `{ 42 }`. Two call sites that disagree cannot both be in one signature.

**Boxing the return alone** is worse — the call site still claimed the per-site type, so every consumer broke rather than just the pointer ones.

## Cost

An Integer-answering lowered yielder loses its unboxed return. That is what one function serving call sites that disagree costs, and the alternative is three of the seven answer types not compiling. Happy to look at specialising per call site instead if you would rather keep the unboxed path — that is a bigger change to the lowering, not to this decision.

## Verification

- Emitted C (`-S --no-line-map`, hashed) across **3646** tests: **3 differ** — exactly `recursive_implicit_yield_value`, `block_forwarded_into_recursive_yielder`, and `yield_beside_block_param_value`, the three that exercise this lowering. All three pass.
- Full suite: **3642 pass, 0 fail, 0 error**.
- New test `test/lowered_yield_block_value_types.rb` covers all seven answer types, one method driven by differently-typed blocks from different call sites, the instance-method form, and pins the "value is its own tail" case as unchanged.

## Note on a follow-up

I have a second, separate fix in hand for block *parameter* typing across call sites. It makes a block's answer poly in ordinary code, which walks straight into this defect — without this PR, that one turns `test/yield_beside_block_param_value.rb` (merged in #4488) red. Filing this first for that reason; the other follows once this lands.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed recursive methods that return a block’s value so results retain the correct type.
  - Improved handling of block-returned values including strings, arrays, hashes, numbers, symbols, booleans, and `nil`.
  - Ensured values remain correct across recursion depths and when the same method is called with differently typed blocks.
  - Preserved existing behavior for methods whose result comes from their own recursive tail.


bartleusink:
Good catch on the dead `is_lowered_yield` disjunct — fixed in 85325d02, though I took it a step further than suggested.

The reason it became unreachable is worth stating: the arm this PR adds (`is_lowered_yield && lowered_carries_block_value`) and the pre-existing one (`is_lowered_yield && !lowered_carries_block_value`) have complementary conditions and the same body, so together they cover every `is_lowered_yield` scope. Deleting the disjunct alone would have left two adjacent arms saying the same thing in opposite ways, so I folded them into one `if (c->scopes[mi].is_lowered_yield) return c->scopes[mi].ret;` and kept both rationales (#4145 and this one) in the comment above it.

Verified behaviour-neutral rather than assumed: emitted C is byte-identical on all 3647 tests, so the suite result is unchanged from the previous commit.

Issue #839, opened 2026-05-26
Array extreme negative index inserts instead of raising IndexError


neidiom:
When assigning to an array with a very negative index that exceeds the array's size, MRI raises IndexError. Spinel silently inserts the element (or in some cases extends the array).

### Ruby proof:

```ruby
a = [1, 2, 3]
a[-999] = 99
puts a.inspect
```

**Expected (MRI):** `IndexError: index -999 too small for array; minimum: -3`
**Spinel:** `[1, 2, 3]` (element not inserted but no error raised either)

**Note:** For extreme negative indices in assignment, Spinel doesn't insert AND doesn't raise — it silently ignores the operation. For less extreme negative indices within range, assignment works correctly.

matz:
Fixed in 5ada915.

Issue #769, opened 2026-05-25
Buffer underflow with very negative index


neidiom:
**File:** `lib/sp_runtime.h:977, 913`

**Ruby proof — IntArray (SEGV):**
```ruby
# bug_repros/02_extreme_negative_index_segfault.rb
# MRI: IndexError (index too small)
# Spinel: Segmentation fault
a = [1, 2, 3]
a[-999] = 42
puts a.inspect
```

**Expected:** `IndexError: index -999 too small for array; minimum: -3`
**Spinel:** Segmentation fault

**Ruby proof — FloatArray (SEGV):**
```ruby
# FloatArray — same SEGV behavior
a = [1.0, 2.0, 3.0]
a[-999] = 4.2
puts a.inspect
```

**Expected:** `IndexError`
**Spinel:** Segmentation fault

PR #4505, opened 2026-09-16
The Integer array's set inlines, with everything but the store cold

bartleusink:
`sp_IntArray_set` is written as a small fast path that defers the grow case to `sp_IntArray_set_slow`. Because `_slow` is a `static` function with exactly one call site, clang inlines it into `sp_IntArray_set` first — nearly free in its cost model, since the original can then be deleted — and the wrapper that results is too big to inline into its own callers.

So the better-factored of the two set helpers was the one left out of line. `sp_FloatArray_set` has no separate slow path, nothing inflates it, and it inlines everywhere:

```
bl _sp_IntArray_set    17 call sites      (tree-ensemble workload)
bl _sp_FloatArray_set   1 call site
bl _sp_IntArray_set    68 call sites      (optcarrot)
bl _sp_FloatArray_set   1 call site
```

`sp_IntArray_push_grow`, a few lines above, already carries `noinline` for exactly this reason.

## Why not just `noinline` on `_slow`

That does fix the inlining, but it costs ~1.6% on optcarrot (n=32 interleaved, t≈3.0): the body that then lands at all 68 sites still carries the null, frozen and `IndexError` paths, including a call to `sp_sprintf`.

Splitting instead on what the hot path actually needs — non-null, not frozen, in range — inlines a single test and store, and moves everything else into `sp_IntArray_set_cold`, which is **the previous body verbatim**, so the semantics of #769 and #839 are unchanged by construction.

The emitted optcarrot binary is 32 bytes *smaller* than before the change, and call sites drop to 0 in both programs. `sp_IntArray_get`, `sp_FloatArray_get` and the `sp_PtrArray` pair were already fully inlined, and `_slow` is the only helper of its kind in the header, so nothing else needs the same treatment.

## Measurements

`benchmark/`, best of 5 rounds × 100 runs, interleaved, output byte-identical to master and to the `.expected` files throughout:

| benchmark | master | this PR | |
|---|--:|--:|--:|
| `bm_loops_times` (yjit-bench) | 31.46 ms | 12.28 ms | **2.56x** |
| `bm_sudoku` | 5.44 ms | 3.90 ms | 1.40x |
| `bm_nqueens` (yjit-bench) | 8.33 ms | 6.11 ms | 1.36x |
| `bm_fannkuch` | 1.95 ms | 1.85 ms | 1.06x |
| `bm_stark_field` | 58.68 ms | 57.18 ms | 1.03x |
| `bm_life` | 12.45 ms | 12.19 ms | 1.02x |

The small figures include ~1–2 ms of process startup, so they understate the compute difference. `bm_loops_times` is the clearest case: its inner loop is `a[i] = a[i] + j % u` on an `Array.new(10000, 0)`, 16M times.

**optcarrot is unchanged** — 2506.6 → 2506.5 fps (n=20 interleaved), checksum 59662. It has the call sites but not in its hot loops.

On a tree-ensemble workload (64 features, predictions byte-identical): 74.99s → 65.28s, 1.15x; three smaller datasets 1.08–1.10x.

## Testing

- `make test`: 3650 pass, 0 fail, 0 error
- `make optcarrot`: OK, checksum 59662
- Every benchmark above produces output identical to master and to its `.expected` file

## Summary by CodeRabbit

* **Performance**
  * Improved the efficiency of integer array updates in common valid-use cases.
  * Preserved existing behavior for invalid indices, frozen arrays, null values, and automatic array growth.


matz:
Confirmed here with clang 18 on x86-64: bm_loops_times 30.5 -> 14.0 ms (20 runs each), with the two call sites moving from sp_IntArray_set to sp_IntArray_set_cold. gcc already inlined the old form on this box, so it reads flat there, which matches your reading of clang's cost model. Merged.

PR #4521, opened 2026-09-17
A poly temporary bound by the setter value-position arm gets the RbVal root (#4520)

### LINKED ISSUE #4520: A poly temporary bound by the setter value-position arm (98b10beb) gets a plain root; the next mark reads its tag word as a pointer
Since 98b10beb (#4516), a hand-written setter called in value position with a non-simple poly argument segfaults in the next GC mark.

**Repro** (32 lines; CRuby prints `100000`, e45df74d prints `100000`, 98b10beb and master c994fa86 die with SIGSEGV in `sp_gc_mark` — `KERN_INVALID_ADDRESS at 0x3`):

```ruby
class Box
  def initialize
    @v = nil
  end

  def v=(x)
    @v = x
  end

  def v
    @v
  end
end

def pick(i)
  i.odd? ? "s#{i}" : nil
end

b = Box.new
n = 0
200_000.times do |i|
  r = (b.v = pick(i))
  n += 1 if r
  s = "x" * 64
  n += 0 if s.empty?
end
puts n
```

**What the C shows.** The value-position arm binds the argument to a temporary that the call reads. Here the argument is a `String?`, so the temporary is an `sp_RbVal` — and it is rooted with the plain root:

```c
sp_RbVal lv___sv7 = pick(...); SP_GC_ROOT(lv___sv7);
sp_RbVal _t8 = lv___sv7;
SP_GC_ROOT_RBVAL(_t8);
```

`SP_GC_ROOT` pushes the slot's address as a `void **`; at the next mark the walker reads the RbVal's first word — the tag — as an object pointer, which is the `0x3` in the crash. The re-emitted argument one line down gets the RbVal root correctly.

**Why.** The arm chooses the root as

```c
if (needs_root(at) && !comp_ty_value_obj(c, at)) SP_GC_ROOT(...)
else if (at == TY_POLY) SP_GC_ROOT_RBVAL(...)
```

and `needs_root()` (codegen_stmt.c:10589) already answers yes for `TY_POLY`, so the second branch is unreachable for the one kind it exists for.

**Where it bit.** On the campfire conformance lane (https://roundbench.showcase.party/campfire-suite/) the run at spinel 98b10beb dropped 284 → 190/288: 11 test files abort with no output, every one at the first collection after an ActiveRecord row hydration (`sp_User_s_from_stmt` binds each nullable text column through `def bio=(v)` in value position). Bisected across 1ed29370..98b10beb with the same emitted tree; 47c17607 and e45df74d are green.

PR to follow: swap the two branches, plus the repro as `test/setter_value_poly_root.rb`.

rubys:
Fixes #4520.

98b10beb's value-position arm for a hand-written setter binds a non-simple argument to a temporary and roots it with

```c
if (needs_root(at) && !comp_ty_value_obj(c, at)) SP_GC_ROOT(lv_..);
else if (at == TY_POLY) SP_GC_ROOT_RBVAL(lv_..);
```

`needs_root()` answers yes for `TY_POLY`, so the RbVal branch was unreachable and a poly temporary (an `sp_RbVal`) got the plain root; the next mark read its tag word as an object pointer. This puts the poly test first. One conditional swapped, plus the issue's repro as `test/setter_value_poly_root.rb` (dies with SIGSEGV on master, `PASS` with the change).

Verified: `make test` 3,660 pass / 0 fail; the 11 campfire conformance files that aborted at 98b10beb link and pass again (52/52 on the five I re-ran).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed garbage collection handling for polymorphic temporary values, preventing incorrect memory interpretation during collection.
  * Improved reliability when assigning strings or `nil` through setters used in value expressions.

* **Tests**
  * Added regression coverage for repeated polymorphic setter assignments and validated the resulting output.


matz:
Merged as 2feb0937. My fault in 98b10beb: the two root tests were in the order that made the RbVal one unreachable, exactly as you read it. Thanks for the bisect and the campfire lane numbers.

PR #4556, opened 2026-09-19
A parse error names its file, line and column, and --emit-types still writes

rubys:
A program that does not parse reports Prism's messages under `Parse errors in 'app.rb':` with no position at all, and under `--emit-types` writes no JSON:

```
$ spinel bad.rb --emit-types -o bad.json
Parse errors in 'bad.rb':
  unexpected 'end'; expected an expression after the operator
spinel: parse failed for 'bad.rb'          # rc 1, no bad.json
```

Prism has the span of every diagnostic (`diag->location`, beside the `message` that was printed). Each error is now printed at it, `file:line:col: message` (the column 0-based, as Prism gives it and `--emit-types` carries it), through the multi-file map when one was built, so an error in a required file names that file at its own line:

```
Parse errors in 'main.rb':
  lib.rb:3:0: unexpected 'end'; expected an expression after the operator
```

Under `--emit-types` the JSON is written too — `types` and `codegen` empty, the errors as `"severity":"error"` diagnostics with their spans — and the exit status stays 1. A buffer is unparseable most of the time it is being typed into, and its marker belongs on the token: rubys/spinel-ide's LSP could say only "parse failed" at line 1 there. The header line stays, since `tools/rubyspec/run.sh` greps it.

- `src/spinel_parse.c`: the error loop, and a small JSON string writer (codegen's is not built yet at that point).
- `docs/emit-types.md`: the parse-error case under `diagnostics`.
- `Makefile`: `infer-test` parses a program whose required file is broken and checks the stderr line and the JSON record; both lines fail on master. `make test` 3686/0/0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Parse errors now report the originating file, line, column, and source range, including errors from required files.
  * Error messages are properly escaped in JSON diagnostics.
  * Parse failures now produce JSON diagnostics with empty type and code-generation output while preserving the expected failure status.
  * Terminal error columns are 1-based, while JSON diagnostic columns remain 0-based.
  * Clear errors are reported when emitted type information cannot be written.

* **Documentation**
  * Updated `--emit-types` documentation to describe parse-failure behavior and diagnostic output.


rubys:
Pushed one follow-up: the stderr column is 1-based (`lib.rb:3:1: …`), as an editor reads a compiler's message and as #4561's `--warn-widen` lines are; the JSON's stays 0-based, as Prism gives it and `--emit-types` documents. The gate line and docs/emit-types.md say so.

rubys:
Taken from the review: when the JSON path cannot be opened on the parse-failure path, stderr now says `spinel: cannot write '<path>'`, the line codegen's `emit_write_file` prints in the same case. Write and close failures are left as codegen leaves them.

matz:
Merged as a4c5bbc5. The branch also carried six build products (dp, env, spinel_analyze, spinel_codegen, spinel_parse, spinel_rbs_extract, about 6 MB); the merge leaves them out, the source changes are in as they were.

Issue #4522, opened 2026-09-17
`--emit-types`: carry what the parser and codegen already know — a span, the node's kind and name, the widened slot, the resolved callee


rubys:
## What exists now, and consumes it

Three out-of-tree consumers of `--emit-types` are running against master
daily, as #4509 suggested they should be:

- **the browser IDE**, https://rubys.github.io/spinel-ide/ — `spinel.wasm`
  in a worker; hover types, refusal and widening markers, the RBS, the C,
  Run (the program built for wasm32-wasi in the tab);
- **spinel-mcp** and **spinel-lsp**, in the same repository under `tools/`
  — subset Ruby over one query core, run by CRuby or compiled by spinel;
  the tracker compiles both with each day's master and requires the two
  forms to answer identically. The LSP publishes diagnostics and answers
  hover, inlay hints (the inferred signature after each `def`) and a
  fast/slow code lens per `def`, verified from VS Code.

All three answer from the JSON `--emit-types` writes, and all three run
into the same four limits of that JSON, listed as the consumers see them
in the repository's README under
[What the compiler doesn't say yet](https://github.com/rubys/spinel-ide#what-the-compiler-doesnt-say-yet).
This is one request to extend the dump with information the compiler has
in hand when it writes the file. No new flag, no API, no
consumer-specific surface; the additions are fields on the records the
dump already emits.

## The four limits, with what a user sees

1. **No end position.** Line 11 of the `point` sample is
   `puts pts.map { |p| p.dist2(pts[0]) }.inspect`. The dump has three
   entries at column 5 — `Array[untyped]`, `Array[Integer]`, `String` —
   for `pts`, `pts.map{…}` and `.inspect`, which all *start* there. A
   hover on `pts` cannot tell which is which, so it lists all three;
   `type_at` in the MCP answers the same. With `end_line`/`end_col` the
   consumer picks the tightest span containing the cursor — the type of
   exactly the expression under it. Prism has the location; the parser
   stamps `node_line`/`node_col` from it already.

2. **No node kind or name.** An entry is a position and a type. Without
   knowing that the node at 11:5 is a `LocalVariableReadNode` named
   `pts`, or that 8:17 is a `CallNode` named `x`, a consumer cannot find
   the other reads of `pts`, cannot tell an identifier from a literal,
   and cannot offer go-to-definition or references — even the purely
   syntactic kind. A `kind` (the Prism node type the AST already carries)
   and, for reads/calls/defs, a `name` would let an editor do the
[...]

Issue #4509, opened 2026-09-16
RFC: the boundary at a cursor — `spinel-mcp`, `spinel-lsp`, a browser IDE, and the playground it leads to


rubys:
## RFC — a question, not a PR yet

2026.09.12 stated the boundary and held it: `docs/limitations.md` says
*what* spinel does not do, and the tests pin *how* it behaves. What a
program's author still cannot see is *where* their file crosses that
boundary — which parameter widened to `untyped` and took the boxed slow
path, which construct is refused and by which row of the limitations
document, what signature the fixpoint settled on. `--emit-rbs` (#1276) and
`--emit-types` (#1298) answer that per program, and #1298 was pitched as
the substrate for an editor integration that never arrived. This RFC is
that consumer side, in the mould #1381 settled on — companion executables
in the spinel subset, compiled by spinel — plus one C-side piece the
executables and a browser build both need. Four separable parts plus the
playground they lead to; the ask at the end is which of them you want and
at what depth.

## Why now, and why from here

The question these tools answer is the one the roundhouse↔spinel loop asks
every day. Twenty-eight test programs in `test/` cite a roundhouse shape as
their trigger, and 281 mention widening; the emitter side of that loop
decides how to write Ruby that spinel types well, and today learns it by
reading `--emit-rbs` output and stderr after a whole-app compile. Roundhouse
itself went through this exercise for its own analyzer — one
position-addressable query layer, then an LSP, an MCP server and a browser
IDE as thin skins over it (rubys/roundhouse#57) — and the skins turned out
to be the cheap part; the substrate is the work. I would like the same
shape here, with spinel-specific answers: *why* did this widen, and *what
won't compile*.

## What exists (probed against f04b62ca)

- `--emit-types` gives per-node `{file, line, col, type, rbs}` and the
  widening warnings. Positions are start-only — nodes carry
  `node_line`/`node_col` and no end position.
- Refusals (`unsupported_feature` / `unsupported`) go to stderr and
  `exit(1)`; they are not in the `--emit-types` JSON, and that path exits 0
  for a program the real compile refuses. `SP_COLLECT_ERRORS` already makes
  them recoverable per method — the foothold for a diagnostics sink.
- **The compiler builds as `wasm32-wasi` with no source changes.**
  `src/*.c` + the vendored prism + `lib/regexp` under the wasi-sdk link to
  an 8.3 MB `spinel.wasm` that produces byte-identical C to the native
[...]

PR #4561, opened 2026-09-19
--warn-widen: a warning at each parameter or return that widened to untyped

rubys:
A plain compile says nothing about a signature that degraded to untyped. The fact lives in `--emit-types` (`"severity":"warning"`, placed on the slot since #4522) and as a `# spinel: widened` comment in `--emit-rbs`, which is what spinel-doctor's inference leg greps — so from the shell, the one line of a method that took the boxed slow path is invisible unless you run an analysis mode and read a file. Refusals go to stderr *and* the JSON (#4509); widenings were the asymmetry.

```
$ spinel point.rb --warn-widen -o point
point.rb:8:13: warning: parameter `o` of `dist2` widened to untyped (boxed poly slow path)
point.rb -> point
```

One line per widened slot, at the slot (the parameter's own node, or the def for a return), through the require map so a widening in a required file names that file. Off by default, since most real programs widen somewhere. The column is 1-based on stderr, as an editor reads a compiler's warning; the JSON's stays 0-based. Positions are forced past `--no-line-map`, as `--emit-types` forces them.

- `src/codegen.c`: the walk over the widened slots is one function (`each_widened_slot`) with two readers, the JSON record and the stderr line, so the two agree by construction. The `--emit-types` JSON is byte-identical before and after.
- `src/main.c`: the flag, and the usage line.
- `docs/emit-types.md`, `README.md`: a paragraph each.
- `Makefile`: `infer-test` checks the three warnings on `emit_types_fields.rb` at their positions, that a plain compile stays quiet, and that `--no-line-map` still places. `make test` 3686/0/0.

This is the surface a `why` would attach to — a `note:` line under each warning naming the argument or write that carried the poly, and the expression it was born at — which #4509 said wants its own node-origin design. That is being scoped separately; nothing of it is here.

## Summary by CodeRabbit

- **New Features**
  - Added the `--warn-widen` compiler option to report when parameter or return type inference widens to `untyped`.
  - Warnings identify each affected parameter or return slot and are emitted during compilation, including with `--emit-rbs`.
  - Warning locations use 1-based columns and remain accurate even when line maps are disabled.

- **Documentation**
  - Added guidance for `--warn-widen` to the README and type-emission documentation.
  - Clarified how widening warnings appear alongside type-emission diagnostics and JSON output.


rubys:
Two of the review's findings taken, one pushed commit: the warnings print once, right after analysis and before the analysis-only modes return, so `--emit-rbs --warn-widen` warns too (checked at both points over the 3,639 programs of the corpus: emission changes no widened slot); the doc's example is a code block. While there, the line takes the form the compiler's other warnings have — `spinel: file:line:col: warning: …` — which the first cut had left off. `make test` 3686/0/0.

Issue #4426, opened 2026-09-11
A call through a `class << self; attr_reader` class-object slot is devirtualized by codegen but typed poly by inference


rubys:
## Summary

A singleton method called through a class object held in a `class << self; attr_reader` slot is devirtualized by codegen to the implementor's direct function, but the call site is typed poly by inference — so the concrete C return (`sp_int`, `const char *`, `sp_bool`, whichever the method has) meets an `sp_RbVal` with nothing boxing the seam, and clang refuses. Spelling the same reader by hand (`def self.adapter; @adapter; end`) compiles and runs.

## Reproduction

```ruby
# m.rb
module Sqlite
  def self.insert(table, attrs)
    attrs.size + 41
  end
end
module AR
  class << self
    attr_accessor :adapter
  end
end
AR.adapter = Sqlite
puts AR.adapter.insert("blobs", { "name" => "x" }) + 1
```

```
$ spinel --version
spinel 97ffe6e8a91c (unreleased) [Apple clang version 21.0.0 (clang-2100.1.1.101)]
$ spinel m.rb
m.rb:18: error: passing 'sp_int' (aka 'long') to parameter of incompatible type 'sp_RbVal'
spinel: C compilation failed
```

## Expected

```
$ ruby m.rb
43
```

## The A/B

Same program, only the reader changes:

| `AR.adapter` is | result |
|---|---|
| `class << self; attr_accessor :adapter; end` | refused |
| `class << self; attr_reader :adapter; end` + explicit `def self.adapter=(a); @adapter = a; end` | refused |
| `def self.adapter; @adapter; end` + `def self.adapter=(a); @adapter = a; end` | **43** |
| `def self.adapter; Sqlite; end` | 43 |
| a local, or an instance's `@adapter`, holding `Sqlite` | 43 |

So it is the synthesized singleton `attr_reader` specifically. Not the return type: `"r" + table` fails the same way with `const char *`, `attrs.size > 1` with `sp_bool`; assigning the call to a local or adding `.to_i` moves the error, it does not clear it. `class Sqlite` vs `module Sqlite` makes no difference.

## Emitted C

```c
sp_AR_s_adapter_set(((sp_Class){0}));
...
sp_poly_puts(sp_poly_add(sp_Sqlite_s_insert("blobs", _t4), sp_box_int(1LL)));
```

[...]

PR #4562, opened 2026-09-19
Why a slot widened: the chain from the slot to the expression the untyped was born at

rubys:
#4509 said the reason a slot widened is the most useful single answer the compiler could give, that a slot-based attempt reached the root cause for 3% of slots and was reverted — because a poly is born at an expression and slots only inherit it — and that it wants its own design, node-origin. This is that design, as a prototype on the surface #4561 added: under each `--warn-widen` warning, the *why*, one `note:` per hop from the value that widened the slot to the expression the untyped was born at, and the same chain as `why` on the `--emit-types` record.

```
$ spinel point.rb --warn-widen -c -o p.c
spinel: point.rb:8:13: warning: parameter `o` of `dist2` widened to untyped (boxed poly slow path)
spinel: point.rb:12:28: note: passed `pts[0]` is untyped
spinel: point.rb:11:7: note: from `(1..5).map { |i| Point.new(i, i * 2) }` is Array[untyped] — born here: no untyped input

spinel: widening.rb:9:18: warning: parameter `p` of `initialize` widened to untyped (boxed poly slow path)
spinel: widening.rb:16:35: note: passed `2.5` is Float, where the slot was Integer (two kinds meet: untyped)
spinel: widening.rb:16:22: note: and `1` is Integer
```

## The design

**Node-origin, at the one funnel.** Every node's type lands at one line, `c->ntype[id] = t` in `infer_type`; there `norigin[id]` is set by one rule: the slot's recorded source for a read of a local or parameter, the callee's return for a call bound to a user def whose return degraded, the tail statement for a body, else the first child carrying a degraded type, else the node itself — the untyped was born here. Nothing at the rules that answer poly; the origin is derived from where the types already are.

**Slots record the flip.** `slot_take` / `slot_set` (compiler.c) replace the idiom `m = ty_unify(slot, t); if (m != slot) { slot = m; changed = 1; }` at the sites that widen a parameter from its call sites (`bind_call_params`, `infer_param_types`), a parameter written in its own body, a local's writes (`infer_write_types`), and the return (`infer_return_types`). A `SlotWhy` holds the node whose value degraded the slot, the node that had given it its concrete type before, both kinds, and the round. The helper *is* the idiom, so inference is unchanged: `make test` 3688/0/0, and the `--emit-types` JSON differs only by the added `why`.

**Six endings**, each named in the last hop: *born here: no untyped input* (a literal array of objects, `map` over them, `&.` — the rule is the compiler's and this is where to look) · *two kinds meet* (with an `and` hop at the other kind's site) · *a transient* (the value is concrete in the end but was untyped on the round the slot took it, and the fixpoint kept the slot there — an imprecision, said as one, with the round) · *never bound* · *by construction* (a `*rest`) · *untraced*.

## Measured

Over the corpus (`test/*.rb`, `test/infer/*.rb`: 6,596 widened slots, no crashes):

| ending | slots | |
|---|---:|---:|
| born here | 2,794 | 42% |
| two kinds meet | 548 | 8% |
| a transient | 615 | 9% |
| never bound | 454 | 7% |
| by construction | 338 | 5% |
| **definitive** | **4,749** | **72%** |
| untraced: a parameter widened at a rule site not yet through `slot_set` | 989 | 15% |
| untraced: a chain reaching a local typed by a pass not yet recorded (block parameters, mostly) | 857 | 13% |

The two untraced rows are lists, not walls: each is one more site routed through the helper, measured the same way. A structural fallback already follows a block parameter to its block's receiver at render time, since the block-parameter inference has ~80 write sites.

- `src/compiler.h`, `src/compiler.c`: `SlotWhy`, `Compiler.norigin`, `slot_take`/`slot_set`, `ty_degraded`, `g_infer_round`.
- `src/analyze_infer.c`: `why_node_origin` at the funnel. `src/analyze.c`: the round counter.
- `src/analyze_pass.c`: the sites, and the return's why in `infer_return_types`.
- `src/codegen.c`: the chain walk and its two renderers; `src/spinel_parse.c`: ends stamped under `--warn-widen` too (the notes quote the expression).
- `docs/emit-types.md`: `why` and the six endings. `Makefile`: `infer-test` pins the chains on `emit_types_fields.rb`, stderr and JSON.

Nothing here is asked of anyone; the sites not yet recorded are the follow-on, and the 3%-vs-72% is the reason to think the node-origin shape is the right one.

## Summary by CodeRabbit

- **New Features**
  - Widening warnings now include clearer note chains explaining how values became untyped.
  - Diagnostics identify degraded method returns, representation-based widening, and conflicting return types.
  - JSON type output now includes structured `why` traces with roles, inferred types, locations, and explanations.

- **Documentation**
  - Expanded guidance for widening warnings, JSON trace details, trace endings, and when tracing is generated.

- **Tests**
  - Added coverage for diagnostic note chains, JSON trace details, and ensuring plain compilation remains free of widening notes.


rubys:
Tried on a real program, and pushed one commit from what it showed: the roundhouse-emitted Rails-shape tree (`rubys.github.io/roundhouse/browse/spinel.tgz`: 166 files, `spinel main.rb --warn-widen -c` in 9.6 s, 1,347 widened slots).

The corpus could not show two things the tree did. A read of an instance variable ended a chain as "born here", which misattributed the Rails-shape case: an ivar is typed from its writes, and the root is what the adapter writes into the row object. And a call on a module constant did not find its singleton def's return. Both follow through now (an ivar read's origin is the first write of the name in its class whose value degraded), a callee's return that met two kinds is explained when the chain reaches it mid-way, the hop cap allows the accessor chains such a tree has, wrappers are not hops of their own, and a value with no type of its own is named as one rather than as a transient. `make test` 3688/0/0; the JSON still differs only by `why`.

A model accessor's chain then runs eleven hops across five files to the expression they all share:

```
spinel: app/models/article.rb:14:3: warning: the return of `body` widened to untyped (boxed poly slow path)
spinel: app/models/article.rb:15:5: note: returned `@body` is untyped
spinel: app/models/article.rb:19:13: note: from `value` is untyped
spinel: app/models/article.rb:85:21: note: from `(row.body).nil? ? nil : (row.body).to_s` is untyped
spinel: app/models/article.rb:85:46: note: from `row.body` is untyped
spinel: app/models/article_row.rb:11:5: note: from `@body` is untyped
spinel: app/models/article_row.rb:15:13: note: from `value` is untyped
spinel: app/models/article_row.rb:56:35: note: from `row["body"]` is untyped
spinel: app/models/article.rb:77:53: note: from `row` is untyped
spinel: runtime/active_record/base.rb:167:68: note: from `row` is untyped (a parameter of the block on:)
spinel: runtime/active_record/base.rb:167:7: note: from `ActiveRecord.adapter.all(table_name)` is Array[untyped]
spinel: runtime/active_record/base.rb:167:7: note: from `ActiveRecord.adapter` is untyped — born here: no untyped input
```

`ActiveRecord.adapter` is a `class << self; attr_accessor :adapter` on a module — the shape #4426 records as typed poly by inference. Grouping every chain by its root ranks the expressions through which untyped enters the program:

| root | slots |
|---|---:|
| `runtime/active_record/base.rb:167` `ActiveRecord.adapter` | 47 |
| `runtime/active_record/relation.rb:508` `@model._hydrate_all(…)` | 31 |
| `runtime/active_storage.rb:539` `ActiveRecord.adapter` | 21 |
| `runtime/action_cable.rb:462` `JSON.parse(text)` | 14 |
| `runtime/rails.rb:393` `Rails.secret_key_base` | 11 |
| `runtime/broadcasts.rb:46` `LOG` | 9 |
| `runtime/action_dispatch/session.rb:14` `{}` | 8 |
| `runtime/action_controller/cookies.rb:20` `{}` | 8 |

Five expressions, ~130 of the warnings; the emitter's author now knows which lines to look at. Endings on this tree: born 26%, two kinds meet 6%, transient 3%, pessimistic (an empty literal the pessimistic round widened) 9%, never bound 21%, by construction 1% — 66% definitive; untraced 34%, of which 239 are parameters widened by a rule not yet through `slot_set` (a class method reached through a boxed send, mostly) and 215 chains end at a local or ivar typed by a pass not yet recorded (`local` 158, `@ivar` 50).

rubys:
Correction to the table above: the roundhouse tree builds with `--rbs .` (its Makefile's `RBS_FLAG`), and the run above did not pass it. The same tree, same binary, with the flag it ships under:

| | without `--rbs` | with `--rbs .` |
|---|---:|---:|
| widened slots | 1,347 | 787 |
| `app/models/article.rb` | 151 | 27 |
| `Article#body` | `sp_RbVal` | `const char *` |
| compile | 9.6 s | 7.2 s |

Endings with `--rbs .`:

| ending | slots | |
|---|---:|---:|
| born here | 143 | 18% |
| never bound | 95 | 12% |
| pessimistic | 31 | 4% |
| by construction | 13 | 2% |
| two kinds meet | 24 | 3% |
| a transient | 4 | 1% |
| **definitive** | **310** | **39%** |
| untraced: a parameter widened at a rule site not yet through `slot_set` | 311 | 40% |
| untraced: a return | 103 | 13% |
| untraced: a local or ivar typed by a pass not yet recorded | 63 | 8% |

Two things change. The funnels are gone: the seeds pin the model accessors and the adapter slot, so the 47 chains that ran to `ActiveRecord.adapter` and the 31 to `_hydrate_all` are not in the table at all — the largest root is now 7 (`runtime/db.rb:394` `SQL.read_ptr(SQL.stmt_out)`), and the top ten born-here roots together account for 41 of 143. What widens under the real build is diffuse, not five expressions. And untraced rises from 34% to 61%, almost all of it (311) parameters with no record — the seed-apply path and the rule sites it feeds are not through `slot_set`, so on a seeded tree the chain mostly cannot start. That is the follow-on the description names, and the number to measure it against is 61%, not 34%.

Two rendering notes from this run, for the next commit: `main.rb:0:1: note: written `` is untyped` — 7 chains (the `__enum_to_a` returns) end at a synthesized node with no source span; and a chain through a poly send follows the receiver rather than the candidate whose return degraded, which is what made `ActiveRecord.adapter` the root of the model chains in the run above (the value that actually degraded them was a user `to_s` returning untyped, which the seeds happen to cover). The origin rule gains that case next.

matz:
Measured on this machine (master 23802372 vs the branch, same base). The generated C is byte-identical over the whole corpus (3630 test programs) and over optcarrot, so the inference is unchanged as claimed.

The cost is not: the origin derivation runs at the type funnel on every plain compile, not only under --warn-widen or --emit-types.

- optcarrot (4309 lines): analysis 0.31 s -> 0.34 s (+10%), peak RSS 15.7 MB -> 16.1 MB.
- a synthetic class with 400 ivars, a reader and a writer each: 0.05 s -> 0.06 s; with 2000: 0.74 s -> 0.91 s (+23%).

The ivar arm of why_node_origin walks every InstanceVariableWriteNode of the program (NT_FOREACH_KIND) once per InstanceVariableReadNode per fixpoint round, and the call arm scans every scope by name, so the added work grows with reads x writes rather than with the program. Two ways to take that off the plain build: derive the origin only when a consumer asked (the SPINEL_WARN_WIDEN / SPINEL_EMIT_TYPES gate nd_stamp already uses), or index the writes by name once per round. Either keeps the numbers above for the analysis modes, where they are fine.

Whether the design goes in tree is matz's call (#4509 named it as wanting its own design); this note is only the measurement.

rubys:
Thank you for the measurement. a43c482a takes it off the plain build the first way: the origin is derived only under `--warn-widen` / `--emit-types` (the same gate `nd_stamp` reads). On this machine optcarrot's plain compile is master's again — 0.23 s user for both, C byte-identical — and the modes keep your numbers. Indexing the writes by name would make the modes cheaper too; I left that for when the seed-apply sites are routed through `slot_set`, since that pass is where the untraced 61% on a seeded tree lives and it will change what the ivar arm has to walk.

The same commit corrects what the roundhouse tree showed once the receiver hop was looked at: a chain through `(row.body).to_s` followed the receiver to `ActiveRecord.adapter`, but `to_s` is a String on anything — what made the send untyped was a user `to_s` whose return is (`Filename#to_s`), and one such makes every `.to_s` on a poly receiver untyped program-wide. The chain now asks for the builtin answer alone (`an_builtin_answer`, the `an_builtin_only` re-derivation exported, at render time only) and, when that is concrete, names the candidate and continues from its return. Two endings said the wrong thing and are fixed: an empty literal filled with objects is "by representation" now, not "pessimistic", and a return that met two kinds names the `return` of the *other* kind. `why_widen_chain.rb` pins the three; 3688/0/0.

matz:
Merged: 24ddc2c0 took the head as first fetched, and 55b251ba brings in the two commits pushed after it (the origin derived only when asked for, the ivar writes through the shared index, a predicate skipped). Between the two, 6a3c60f0 had added the same gate as a why_wanted() helper; the merge keeps the branch's form and drops the helper. A plain optcarrot compile is back at its old analysis time and its C is unchanged; the chains and the JSON are as the PR wrote them.

PR #4573, opened 2026-09-19
A rule that widened a slot says so in its own words

rubys:
The follow-on #4562 named. Measured first: on the roundhouse tree, the untraced remainder was **not** block parameters (2 of 1,347 — the render-time fallback already follows one to its block's receiver) but parameters a rule sets to untyped directly, outside the widening idiom. A watch after each pass of the fixpoint named them: the pass that makes a never-bound parameter poly before codegen (112 on the tree), the `= nil` default backstop (48), `infer_default_param_types` (20), the empty-literal argument rules (5).

A `SlotWhy` now carries a `reason`, the rule's words, set by `slot_rule` at those sites with the rule's subject when it has one, and the chain ends there:

```
spinel: app.rb:5:12: warning: parameter `b` of `opt` widened to untyped (boxed poly slow path)
spinel: app.rb:5:16: note: by `nil` is nil — a `= nil` default and no call site typing it: the parameter holds nil or a value, untyped

spinel: app.rb:6:9: warning: parameter `h` of `bag` widened to untyped (boxed poly slow path)
spinel: app.rb:6:13: note: by `{}` is untyped (no type of its own) — an empty literal default, which has no element type: the parameter is the untyped container

spinel: app.rb:8:12: warning: parameter `q` of `lonely` widened to untyped (boxed poly slow path)
spinel: note: never bound: no call site gives it a type
```

A rule with no subject is a JSON hop of role `rule` with only its note; a rule with one is a hop of role `by` at the subject. A chain that reaches a *read* of such a parameter ends the same way (`from `b` is untyped` → the `= nil` line). The two endings `--warn-widen` had said on stderr alone — a splat, never bound — are recorded the same way now, so they reach the JSON too.

| | #4562 | now |
|---|---:|---:|
| corpus (6,604 widened slots), definitive | 72% | **91%** |
| roundhouse tree (1,347), definitive | 66% | **87%** |

Left, by the same measurement: a few more rules touching parameters (`infer_block_params`' method parameters, `bind_coerce_operator_params`), reads of locals typed by passes not yet recorded (multiple and operator assignment), and reads of instance variables no write of which is untyped.

- `src/compiler.h`, `src/compiler.c`: `SlotWhy.reason`, `slot_rule`.
- `src/analyze.c`, `src/analyze_pass.c`, `src/analyze_scope.c`: the sites.
- `src/codegen.c`: a reason ends the chain, on stderr and in the JSON; the stderr-only special cases fold into it.
- `docs/emit-types.md`: the `by` and `rule` roles and the endings. `test/infer/why_widen_rules.rb` + `infer-test` pin them.

Inference is unchanged; `make test` 3689/0/0.

## Summary by CodeRabbit

- **New Features**
  - Widening warnings now explain the specific rule responsible, including nil defaults, empty literals, splat parameters, and parameters without typed call sites.
  - Widening explanations now trace through related values and parameter reads for clearer diagnostics.
  - Type-diagnostic JSON now includes rule-based widening explanations and their relationships.

- **Documentation**
  - Updated `emit-types` documentation with new widening explanation roles and scenarios.

- **Tests**
  - Added coverage for rule-based widening warnings and emitted type diagnostics.


rubys:
Cost on a plain compile, measured as #4562's was (same base, optcarrot's 4,309 lines through `-c --no-line-map`, 30 interleaved runs each, this machine): master min 234.4 ms / median 238.1 ms, this branch min 234.0 ms / median 238.7 ms — +0.3% on the median, inside the spread; peak RSS 18.6–19.6 MB either way; the C byte-identical. Nothing here runs per node or per round: six `slot_rule` calls that each write a few fields once, and the rendering only under `--warn-widen` / `--emit-types`.

rubys:
Taken from the review, one commit: the optimistic re-narrow reset a slot's type to UNKNOWN and left its why, so a rule's words could stand under a subject a later value recorded (and a stale ordinary why could keep a rule from recording). The reset clears the why with the type, and `slot_set` clears the reason when a value degrades the slot. The chains on the corpus and the roundhouse tree are unchanged; `make test` 3689/0/0.

PR #4595, opened 2026-09-20
A bigint result that fits the inline Integer is one (#4594)

### LINKED ISSUE #4594: A bigint operation whose result is small still answers a Bignum, and tag-dispatching consumers refuse it (`set_value(:u32, big & 0xffff_ffff)`)
## Symptom

A Bignum operation whose result is small still answers a Bignum, and every consumer that decides on the representation rather than the value refuses it:

```ruby
buf = IO::Buffer.new(64)
v = 18446744073709551615 & 0xffff_ffff
p v                        # 4294967295 — the value is right
buf.set_value(:u32, 0, v)  # CRuby stores it
                           # Spinel: RangeError: bignum too big to convert into 'unsigned int'
```

Passing the literal `4294967295` works. It is not the mask: every bigint operation does it, including ones whose result is tiny.

```ruby
B = 18446744073709551615
B & 0xffff_ffff   # 4294967295
B % 1000          # 615
B - (B - 7)       # 7
B * 0             # 0
B ** 0            # 1
```

Each of those prints correctly and compares correctly, and each is refused by `set_value(:u32, ...)` the same way. Both overflow modes.

## Cause

Ruby has one Integer type; CRuby normalizes an operation's result to a Fixnum whenever it fits, so `(2**64-1) & 0xffff_ffff` *is* a Fixnum. Spinel's `sp_box_bigint` boxes whatever pointer it is given with `SP_TAG_BIGINT`, so the tag records how the value was computed rather than what it is. Consumers that dispatch on the value (`p`, `==`, arithmetic) are unaffected — which is why the value looks right — and consumers that dispatch on the tag are wrong. IO::Buffer's u32 lane is one: it sees `SP_TAG_BIGINT` and reports "too big" without looking at the magnitude.

## Fix

Normalize at the boxing boundary: `sp_box_bigint` answers an inline Integer when the magnitude fits (`|v| <= INTPTR_MAX`, the same range test `sp_box_i64` already makes) and a Bignum otherwise. `INTPTR_MIN` deliberately stays a Bignum — the inline representation spells that value nil.

One consumer assumed the old invariant and is fixed with it: the Marshal ivar reload for a `TY_BIGINT` slot answered `NULL` for anything not `SP_TAG_BIGINT`, so a normalized small value would come back as no value at all; it converts instead.

Note this depends on #4590 — with normalization an int-tagged poly reaches a Bignum-typed slot far more often, and the bare pointer cast that #4590 replaces would segfault on it. PR follows, stacked on #4591.

makenowjust:
Fixes #4594.

**Stacked on #4591** — its first commit is that PR's. With normalization an int-tagged poly reaches a Bignum-typed slot far more often, and the bare pointer cast #4591 replaces would segfault on it.

Ruby has one Integer type, and CRuby normalizes an operation's result to a Fixnum whenever it fits: `(2**64-1) & 0xffff_ffff` *is* a Fixnum there. `sp_box_bigint` boxed whatever pointer it was handed with `SP_TAG_BIGINT`, so the tag recorded how a value was computed rather than what it is. Consumers that dispatch on the **value** were fine — which is why such a number printed and compared correctly — and consumers that dispatch on the **tag** were not: IO::Buffer's u32 lane called `big & 0xffff_ffff` *"bignum too big to convert into 'unsigned int'"* without looking at the magnitude. Not just the mask — `big % 1000`, `big - (big - 7)`, `big * 0` and `big ** 0` all answered a Bignum holding 615, 7, 0 and 1.

`sp_box_bigint` now answers an inline Integer when the magnitude fits, the same range test `sp_box_i64` already makes (`sp_bigint_mag_u64` + `sp_bigint_sign`, both already in the runtime). `INTPTR_MIN` deliberately stays a Bignum: the inline representation spells that value nil.

One consumer assumed the old invariant and moves with it — the Marshal ivar reload for a `TY_BIGINT` slot answered `NULL` for anything not `SP_TAG_BIGINT`, so a normalized small value came back as no value at all; it converts now, the same way #4591's unbox does.

The new test pins ten operations through the u32 lane that refused them, the results that genuinely need a Bignum staying Bignums, and both ends of the inline range (`2**63-1` inline, `-(2**63)` still a Bignum).

`make test` on this branch: raise 3703 pass / 0 fail / 0 error; promote 80 ERR + 25 FAIL — the *identical* failure set to the same-machine baseline, nothing added or masked. (Two entries in that baseline differ from an older measurement and neither is this change's: `nullable_num_compare` fails the same way on clean master — verified on 5680dddd — and `int_arg_into_bigint_param` is flaky in a full parallel run, passing 3/3 when run alone.)

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed handling of integer results that fit within the inline integer range, ensuring they behave as regular `Integer` values.
  - Corrected conversion of inline integers and large integers in buffer and variable operations.
  - Preserved Bignum behavior for values exceeding the inline integer range.

- **Tests**
  - Added coverage for arithmetic, bitwise operations, boundary values, signed wrapping, and type conversions involving large integers.


matz:
Merged as b6fca4a7 with one line added in the merge: the new test carries `# spinel: int64`, since its literals (2**63-1, a 2**32 divisor) overflow the 32-bit sp_int under raise mode, which is what the gcc -m32 lane reported (bigint_result_normalizes: RangeError in **). The marked tests are left out on that target, as the 32 others are.

Issue #4528, opened 2026-09-18
A green thread parked on a socket read stalls I/O in every other thread


ryudoawaru:
One thread blocked on a socket that never receives data is enough to stop socket I/O
everywhere else in the process. The thread does nothing but sit in `readpartial`.

Without it, Spinel's wakeup latency is a little better than CRuby's. With it, the
program never makes progress again. This reproduces on macOS/arm64, Linux/aarch64 and
Linux/x86_64, first on `6a19430b` and still on `d10ef3dd` (87 commits later).

A thread that only `sleep`s costs nothing, so the timer path looks fine and the fd
readiness path does not.

## Reproduction

```ruby
require 'socket'
$stdout.sync = true

ROUNDS = (ENV['ROUNDS'] || 200).to_i
IDLE = ENV.fetch('IDLE', '1') == '1'

# A thread that blocks on a socket which never receives data.
def idle_blocked_reader(sock)
  Thread.new do
    buf = +''
    loop { sock.readpartial(8192, buf) }
  rescue StandardError
    nil
  end
end

server = TCPServer.new('127.0.0.1', 0)
idle_server = TCPServer.new('127.0.0.1', 0)

idle_client = TCPSocket.new('127.0.0.1', idle_server.addr[1])
idle_side = idle_server.accept
writer = TCPSocket.new('127.0.0.1', server.addr[1])
reader = server.accept

idle_blocked_reader(idle_side) if IDLE
sleep 0.2

# Round trip: write one small message, read it back, measure the wall clock.
# The reader re-parks between rounds, so each round measures one I/O wakeup.
latencies = []
buf = +''
round = 0
while round < ROUNDS
  started = Time.now
  writer.write('x' * 64)
  reader.readpartial(8192, buf)
  latencies << (Time.now - started)
  round += 1
  sleep 0.005
end

sorted = latencies.sort
at = ->(p) { sorted[[(sorted.length * p).to_i, sorted.length - 1].min] * 1e6 }
puts format('wakeup latency: median=%.0fus p90=%.0fus p99=%.0fus max=%.0fus (idle blocked reader: %s)',
            at.call(0.5), at.call(0.9), at.call(0.99), sorted.last * 1e6, IDLE)
```

```sh
ruby repro.rb                  # CRuby, with the idle reader
spinel repro.rb -o repro
IDLE=0 ./repro                 # Spinel, no idle reader
./repro                        # Spinel, with the idle reader
```

## Measurements

[...]

PR #4612, opened 2026-09-20
A socket write no longer sleeps in the kernel with the collector waiting on it

ryudoawaru:
Under sustained socket traffic the process freezes: **0% CPU, no further output**, every
green thread stopped, only `kill -9` gets out.

`sample` while frozen:

```
one thread   sp_sock_write → sp_File_write_len → write (in libsystem_kernel.dylib)
one thread   sp_str_alloc  → sp_stw_collect_impl → _pthread_cond_wait
the rest     sp_safepoint  → sp_stw_park_locked  → _pthread_cond_wait
main         sp_sched_sleep → sp_sched_pump      → _pthread_cond_wait
```

This is #4528 on the write side — the read side was fixed in 1e0dba06.

## Why the existing park was not enough

`sp_sock_write` does park on `poll(POLLOUT)` first, but *writable* means there is room,
not room for all of `n`. A 64 KB write into a nearly full send buffer still slept in the
kernel, and a worker in a syscall never reaches a safepoint — so the first collection
stopped the world and waited for it forever. The peer that would have drained the buffer
was itself parked at a safepoint. Deadlock.

## The fix

The send is capped at the free space the kernel reports — `SO_SNDBUF − SO_NWRITE` on
Darwin, `SO_SNDBUF − SIOCOUTQ` on Linux — so the bytes handed to the kernel always fit
and the call returns. The descriptor's flags are untouched, so the stdio read side that
shares it (where `fgets` treats `EAGAIN` as EOF, per `sp_io_fdopen_sock`) is unaffected.
Where neither query is available the behaviour is unchanged.

`MSG_DONTWAIT` rides along for the platforms that honour it, but **it is not sufficient
on its own**: Darwin and the BSDs accept the flag on `send(2)` and sleep regardless.
Verified in plain C — a 64 KB `send(…, MSG_DONTWAIT)` on a blocking socket whose peer
never reads blocks indefinitely on macOS 26.6.2.

## Reproduction

Freezes before finishing a single round on a9089ec; completes 30/30 rounds with this
patch, on three consecutive runs. CRuby passes it either way.

```ruby
# Spinel：高吞吐 socket I/O 之下，整個 process 會凍住（0% CPU）
#
#   ruby repro.rb                        → SURVIVED 30 rounds
#   spinel repro.rb -o repro && ./repro  → 某一輪之後再也沒有輸出
#
# 形狀取自 Lens：數條 thread 在 socket 上雙向搬大量資料（等同 guacd 轉發），
# 同時另外數條反覆做 Net::HTTP 請求（等同回報狀態）。全部都在本 process 內，
# 不需要外部服務。
#
# 凍住時 `sample <pid>` 看到的現場：
#   一條  sp_sock_write → write(2)                     （阻塞在 syscall 裡）
#   一條  sp_str_alloc  → sp_stw_collect_impl           （等世界停下來收）
#   其餘  sp_safepoint  → sp_stw_park_locked            （已停在 safepoint）
#   主執行緒 sp_sched_sleep
# 也就是 STW barrier 等不到那條卡在 write(2) 的 thread 進 safepoint。
# 與已修的 #4528（readpartial 下裸的 read(2)）是同一個形狀。
require 'net/http'
require 'socket'

FEED_PORT = 18089
HTTP_PORT = 18088
FEEDERS = 5
CLIENTS = 5
ROUNDS = 30
CHUNK = 'x' * 65_536

# 一直送資料的 server，與一直讀的 client：兩邊都不停
feeder = TCPServer.new('127.0.0.1', FEED_PORT)
Thread.new do
  loop do
    sock = feeder.accept
    Thread.new do
      loop { sock.write(CHUNK) }
    rescue StandardError
      nil
    end
  end
end
FEEDERS.times do
  Thread.new do
    sock = TCPSocket.new('127.0.0.1', FEED_PORT)
    loop { sock.readpartial(65_536) }
  rescue StandardError
    nil
  end
end

# 最小 HTTP server，只回 200
http_server = TCPServer.new('127.0.0.1', HTTP_PORT)
Thread.new do
  loop do
    sock = http_server.accept
    Thread.new do
      sock.readpartial(4096)
      sock.write("HTTP/1.1 200 OK\r\nContent-Length: 2\r\nConnection: close\r\n\r\nok")
      sock.close
    rescue StandardError
      nil
    end
  end
end
sleep 1

ROUNDS.times do |round|
  started = Time.now
  threads = CLIENTS.times.map do
    Thread.new do
      Net::HTTP.start('127.0.0.1', HTTP_PORT) do |http|
        http.request(Net::HTTP::Get.new('/'))
      end
    end
  end
  threads.each { |t| t.join(10) }
  stuck = threads.count(&:alive?)
  puts "round #{round}: #{CLIENTS - stuck}/#{CLIENTS} ok (#{(Time.now - started).round(1)}s)"
  $stdout.flush
  next unless stuck.positive?

  puts "STUCK at round #{round}"
  exit 1
end
puts "SURVIVED #{ROUNDS} rounds"
```

## No test is included, deliberately

I tried four shapes — a peer that never reads, a peer that drains slowly, a shrunken
`SO_SNDBUF`, and `SPINEL_WORKERS=1` — and all four pass on the unpatched runtime: the
fully-full case parks correctly, and the scheduler grows a worker when one blocks, so
nothing deadlocks until a collection happens to land inside the partial-room window.
The script above hits it reliably but by timing, which would make a flaky addition to
the suite. Happy to add one if you see a deterministic shape I missed.

## Environment

spinel a9089ec, macOS 26.6.2 (Apple M3), Apple clang 21.0.0.
Full suite after the patch: 3718 pass, 0 fail, 0 error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved nonblocking stream-socket writes by respecting available buffer space and retrying safely when the buffer is full.
  * Preserved complete message boundaries for datagram and sequenced-packet sockets, including appropriate handling of oversized messages.
  * Preserved existing socket descriptor settings and read-side behavior.


ryudoawaru:
Good catch on the message-oriented sockets — it was a real hole. `UDPSocket` handles carry `is_sock = 1` (sp_io.c:664), so they went through the same path and the cap could have split one datagram into two, or hidden the `EMSGSIZE` the kernel owes an over-long payload.

Fixed in ac05c47: the socket's `SO_TYPE` is asked once and cached beside `wnonblock`, and anything that is not `SOCK_STREAM` keeps the whole payload in one send.

Checked against CRuby, both now agree:

| | CRuby | Spinel |
|---|---|---|
| `UDPSocket#write` of 20 KB | `Errno::EMSGSIZE` | `Errno::EMSGSIZE` |
| 8 KB datagram | 1 message, intact | 1 message, intact |

The repro still completes 30/30 rounds, and the suite is 3718 pass, 0 fail, 0 error.

matz:
Merged as 171e0ff7 (your second head, with the datagram guard). One line differs from the branch: on Linux the room query is skipped, since MSG_DONTWAIT is honoured there and send answers what fit or EAGAIN, which the park already handles, so a socket write pays no extra syscalls on that path; the query stays where the flag is not enough (Darwin, the BSDs). Note for the repro itself: on Linux it never printed SURVIVED on master or on your branch, but that is a different thing from the freeze you saw. Spinel's main drains the runnable threads at exit (sp_sched_drain), where CRuby kills them, and your feeders loop forever, so the process runs on at 500% CPU rather than exiting; the final line sits in stdio's buffer. That exit-time divergence is separate from this PR and worth its own issue if it bites Lens.

matz:
One more data point from CI after the merge: the macOS lane hung once in test/io_close_wakes_parked_reader.rb's churn phase (two sockets pumped by threads, cross-thread closes, 2 seconds of connection churn) and passed on a rerun; the same test hung once on 3846dc50 before your change too. It is the shape of your freeze on a smaller scale and it has never hung on Linux here, so if you have a minute on your Mac: `bin/spinel test/io_close_wakes_parked_reader.rb -o t && for i in $(seq 20); do ./t | tail -1; done` would say whether the room cap closes it or something else is left in the Darwin write path.

ryudoawaru:
Ran it on the Mac. Short version: **I cannot reproduce that hang here, on either side of the room cap** — so this does not tell us whether the cap closed it.

| build | conditions | runs | hangs |
|---|---|---|---|
| dca3798 (with the cap) | back to back, 3 s apart | 100 | 0 |
| a9089ec (before it) | same | 100 | 0 |
| dca3798 | + 8 spinning CPU hogs | 20 | 0 |
| a9089ec | same | 20 | 0 |
| dca3798 / a9089ec | 25 each, quiet machine | 50 | 0 |

290 runs, no hang either way. The loop had a 60 s watchdog that would have run `sample` on the stuck process and kept the stacks; nothing fired.

One thing that did show up, and it may be worth ruling out on the CI box: with **no pause between runs**, run 15 onwards fails with

```
Connection refused (Errno::ECONNREFUSED)
```

and four copies in parallel makes nearly every run fail that way. That is this machine running out of ephemeral ports under the churn, not a hang — but if the macOS lane runs the corpus back to back, a `connect` that gets refused at the wrong moment might surface as something other than a clean failure. Spacing the runs 2–3 s apart made it go away completely here.

If it hangs again in CI and the job can keep an artifact, `sample <pid> 3` on the stuck process is what identified the write-side freeze — a thread sitting in `sp_sock_write → write`/`__sendto` with the collector in `sp_stw_collect_impl` and everything else parked at `sp_safepoint` would say it is the same shape again. If it is something else, the stacks will say that too. Happy to keep a longer loop running here overnight if that is more useful than guessing.

Environment: macOS 27.0 (26A428), Apple M3, 8 cores, Apple clang 21.0.0. (The 26.6.2 in the PR description was this machine before it updated mid-way; the patch itself was written and tested on both.)

Issue #4623, opened 2026-09-20
IO#pwrite drops every byte after the first NUL, and IO#seek reports a refused seek as -1 instead of raising


makenowjust:
## Symptom

**`IO#pwrite` loses every byte after the first NUL, silently.**

```ruby
io.pwrite("abcd", 0)              # CRuby 4 / Spinel 4
io.pwrite("\x00\x01\x02\x03", 4)  # CRuby 4 / Spinel 0
io.pwrite("ab\x00cd", 8)          # CRuby 5 / Spinel 2
io.write("ef\x00gh")              # CRuby 5 / Spinel 5   — the write side is right
```

No exception; the bytes that were not written keep their old contents. Any program writing binary through `pwrite` loses data.

**`IO#seek` reports a refused seek as a return value, not an error.**

```ruby
io.seek(-1, IO::SEEK_SET)   # CRuby raises Errno::EINVAL / Spinel returns -1
```

A caller that rescues `Errno` — the shape generated code uses for a seek syscall — reads that as a successful seek.

## Cause

- `sp_File_pwrite` sizes its operand with `strlen`. The write side stopped doing that in #3540: `sp_File_write_bin` sizes with the header length (`sp_str_byte_len`) so an embedded NUL reaches the descriptor, and the emitter picks it for a String operand and the `strlen` entry for a converted value. `pwrite` is that call's positional twin and never got the split.
- `sp_File_seek` returns `fseeko`'s result unchanged, so a failure arrives as `-1` rather than the descriptor's `errno`.

## Fix

`sp_File_pwrite_bin` beside `sp_File_pwrite`, with the emitter making the same String/non-String choice the write arm makes (both the typed and the boxed-receiver arm). `sp_File_seek` raises through `sp_file_raise_errno` when `fseeko` fails and answers 0 otherwise, as CRuby does.

A sweep of the neighbouring write paths found `write`, `syswrite`, `write_nonblock`, `File.write`, `File.binwrite` and `IO#<<` all binary-clean — but `print`, `puts` and `printf` cut at the first NUL the same way, in both the typed and the boxed forms. That is the same defect in different entries and is fixed separately. PR follows.

Issue #3540, opened 2026-08-08
IO#write truncates at null bytes


rofl0r:
## Title
`IO#write` with strings containing null bytes (`\0`) truncates the
write at the first null byte

## Category
I/O bug / C string handling defect

## Description
Under Spinel, writing a Ruby string that contains null bytes (`\0`)
via `IO#write` or `File#write` results in only the bytes before the
first `\0` being written. The C runtime's null-terminated string
semantics are leaking into Spinel's I/O layer.

This is critical for the agent backend, which uses null bytes as
sentinels in the response pipe (`os.write(self._pipe_w, b'\0')` to
signal end-of-response).

## Test case

```ruby
r, w = IO.pipe
w.write("hello\0world")
w.close
data = r.read(100)
puts "got " + data.length.to_s + " bytes"
r.close
```

## Expected
```
got 11 bytes
```

## Actual
```
got 5 bytes
```

## Workaround
Not yet determined. Writing byte-by-byte also truncates. The null
byte is silently dropped, meaning end-of-response markers cannot be
transmitted through pipes.

Possible workarounds being investigated:
- `IO#write_nonblock` (may bypass the truncation)
- Writing through `File.open` with binary mode (may not help)

## Workaround
Use `"\1"` (SOH, byte value 1) as the end-of-response sentinel instead
of `"\0"` (NUL). SOH is not a valid UTF-8 character and will not appear
in LLM-generated text, making it a safe sentinel that survives Spinel's
C-string truncation.

```ruby
# Instead of:  w.write("\0")
# Use:         w.write("\1")
```

PR #4630, opened 2026-09-20
puts and print size a String operand by its header, as write does (#4629)

### LINKED ISSUE #4629: IO#puts and IO#print stop at the first NUL byte, where IO#write does not
## Symptom

`IO#puts` and `IO#print` stop at the first NUL byte:

```ruby
s = "ab\x00cd"
io.print(s)   # CRuby writes 5 bytes / Spinel writes 2
io.puts(s)    # CRuby writes 6 (with the newline) / Spinel writes 3
io.write(s)   # 5 — the write side is right
```

No exception, the rest of the bytes simply never reach the descriptor. Both the typed and the boxed-receiver forms.

`printf` writes short too, but for a different reason — see the note below.

## Cause

`sp_File_puts` and the print path size their operand with `strlen`. `IO#write` stopped doing that in #3540: `sp_File_write_bin` sizes with the header length (`sp_str_byte_len`) so an embedded NUL reaches the descriptor, and the emitter picks it for a String operand and the `strlen` entry for a converted value. `puts` and `print` never got the split. (`IO#pwrite` had the same gap; that one is #4623.)

## Fix

`sp_File_puts_bin` beside `sp_File_puts` — header length, and the trailing newline decided by the real last byte — and `print`, which already emits through `sp_File_write`, picks `sp_File_write_bin` on the same String/non-String test. Both the typed arm and the boxed-receiver arm.

### printf is only half fixed by this, and the remaining half is not IO's

`printf` formats and then writes, so its write picks up the binary-safe entry here. But the formatting itself already lost the bytes:

```ruby
s = "ab\x00cd"
p format("%s", s).bytesize   # CRuby 5 / Spinel 2
p "#{s}!".bytesize           # CRuby 6 / Spinel 3
p ("x" + s).bytesize         # 6 — concat is fine
```

`format` / `sprintf` and string interpolation truncate at a NUL on their own, everywhere — not just on the way to an IO. That is a separate defect in the string layer (the formatter's `%s` and the interpolation the emitter writes inline), filed separately. **This PR does not claim printf is fixed.**

makenowjust:
Fixes #4629. Independent of the other open PRs — this branch sits directly on master.

`sp_File_puts` and the print path sized their operand with `strlen`, so a String with an embedded NUL was cut at the first one — no exception, the rest of the bytes never reaching the descriptor:

```ruby
s = "ab\x00cd"
io.print(s)   # wrote 2 bytes — CRuby writes 5
io.puts(s)    # wrote 3       — CRuby writes 6
io.write(s)   # 5 — the write side was already right
```

`IO#write` stopped doing that in #3540: `sp_File_write_bin` sizes with `sp_str_byte_len` and the emitter picks it for a String operand, the `strlen` entry for a converted value. `puts` and `print` never got the split (`IO#pwrite` had the same gap — #4623).

`sp_File_puts_bin` is puts's half — header length, and the trailing newline decided by the real last byte — and `print`, which already emits through `sp_File_write`, now picks `sp_File_write_bin` on the same test. Both the typed arm and the boxed-receiver arm.

### printf's write joins them, but printf is *not* binary-safe yet

`format` / `sprintf` and string interpolation truncate at a NUL on their own:

```ruby
p format("%s", "ab\x00cd").bytesize   # CRuby 5 / Spinel 2
p "#{s}!".bytesize                    # CRuby 6 / Spinel 3
p ("x" + s).bytesize                  # 6 — concat is fine
```

That is a defect in the string layer — the formatter's `%s`, and the interpolation the emitter writes inline — affecting every use, not just the way to an IO. It is filed separately, and **this PR does not claim printf is fixed**.

The new test pins both names through both receiver forms, a multi-argument call, a converted (non-String) operand, and a string that already ends in a newline — against CRuby's output in both modes.

`make test`: raise 3728 pass / 0 fail / 0 error; promote 79 ERR + 25 FAIL, the identical failure set to the same-machine master baseline.

## Summary by CodeRabbit

- **New Features**
  - `IO#print` and `IO#puts` now preserve embedded NUL bytes in strings instead of truncating output.
  - Binary-safe output is supported across relevant string-writing paths.

- **Bug Fixes**
  - Output now uses the full byte length of strings, including embedded NULs and trailing newline data.

- **Tests**
  - Added coverage for scalar, repeated, integer, newline-terminated, and boxed-receiver output cases.


matz:
Merged in a2b748d0 with one adjustment: the binary puts entry now receives the String value itself instead of the `""` stand-in the plain entry gets for nil. A bare literal has no header in front of it, so `sp_str_byte_len("")` would have read the byte before a C string constant; `sp_File_puts_bin(NULL)` prints the newline `puts nil` prints instead. Thanks.

Issue #4412, opened 2026-09-09
An ivar's array literal locks its element type: a mismatched push is silently coerced, dropped, or refused to compile


rubys:
An ivar initialized with a *homogeneous* array literal is locked to that
literal's element type. A later push of a different type through the reader is
not widened to a poly array and not rejected -- it is silently wrong, and which
way it is wrong depends on what else is in the program.

Found on `ba502fba` while probing the #4410 fix. Pure locals get this right, so
it is the ivar/reader path specifically.

## Reproducer

```ruby
class Model
  def initialize; @a = ["seed"]; end
  def a; @a; end
end

m = Model.new
e = m.a
p e.equal?(m.a)
e << 7
p m.a
p e
```

```
cruby   true    ["seed", 7]     ["seed", 7]
spinel  false   ["seed"]        ["seed", 7]
```

The reader handed back a copy, so the push landed only on `e`. Nothing raised.

## Three symptoms, one root

The same defect surfaces three different ways. All three are on `ba502fba`.

**1. The reader returns a copy and the push is lost** -- the reproducer above.
`equal?` is `false` where CRuby says `true`.

**2. The alias is preserved and the VALUE is coerced.** Add one direct
mismatched push anywhere earlier in the program and the symptom changes:

```ruby
Model.new.a << 7          # <- only difference
m = Model.new
e = m.a
p e.equal?(m.a)           # now true
e << 7
p m.a                     # ["seed", ""]   <- the 7 became an empty string
```

**3. A hard compile error,** when `require "set"` shifts inference. The
int-typed array meets a `char *`:

```ruby
require "set"
class Model
  def initialize; @errors = []; end
  def errors; @errors; end
  def validate; self.errors << "blank"; end
end
m = Model.new; m.validate; p m.errors
```

```
error: incompatible pointer to integer conversion passing 'char *' to
       parameter of type 'sp_int' (aka 'long') [-Wint-conversion]
```

This is the direction `[]` defaults to int and a String arrives. It is the same
lock seen from the other side.

## What it takes, and what it does not

Holding the program shape fixed and varying only the ivar's initializer, with
`e = m.a; e << 7`:

| initializer | result |
|---|---|
| `["seed"]` | **wrong** -- `["seed"]` |
| `[]` (and pushing `"s"`) | **wrong** -- `[]` |
| `Array.new` | correct |
| `[].dup` | correct |
| `["seed", 1]` | correct |

* **It needs the ivar + reader.** The same aliasing with pure locals
[...]

PR #4639, opened 2026-09-20
A typed-array read is judged by the receiver's class, not by every class that spells the method

rubys:
`conv_reads_shared_storage` (the copy refusal from #4412) decides whether a zero-argument call reads storage something else holds by scanning every scope in the program for a method of that *name* whose body is a single ivar read. The receiver is never consulted, so `(1..max).to_a` — a Range making a fresh array — is refused as a read of `@rows` whenever a `Result#to_a` answering `@rows` exists anywhere in the program.

Repro (CRuby prints `true`, `[1, 2]`, `["...", 4, 5, 6]`; spinel HEAD refuses):

```ruby
class Result
  def initialize(rows); @rows = rows; end
  def to_a; @rows; end
end

def pages(max, cur)
  return (1..max).to_a if max <= 3
  pages = (cur - 1..cur + 1).to_a
  pages.unshift "..." if pages[0] > 1
  pages
end

r = Result.new([1, 2]); a = r.to_a; a << 3
p r.to_a.equal?(a)
p pages(2, 1)
p pages(9, 5)
```

```
spinel: main.rb:17: unsupported widening a typed array READ from an object into a poly slot (the conversion copies, so writes would not be shared): node 34 (CallNode `to_a`) recv=RangeNode/ty10 argc=0
```

This is lobsters' pagination helper beside the `ActiveRecord::Result` stand-in every transpiled tree carries; c14bbb73 named "a range" as one of the four safe shapes the refusal was meant to keep.

The fix asks the receiver's class: an object-typed receiver resolves the name through its own chain (`comp_method_in_chain`) and is a read only if *that* method is an ivar read; a receiver the compiler knows to be a builtin reaches no user method and is never one. Only a receiver whose class is not known here (unknown, poly, or an implicit self) keeps the name-only scan, so #4412's shapes are unchanged.

- `test/typed_array_read_asks_the_receiver_class.rb` carries the shape (expected output is CRuby's).
- `test/ivar_array_widens_on_read.rb` (#4412) still passes.
- `make test`: 3731 pass, 0 fail.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of method calls on typed receivers so storage reads are determined by the receiver’s class.
  - Prevented methods with the same name on unrelated classes from affecting built-in types such as ranges, arrays, and hashes.
- **Tests**
  - Added coverage confirming class-specific behavior for array conversion and preserving unaffected range behavior.


rubys:
The macOS lane failed on `hash_store_operand_gc_root` with no stdout at all (78 expected lines, 0 actual, nothing on stderr); the three Linux lanes passed. The generated C for that test is byte-identical between master and this branch (`spinel test/hash_store_operand_gc_root.rb -S` on each, `cmp` clean), and it passes locally on macOS in 1.8s, so this looks like the GC-stress test being killed on the slower runner rather than anything the change touches. Rebased onto current master and pushed to re-run.

Issue #4637, opened 2026-09-20
An out-of-int64 integer literal is re-parsed and re-allocated at every evaluation


makenowjust:
An integer literal too large for `int64` is emitted as `sp_bigint_new_str("...", 10)` **at its use site**, so every evaluation re-parses the decimal string and allocates a fresh Bignum. In a loop that is once per iteration, for a value that cannot change — an `Integer` is immutable.

The 64-bit mask idiom makes this easy to hit: a tag mask like `0xFFFFFFFF_00000000` appears in the hot path of anything that packs a payload into the low half of a word.

```ruby
# the 64-bit-tag idiom: one out-of-int64 mask, used in a hot loop
acc = 0
i = 0
while i < 2_000_000
  acc += 1 if (i & 0xFFFFFFFF_00000000) == 0
  acc += 1 if (i | 0xFFFFFFFF_00000000) != 0
  i += 1
end
p acc
```

The loop allocates 4,000,000 Bignums and parses the same 20-character decimal string 4,000,000 times, purely to obtain a constant.

On an M-series mac (`--int-overflow=raise`, three runs, best of):

| | time |
|---|---|
| spinel (today) | 0.66 s |
| CRuby 4.0.7 | 0.36 s |

So the generated program is ~1.8x *slower* than the interpreter on a loop that is nothing but integer arithmetic — the mask allocation dominates.

Nothing is wrong with the answers; this is purely the cost of rebuilding a constant. The same literal written in several places pays it several times over, and the emitted C also carries one `sp_bigint_new_str("18446744069414584320", 10)` per use site.

The value is immutable and known at compile time, so it can be built once per distinct literal per translation unit and shared — the slot filled on first use and marked with the other TU globals.

PR #4643, opened 2026-09-20
The literal table is filled at startup, not on first use

### LINKED ISSUE #4641: The shared table for out-of-int64 literals is filled lazily, which races between threads
An out-of-int64 integer literal is now built once per distinct value and shared through a table slot (#4637). The slot is filled on first use:

```c
static sp_Bigint *sp_bigl_get(int i) {
  if (!sp_bigl[i]) sp_bigl[i] = sp_bigint_new_str(sp_bigl_s[i], 10);
  return sp_bigl[i];
}
```

Spinel runs `Thread`s as true parallelism with no GVL (`docs/thread.md`), so that read-test-write on a shared slot is a data race. Two threads that reach the same literal first both see `NULL`, both call `sp_bigint_new_str`, and both store — an unsynchronized write to a pointer another thread is reading, which is undefined behavior, and one of the two Bignums is dropped while it is still the value a thread is about to return.

```ruby
# every worker reaches the same literal for the first time at once
ts = 8.times.map { Thread.new { 0xFFFFFFFF_00000000 + 1 } }
p ts.map { |t| t.value }.uniq
```

The GC makes the window worse rather than better: the marker walks `sp_bigl` for the filled slots, and a Bignum built but not yet stored is reachable only from the building thread's frame.

The value is a compile-time constant, so nothing requires it to be built lazily: filling the table at startup removes the race by construction.

makenowjust:
Fixes #4641. Follows #4638.

The shared table for out-of-int64 literals is filled on first use. Spinel runs `Thread`s as true parallelism with no GVL, so that read-test-write on a shared slot is a data race: two threads that reach the same literal first both build a Bignum and both store it, and the one that loses is dropped while a thread is still about to return it.

## The change

`sp_tu_init` fills the table:

```c
for (int _i = 0; _i < N; _i++) sp_bigl[_i] = sp_bigint_new_str(sp_bigl_s[_i], 10);
```

Emitted **after** the globals mark hook is installed, so a collection triggered by the fill itself marks the slots already written; the marker keeps its NULL test for the slots past the one being built. `sp_tu_init` is always emitted and called when the table exists, since `g_bigl_n` is what puts the mark line into `sp_mark_user_globals` and that sets `g_re_init_needed`.

`emit_expr` then names the slot directly — `sp_bigl[i]` rather than `sp_bigl_get(i)`. The accessor existed only to sequence the lazy store; two reads of one slot in an argument list are ordinary reads, so `A - A` emits `sp_bigint_sub(sp_bigl[0], sp_bigl[0])` and `-Wunsequenced` has nothing to say. The generated code loses a call and a branch per use.

Nothing is built that was not built before — every distinct literal in the AST already had a slot; only the moment changes.

## Test

`test/bigint_literal_shared.rb` gains the threaded case, where several workers reach the same literal for the first time at once:

```ruby
ts = 4.times.map { Thread.new { 0xFFFFFFFF_00000000 + 1 } }
p ts.map { |t| t.value }.uniq
```

Also checked by hand: `SPINEL_WORKERS=4` over repeated runs, `--int-overflow=promote`, and `SPINEL_GC_STRESS=1`.

## Verification

- `raise`: **3735 pass, 0 fail, 0 error**
- The #4637 benchmark is unchanged at 0.32 s (from 0.66 s before the literal was shared)


makenowjust:
Superseded by b1860a46, which landed the same fix — the table filled in `sp_tu_init` after the globals marker is installed, and the reader reduced to a load. `test/bigint_literal_table_startup.rb` covers the threaded case this branch had added to `bigint_literal_shared.rb`. Closing; nothing here is left to take.

PR #4645, opened 2026-09-20
The typed IO::Buffer lowering takes a boxed offset and a boxed value

### LINKED ISSUE #4644: IO::Buffer#get_value / #set_value lose the typed lowering whenever the offset is boxed, so --int-overflow=promote never gets it
`IO::Buffer#get_value` / `#set_value` with a **literal** type symbol lower to a typed accessor — no symbol decode, no boxing. That is the whole point of the class: reading a binary layout in a loop.

The lowering also requires the **offset** to be a statically known machine integer, and `set_value` requires the same of the stored value. Neither is true as soon as the value reaches the call through a parameter that is also given something else — and under `--int-overflow=promote` it is never true, because every int local and parameter is boxed there. The loop the fast path exists for is exactly where it is lost.

```ruby
require "io/buffer"

buf = IO::Buffer.new(64)
i = 0
while i < 8
  buf.set_value(:U32, i * 4, i * 111)
  i += 1
end
i = 0
sum = 0
while i < 8
  sum += buf.get_value(:U32, i * 4)
  i += 1
end
p sum   # 3108 either way -- only the route differs
```

Counting the accessors in the generated C (`spinel -c`) for that one source:

| | `get_i` | `get_value` | `set_i` | `set_value` |
|---|---|---|---|---|
| `--int-overflow=raise` | 2 | 3 | 2 | 1 |
| `--int-overflow=promote` | 1 | 4 | 1 | 2 |

(The remaining `get_value`/`set_value` are the library's own, where the type really is a variable.)

A larger version of the same loop, 10.24M reads, on an M-series mac:

| | time |
|---|---|
| spinel `promote` | 0.30 s |
| spinel `raise` | 0.03 s |

A tenfold difference on a class whose reason to exist is the fast path.

Neither requirement is inherent. An offset can be unboxed on the way into the accessor — the compiler already does that for typed-array indices, raising CRuby's `TypeError` for a non-Integer. The stored value is the subtler one: it cannot simply be unboxed, because an integer type accepts a `Float` (truncating) and `u64`/`s64` accept a value too large for a machine word, so a boxed value has to reach the same store the generic path uses.

makenowjust:
Fixes #4644.

`IO::Buffer#get_value` / `#set_value` with a literal type symbol lower to the typed accessor, but only when the offset is a statically known machine integer — and under `--int-overflow=promote` every int local and parameter is boxed, so the loop the fast path exists for never got it.

## The offset

It needs no static type. `emit_int_expr` already handles a boxed node: it routes it through `sp_poly_arg_int_chk`, which raises CRuby's `TypeError` for a non-Integer — the same thing the compiler does for a typed-array index. So the condition, in the fold and in the analyzer twin that has to agree with it, just widens to accept a boxed offset.

## The stored value

This one cannot be unboxed the same way, and that is the interesting half. For an integer type CRuby accepts a `Float` and truncates it, and `u64`/`s64` accept a value wider than a machine word. Putting either through `sp_int` changes what the program answers, so `sp_poly_arg_int_chk` is the wrong tool:

```ruby
buf.set_value(:U32, off, 3.9)                    # 3, not a TypeError
buf.set_value(:U64, off, 18446744073709551615)   # stays a Bignum
```

Instead `sp_IOBuffer_set_v` hands the boxed value to `iob_set_core` — the same store the generic binding reaches, with only the symbol decode skipped:

```c
sp_int sp_IOBuffer_set_v(sp_IOBuffer *b, sp_int ty, sp_int off, sp_RbVal v) {
  return iob_set_core(b, ty, off, v);
}
```

## Effect

Accessor counts in the generated C for one unchanged source (`spinel -c`):

| | `get_i` | `get_value` | `set_i` | `set_v` | `set_value` |
|---|---|---|---|---|---|
| `promote`, before | 1 | 4 | 1 | – | 2 |
| `promote`, after | 2 | 3 | 1 | 1 | 1 |

which is what `raise` already emitted. The remaining `get_value`/`set_value` are the library's own, where the type really is a variable.

10.24M reads through a loop-computed offset, M-series mac, best of three:

| | time |
|---|---|
| spinel `promote`, before | 0.30 s |
| spinel `promote`, after | **0.10 s** |
| spinel `raise` (already folded) | 0.03 s |
| CRuby 4.0.7 | 0.49 s |

## Test

`packages/io/test/io_buffer_poly_offset_value.rb`. Its expectations are CRuby's own output, and it covers what the boxed route must not change: `TypeError` for a String offset and for a String value, a `Float` truncating into `:U32`, `:U64` round-tripping a value above `2**63-1` as a Bignum, `:S32` keeping its sign, `ArgumentError` for an out-of-range offset, and the loop-computed offset the whole thing is for. Every parameter in it is reached by two argument kinds, so the test exercises the boxed route in `raise` mode too, not only under `promote`.

## Verification

- `raise`: **3738 pass, 0 fail, 0 error**

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `IO::Buffer#get_value` and `set_value` with literal type symbols when offsets or values are dynamically typed.
  - Preserved correct handling for floating-point values, large integers, signed values, and boxed numeric values.
  - Maintained expected validation for invalid argument types and out-of-bounds offsets.
  - Improved compatibility with `--int-overflow=promote`, including loop-computed offsets.


matz:
Merged in 85a45274 with one addition on the way in: the new test carries `# spinel: int64`, since its U32 value above 2**31-1 and the U64 round trip are the 32-bit gap you filed as #4647, and the -m32 lane was red on exactly those lines.

makenowjust:
The `gcc -m32` job caught something real, and it was not this change: the new test exposed a pre-existing 32-bit bug, filed as #4647 and fixed in the second commit here.

`iob_int_operand` refused every Bignum for a 32-bit type, reasoning that a Bignum is out of range by definition. `sp_int` is `intptr_t`, so that holds only where a pointer is 64 bits — under `-m32` every value above `2**31-1` is already a Bignum, and `0xCAFEBABE` is an ordinary `U32`. Most of the `U32` range was unreachable there, and `buf.set_value(:U32, 0, 0xCAFEBABE)` raised `RangeError` on master with nothing but a literal. The type's own width now decides; a magnitude that does not fit still raises.

64-bit builds reach none of it — a Bignum that survives normalization there has `|v| > INT64_MAX`, so it is out of a 32-bit range anyway — and the suites confirm: `raise` 3739 pass / 0 fail / 0 error, `promote` identical to the 104-entry baseline. The test gained the width boundaries (`4294967295` stored, `4294967296` and `2**100` refused, `-2147483648` stored, `2147483648` refused), with CRuby's own output as the expectation.

Issue #4647, opened 2026-09-20
IO::Buffer refuses any U32/S32 value above 2**31-1 on a 32-bit build: a Bignum is judged out of range by definition


makenowjust:
`iob_int_operand` refuses any Bignum for a 32-bit `IO::Buffer` type, on the grounds that a Bignum is out of range by definition:

```c
/* 32-bit and s64: a Bignum is out of range by definition (|v| >= 2^63
   for any heap Bignum the program builds; smaller ones stay sp_int) --
   except s64 accepting the negative bound. */
```

`sp_int` is `intptr_t` (`lib/sp_types.h:90`), so that reasoning holds only where a pointer is 64 bits. On a 32-bit build every value above `2**31 - 1` is already a heap Bignum, and plenty of them are perfectly ordinary `U32`s:

```ruby
require "io/buffer"
buf = IO::Buffer.new(8)
buf.set_value(:U32, 0, 0xCAFEBABE)   # 3405691582
p buf.get_value(:U32, 0)
```

Built with `gcc -m32` this raises

```
bignum too big to convert into 'unsigned int' (RangeError)
```

where CRuby stores the value and prints `3405691582`. Any `U32` above `0x7FFFFFFF` is unreachable there — which is most of the range that makes `U32` worth asking for, and exactly the range a binary format uses for a 32-bit field.

The type's own width is what decides, not how the value happens to be represented: a Bignum whose magnitude fits the target should be accepted, and only one that does not should raise.

(64-bit builds are unaffected: a Bignum that survives normalization there has `|v| > INT64_MAX`, so it is out of a 32-bit range anyway.)

matz:
Fixed in 0224b789. The 32-bit types now decide by the field's own width: a Bignum whose magnitude fits (U32 up to 2**32-1, S32 within [-2**31, 2**31-1]) is stored, one past it raises as before. packages/io/test/io_buffer_u32_full_range.rb carries 0xCAFEBABE, the S32 bounds and the three out-of-range cases, and runs on the -m32 lane (no int64 marker), which is the build it is for. This box has no 32-bit libcrypt to link against, so the CI lane is the check.

PR #4653, opened 2026-09-20
A Bignum shift moves limbs instead of building 2**n

### LINKED ISSUE #4652: A Bignum shift builds 2**n and multiplies or divides by it, where a limb shift would do
`Integer#<<` and `Integer#>>` on a Bignum build `2**n` as a full Bignum and then run a general multiply or divide against it:

```c
sp_Bigint *sp_bigint_shl(sp_Bigint *a, int64_t n) {
  if (n < 0) return sp_bigint_shr(a, -n);
  return sp_bigint_mul(a, sp_bigint_pow(sp_bigint_new_int(2), n));
}

sp_Bigint *sp_bigint_shr(sp_Bigint *a, int64_t n) {
  if (n < 0) return sp_bigint_shl(a, -n);
  return sp_bigint_div(a, sp_bigint_pow(sp_bigint_new_int(2), n));
}
```

A shift is a limb `memmove` plus a bit rotate. The pow allocates and fills `n/64` limbs, and the multiply or divide then walks both operands — so the work grows with the *shift distance* as well as the operand, for an answer that never needed either.

The file already has the primitives: `mpz_mul_2exp` and `mpz_div_2exp`, declared at `lib/sp_bigint.c:90-91` and used elsewhere in it.

```ruby
# a wide value shifted by a varying count: the bit-field idiom on a Bignum
x = 3**2000
acc = 0
i = 0
while i < 20000
  acc += ((x >> (i % 997)) & 0xFF)
  i += 1
end
p acc
```

On an M-series mac:

| | time |
|---|---|
| spinel | 0.09 s |
| CRuby 4.0.7 | 0.04 s |

and a round-trip shape, `((x << (i % 251)) >> (i % 251)) == x` over `x = 7**4000`, 40000 iterations:

| | time |
|---|---|
| spinel | 0.27 s |
| CRuby 4.0.7 | 0.06 s |

Spinel is 2x to 4x slower than the interpreter at shifting a Bignum, which is arithmetic it should win outright.

One thing the swap is not free of: `mpz_div_2exp` shifts the magnitude and keeps the sign, so it truncates toward zero, where Ruby's `>>` floors. `-7 >> 1` is `-4`, not `-3`. A negative value that loses a set bit needs one subtracted — the current `sp_bigint_div` form gets this right by flooring, and a limb shift has to keep getting it right.

makenowjust:
Fixes #4652.

`sp_bigint_shl` / `sp_bigint_shr` built `2**n` as a full Bignum and then ran a general multiply or divide against it, so the work grew with the shift distance as well as with the operand. `mpz_mul_2exp` and `mpz_div_2exp` were already in the file (`lib/sp_bigint.c:90-91`): a limb `memmove` and a bit rotate.

## The floor

`mpz_div_2exp` shifts the **magnitude** and keeps the sign, so it truncates toward zero where Ruby's `>>` floors — `-7 >> 1` is `-4`, not `-3`. The old `sp_bigint_div` form got this right for free. So the low bits are examined and one subtracted when any was set:

```c
static int mpz_low_bits_set(const mpz_t *x, int64_t e) {
  size_t digs = (size_t)(e / DIG_SIZE), bs = (size_t)(e % DIG_SIZE);
  size_t whole = digs < x->sz ? digs : x->sz;
  for (size_t i = 0; i < whole; i++) if (x->p[i]) return 1;
  if (bs && digs < x->sz && (x->p[digs] & ((((mp_limb)1) << bs) - 1))) return 1;
  return 0;
}
```

That scan is over the limbs being discarded, which the old division walked anyway.

## Effect

M-series mac, best of three:

| | before | after | CRuby 4.0.7 |
|---|---|---|---|
| `3**2000` shifted by a varying count, 20000x | 0.09 s | **0.01 s** | 0.04 s |
| `7**4000` through `<< n >> n`, 40000x | 0.27 s | **0.03 s** | 0.06 s |

About 9x, and from losing to the interpreter to beating it.

Behaviour at the edges is unchanged: a shift past `MRB_BIGINT_BIT_LIMIT` still raises `RangeError: shift width too large`, since the old path reached `mpz_mul_2exp` through `sp_bigint_pow` as well.

## Test

`test/bigint_shift_limbs.rb`, with CRuby's own output as the expectation: the floor at every place a bit can be lost (`>> 1`, `63`, `64`, `65`, `100`, `101`, `1000` on `-(2**100) - 1`), exact negative shifts that lose nothing, a negative count routing to the other shift, and a wide round trip in both signs.

Beyond the test, I ran a differential sweep against CRuby — 30 Bignum receivers (positive, negated, and negated-minus-one across `2**64`, `2**100 ± 1`, `3**200`, `2**127`, `2**128 - 1`, `2**200 + 2**37 - 1`) by 17 shift counts in both directions, `<<` and `>>`. **2040 lines, all identical to CRuby.**

## Verification

- `raise`: **3738 pass, 0 fail, 0 error**

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved large-integer left and right shift operations for more accurate results.
  - Corrected right-shift behavior for negative values by applying floor-division semantics.
  - Supports shifts across limb boundaries and with large values.

- **Tests**
  - Added coverage for positive and negative shifts, large integers, boundary counts, opposite-direction shifts, and round-trip operations.


matz:
Merged in 90437be3. Its -m32 lane was red on packages/io/test/io_buffer_u32_full_range.rb, which is master's own test from 0224b789 (the 32-bit half of #4647); the load side is fixed in 4620b215 on top, nothing in this PR.

Issue #2024, opened 2026-07-11
[Runtime] Integer arithmetic overflow raises `RangeError` instead of promoting to Bignum


KieranP:
> (Note: I asked Claude Opus 4.8 to find issues with common Ruby patterns present in my Ruby codebases. It found the following issue, which I manually verified before publishing this report)

_Reproduced with `spinel 6fcc370c9418` vs `ruby 4.0.3`._

Ruby integers are unbounded, so an arithmetic result that exceeds a machine word is silently promoted to arbitrary precision. Spinel keeps runtime integer values as 64-bit fixnums; when an operation on fixnum values overflows int64 it raises `RangeError: integer overflow in <op>` at run time instead of producing the large integer. The program compiles fine — the error occurs during execution. A constant-folded literal expression such as `2 ** 100` (no variables) does produce the correct Bignum; the bug is specifically a fixnum-by-fixnum operation on runtime values.

## Repro

```ruby
a = 1000000000000; b = 1000000000000; p(a * b)
```

- **Ruby 4.0.3:** `1000000000000000000000000`
- **Spinel:** `integer overflow in * (RangeError)`

The same overflow-instead-of-Bignum shows up across the arithmetic operators:

```ruby
a = 9000000000000000000; p(a + a)          # Ruby: 18000000000000000000                    Spinel: integer overflow in + (RangeError)
a = 1; p(a << 70)                          # Ruby: 1180591620717411303424                  Spinel: integer overflow in << (RangeError)
b = 2; e = 100; p(b ** e)                  # Ruby: 1267650600228229401496703205376         Spinel: integer overflow in ** (RangeError)
f = 1; (1..25).each { |k| f = f * k }; p(f) # Ruby: 15511210043330985984000000              Spinel: integer overflow in * (RangeError)
```

A constant literal with no variables is folded correctly:

```ruby
p(2 ** 100)   # Ruby: 1267650600228229401496703205376   Spinel: 1267650600228229401496703205376
```

KieranP:
This looks to be intentional, per README, but I would recommend perhaps enabling BigInt by default, instead of having it disabled?

matz:
Status update rather than a fix, with the honest findings:

[...]

PR #4658, opened 2026-09-20
An out-of-range Float reaching an sp_int slot raises, rather than casting

### LINKED ISSUE #4657: Float#floor / #ceil / #round / #truncate / Integer(f) cast out-of-range values to sp_int: the UB lets -O delete the statement
`Float#floor`, `#ceil`, `#round`, `#truncate`, `Kernel#Integer(f)`, `Float#div` and `Float#divmod` cast the result straight to `sp_int`:

```c
{ double _t2 = (lv_f);
  if (isinf(_t2)) sp_raise_cls("FloatDomainError", ...);
  if (isnan(_t2)) sp_raise_cls("FloatDomainError", "NaN");
  (sp_int)floor(_t2); }
```

The non-finite cases are handled. A **finite** value whose integer part escapes `int64` is not, and casting it is undefined behavior — which the optimizer is entitled to act on:

```ruby
f = 2.0**70
p f.floor
p Integer(f)
p f.to_i
```

| | `-O0` | `-O1` and `-O2` (the default) |
|---|---|---|
| `p f.floor` | `9223372036854775807` | **nothing at all** |
| `p Integer(f)` | `9223372036854775807` | **nothing at all** |
| `p f.to_i` | `RangeError` | `RangeError` |

At `-O0` the cast saturates and the program prints a wrong number. From `-O1` the optimizer takes the UB as a promise that the branch is unreachable and deletes the statements around it: the program prints **nothing** and exits 0, where CRuby prints `1180591620717411303424` three times. An earlier build of the same source printed `4373126616` and then `4304084488` on a rerun — the value is not even stable.

`Float#to_i` is the one method that is honest about it. It goes through `sp_float_to_i_checked` (`lib/sp_cold.c`), which range-tests and raises:

```c
if (f >= 9223372036854775808.0 || f < -9223372036854775808.0)
  sp_raise_cls("RangeError", "float out of Integer range (Bignum promotion pending)");
```

Its six siblings skip that test. So whether an out-of-range Float conversion is a clear error, a wrong number, or a statement that silently does not run depends on which method was called and what `-O` the program was built at.

Separately from the UB: CRuby answers the Bignum for all seven, and spinel raises. Promoting a statically-int result is the larger question tracked on #2024 — but a `RangeError` is a defined outcome, and a deleted statement is not.

makenowjust:
Fixes #4657.

`Float#floor` / `#ceil` / `#round` / `#truncate`, `Kernel#Integer(f)`, `Float#div` and `#divmod` checked the non-finite cases and then cast straight to `sp_int`. A **finite** value whose integer part escapes `int64` made that cast undefined behavior, and the optimizer acted on it:

| `p (2.0**70).floor` | result |
|---|---|
| `-O0` | `9223372036854775807` |
| `-O1`, `-O2` (the default) | **nothing printed, exit 0** |

At `-O0` the cast saturates and the program answers a wrong number; from `-O1` the optimizer takes the UB as a promise that the branch is unreachable and deletes the statements around it. Two earlier builds of the same source printed `4373126616` and `4304084488` — not even stable.

`Float#to_i` was the one method that range-tested first (`sp_float_to_i_checked`), which is why it alone raised.

## The change

One helper beside the domain check it pairs with:

```c
static inline sp_int sp_float_fit_i(sp_float v) {
  if (v >= 9223372036854775808.0 || v < -9223372036854775808.0)
    sp_raise_cls("RangeError", "float out of Integer range (Bignum promotion pending)");
  return (sp_int)v;
}
```

and every cast that can be reached with an out-of-range value goes through it — twelve sites: the static `floor`/`ceil`/`round`/`truncate` in all three ndigits shapes (none, a literal negative, a runtime value), `div`, `divmod`, `Integer(f)`, and the poly runtime's `sp_poly_floor` / `ceil` / `round` / `truncate` / `round_n`.

The message is the one `to_i` already raises, so the seven methods now agree with each other.

## What this does not do

CRuby answers the Bignum for all of these; spinel raises. Promoting a statically-int result is the larger question on #2024 — the slot holds `sp_int`, and widening it would box every float floor in every hot loop. This PR only makes the outcome defined. The boxed (`sp_poly_*`) route could answer a Bignum without that cost, since it boxes anyway; that needs a `double` -> Bignum constructor (`frexp` and a shift, exact above 2^53) and is worth doing separately.

## Cost

None measurable: a compare against a constant on a path that already branches twice for NaN/Inf. 50M iterations of `(i * seed).floor` with a runtime-seeded, unfoldable multiplier: 0.05 s before, 0.05 s after.

## Test

`test/float_to_int_out_of_range.rb`. Every route out of range (each method directly, through `send`, through a boxed receiver, with a literal and a runtime `ndigits`, negated, `div`, `divmod`, `Integer()`); `FloatDomainError` still named as CRuby names it for `Infinity` / `-Infinity` / `NaN`; the largest Float that fits and the first that does not; and the in-range answers, which I checked line-for-line against CRuby (24 lines, identical). **The test output is byte-identical at `-O0` and `-O2`** — before the change those differed.

## Verification

- `raise`: **3745 pass, 0 fail, 0 error**
- `promote`: no new failures against the 104-entry same-machine master baseline. One baseline entry, `bignum_recv_int_ord_float_toi`, is absent — but it passes 3/3 alone on master too, so it is a parallel-run flake, not something this change repaired.

## Adjacent, not fixed here

`(-9223372036854775808.0).to_i` answers `nil` rather than `-9223372036854775808`: `INT64_MIN` is the `SP_INT_NIL` sentinel. That is the same on master and unrelated to this change, so I left it alone and kept it out of the test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Float-to-integer conversions now reliably raise an error when values exceed the supported integer range.
  * Conversions of infinity and NaN now report the appropriate floating-point domain error.
  * Updated rounding, truncation, division, and related operations to handle out-of-range floats consistently.
  * Valid in-range conversions continue to return the expected integer results.


matz:
Merged as bbfe1ee5, thank you. Two adjustments on the way in:

- The 32-bit lane was red on the new test. `sp_int` is `int32` there, so the int64 literals let `2.0**40` through to the same UB cast. `sp_float_fit_i` and the `sp_float_to_i_checked` it mirrors now test against `(sp_float)INTPTR_MIN`, which is exact at either width; `(2.0**40).floor` raises RangeError on the 32-bit build too.
- The four int64 boundary lines (largest fitting Float, first that does not) moved to `test/float_to_int_boundary.rb` under the `# spinel: int64` marker, so `float_to_int_out_of_range` itself still runs on all four lanes.

Issue #4659, opened 2026-09-20
promote: a block_given? else-arm emits sp_RbVal into an sp_int slot and the C build fails (bisected to b29ec9a7)


makenowjust:
Under `--int-overflow=promote`, this stops the C build on master:

```ruby
def maybe_yield(x)
  if block_given?
    yield x
  else
    x * 2
  end
end
puts maybe_yield(5)
maybe_yield(5) { |n| n * 3 }
```

```
reg.rb:5: error: assigning to 'sp_int' (aka 'long') from incompatible type 'sp_RbVal'
spinel: C compilation failed
```

Line 5 is the `else` arm. The method's inline result slot is declared `sp_int`, and the arm emits `sp_poly_mul(...)`, which is an `sp_RbVal`:

```c
sp_int _t3 = 0;
...
  _t3 = sp_poly_mul(lv__y2_x, sp_box_int_or_nil(2LL));
```

`--int-overflow=raise` is unaffected — it compiles and prints `10`.

**Bisected to b29ec9a7** ("Enumerable#each_with_object is written in Ruby and compiled with the program"). Its parent `4620b215` compiles the same program and prints `10`.

Two tests in the corpus go with it, both failing to build in promote mode from that commit and passing at its parent:

- `test/block2.rb:7`
- `test/block_given_tail_typed_per_call.rb:28`

Both are the same shape: a `block_given?` guard whose `else` arm is arithmetic on the parameter, called once without a block and once with one. Writing `each_with_object` in Ruby presumably brings a new yielding method into every program's analysis, which moves what the parameter's type unifies to — the arm is then poly where the slot was settled as `sp_int`.

The promote lane is not in CI, which is why this landed green.

matz:
Fixed in 273cb8ac. The promote pass widens every Integer local and return to a boxed slot after the fixpoint, and the blockless answer of an `if block_given?` tail (Scope.ret_noblock, new in b29ec9a7) was not on that list, so the inline result slot stayed sp_int while the else arm ran on the widened parameter. It now widens under the same exclusions as the return. Your program prints 10 in promote mode, and `test/block2.rb` and `test/block_given_tail_typed_per_call.rb` build there again.

Test: `test/block_given_else_arm_overflow_modes.rb`, checked under both modes by hand since the promote lane is not in CI.

makenowjust:
A second one of the same shape, from a different commit in the same family.

[...]

PR #4661, opened 2026-09-20
A block that always leaves fills the value slot its splice feeds

### LINKED ISSUE #4660: A method called with both a value-returning block and an always-raising one does not compile: the splice leaves void in a value slot
A method called with one block that returns a value and another that always leaves — `raise` or `throw` — does not compile. Not a wrong answer: the C build fails and there is no program.

```ruby
class C
  def set; @v = yield; self; end
  def v; @v; end
end
p C.new.set { 5 }.v
begin; C.new.set { raise "x" }; rescue => e; p e.message; end
```

```
m.rb:2: error: assigning to 'sp_int' (aka 'long') from incompatible type 'void'
spinel: C compilation failed
```

Three shapes reach it, all with the same cause:

```ruby
@v = yield                              # assigning to 'sp_int' from 'void'
3.times { |i| t += yield i }            # passing 'void' to parameter of incompatible type 'sp_RbVal'
def deflt; yield; rescue; -1; end       # assigning to 'sp_int' from 'void'
```

A plain `v = yield; v + 1` is fine, so whether the program builds depends on where the yield's value goes.

The inlined block body is spliced into a C statement expression whose value the slot reads:

```c
_t9->iv_v = ({
  sp_raise(sp_exc_msg_given(...));
});
```

The raising block never produces a value, so the statement expression is `void` — but the slot was typed `sp_int` from the *other* call site's block, which returns 5.

`{ return e }` is the same hole and is already filled: the splice appends `default_value(want_ty)` after an unreachable `return`, with the comment "C still needs a value expression there (a trailing `return;` makes the `({...})` void)". A diverging tail wants exactly that treatment, and `stmts_diverge` (codegen_stmt.c) already recognizes all three tails — `return`, `raise`, `throw`.

One wrinkle: the filler's type must stay the consumer's. Only a `return` carries a value whose own type can stand in — a raise's argument is its *message*, so taking it puts a `const char *` where an `sp_int` was wanted.

makenowjust:
Fixes #4660.

A method called with one block that returns a value and another that always leaves — `raise` or `throw` — did not compile. Not a wrong answer: the C build failed and there was no program.

```
error: assigning to 'sp_int' (aka 'long') from incompatible type 'void'
```

The inlined block body is spliced into a C statement expression whose value the yield's slot reads. A diverging block never produces one, so the splice left a `void` where the slot — typed from the *other* call sites' blocks — wanted an `sp_int`:

```c
_t9->iv_v = ({
  sp_raise(sp_exc_msg_given(...));
});
```

Three shapes reached it: `@v = yield`, `t += yield i` inside a block, and a bare `yield` in a method carrying a `rescue`. A plain `v = yield; v + 1` was fine, so whether a program built depended on where the yield's value went.

## The change

`{ return e }` is the same hole and was already filled — the splice appends `default_value(want_ty)` after the unreachable tail, because "C still needs a value expression there". `stmts_diverge` (codegen_stmt.c) already recognizes all three tails, so it now gates the same filler; it is exported for that.

The filler's type stays the **consumer's**. The existing code falls back to the return's own argument type when `want_ty` is unusable, and that fallback is now kept for a `return` only: a raise's argument is its *message*, so taking it would put a `const char *` where an `sp_int` was wanted — the bug one level over.

## Test

`test/yield_block_tail_diverges.rb`, expectations taken from CRuby: the three failing shapes; the slot typed `String`, `Float` and `Array` by the other site; the raising site placed *first*, before any site that gives the slot its type; two raising sites against one that answers; `throw` as well as `raise`; and `raise TypeError, "t"` with a class rather than a message.

## Verification

- `raise`: **3751 pass, 0 fail, 0 error**
- `promote`: failure set **identical** to a same-machine master baseline taken at `bbfe1ee5` (107 entries)

A note on that baseline: it is 107, not the 104 this branch's earlier runs used. Master itself gained promote failures in the meantime — `block2` and `block_given_tail_typed_per_call` stop the C build there, which I bisected to b29ec9a7 and filed as #4659, and `yield_result_through_forwarding_block` does the same. None of them involve this change: they fail identically with the patch reverted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed block-yield behavior when a block ends with `raise` or `throw`.
  * Improved handling of these diverging block tails in assignments, compound assignments, typed values, and rescued method calls.
  * Prevented incorrect results or type errors when different yield call sites expect different value types.
  * Preserved expected exception messages and raised exception classes.

* **Tests**
  * Added coverage for divergent block tails across common value and exception-handling scenarios.


matz:
Merged as 17f48b90, thank you. Gate 3757 pass here on the merged tree, and the expected file matches CRuby line for line.

Issue #4662, opened 2026-09-21
Enumerable#partition in Ruby (50e2d7bf): campfire's `a, b = relation.partition { }` fails the C build on Linux, and analyze does not finish on macOS


rubys:
Two things on the same program, both bisected to 50e2d7bf (Enumerable#partition and #group_by written in Ruby); 28aa0795 just before it is fine on both counts.

## Repro

A dated, immutable emit of Campfire (roundhouse 7a303170, campfire 977cbcd1), built on our box at that stamp with spinel db46a572:

```sh
curl -sSLO https://roundbench.showcase.party/archive/campfire/20260920-110811/spinel-tree.tgz
tar xzf spinel-tree.tgz && cd transpiled-campfire-spinel
SP_FIXPOINT_LOG=1 spin build
```

| spinel | macOS arm64, apple clang 21 | Linux x86_64, gcc 13 (GitHub runner) |
|---|---|---|
| 3cfd775e | 107 s, `[fp] rounds=12`, binary built | — |
| 17f48b90 (master) | not finished at 300 s; a test program from the same emit was killed at 28 min CPU (3cfd775e: 23 s) | front-end finishes in its usual time, C fails (below) |

## 1. The C build fails (Linux)

From CI on 17f48b90's artifact — both sites are `a, b = relation.partition { … }` where `Relation#partition` is `to_a.partition { |x| yield x }` and `to_a` is RBS-typed `Array[untyped]`:

```
app/controllers/users/profiles_controller.rb:55: error: assignment to ‘sp_PolyArray *’ from ‘sp_int’ {aka ‘long int’} makes pointer from integer without a cast [-Werror=int-conversion]
app/controllers/rooms/closeds_controller.rb:81: error: assignment to ‘sp_PolyArray *’ from ‘sp_int’ {aka ‘long int’} makes pointer from integer without a cast [-Werror=int-conversion]
```

The two Ruby sites (app/controllers/users/profiles_controller.rb, app/controllers/rooms/closeds_controller.rb in the archive):

```ruby
@direct_memberships, @shared_memberships = Membership.with_ordered_room(...).partition { |m| m.room.direct? }
@selected_users, @unselected_users = User.ordered(User.active).partition { |user| selected_user_ids.include?(user.id) }
```

A 40-line standalone with the same shape (a Relation class whose `partition` forwards to an `Array[untyped]` `to_a`, destructured into two ivars) compiles and runs correctly on master, so whatever it needs is elsewhere in the program.

[...]

PR #4663, opened 2026-09-21
A compiler that cannot find builtins/enumerable.rb says so and stops

rubys:
Follow-up to #4662. After bc32aa72 and 2eb4bf21 our CI was still red on the same two `sp_int` → `sp_PolyArray *` lines, and the reason turned out to be ours: the job stages the compiler by copying `spinel`, `spin`, `spinel_rbs_extract`, `lib/` and `packages/` out of the build — the list that was complete before `builtins/` existed. Without the file, `sp_splice_builtins` returns the source unchanged, there is no `Array#partition` anywhere since its emitter left in 50e2d7bf, and campfire's `a, b = relation.partition { }` compiles to an unconditional `sp_raise_nomethod("undefined method 'partition' for an instance of Array")` whose carrier fails the C build. A full checkout of the same commit built the same program clean, which is what made it look like a compiler regression.

This makes the missing file a diagnostic instead of a silent fallback:

```
$ mkdir dist && cp -L spinel dist/ && cp -R lib packages dist/   # no builtins/
$ dist/spinel x.rb
spinel: builtins/enumerable.rb not found beside the compiler (looked under /…/dist and its parent);
        the toolchain is incomplete -- it ships with `make install`. SPINEL_NO_BUILTINS=1 compiles without it.
```

`make install` already ships the directory; `SPINEL_NO_BUILTINS=1` keeps the A/B path. `builtins_partition_group_by`, `builtin_forwarded_yield_proc_form` and `cli-opts-test` pass on the patched compiler. The roundhouse side (staging `builtins/`) is fixed in rubys/roundhouse@0822829d.

## Summary by CodeRabbit

* **Bug Fixes**
  * Installation now reports a clear error and exits when the required built-in definitions cannot be found.
  * Error messages identify the directories searched and provide guidance for resolving the issue or bypassing built-in loading.


matz:
Merged as 34b4590b, thank you. Gate 3762 pass on the merged tree; all four lanes green.

PR #4664, opened 2026-09-21
sp_PolyArray_from_*_array roots its typed argument like the typed dup helpers

blacktm:
## Why

`sp_PolyArray_from_int_array`, `sp_PolyArray_from_str_array` and `sp_PolyArray_from_float_array` in `lib/spinel_rt.h` rebuild a typed array as a poly array with every element boxed. Each allocates its result first and reads the typed array's elements after. Nothing holds the typed array during that allocation but the C argument, which is not a root. The typed copy helpers with the same shape, `sp_IntArray_dup`, `sp_StrArray_dup` and `sp_FloatArray_dup` in `lib/sp_array.c`, each open with `SP_GC_ROOT(a)`. These three did not.

Codegen emits them wherever a typed array is widened into a poly slot: a local write whose slot has widened (`x = ints; x << "tail"`), the same write as an expression, a typed array written into a poly-array ivar slot, a typed array returned through a poly-array return slot, a typed array passed to a user method's general Array parameter, and a typed array passed to `concat` on a poly array. Every one of the six sites passes the typed array expression straight to the helper, so a fresh array from a method call is held by nothing else. With `ints` a method returning four hundred integers, `floats` one returning four hundred floats and `strs` one returning forty strings, this

```ruby
x = ints
x << "tail"
```

the same two lines with `floats`, and `x = strs; x << 7`, run 4000 times each in one program, are wrong on bc32aa72 129, 266 and 1 times respectively in a plain build, 3 of 3 runs, and that program crashes under `SPINEL_GC_STRESS=1`, 3 of 3. With the three helpers rooting their argument every count is 0 in both builds and the output matches CRuby.

## What

Three lines in `lib/spinel_rt.h`: each helper opens with `SP_GC_ROOT(a)`, the same root in the same position as the typed dup helpers. The root is pushed before the helper's `NULL` check; the root stack skips a `NULL` slot when it marks. The generated C of every program is unchanged, since the helpers live in the runtime header the generated C includes. optcarrot calls `sp_PolyArray_from_int_array` once per frame, through `@output.concat(@buffer)` at the end of `Optcarrot::APU#flush_sound`, where the argument is `self->iv_buffer`, an ivar the object holds. Its binary gains one root push and pop per frame; its generated C is byte-identical and it runs to its checksum below.

## How

The root holds the typed array from the helper's entry to its return, so the result allocation cannot free it. The result array is pushed into with `sp_PolyArray_push`, whose growth goes through the payload allocator and does not collect, so the result needs no root of its own, as in the typed helpers.

## Validation

`test/poly_from_typed_array_widen_root.rb` widens a fresh typed array from a method into a poly slot 4000 times for each of the int, float and string helpers, and 4000 times passes a fresh int array to `concat` on a poly array, optcarrot's site, counting wrong answers. It then prints one widened array of each kind and one `concat` result, and shows that a frozen typed array widened through the int helper stays frozen. On bc32aa72 its first line reads `[129, 266, 1, 17]` in a plain build, 3 of 3, and the stress build exits with a segmentation fault before printing, 3 of 3. On the branch it matches CRuby with and without stress, 3 of 3 each, and asan-plain and asan-stress are clean.

Deletion pin. The test's generated C is byte-identical on bc32aa72 and the branch. Built against master's header and runtime it gives master's two results, the wrong counts and the crash; built against the branch's header, `[0, 0, 0, 0]` plain and under stress, 3 of 3 each. Nothing but the three roots differs between the two builds.

Two of the emit sites were also measured in 4000-round programs of their own. A typed array from a method written into a poly-array ivar slot, `@a = ints`, is wrong 111 times plain and 4000 times under stress on bc32aa72, 0 and 0 on the branch, 3 of 3 each. A typed array from a method passed to `concat`, `x.concat(ints)`, is wrong 0 times plain and 4000 under stress, 0 and 0 on the branch. The `concat` site roots the helper's result, not its argument.

Element count matters. In 4000-round programs under stress on bc32aa72: a forty-element int or float array alone 0 of 4000, a four-hundred-element int or float array alone 4000 of 4000, a forty-element string array alone 92 of 4000, and the same forty-element string array 2873 of 4000 when the program runs a forty-element int loop before it. With the roots, 0 for each.

Gate on bc32aa72: 3761 tests and 62 benchmarks pass with no failure, optcarrot runs to its checksum 59662, ext-cruby, rbs-seed, rubyspec and the property checks pass, and the fixpoint leg reports no new non-convergence (the four programs that hit the round cap hit it at the merge-base too). The compatibility corpus holds at 938 of 2131 with nothing lost. The generated C of all 3761 swept programs, optcarrot included, is identical to master's. The one red leg, spin-e2e "pack: make in the pack", is red the same way on a pristine bc32aa72 checkout.

## Left alone

The runtime's own callers of the three helpers, the `keys` and `values` arms of the boxed hash dispatch in `lib/spinel_rt.h`, already root the typed array they pass, and keep those roots. The six codegen emit sites are not touched. `sp_ptr_array_box` in `lib/spinel_rt.h` has the inverse shape, a root on its result and none on its argument, and no program was made to go wrong through it here; it is not touched. `sp_StrArray_join` in `lib/sp_array.c`, noted in #4633 the same way, is not touched.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented data corruption when converting typed arrays into mixed-value arrays during repeated widening operations.
  - Preserved array contents and frozen status during these conversions.

- **Tests**
  - Added regression coverage for integer, float, string, mixed-value, concatenated, and frozen arrays.
  - Added validation for representative widened values and conversion stability.


matz:
Merged as f5a2c126, thank you. The expected file matches CRuby here and the gate is 3762 pass on the merged tree; all four lanes green.

